PR #16, opened 2018-08-24
[exp] Update xdl dependency of exp

sjchmiela:
Per https://exponent-internal.slack.com/archives/C1QP38NQ5/p1535106617000100. 🙂


fson:
Thanks!

PR #21, opened 2018-08-31
[xdl] add sdkVersions to unimodules config

esamelson:
# Why

Fixes https://github.com/expo/universe/issues/3059 . The current latest version of `xdl` generates bad Podfiles on new detached projects because it includes unimodules that are not dependencies of `expo@29.0.0`.

# How

Added a new `sdkVersions` field to module configs that is a semver range; when generating a Podfile, filter out the modules for which the `sdkVersion` of the project does not satisfy this range.

I filled out the `sdkVersions` of the current modules based on my knowledge, but @tsapeta or someone else should definitely double check them 😅 

# Test Plan

Generated a new SDK 29 project, detached it, and successfully ran `pod install` 🎉

### DIFF
--- a/packages/xdl/src/detach/IosWorkspace.js
+++ b/packages/xdl/src/detach/IosWorkspace.js
@@ -12,7 +12,6 @@ import {
   isDirectory,
   rimrafDontThrow,
   parseSdkMajorVersion,
-  spawnAsync,
   spawnAsyncThrowError,
   transformFileContentsAsync,
 } from './ExponentTools';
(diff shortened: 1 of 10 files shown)


packages/xdl/src/detach/IosWorkspace.js
```diff
@@ -182,7 +182,10 @@ async function _renderPodfileFromTemplateAsync(
     TARGET_NAME: projectName,
   };
   let reactNativeDependencyPath;
-  const detachableUniversalModules = Modules.getDetachableModulesForPlatform('ios');
+  const detachableUniversalModules = Modules.getDetachableModulesForPlatformAndSdkVersion(
+    'ios',
+    sdkVersion
+  );
```
tsapeta: is it the only place that we should update? I guess we also need to change sth in `android-tasks.js` and `ios-tasks.js`

PR #25, opened 2018-09-03
[exp] Add a command for generating a unimodule

Szymon20000:
Fixes https://github.com/expo/universe/issues/2837


packages/exp/src/commands/generate-unimodule.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * @flow
+ */
+
+export default (program: any) => {
+  program
+    .command('generate-unimodule')
+    .alias('gum')
```
fson: Please avoid adding new aliases.

packages/exp/src/commands/generate-unimodule.js
```diff
@@ -0,0 +1,111 @@
...
+    .description('Generate new unimodule.')
+    .asyncAction(async () => {
+      const readline = require('readline');
+      const proc = require('child_process');
+      const path = require('path');
```
fson: Imports should be in the beginning of the file and use the ES2015 syntax like this:
```js
import readline from 'readline';
import childProcess from 'child_process';
import path from 'path';
```

packages/exp/src/commands/generate-unimodule.js
```diff
@@ -0,0 +1,111 @@
...
+      const npmVersionOfTemplate = '1.0.1';
+
+      const rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+      });
+
+      const askQuestionAsync = (question, resultMerger) =>
+        new Promise((resolve, reject) => {
+          rl.question(`${question} `, answer => {
```
fson: Please use the function in [`prompt.js`](https://github.com/expo/expo-cli/blob/master/packages/exp/src/prompt.js) that returns a promise and also handles non-interactive mode.

packages/exp/src/commands/generate-unimodule.js
```diff
@@ -0,0 +1,111 @@
...
+        'android',
+        'src',
+        'main',
+        'java',
+        ...configuration.javaModule.split('.')
+      );
+
+      proc.execSync(`mkdir -p ${javaDir}`);
+      proc.execSync(`echo "package ${configuration.javaModule};" > ${javaDir}/Placeholder.java`);
+      proc.execSync(`echo "class Placeholder {}" >> ${javaDir}/Placeholder.java`);
```
fson: All the code above using Unix commands needs to be changed to use the JavaScript equivalents so that it also works on Windows.

packages/exp/src/expo_commands/generate-unimodule.js
```diff
@@ -0,0 +1,146 @@
...
+import fsExtra from 'fs-extra';
+import replace from 'replace';
+import os from 'os';
+const npmVersionOfTemplate = '1.0.1';
+
+const decompress = async() => {
+  return new Promise((resolve, reject) => {
+    targz.decompress({
+        src: `expo-module-template-${npmVersionOfTemplate}.tgz`,
+        dest: 'temp-expo-module-template',
```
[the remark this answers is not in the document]
tsapeta: and also replace all occurrences of `temp-expo-module-template` with that const

tsapeta:
@fson if you think this PR is fine then feel free to merge 😉

packages/exp/src/expo_commands/generate-unimodule.js
```diff
@@ -0,0 +1,148 @@
...
+        message: 'How would you like to call your module in Cocoapods? (eg. EXCamera) (leave empty to not include iOS part)',
+        name: 'podName'
+      })).podName;
+
+      configuration.javaModule = (await prompt({
+        message: 'How would you like to call your module in Java? (eg. expo.modules.camera)',
+        name: 'javaModule'
+      })).javaModule;
+
+      proc.execSync(`npm pack expo-module-template@${NPM_TEMPLATE_VERSION}`);
```
fson: Shouldn't we use the latest version? Otherwise updating the template will require us to release a new version of Expo CLI and users to update to it.

packages/exp/src/expo_commands/generate-unimodule.js
```diff
@@ -0,0 +1,148 @@
...
+        'java',
+        ...configuration.javaModule.split('.')
+      );
+      fsExtra.mkdirpSync(javaDir);
+
+      const placeholderPath = path.join(
+        `${javaDir}`,
+        `Placeholder.java`
+      );
+      fs.appendFileSync(`${placeholderPath}`, `package ${configuration.javaModule};${os.EOL}`);
```
fson: I think instead of platform specific `os.EOL` (`\n` or `\r\n`) you should just use `\n` since that's the line break character the other files in the template use too.

packages/exp/src/expo_commands/generate-unimodule.js
```diff
@@ -0,0 +1,148 @@
...
+
+          fs.renameSync(
+            path.join(`${configuration.jsName}`, `ios`, `EXModuleTemplate.xcodeproj`),
+            path.join(`${configuration.jsName}`, `ios`, `${configuration.podName}.xcodeproj`)
+          );
+      } else {
+        await fsExtra.remove(path.join(`${configuration.jsName}`,`ios`));
+      }
+
+      await replace({
```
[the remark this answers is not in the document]
Szymon20000: I removed await word. I think that this function is just synchronous. What's wrong with it?

packages/exp/src/expo_commands/generate-unimodule.js
```diff
@@ -0,0 +1,148 @@
...
+import fsExtra from 'fs-extra';
+import replace from 'replace';
+import os from 'os';
+
+const NPM_TEMPLATE_VERSION = '1.0.1';
+const TEMP_DIR_NAME = `temp-expo-module-template`;
+
+const decompress = async() => {
+  return new Promise((resolve, reject) => {
+    targz.decompress({
```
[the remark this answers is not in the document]
Szymon20000: I read the code of extractAsync function. There is insurance function call from tar package so probably you had problem with tar. I tried to use Extract.extractAsync but could not make it work.

packages/exp/src/expo_commands/generate-unimodule.js
Szymon20000: I added [ ^ ] before version so that we could fix the bugs in template without releasing a new version of Expo CLI.

packages/exp/package.json
```diff
@@ -79,9 +79,12 @@
     "ora": "^1.3.0",
     "progress": "^2.0.0",
     "qrcode-terminal": "^0.11.0",
+    "recursive-copy": "^2.0.9",
```
fson: Unused dependency

packages/exp/src/expo_commands/generate-unimodule.js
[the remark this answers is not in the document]
fson: Ok, sorry, I saw this package was using the async versions of `fs` functions, but on a closer look, it only uses them when passed `async: true` in the options. 🤔 

This seems fine now.

packages/exp/src/expo_commands/generate-unimodule.js
fson: 👍

PR #27, opened 2018-09-05
[xdl] add expo-print-callbacks to module config

esamelson:
Companion PR to https://github.com/expo/universe/pull/3113 . Adds new module to XDL config.

Tested by running `./generate-files-ios.sh` and ensuring that this new module was not added to the podfile (it's android only).


esamelson:
decided not to merge this and use a different solution instead -- sorry for the noise!

PR #30, opened 2018-09-12
[xdl][exp] export:merge for multimanifest support

quinlanj:
As new SDK versions are released, you may want to serve multiple versions of your app from your server endpoint. For example, if you first released your app with SDK 29 and later upgraded to SDK 30, you'd want users with your old standalone binary to receive the SDK 29 version, and those with the new standalone binary to receive the SDK 30 version.  

In order to do this, you can run `exp export:merge` to merge previously exported apps into a single multiversion app which you can serve from your servers.

Here is what the workflow would look like:

1. Release your app with SDK 29
2. Run `exp export --output-dir sdk29`. This exports the sdk 29 version of the app in the `sdk29` directory. 
3. Release your app with SDK 30
4. Run `exp export --output-dir sdk30`. This exports the sdk 30 version of the app in the `sdk30` directory. 
5. Merge into a single multiversion app by running `exp export --merge-src-dir sdk29 --merge-src-dir sdk30`. This dumps a multiversion app into the `dist` output directory. The asset and bundle folders contain everything that the source directories had. The `index.json` files are an array of the the indexes found in the source directories.

In a CI usecase where people wont have the previous output directories on the local filesystem, they can specify a remote URL to download a `tar.gz`, which we will unpack and merge. For example, `exp export --merge-src-url https://myserver.com/sdk29.tar.gz`


fson:
Re: the workflow:
How do you envision the full workflow would look like in a setup where an app is regularly released (published), for example using a CI server?

If I’ve understood correctly, each time you’re releasing, you need to have access to the files of all the previous releases that you want to still include in your manifest (for older binary versions of your app out in the wild). This seems a bit complicated, as the `dist` folders of old releases are not typically stored in the codebase.

I wonder if this scenario could be made easier by being able to automatically fetch an existing manifest (which can itself be a multimanifest) to merge in the new multimanifest. Basically the command could take a manifest URL as an input instead of (or in addition to) input folder.

Another complication is that merging is another step that you need to perform each time. I wonder if we could go as far as make this just an option of `expo export`, where you could optionally give it an existing manifest (or build) to merge in the resulting (multi) manifest?

quinlanj:
sorry, i thought i had dealt with this PR (oops!)

I've made the following changes:
1. I've added `merge` flags to `expo export`, instead of having a separate command
2. I've added a `merge-src-url`, so CI workflows can specify a `tar.gz` remote url to merge (which we will download and decompress). We need to grab the `tar.gz` of a project rather than just the manifest url because we also have to merge the assets and bundles, and you cannot know all the asset locations just from the manifest URL (well technically u could from unpacking the metro asset map in the bundle but that seems painful). 
3. merged in projects can also be multimanifests themselves

packages/expo-cli/src/commands/export.js
```diff
@@ -56,9 +69,88 @@ export async function action(projectDir: string, options: Options = {}) {
...
+
+  // Merge src dirs/urls into a multimanifest if specified
+  const mergeSrcDirs = [];
+
+  // src urls were specified to merge in, so download and decompress them
+  if (options.mergeSrcUrl.length > 0) {
+    // delete .tmp if it exists and recreate it anew
+    const rimrafP = util.promisify(rimraf);
+    const tmpFolder = path.resolve(projectDir, path.join('.tmp'));
+    await rimrafP(tmpFolder); // rm -rf tmpFolder
```
fson: Lately I've been trying to get rid of the separate `rimraf` and `mkdirp` libraries and to just use `fs-extra` since it can also do a `rm -rf` and returns promises from all functions without having to promisify anything:
```suggestion
    await fs.remove(tmpFolder);
```

packages/expo-cli/src/commands/export.js
```diff
@@ -56,9 +69,88 @@ export async function action(projectDir: string, options: Options = {}) {
...
+      options.outputDir
+    );
+    log(`Project merge was successful. Your merged files can be found in ${options.outputDir}`);
+  }
   log(`Export was successful. Your exported files can be found in ${options.outputDir}`);
 }
 
+const download = async (uri, filename) => {
+  return new Promise((resolve, reject) => {
+    request(uri)
```
fson: `xdl` is using `axios` for almost all HTTP stuff, so could we also use it here?

packages/xdl/src/Project.js
```diff
@@ -303,6 +304,96 @@ export async function getLatestReleaseAsync(
...
+      const outputBundleDir = path.resolve(projectRoot, path.join(outputDir, 'bundles'));
+      const bundlePromise = fs.copy(sourceBundleDir, outputBundleDir);
+      promises.push(bundlePromise);
+
+      await Promise.all(promises);
+    }
+
+    // put index.jsons into memory
+    const putJsonInMemory = async (indexPath, accumulator) => {
+      const index = await new JsonFile(indexPath).readAsync();
```
fson: You can also use `JsonFile` without creating an instance:
```suggestion
      const index = await JsonFile.readAsync(indexPath);
```

packages/xdl/src/Project.js
```diff
@@ -1,6 +1,8 @@
 /**
  * @flow
  */
+import JsonFile from '@expo/json-file';
+import bodyParser from 'body-parser';
```
fson: This line can be removed, seems to have come from a merge.
```suggestion
```

packages/expo-cli/src/commands/export.js
```diff
@@ -76,8 +168,10 @@ export default (program: any) => {
     )
     .option('-d, --dump-assetmap', 'Dump the asset map for further processing.')
     .option('--dev', 'Configures static files for developing locally using a non-https server')
-    .option('--s, --dump-sourcemap', 'Dump the source map for debugging the JS bundle.')
+    .option('-s, --dump-sourcemap', 'Dump the source map for debugging the JS bundle.')
     .option('-q, --quiet', 'Suppress verbose output from the React Native packager.')
+    .option('--merge-src-dir [dir]', 'A repeatable source dir to merge in.', collect, [])
+    .option('--merge-src-url [url]', 'A repeatable source tg.gz dir to merge in.', collect, [])
```
fson:
```suggestion
    .option('--merge-src-url [url]', 'A repeatable source tar.gz file URL to merge in.', collect, [])
```

packages/expo-cli/src/commands/export.js
```diff
@@ -76,8 +168,10 @@ export default (program: any) => {
     )
     .option('-d, --dump-assetmap', 'Dump the asset map for further processing.')
     .option('--dev', 'Configures static files for developing locally using a non-https server')
-    .option('--s, --dump-sourcemap', 'Dump the source map for debugging the JS bundle.')
+    .option('-s, --dump-sourcemap', 'Dump the source map for debugging the JS bundle.')
     .option('-q, --quiet', 'Suppress verbose output from the React Native packager.')
+    .option('--merge-src-dir [dir]', 'A repeatable source dir to merge in.', collect, [])
```
fson: Is this command documented more thoroughly somewhere? I wonder if people will understand what "a repeatable source dir" means.

quinlanj: i've got a sibling docs pr in universe: https://github.com/expo/universe/pull/3139

PR #31, opened 2018-09-12
[exp] Show a message about upgrading to expo-cli

fson:
screen shot 2018-09-12 at 20 58 28


brentvatne:
medium is really bad at exploring old posts so we should probably link to a specific post

PR #34, opened 2018-09-14
[xdl] fix versions endpoint scripts

esamelson:
Currently, all XDL and powertools scripts that modify the versions endpoint are broken. Since all of these scripts (but one) read and modify the staging endpoint, but `Api` includes the prod session id with the request, the endpoint returns a 401.

Worse, because the `_versionCache` does not differentiate between staging and prod versions, this error is swallowed and a cached copy of versions is returned (likely from prod). This causes any previous changes on the staging endpoint to be overwritten. We ran into this multiple times while doing the SDK 30 release and had to resort to manually overwriting the entire endpoint whenever we wanted to make any change 😱

This PR explicitly sets the `Expo-Session` header to empty on Versions requests to avoid this problem. No auth is needed to read the endpoint, and a separate secret is used to write to it, so this header is unnecessary. It also removes the `_versionCache` for the reasons above; AFAIK there is no need for it since the versions endpoint always has `max-age=0`.

Tested by running `pt-dev update-turtle-sdk-version` and `pt-dev promote-versions-to-prod` with linked XDL; both worked as expected after this change.

### DIFF
--- a/packages/xdl/src/Api.js
+++ b/packages/xdl/src/Api.js
@@ -224,7 +224,12 @@ export default class ApiClient {
   static port: number = Config.api.port || 80;
 
   static _versionCache = new Cacher(
-    () => ApiClient.callPathAsync('/--/api/v2/versions'),
+    () =>
+      ApiClient.callPathAsync('/--/api/v2/versions', null, null, {
+        headers: {
+          'Expo-Session': '',
+        },
+      }),
     'versions.json',
     0,
     path.join(__dirname, '../caches/versions.json')


fson:
I don't think removing the cache altogether is the solution. As far as I know, it's there so that Expo CLI (and previously CRNA) can work offline. For example, the `Doctor` checks require this information so it's requested when you run `expo start`. Because of the cache, `expo start` works even when the request fails, because you're not connected to internet. The cache is also warmed up by version data that's fetched when we publish xdl.

I think a better fix would be to key the cache filename by API hostname, for example:
```js
static _versionCache = new Cacher(
  () => ApiClient.callPathAsync('/--/api/v2/versions'),
  Config.api.host === 'exp.host' ? 'versions.json' : `versions.${Config.api.host}.json`,
  0,
  path.join(__dirname, '../caches/versions.json')
);
```

And to fix the problem of `Api.versionsAsync` (and all the other methods in `Api`!) sending a production session secret to staging API: we need to fix the authentication code in xdl so that it never mixes production and staging sessions. [The session secret is read here.](https://github.com/expo/expo-cli/blob/5df9f83baf3a7ee25256a6420617ea4252355b61/packages/xdl/src/User.js#L198) It's stored in the [`state.json` file managed by `UserSettings`](https://github.com/expo/expo-cli/blob/5df9f83baf3a7ee25256a6420617ea4252355b61/packages/xdl/src/UserSettings.js#L14). I think to really fix the underlying problem, we should use a different file for staging, using the same approach as with the versions cache above.

fson:
Thanks for the clarification. Given that this is a rather special case (a script that needs to access both production and staging APIs), I would not invest effort to make a too general solution to it. So I think your initial solution of skipping authentication for fetching versions is fine here.

So now we just need to make sure the versions cache continues to work as intended for offline usage.

esamelson:
Okay, I've updated this to keep the cache and just skip authentication when refreshing the cache. This should be fine as long as there are no errors (since those might get erroneously swallowed) but since there is only one command that writes to prod versions and it's pretty safe already, I'm fine with this.

PR #35, opened 2018-09-15
[xdl] default to yarn when installing new packages, if it exists

esamelson:
If you run `expo init` and then eject to ExpoKit immediately, the current script installs the `expokit` npm package using npm by default (since `yarn.lock` doesn't exist). However, `npm` does something funky here and totally erases the `node_modules/react-native` folder (aside from its own `node_modules`). This makes both `expo start` and, in the ios directory, `pod install` fail. While this is fixable by simply reinstalling node_modules, the fix may not be obvious from the error message, and this extra step should not be necessary.

Using `yarn` as the default fixes this at least for people who have `yarn` installed.

I'm not tied to this PR, just wanted to suggest this in case it's worth it!

### DIFF
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -40,6 +40,7 @@
     "axios": "0.16.2",
     "body-parser": "^1.15.2",
     "chalk": "^2.4.1",
+    "command-exists": "^1.2.7",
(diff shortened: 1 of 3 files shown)


fson:
Thanks! LGTM.

PR #36, opened 2018-09-17
[xdl] Change iOS deployment target version from 9.0 to 10.0

Szymon20000:


fson:
@tsapeta can you review this one?

tsapeta:
So, I guess only these two files need to be changed here: `packages/xdl/src/detach/IosAssetArchive.js`, `packages/xdl/src/detach/IosPodsTools.js`. There is no need to touch other files, especially `packages/xdl/caches/versions.json` 😛 
I'll checkout other files and merge it but before that we need to merge PR which is still in universe: https://github.com/expo/universe/pull/3162

PR #46, opened 2018-09-25
Accept an array of objects as the data for Android intent filters

mbillard:
Follow-up to fix a small mistake in https://github.com/expo/xdl/pull/30

The intent filters is intended to accept an array as "data" or an object however the array type wasn't included in the accepted types.

The underlying logic already works with arrays (a test covers that use case already).

Thanks!

### DIFF
--- a/packages/xdl/caches/schema-29.0.0.json
+++ b/packages/xdl/caches/schema-29.0.0.json
@@ -733,7 +733,7 @@
               "properties": {
                 "action": { "type": "string" },
                 "data": {
-                  "type": "object",
+                  "type": ["array", "object"],
                   "properties": {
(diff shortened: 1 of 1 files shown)


packages/xdl/caches/schema-29.0.0.json
```diff
@@ -733,7 +733,7 @@
               "properties": {
                 "action": { "type": "string" },
                 "data": {
-                  "type": "object",
+                  "type": ["array", "object"],
                   "properties": {
```
mbillard: Does properties work with an attribute that can either be an object or an array? I wasn't able to find any other similar scenario.

PR #53, opened 2018-09-27
Add Firebase Core package to client

EvanBacon:
Just opening this to have it open... 🙃

Ideally we would add Firebase when a dev has the option to remove modules in the turtle build.


tsapeta:
@fson iirc this is not going to be merged soon, as for now we don't want to include firebase modules into Expo Client

PR #56, opened 2018-09-28
[xdl] update release channel when publishing an ExpoKit project

esamelson:
Currently, the release channel of an ExpoKit project is set only once, at the time of ejecting. If the developer then publishes to a different release channel, the new manifest + bundle will be copied to the filesystem and embedded in the app, but the app will still be pointing to the old release channel. This can lead to a confusing situation in which it appears that OTA updates are broken.

This PR makes `publishAsync` also update the release channel of an ExpoKit project. This means that the embedded manifest + bundle will always match the hardcoded release channel after publishing, which should mitigate the situation above. I don't think this is unexpected behavior (and if developers did not want to change the release channel of their app, it would be easy enough to notice this happened using version control).

I tested this by publishing to a new release channel and then to `default` (by not specifying a channel), and saw that the constant was updated correctly both times.

### DIFF
--- a/packages/xdl/src/Project.js
+++ b/packages/xdl/src/Project.js
@@ -39,10 +39,13 @@ import * as ExponentTools from './detach/ExponentTools';
 import * as Exp from './Exp';
 import * as ExpSchema from './project/ExpSchema';
 import FormData from './tools/FormData';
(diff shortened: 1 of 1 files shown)


stdavis:
Just wanted to say thanks for the PR. It saved my bacon today.

PR #58, opened 2018-09-30
[Unimodule] Localization

EvanBacon:


packages/xdl/src/modules/config.js
```diff
@@ -205,6 +205,11 @@ const expoSdkUniversalModules = [
     libName: 'expo-font',
     sdkVersions: '>=30.0.0',
   },
+  {
+    podName: 'EXLocalization',
+    libName: 'expo-localization',
+    sdkVersions: '>=31.0.0',
+  },
```
tsapeta: Can we keep this array sorted alphabetically by `libName`? 😄 Sorry for nitpicking 😅

PR #69, opened 2018-10-03
Fixed eject without Jest

MatthiasKunnen:
This fixes
https://github.com/react-community/create-react-native-app/issues/554.

Original PR:
https://github.com/react-community/create-react-native-app/pull/748.

### DIFF
--- a/packages/expo-cli/src/commands/eject/Eject.js
+++ b/packages/expo-cli/src/commands/eject/Eject.js
@@ -206,16 +206,19 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
     pkgJson.scripts.ios = 'react-native run-ios';
     pkgJson.scripts.android = 'react-native run-android';
 
-    if (pkgJson.jest && pkgJson.jest.preset === 'jest-expo') {
-      pkgJson.jest.preset = 'react-native';
-      newDevDependencies.push('jest-react-native');
-    } else if (pkgJson.jest) {
-      log(
-        `${chalk.bold(
-          'Warning'
-        )}: it looks like you've changed the Jest preset from jest-expo to ${pkgJson.jest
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eject/Eject.js
```diff
@@ -206,16 +206,19 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
...
-      newDevDependencies.push('jest-react-native');
-    } else if (pkgJson.jest) {
-      log(
-        `${chalk.bold(
-          'Warning'
-        )}: it looks like you've changed the Jest preset from jest-expo to ${pkgJson.jest
-          .preset}. We recommend you make sure this Jest preset is compatible with ejected apps.`
-      );
+    if (pkgJson.jest !== undefined) {
+      pkgJson.scripts.test = 'jest';
```
fson: If the developer has changed the `test` script from the original `jest` (e.g. added arguments to Jest), I think we shouldn't overwrite it here. Let's just remove this line?

packages/expo-cli/src/commands/eject/Eject.js
```diff
@@ -206,16 +206,19 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
...
+
+      newDevDependencies.push('jest');
+
+      if (pkgJson.jest.preset === 'jest-expo') {
+        pkgJson.jest.preset = 'react-native';
+        newDevDependencies.push('jest-react-native');
```
fson: I think we can remove this line since the preset is `react-native` instead of `jest-react-native` now?

packages/expo-cli/src/commands/eject/Eject.js
MatthiasKunnen: Agree, the original author of this line wasn't sure about this either. See https://github.com/react-community/create-react-native-app/pull/473.

PR #93, opened 2018-10-16
[xdl] '/modules' -> '/packages' in detached workspaces

nicknovitski:
expo/expo#2380 moved all universal modules to `/packages`, but expo/expo#2425 is blocked until the conflict is resolved some way.

Are these paths safe to change?  Or is there a better approach?


nicknovitski:
Thank you right back.  I'll be waiting for the next release.

PR #100, opened 2018-10-18
[xdl] iOS shell app builds propagate xcodebuild errors

nicknovitski:
As documented in [xcpretty's usage instructions](https://github.com/supermarin/xcpretty/tree/456604313faf4a1de6a4e687c2412eaf6097c4b0#usage), it doesn't exit with the same status code as the xcodebuild command.

This unblocks expo/expo#2425.

### DIFF
--- a/packages/xdl/src/detach/IosShellApp.js
+++ b/packages/xdl/src/detach/IosShellApp.js
@@ -88,10 +88,11 @@ async function _buildAsync(
   relativeBuildDestination,
   verbose
 ) {
-  let buildCmd, pathToArtifact;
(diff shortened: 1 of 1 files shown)


nicknovitski:
Another annoying part of this is that xcpretty hides any output it doesn't recognize the format of, for example arbitrary output from build scripts.  This makes debugging build errors more difficult.  The way fastlane handles this is by adding `| tee /var/log/some.log |` to the invocation of xcodebuild.  We could copy that, but it's starting to seem that we'd benefit from having this function depend on fastlane rather than just xcpretty, and then use `fastlane run xcbuild {args}`.

PR #128, opened 2018-10-31
[expo-cli] Add a few manual pre-release checks

fson:
# Why

The `master` branch should be releasable at any time. It should be as straightforward as possible for Expo CLI maintainers to publish a new release. At the same time, we should make sure that when releasing new versions, we don't break the developer experience of Expo CLI users.

Offline development, logged-out development and ejecting has previously had many regressions, so these manual tests target those features specifically. To make releasing as smooth as possible, most of these manual tests should be automated in the future.

# How

Added a `preversion` script to the `expo-cli` package, which Lerna will run before bumping the package version when we run `lerna publish`. The script shows instructions for manual tests to perform before publishing.

# Test Plan

Run `lerna publish`. Answering "no" to the "Have you completed all the checks?" prompt will interrupt publishing.

### DIFF
--- a/packages/expo-cli/scripts/preversion.js
+++ b/packages/expo-cli/scripts/preversion.js
@@ -9,16 +9,16 @@ console.log(
     `
 ${bold('1) Offline development')}:
   Check that you're logged in (\`expo whoami\`) and then make sure
-  \`expo start\` works without internet connection.
+  \`<repo-dir>/packages/expo-cli/bin/expo.js start --offline\` works without internet connection.
   (Use Network Link Conditioner with the the "100% Loss" profile, or turn off your wi-fi.)
 ${bold('2) Offline development, logged out')}:
-  Same as (1), but first run \`expo logout\`
+  Run \`expo logout\` and then \`<repo-dir>/packages/expo-cli/bin/expo.js start\`.
 ${bold('3) Eject')}: Create an app and eject it immediately. Check that it builds.
 
-    expo init testapp --template blank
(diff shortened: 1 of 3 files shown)


packages/expo-cli/scripts/preversion.js
```diff
@@ -0,0 +1,49 @@
...
+let inquirer = require('inquirer');
+
+console.log(
+  boxen(bold("Please complete these checks before publishing the 'expo-cli' package:"), {
+    padding: 1,
+  }) +
+    `
+${bold('1) Offline development')}:
+  Check that you're logged in (\`expo whoami\`) and then make sure
+  \`expo start\` works without internet connection.
```
esamelson: If `expo start --offline` is the correct workflow for offline development, could you change the text here to reflect that, as well?

packages/expo-cli/scripts/preversion.js
```diff
@@ -29,7 +29,7 @@ ${bold('3) Eject')}: Create an app and eject it immediately. Check that it build
 
     # Test that it builds for Android
     cd ../android
-    ./gradlew installDevMinSdkDevKernelDebug`
+    ./run.sh`
```
fson: @esamelson I got the command from https://docs.expo.io/versions/v31.0.0/expokit/expokit#4-android-build-and-run, should that page in docs also be updated to use `run.sh`?

esamelson: ah yes -- thanks for pointing that out!

PR #129, opened 2018-10-31
Fix Android detach not being able to load embedded manifest

sjchmiela:
Should fix https://forums.expo.io/t/could-not-load-embedded-manifest-android/14713/6?u=sjchmiela


sjchmiela:
Ah, ok, wasn't sure whether we're moving towards or away from `/index.exp` 😃 Doesn't latest XDL still use `/index.exp`? https://github.com/expo/expo-cli/blob/master/packages/xdl/src/Project.js#L666

esamelson:
😮 yes, I believe that should be the proper fix. @quinlanj can you verify?

PR #140, opened 2018-11-02
[xdl] Fix code assuming node_modules is under the project root

fson:
A few places in the code assumed `node_modules` is directly under the root
folder of the app. This breaks, for example, when the project is using
Yarn workspaces and some modules are hoisted to the `node_modules` folder
in the root of the monorepo.

To fix this, instead of joining paths and assuming a certain folder structure,
we now use `require.resolve` to find packages using the Node package resolution
algorithm that will also look for them in the `node_modules` folders above.


brentvatne:
my only feedback is we need to be sure to test this on windows, things with file paths can be problematic sometimes

fson:
This has been released in `expo-cli@2.6.7`.

PR #150, opened 2018-11-06
[xdl] Skip logging POST network requests

gerardo15:
Because all network Request include "HTTP/1.1"

These solution will also hide POST Request as following:

::ffff:192.168.1.44 - - [05/Nov/2018:22:39:00 +0000] "POST /symbolicate HTTP/1.1" 200 - "-" "okhttp/3.6.0"


fson:
This is great, thanks @gerardo15!

PR #178, opened 2018-11-13
Add modules for background tasks

tsapeta:
Related PR: https://github.com/expo/expo/pull/2338

It also fixes some problems on expo/expo CI with the shell app, as it was not including modules that were not released yet (like these one that has `sdkVersion: '>=32.0.0'`), however we need to include them to properly test new modules with unversioned shell apps.


fson:
@tsapeta is this PR ready to be merged?

PR #220, opened 2018-11-28
[android] Shell and ejected apps unified

sjchmiela:
Step of https://github.com/expo/expo/issues/2621, prerequisite for https://github.com/expo/expo/pull/2836 and https://github.com/expo/universe/pull/3316.

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -375,6 +375,8 @@ export async function runShellAppModificationsAsync(
   let url: string = context.published.url;
   let manifest = context.config; // manifest or app.json
   let releaseChannel = context.published.releaseChannel;
+  // In SDK32 we've unified build process for shell and ejected apps
+  const shellAndEjectedAppsAreUnified = ExponentTools.parseSdkMajorVersion(sdkVersion) >= 32;
 
   if (!context.data.privateConfig) {
     fnLogger.warn('Warning: No config file specified.');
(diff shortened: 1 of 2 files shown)


packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -364,6 +364,8 @@ export async function runShellAppModificationsAsync(
   let url: string = context.published.url;
   let manifest = context.config; // manifest or app.json
   let releaseChannel = context.published.releaseChannel;
+  // In SDK32 we've unified build process for shell and ejected apps
+  const shellAndEjectedAppsAreUnified = ExponentTools.parseSdkMajorVersion(sdkVersion) >= 32;
```
esamelson: wonder if it's worth just setting `isDetached` to this instead? That way we don't have to worry about checking two separate variables in the future. Not sure if it's more or less confusing that way 😄

sjchmiela: https://github.com/expo/expo-cli/pull/220/commits/1ad19786d2fdfacae174dbe09def14c343938560:
-  changed the meaning of `isDetached` to `sdkVersion > 32 || user`, so when `isDetached == true` we know that `expoview` is precompiled and we should use `MainActivity`
- introduced `isRunningInUserContext`, so when it's true we know we should use paths instead of URLs

sjchmiela: I would say the changes up to https://github.com/expo/expo-cli/pull/220/commits/1ad19786d2fdfacae174dbe09def14c343938560 are safe, the last three commits are an attempt at not copying obsolete files to `android-shell-app` so we can eventually get rid of them also in https://github.com/expo/universe/blob/master/server/turtle/scripts/android/buildAndroidTarballLocally.sh#L15. I didn't have time to test these changes, unfortunately, I'll do that next week, sorry! 😅

PR #237, opened 2018-12-05
feat: output SHA-256 hash for universal links

jaulz:
In order to use universal links on Android the SHA-256 hash is needed (see https://docs.branch.io/pages/deep-linking/android-app-links/).


fson:
Seems useful. Thanks for adding this!

fson:
This has been released in `expo-cli@2.6.7`.

PR #267, opened 2018-12-18
Add an option to `expo init` using the advanced workflow with ExpoKit

fson:
# Why

We want to support [ExpoKit as a first class workflow](https://blog.expo.io/expokit-2019-1e5cb02106f8). Part of this is being able to create a project directly as an ExpoKit project ("advanced workflow") without having to first create a managed project and then eject it.

# How

This PR adds workflow choices to `expo init`. The user can choose between "managed" and "advanced" workflows. For now, the "advanced" workflow is flagged as "experimental".

A new UI for entering required options is included:
screen shot 2018-12-18 at 18 28 17
It aims to make it easier to see how the entered options map to `app.json`. For the advanced workflow we need to ask a few more questions, for the managed flow we only ask for the name and slug (slug now defaults to directory name).

I've also cleaned up the "detach" log output XDL prints when creating an advanced flow project from init or by ejecting.

# Test Plan

Create project with `expo init` (choose advanced workflow) and make sure the project builds and works.

Fixes #267.


fson:
Still getting this error `ENOENT: no such file or directory, lstat './assets/splash.png'`... looks like the path is assumed to be relative to the current working directory. I'll fix that.

fson:
The `'./assets/splash.png'` issue is also fixed now.

13013SwagR:
What is the option to fill the snippet in `--non-interactive` mode?

PR #268, opened 2018-12-18
[expo-xli][xdl][Android] Splash screen standalone app fixes

bbarthec:
# Why

It fixes messed up `resizeMode` on standalone Android apps.
Moreover when no `mdpi` to `xxxhdpi` image is provided it copies `splash.image` to `mdpi` directory as baseline resolution (was copied to `xxxhdpi` previously).

# Test

- [x] Needs to be tested on turtles staging.

### DIFF
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -70,6 +70,7 @@
     "p-timeout": "2.0.1",
     "pacote": "9.2.3",
     "plist": "2.1.0",
+    "probe-image-size": "^4.0.0",
     "querystring": "0.2.0",
     "raven": "2.6.3",
     "read-chunk": "2.1.0",
--- a/packages/xdl/src/Android.js
+++ b/packages/xdl/src/Android.js
@@ -7,6 +7,7 @@ import fs from 'fs-extra';
 import spawnAsync from '@expo/spawn-async';
 import path from 'path';
 import semver from 'semver';
+import probeImageSize from 'probe-image-size';
 
 import * as Analytics from './Analytics';
 import * as Binaries from './Binaries';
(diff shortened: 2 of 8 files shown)


bbarthec:
@esamelson Here's the issue: https://github.com/expo/expo/issues/2533 and corresponding PR with docs and one small change in handling initial color:  https://github.com/expo/expo/pull/3029
Explanation: After checking all the logic, it turned out that in standalone Android apps `resizeMode` behaviour was swapped. `contain` was behaving as `cover` and vice-versa (according to docs).
I've fixed that and adjusted docs accordingly (Android was not described well at all).

packages/xdl/src/Android.js
```diff
@@ -329,3 +330,90 @@ async function adbReverseRemove(port: number) {
...
+    dpi: 'xxhdpi',
+    sizeMultiplier: 3,
+  },
+  {
+    dpi: 'xxxhdpi',
+    sizeMultiplier: 4,
+  },
+];
+
+async function checkAssetImageSize(
```
esamelson: could you unify this and the `getImageDimensionsMacOS` function in ImageUtils.js that provides the same functionality?

packages/xdl/src/Android.js
```diff
@@ -329,3 +330,90 @@ async function adbReverseRemove(port: number) {
...
+        expectedHeight,
+        sizeMatches: width === expectedWidth && height === expectedHeight,
+      });
+    }
+  }
+
+  if (androidSplashImages.length === 0) {
+    Logger.global
+      .warn(`Splash resizeMode is set to 'cover', but you haven't provided any images for different DPIs.
+Be aware that your splash image will be used as xxxhdpi asset and it's ACTUAL SIZE WOULD BE DIFFERENT depending on device's DPI.
```
esamelson: it's -> its  /  WOULD -> WILL

packages/xdl/package.json
```diff
@@ -71,6 +71,7 @@
     "p-timeout": "2.0.1",
     "pacote": "9.2.3",
     "plist": "2.1.0",
+    "probe-image-size": "^4.0.0",
```
esamelson: @fson is it ok with you to add this library as an xdl dep?

fson: It should be fine, xdl even already depends on it indirectly, through `@expo/schemer`.

packages/xdl/src/Android.js
fson: Please use **bold** (with `chalk`) for emphasis instead of SCREAMING.

PR #270, opened 2018-12-19
[expo-cli][xdl] Check android.package before starting build

### LINKED ISSUE #256: `expo build:android` should check configuration more strictly before actual building process
## Description

screenshot 2018-12-13 at 11 51 52

## Expected Behavior

Information about missing configuration in project/`app.json` should be provided at the beginning.
`expo build:ios` is prompting about missing configuration in `app.json`.

## Observed Behavior

Project is being checked before building process and despite the fact configuration is not meeting requirements we're prompted with `everting is okay - proceeding`.

## Reproducible Demo

* `expo init`
* navigate to project and `yarn`
* `expo build:android`

jakubste:
Closes #256 

zrzut ekranu 2018-12-19 o 11 40 19

### DIFF
--- a/packages/expo-cli/src/commands/build/AndroidBuilder.js
+++ b/packages/expo-cli/src/commands/build/AndroidBuilder.js
@@ -5,7 +5,7 @@
 import fs from 'fs-extra';
 import path from 'path';
 import untildify from 'untildify';
-import { Exp, Credentials } from 'xdl';
+import { Exp, Credentials, XDLError, ErrorCode } from 'xdl';
 import chalk from 'chalk';
 import log from '../../log';
 
@@ -212,11 +212,20 @@ export default class AndroidBuilder extends BaseBuilder {
   async validateProject(options) {
     const publicUrl = options.publicUrl;
     const {
-      args: { sdkVersion },
+      args: { sdkVersion, androidPackage },
     } = publicUrl
-      ? await Exp.getThirdPartyInfoAsync(publicUrl)
-      : await Exp.getPublishInfoAsync(this.projectDir);
+      ? await Exp.getThirdPartyInfoAsync(publicUrl, 'android')
+      : await Exp.getPublishInfoAsync(this.projectDir, 'android');
     await this.checkIfSdkIsSupported(sdkVersion, 'android');
+
+    if (!androidPackage) {
+      throw new XDLError(
+        ErrorCode.INVALID_OPTIONS,
+        `Your project must have an Android package set in app.json.
+See https://docs.expo.io/versions/latest/guides/building-standalone-apps.html`
+      );
+    }
+
     return sdkVersion;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/build/AndroidBuilder.js
```diff
@@ -212,11 +212,20 @@ export default class AndroidBuilder extends BaseBuilder {
   async validateProject(options) {
     const publicUrl = options.publicUrl;
     const {
-      args: { sdkVersion },
+      args: { sdkVersion, androidPackage },
     } = publicUrl
-      ? await Exp.getThirdPartyInfoAsync(publicUrl)
-      : await Exp.getPublishInfoAsync(this.projectDir);
+      ? await Exp.getThirdPartyInfoAsync(publicUrl, 'android')
+      : await Exp.getPublishInfoAsync(this.projectDir, 'android');
```
fson: `Exp.getThirdPartyInfoAsync` also needs to be updated to return the `androidPackage` field.

packages/xdl/src/Exp.js
```diff
@@ -284,16 +284,20 @@ export async function getPublishInfoAsync(root: string): Promise<PublishInfo> {
   const remotePackageName = name;
   const remoteUsername = username;
   const remoteFullPackageName = `@${remoteUsername}/${remotePackageName}`;
-  const bundleIdentifierIOS = exp.ios ? exp.ios.bundleIdentifier : null;
+
+  const platformSpecific =
+    platform === 'android'
+      ? { androidPackage: _.get(exp, 'android.package', null) }
+      : { bundleIdentifierIOS: _.get(exp, 'ios.bundleIdentifier', null) };
```
fson: To keep things simple, I suggest we just always return both of these fields (or `null` if they're not defined in the manifest) and remove the optional `platform` argument.

packages/expo-cli/src/commands/build/AndroidBuilder.js
```diff
@@ -212,11 +212,20 @@ export default class AndroidBuilder extends BaseBuilder {
...
     await this.checkIfSdkIsSupported(sdkVersion, 'android');
+
+    if (!androidPackage) {
+      throw new XDLError(
+        ErrorCode.INVALID_OPTIONS,
+        `Your project must have an Android package set in app.json.
```
fson: The error message should mention the field is `android.package`, or even better, prompt the user for the package name and write it to `app.json`, example: https://github.com/expo/expo-cli/blob/371be04175a676ae789544836ada6421a5d95937/packages/xdl/src/detach/Detach.js#L213-L225

wkozyra95: updating a local android package would be confusing for users using self-hosted apps. We would update local app.json and deploy remote one.

wkozyra95:
done in #367

Issue #262, opened 2018-12-17
expo url's do not respect REACT_NATIVE_PACKAGER_HOSTNAME


joshball:
## Description

REACT_NATIVE_PACKAGER_HOSTNAME listens on **correct hostname**, but expo still creates **incorrect** exp:// URL (apparently still based off of primary network interface).

The workaround is to construct your own exp://url on the device.

## Network Context
Assuming you are overriding a primary interface of 10.0.75.1 with a different interface (say 192.168.86.100) using REACT_NATIVE_PACKAGER_HOSTNAME 
```
os.networkInterfaces():
{
    "vEthernet (DockerNAT)": [{ "address": "10.0.75.1" }],
    "Ethernet": [{ "address": "192.168.86.78" }],
    "Wi-Fi": [{ "address": "192.168.86.100" }],
}
```

## Expected Behavior

```
set REACT_NATIVE_PACKAGER_HOSTNAME=192.168.86.100
$ yarn start
yarn run v1.12.1
$ expo start
[05:58:45] Starting project at W:\src\react-native\my-first-app
[05:58:46] Expo DevTools is running at http://localhost:19002
[05:58:46] Opening DevTools in the browser... (press shift-d to disable)
[05:58:51] Starting Metro Bundler on port 19001.
[05:58:52] Tunnel ready.
```
```
  exp://192.168.86.100:19000   *** EXPECTED ***
```

## Observed Behavior

```
$ yarn start
yarn run v1.12.1
$ expo start
[05:58:45] Starting project at W:\src\react-native\my-first-app
[05:58:46] Expo DevTools is running at http://localhost:19002
[05:58:46] Opening DevTools in the browser... (press shift-d to disable)
[05:58:51] Starting Metro Bundler on port 19001.
[05:58:52] Tunnel ready.
```
```
  exp://10.0.75.1:19000    *** ACTUAL***
```

## Environment

```
Expo CLI 2.6.13 environment info:
    System:
      OS: Windows 10
    Binaries:
      Yarn: 1.12.1 - C:\Program Files\nodejs\yarn.CMD
      npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD

{
    "vEthernet (DockerNAT)": [{ "address": "10.0.75.1" }],
    "Ethernet": [{ "address": "192.168.86.78" }],
    "Wi-Fi": [{ "address": "192.168.86.100" }],
}
```

## Reproducible Demo

[...]

Issue #259, opened 2018-12-14
expo gives a wrong LAN address link


PeichengLiu:
Having Docker installed on my machine, expo gives a wrong LAN address link.

![Screenshot](https://i.loli.net/2018/12/14/5c13aed098673.png)

My environment:

>  Expo CLI 2.6.13 environment info:
    System:
      OS: Windows 10
    Binaries:
      Yarn: 1.12.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
      Watchman: 4.9.4 - E:\Software\Util\watchman\watchman.EXE

fson:
Expo CLI internally uses `os.networkInterfaces()` to get a list of network interfaces and then tries to sort them according to some heuristics and finally returns an IPv4 address.

Could you copy paste the full output of this command from your computer:
```sh
node -e 'console.log(os.networkInterfaces())'
```
Then hopefully we can use this to information to improve the heuristics used to select the external network interface.

joshball:
@PeichengLiu It is because the Docker NAT interface has a higher priority, so it (node, using `os.networkInterfaces()`) is connecting to 10.0.75.1.
What you need to do is set 
`set REACT_NATIVE_PACKAGER_HOSTNAME=192.168.1.100`
Note, there still seems to be a bug where expo will continue to REPORT that it is listening on the primary (even though you have set the HOSTNAME). So you will have to manually construct your url:
exp://192.168.1.100:19000 in the expo client

PeichengLiu:
@joshball Thanks for your note. My temporary solution is shutting down the Docker, and waiting for this issue to be fixed.

fson:
We just released `expo-cli` version 3.3.0, which drastically improves the way we detect the correct LAN IP address. This should help with issues like this one.

If you're still having issues with IP address detection with the latest Expo CLI version, please open a another issue.

PR #288, opened 2019-01-06
Update ip.js

trimalcione:
It fixes this issue https://github.com/react-community/create-react-native-app/issues/60.

My purpose is to detect a virtual adapter simply looking up his OUI, not sure it is the most reliable way but after searching it seems to be.
Suggestions or improvements are well accepted.
Thank you.


packages/xdl/src/ip.js
```diff
@@ -44,6 +44,13 @@ function isPublic(addr) {
   return !isPrivate(addr);
 }
 
+function isVirtualNetworkAdapter(macAddr) {
+  var OUIVendors = [
+    '00:50:56' // VMware Network Adapter
+  ];
+  return OUIVendors.map( (oui) => {return macAddr.startsWith(oui)} )[0];
```
fson: This seems a bit more complicated than necessary, can you do just:
```js
const VMWARE_OUI = '00:50:56';
// ...
return macAddr.startsWith(VMWARE_OUI);
```

Or if the idea is to have a list of OUIs that can be later added to, then it should check each of them, like:
```js
const VIRTUAL_OUIS = [
  '00:50:56', // VMware
];
// ...
return VIRTUAL_OUIS.some(oui => macAddr.startsWith(oui));
```

packages/xdl/src/ip.js
```diff
@@ -92,7 +99,7 @@ function address(name, family) {
       //
       var addresses = interfaces[nic].filter(function(details) {
         details.family = details.family.toLowerCase();
-        if (details.family !== family || isLoopback(details.address)) {
+        if (details.family !== family || isLoopback(details.address) || isVirtualNetworkAdapter(details.mac)) {
```
fson: Since this filters out the interface altogether if it's a VMWare virtual network adapter, will this work for people who are running Expo CLI within a VMWare virtual machine?

Maybe we could tweak the [`priority`](https://github.com/expo/expo-cli/blob/371be04175a676ae789544836ada6421a5d95937/packages/xdl/src/ip.js#L9) function instead, so that other NICs would be sorted before them, but if no regular NIC exists, the virtual NIC address could still be returned?

trimalcione: Theoretically it should work both on virtualized systems and not.
I tried it practically (on a vm with VMware Workstation and Win10 as OS) and the result was what I expected basing on my knowledge. Launching an `ipconfig /all` on the vm it gives this result:
cattura

Virtualization is a job made from hypervisors which virtualize hardware, so NICs too.
For this, the virtual NIC you see on the container, will be virtualized once you boot the guest. So, basically, from the container you see a virtual NIC - having the VMware OUI - but from guest you see a virtualized NIC - having Intel OUI in my test.

The cool thing is that it should work even on a virtualized machine running on a virtualized machine, in a Russian dolls style.
Then fix affects only virtualized NICs created with VMware, which are the ones created using VMware hypervisor.

packages/xdl/src/ip.js
trimalcione: Definitely better, especially the second one. Thank you 👍

fson:
While we're fixing this, can we come up with a solution that would fix not only VMWare, but also some other virtual network adapters that people have reported having problems with?

Here are the ones that I found with a quick search:
- `VirtualBox Host-Only Network`: https://github.com/react-community/create-react-native-app/issues/60#issuecomment-287045080
- `vEthernet (DockerNAT)`: https://github.com/expo/expo-cli/issues/259#issuecomment-447557546, https://github.com/expo/expo-cli/issues/262#issue-391709092

I'm not sure if we should detect these by OUI. For example, [`vEthernet (DockerNAT)`](https://github.com/expo/expo-cli/issues/259#issuecomment-447557546) seems to use `00:15:5d` (Microsoft). VirtualBox previously used `08:00:27` (Cadmus Computer Systems 🙄), but now seems to have switched to `0a:00:27` (unregistered). I'm afraid filtering these out altogether may result in false positives and filter out some actual network interfaces too.

Perhaps there's some heuristic we can use instead, like in this function,
https://github.com/expo/expo-cli/blob/fd1fae89cca71d9caeeded489adb15c7a5f5d285/packages/xdl/src/ip.js#L9-L18
we could give the interface a lower priority if the name (converted to lowercase) starts with "virtual" or "vethernet"? Or even changing the current priority heuristic (prioritize "en*", "eth* and "wlan*") could be sufficient if we change it to be case insensitive? (On Windows the wi-fi interfaces seem to be named "WLAN", not "wlan", so the priority doesn't get applied to them.)

Also since you have the VMware installed, could you run this in the Node.js REPL and copy-paste the output here? This would be another data point in figuring out how to best recognize these interfaces:
```
var os = require('os')
os.networkInterfaces()
```

trimalcione:
```{
    Ethernet0: [
        {
            address: 'fe80::c4c2:a91e:2522:65fa',
            netmask: 'ffff:ffff:ffff:ffff::',
            family: 'IPv6',
            mac: '00:0c:29:c4:23:c2',
            scopeid: 8,
            internal: false,
            cidr: 'fe80::c4c2:a91e:2522:65fa/64'
        },
        {
            address: '192.168.1.245',
            netmask: '255.255.255.0',
            family: 'IPv4',
            mac: '00:0c:29:c4:23:c2',
            internal: false,
            cidr: '192.168.1.245/24' 
        } 
    ],
    'Loopback Pseudo-Interface 1': [
        {
            address: '::1',
            netmask: 'ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff',
            family: 'IPv6',
            mac: '00:00:00:00:00:00',
            scopeid: 0,
            internal: true,
            cidr: '::1/128' 
        },
        {
            address: '127.0.0.1',
            netmask: '255.0.0.0',
            family: 'IPv4',
            mac: '00:00:00:00:00:00',
            internal: true,
            cidr: '127.0.0.1/8'
        }
    ] 
}

**Comment by trimalcione** [comment]
>Also since you have the VMware installed, could you run this in the Node.js REPL and copy-paste the output here?

[This](https://hastebin.com/gusigojelo.cs) is the output on a VMware clean setup (Windows 10 x64) with a bridged network connection - bridged or not, that should not be a problem.
While [this](https://hastebin.com/eguduqibur.cs) is the output on a Ubuntu vm - same settings as the Windows one.

I think your purpose is more safe, as it's a generic solution.

> (On Windows the wi-fi interfaces seem to be named "WLAN", not "wlan", so the priority doesn't get applied to them.)

On my machine with Windows 10 x64 1809 installed, it's called "Wi-Fi".

**Comment by fson** [comment]
It looks like the hastebin links in the message don't work anymore. Could you paste the outputs again? (You can add them directly in the comment here.)

**Comment by trimalcione** [comment]
Sorry.
So, this is the output on a VMware clean setup (Windows 10 x64) with the features I write above: 
```
{ Ethernet0:
   [ { address: 'fe80::c4c2:a91e:2522:65fa',
       netmask: 'ffff:ffff:ffff:ffff::',
       family: 'IPv6',
       mac: '00:0c:29:c4:23:c2',
       scopeid: 8,
       internal: false,
       cidr: 'fe80::c4c2:a91e:2522:65fa/64' },
     { address: '192.168.23.134',
       netmask: '255.255.255.0',
       family: 'IPv4',
       mac: '00:0c:29:c4:23:c2',
       internal: false,
       cidr: '192.168.23.134/24' } ],
  'Loopback Pseudo-Interface 1':
   [ { address: '::1',
       netmask: 'ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff',
       family: 'IPv6',
       mac: '00:00:00:00:00:00',
       scopeid: 0,
       internal: true,
       cidr: '::1/128' },
     { address: '127.0.0.1',
       netmask: '255.0.0.0',
       family: 'IPv4',
       mac: '00:00:00:00:00:00',
       internal: true,
       cidr: '127.0.0.1/8' } ] }
```

and this is the output on an Ubuntu vm:
```
{ lo: 
   [ { address: '127.0.0.1',
       netmask: '255.0.0.0',
       family: 'IPv4',
       mac: '00:00:00:00:00:00',
       internal: true,
       cidr: '127.0.0.1/8' },
     { address: '::1',
       netmask: 'ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff',
       family: 'IPv6',
       mac: '00:00:00:00:00:00',
       scopeid: 0,
       internal: true,
       cidr: '::1/128' } ],
  ens33: 
   [ { address: '192.168.23.133',
       netmask: '255.255.255.0',
       family: 'IPv4',
       mac: '00:0c:29:1d:2b:f7',
       internal: false,
       cidr: '192.168.23.133/24' },
     { address: 'fe80::16d5:8292:8120:60d3',
       netmask: 'ffff:ffff:ffff:ffff::',
       family: 'IPv6',
       mac: '00:0c:29:1d:2b:f7',
       scopeid: 2,
       internal: false,
       cidr: 'fe80::16d5:8292:8120:60d3/64' } ] }
```

trimalcione:
I've just find out that for Linux, since version v197 of systemd - so since 2013 ([see here](https://github.com/systemd/systemd/releases/tag/v197)) - there's like a standard for network devices: you can find a good scheme of the standard [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-understanding_the_predictable_network_interface_device_names).

mekwall:
Just use https://github.com/sindresorhus/internal-ip. It solves all these problems.

PR #291, opened 2019-01-07
Allow using EXPO_CLI_PASSWORD env var for expo password

### LINKED ISSUE #126: Remove `-p` option in `expo login`, use a token env var instead
<table><tr><td>🚧</td><td>Moved from <a href="https://github.com/expo/expo/issues/172">https://github.com/expo/expo/issues/172</a></td></tr></table>
Currently not passing a password to `exp login` via `--password` in non-interactive mode throws an error: <a href="https://github.com/expo/exp/blob/1a89b83ba9ee921a4feabb4dc060271c40abbeea/src/accounts.js#L82-L89">https://github.com/expo/exp/blob/1a89b83ba9ee921a4feabb4dc060271c40abbeea/src/accounts.js#L82-L89</a>

So instead of the `--password`  (`-p`) option, we should change it to look for `EXPO_CLI_PASSWORD` env variable and remove the `--password` option from the CLI.

`EXPO_CLI_PASSWORD` needs to be added to the [blacklist](https://github.com/expo/xdl/blob/a67d52482e6246250d4e22ed2d95e0965c9310d6/src/Project.js#L1236) in xdl, so it's never exposed in the manifest of the development server.

epicfaace:
Fixes #126.

Usage:
Set environment variable `EXPO_CLI_PASSWORD` to expo password.
```bash
expo login --non-interactive -u username
```

### DIFF
--- a/packages/expo-cli/src/accounts.js
+++ b/packages/expo-cli/src/accounts.js
@@ -90,10 +90,12 @@ export async function login(options: CommandOptions) {
     return _usernamePasswordAuth(options.username, options.password);
   } else if (options.username && options.password) {
     return _usernamePasswordAuth(options.username, options.password);
+  } else if (options.username && process.env.EXPO_CLI_PASSWORD) {
+    return _usernamePasswordAuth(options.username, process.env.EXPO_CLI_PASSWORD);
   } else {
     throw new CommandError(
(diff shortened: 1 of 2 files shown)


fson:
This is really good. Thanks for the pull request.

PR #294, opened 2019-01-09
Minify production source code from `expo export`

### LINKED ISSUE #293: expo export does not minify code, only bundles it.
## Description

When using "expo export --dump-sourcemap", the resulting bundle is not minified/uglified, but merely transpiled/bundled.  "expo publish" minifies its bundles.

## Expected Behavior

When running "expo export" Resulting bundles should be minified/uglified to save space and obfuscate code.

## Observed Behavior

an example below...

```
Tims-MacBook-Pro:current tim$ head -20 dist/bundles/ios-23da2d567c84c33654551fcc725d65d4.js 
var __DEV__=false,__BUNDLE_START_TIME__=this.nativePerformanceNow?nativePerformanceNow():Date.now(),process=this.process||{};process.env=process.env||{};process.env.NODE_ENV='production';
(function (global) {
  'use strict';

  function _toConsumableArray(arr) {
    if (Array.isArray(arr)) {
      for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) {
        arr2[i] = arr[i];
      }

      return arr2;
    } else {
      return Array.from(arr);
    }
  }

  global.require = _require;
  global.__d = define;
  var modules = Object.create(null);

```

## Environment

I'm seeing it with expo-cli version 2.6.14, however, it has never worked since the release of "export" for expo-cli.

## To Reproduce

run "expo export" in a project dir.  Examine the resulting bundles in /dist

fson:
Fixes #293.

### DIFF
--- a/packages/xdl/src/Project.js
+++ b/packages/xdl/src/Project.js
@@ -415,10 +415,7 @@ export async function exportForAppHosting(
   await fs.ensureDir(bundlesPathToWrite);
 
   // build the bundles
-  let packagerOpts = {};
-  if (options.isDev) {
-    packagerOpts = { dev: true, minify: true };
-  }
+  let packagerOpts = options.isDev ? { dev: true, minify: false } : { dev: false, minify: true };
   const { iosBundle, androidBundle } = await _buildPublishBundlesAsync(projectRoot, packagerOpts);
   const iosBundleHash = crypto
     .createHash('md5')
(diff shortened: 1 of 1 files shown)


isthaison:
https://forums.expo.io/t/host-static-file-iis-not-working/20423/3.
This is reason! Can  you  add option  for me select 
minify on `expo export` 
@quinlanj

PR #317, opened 2019-01-21
Skip initial prompt on Build iOS app with bi --managed me | bi --managed expo

dimovdaniel:
if run
expo bi --managed me
or
expo bi -m expo

then the initial prompt will be skipped when building ios app


brentvatne:
hi! sorry nobody got around to looking at this. it looks like there are some merge conflicts that are preventing this, but also i think i need to understand the use case better before landing this change. if someone wants to pick this up and resolve merge conflicts and explain the use case i can look at it quickly. please open a new pr for it. thanks! :)

Issue #338, opened 2019-01-29
Broken preview link


SaadRegal:
## Description

After running `expo start` on a blank project, the cli generate a broken link  

## Expected Behavior

exp://192.168.137.1:19000

## Observed Behavior

exp69d2818b57134987cbe5d00a0da9cb27://192.168.137.1:19000

## Environment

Expo CLI 2.7.0 environment info:
    System:
      OS: Windows 10
    Binaries:
      Yarn: 1.10.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD

swang:
I too am experiencing this behavior. Cost me over an hour of time trying to figure out why creating a new app couldn't even get loaded. Have to revert to v2.7.0 to get it working again. :|

PR #372, opened 2019-02-12
[expo-cli] Remove the experimental workflow prompt from `expo init`

fson:
We previously added a prompt to `expo init` that allowed choosing between the `managed` (a regular Expo project) and `advanced` (an ExpoKit project) workflows.

The `advanced` workflow was marked as experimental, but it seems that people sometimes choose it by mistake, which results in confusing problems: https://forums.expo.io/t/really-stuggling-to-run-in-simultor/18743/2, https://github.com/expo/expo-cli/issues/338. Also once the current work on [universal modules](https://blog.expo.io/expokit-2019-1e5cb02106f8) is done, the upcoming `custom` workflow will likely be a better option for many use cases than the current `advanced` workflow.

This PR removes the experimental `advanced` workflow prompt from `expo init`. However, you can still initialize a project using the advanced workflow, if you pass the `--workflow advanced` flag to the command.

### DIFF
--- a/packages/expo-cli/src/commands/init.js
+++ b/packages/expo-cli/src/commands/init.js
@@ -95,7 +95,7 @@ async function action(projectDir, options) {
       );
     }
   } else {
-    workflow = await promptForWorkflowAsync();
+    workflow = 'managed';
   }
 
   let initialConfig = await promptForInitialConfig(parentDir, dirName, workflow, options);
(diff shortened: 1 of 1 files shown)


mohsinriaz17:
i am on expo-cli 2.11.6
--workflow advanced is deprecated,
the workflow is chosen based on the template selected
and there is no template which gives the detached project from the beginning like it was in the experimental feature

mohsinriaz17:
> We previously added a prompt to `expo init` that allowed choosing between the `managed` (a regular Expo project) and `advanced` (an ExpoKit project) workflows.
> 
> The `advanced` workflow was marked as experimental, but it seems that people sometimes choose it by mistake, which results in confusing problems: https://forums.expo.io/t/really-stuggling-to-run-in-simultor/18743/2, #338. Also once the current work on [universal modules](https://blog.expo.io/expokit-2019-1e5cb02106f8) is done, the upcoming `custom` workflow will likely be a better option for many use cases than the current `advanced` workflow.
> 
> This PR removes the experimental `advanced` workflow prompt from `expo init`. However, you can still initialize a project using the advanced workflow, if you pass the `--workflow advanced` flag to the command.

instead of removing a good feature, document it and then people can understand what is the advances / experimental features are. everyone just likes to select 'Advanced' and then usually get confused but that doesn't mean the features should be removed

if we started removing advanced features for people getting confused, we would end up removing all the features eventually 

one example is react-navigation
i have been using it since V1
and i still don't understand the flow of V3 completely and for that, there is V2 which i can use or i can push myself to learn V3, but that doesn't mean V3 should be removed because it was 'confusing' for someone like me

brentvatne:
@mohsinriaz17 - this PR removes initializing the 'advanced' workflow, which was just a temporary name that we had for ExpoKit projects. we are replacing ExpoKit with the 'bare' workflow, which just means using normal react-native and installing whichever expo packages you'd like to use. it's like ExpoKit but modular instead of monolithic. we removed it because we don't plan on supporting ExpoKit projects at some point in the nearish future, users will be encouraged to migrate to a 'bare' project. you can read more here: https://blog.expo.io/you-can-now-use-expo-apis-in-any-react-native-app-7c3a93041331

re: react-navigation, i don't know what to tell you - we had to change things to make the library better. if v1 works for you then great, don't bother updating! but probably you would benefit from spending an hour reading through the fundamentals on https://reactnavigation.org/ and the release notes for v2 and v3.

PR #375, opened 2019-02-13
[expo-cli] add select itc team in upload:ios command

wkozyra95:
- run fastlane with stdin
- add `--itc-team-id` option to upload ios

Closing #182

### DIFF
--- a/packages/expo-cli/src/commands/upload.js
+++ b/packages/expo-cli/src/commands/upload.js
@@ -43,7 +43,7 @@ export default program => {
     )
     .option(
       '--itc-team-id <itc-team-id>',
-      'iTunes connect Team ID (optional if there is only one team available)'
+      'App Store Connect Team ID (optional if there is only one team available)'
     )
     .option(
       '--apple-id-password <apple-id-password>',
--- a/packages/expo-cli/src/commands/upload.js
+++ b/packages/expo-cli/src/commands/upload.js
@@ -41,6 +41,9 @@ export default program => {
       '--apple-id <apple-id>',
       'your Apple ID username (you can also set EXPO_APPLE_ID env variable)'
     )
+    // apple unified App Store Connect and Developer Portal teams, this is temporary solution until fastlane implements those changes
+    // https://github.com/fastlane/fastlane/issues/14229
+    // after updating fastlane this value will be unnecessary
     .option(
       '--itc-team-id <itc-team-id>',
       'App Store Connect Team ID (optional if there is only one team available)'
--- a/packages/expo-cli/src/commands/upload.js
+++ b/packages/expo-cli/src/commands/upload.js
@@ -28,6 +28,7 @@ export default program => {
     ...COMMON_OPTIONS,
     'appleId',
     'appleIdPassword',
+    'itcTeamId',
     'appName',
     'sku',
     'language',
@@ -39,6 +40,10 @@ export default program => {
     .option(
       '--apple-id <apple-id>',
       'your Apple ID username (you can also set EXPO_APPLE_ID env variable)'
+    )
+    .option(
+      '--itc-team-id <itc-team-id>',
+      'iTunes connect Team ID (optional if there is only one team available)'
     )
     .option(
       '--apple-id-password <apple-id-password>',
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/commands/upload.js
```diff
@@ -40,6 +41,10 @@ export default program => {
       '--apple-id <apple-id>',
       'your Apple ID username (you can also set EXPO_APPLE_ID env variable)'
     )
+    .option(
+      '--itc-team-id <itc-team-id>',
```
fson: Can we unify this with `expo build:ios` and name this flag just `--team-id`?

wkozyra95: This is not the same team id as in `build:ios` recently apple unified iTunes connect app developer portal teams, but fastlane still requires ITC team.

fson: In case I need to help someone use this feature in the future, where does one find the iTunes Connect (now called App Store Connect) team ID? I searched, but couldn't find it on https://appstoreconnect.apple.com.

wkozyra95: Actually, I couldn't find it either but if you call upload:ios without this flag fastlane will list your teams interactively with those ids. I'm assuming that those ids were removed from App Store Connect after  February 12 (https://developer.apple.com/support/teams/)

packages/expo-cli/src/commands/upload.js
```diff
@@ -40,6 +41,10 @@ export default program => {
...
+      'iTunes connect Team ID (optional if there is only one team available)'
```
fson: iTunes Connect has been renamed to App Store Connect.

wkozyra95: I'm leaving --itc-team-id because this name is also used in interactive fastlane prompt and don't want to confuse users

packages/expo-cli/src/commands/upload.js
fson: Maybe we could have a team name flag instead of team ID flag (it would set the [FASTLANE_ITC_TEAM_NAME](https://github.com/fastlane/fastlane/blob/d738d74a1070b8a8eb8ffb69374589d929364ef2/spaceship/lib/spaceship/tunes/tunes_client.rb#L63) environment variable). I believe team names are usually the same between App Store Connect and Apple Developer website?

packages/expo-cli/src/commands/upload.js
fson: If you have many teams and you didn't pass `--itc-team-id`, what does Fastlane print (in non-interactive mode)? Can you copy-paste the output here?

I think the error message we show needs to tell that you should pass the `--itc-team-id` flag to the command in that case. (I believe the default message in Fastlane might refer to the `FASTLANE_ITC_TEAM_ID` environment variable instead.)

wkozyra95:
```
Ensuring the app exists on App Store Connect, this may take a while...

+----------------+--------------------------------------+
|              Summary for produce 2.112.0              |
+----------------+--------------------------------------+
| app_identifier | com.testExpo.testProject1233iut1ut2  |
|                | fre7r32r                             |
| app_name       | testProject                          |
| username       | turtle-dev@sixfivezero.net           |
| language       | English                              |
| sku            | 1550147107                           |
| platform       | ios                                  |
| skip_itc       | false                                |
| skip_devcenter | false                                |
+----------------+--------------------------------------+

[13:25:09]: [DevCenter] App 'com.testExpo.testProject1233iut1ut2fre7r32r' already exists, nothing to do on the Dev Center
Multiple App Store Connect teams found, please enter the number of the team you want to use: 
Note: to automatically choose the team, provide either the App Store Connect Team ID, or the Team Name in your fastlane/Appfile:
Alternatively you can pass the team name or team ID using the `FASTLANE_ITC_TEAM_ID` or `FASTLANE_ITC_TEAM_NAME` environment variable

  itc_team_id "119590309"

or

  itc_team_name "650 INDUSTRIES INC. - M7Y5F9WYK2"

1) "650 INDUSTRIES INC. - M7Y5F9WYK2" (119590309)
2) "650 Industries, Inc." (17102800)
3) "Alicja Warchał" (118989844)
4) "SOFTWARE MANSION SP Z O O SPOLKA KOMANDYTOWA" (118729549)
```

wkozyra95:
```
[14:16:22] Failed to upload the standalone app to the app store.
[14:16:22] Multiple App Store Connect Teams found; unable to choose, terminal not interactive!
```

packages/expo-cli/src/commands/upload.js
wkozyra95: names would solve that problem but it would be problematic for users and probably not portable between different shells (especially escaping spaces)

PR #382, opened 2019-02-16
Don't use a simple spinner when waiting for the build

TPXP:
The simple spinner can cause issues in basic terminals (e.g. CI terminals), in which it will create a new line per spinning tick. By using a more advanced spinner, we keep the spinner but avoid issues with these terminals.

Another approach could be to detect CI terminals and to disable the loader, but I think the ora spinner is cool, so why not use it here as well?

### DIFF
--- a/packages/expo-cli/src/commands/build/BaseBuilder.js
+++ b/packages/expo-cli/src/commands/build/BaseBuilder.js
@@ -6,6 +6,7 @@ import { Project, ProjectUtils, Exp, User } from 'xdl';
 import chalk from 'chalk';
 import fp from 'lodash/fp';
 import get from 'lodash/get';
+import ora from 'ora';
 import simpleSpinner from '@expo/simple-spinner';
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/build/BaseBuilder.js
```diff
@@ -6,6 +6,7 @@ import { Project, ProjectUtils, Exp, User } from 'xdl';
 import chalk from 'chalk';
 import fp from 'lodash/fp';
 import get from 'lodash/get';
+import ora from 'ora';
 import simpleSpinner from '@expo/simple-spinner';
```
tsapeta:
```suggestion
```

TPXP:
@fson May I bump this PR? According to what I know, we're awaiting your review :pray:

fson:
Looks good to me. Thanks 🙏

PR #398, opened 2019-02-26
[xdl] switch to versions/latest and use ApiV2 client

esamelson:
blocked by https://github.com/expo/universe/pull/3431

Move to new `versions/latest` endpoint that correctly conforms to the APIv2 protocol (response body should be wrapped in a root level object under a `data` key if there are no errors), and use APIv2 to fetch versions. The current endpoint does not conform to the protocol and cannot be used with the APIv2 client even though it has `api/v2` in the url.

### DIFF
--- a/packages/xdl/src/Versions.js
+++ b/packages/xdl/src/Versions.js
@@ -3,16 +3,35 @@
  */
 
 import _ from 'lodash';
+import path from 'path';
 import semver from 'semver';
 
 import Api from './Api';
 import ApiV2Client from './ApiV2';
+import { Cacher } from './tools/FsCache';
 import ErrorCode from './ErrorCode';
 import XDLError from './XDLError';
 import UserManager from './User';
 
 export async function versionsAsync() {
-  return await Api.versionsAsync();
+  const api = new ApiV2Client();
+  const versionCache = new Cacher(
+    () => api.getAsync('versions/latest'),
+    'versions.json',
+    0,
+    path.join(__dirname, '../caches/versions.json')
+  );
+  return await versionCache.getAsync();
+}
+
+export async function sdkVersionsAsync() {
+  const { sdkVersions } = await versionsAsync();
+  return sdkVersions;
+}
+
+export async function turtleSdkVersionsAsync() {
+  const { turtleSdkVersions } = await versionsAsync();
+  return turtleSdkVersions;
 }
 
 export async function setVersionsAsync(value: any) {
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Versions.js
```diff
@@ -3,16 +3,35 @@
...
+import { Cacher } from './tools/FsCache';
 import ErrorCode from './ErrorCode';
 import XDLError from './XDLError';
 import UserManager from './User';
 
 export async function versionsAsync() {
-  return await Api.versionsAsync();
+  const api = new ApiV2Client();
+  const versionCache = new Cacher(
+    () => api.getAsync('versions/latest'),
```
fson: Can we use a URL that doesn’t do a redirect here? I previously had to switch from `/--/versions` to `/--/api/v2/versions` (which the former redirects to) after I discovered the redirect added a 1-2 second round trip to our servers (at least at the time). Because many commands call this API, so it should be as fast as possible.

packages/xdl/src/Versions.js
```diff
@@ -3,16 +3,35 @@
...
+    'versions.json',
+    0,
+    path.join(__dirname, '../caches/versions.json')
```
fson: If the format of the result has changed, maybe we should use a different filename here? (E.g. `latest-versions.json` or `versions-v2.json`). This way any data cached by old versions wouldn’t break things.

Also look at the XDL gulpfile which saves this file to `xdl` folder (when publishing), for offline mode. Might have to change the name there as well.

packages/xdl/src/Versions.js
esamelson: I think this won't redirect bc the ApiV2Client adds the `/--/api/v2` part for us: https://github.com/expo/expo-cli/blob/91239153591d989b37bef525c20dbaeb93573de8/packages/xdl/src/ApiV2.js#L23
I confirmed this by logging the url in the axios request object.

packages/xdl/src/Versions.js
esamelson: The format of the raw response has changed but that's handled by ApiV2Client, so no changes should be necessary here. Tested by running `Versions.versionsAsync()` before and after this change, format of `versions.json` remained the same.

packages/xdl/src/Versions.js
fson: Ah, no problem then. I was just confused because I saw the redirect here: https://github.com/expo/universe/pull/3431/files#diff-483720538a0595eb9edfa7d2d9ce5316R317.

packages/xdl/src/Versions.js
fson: 👍

PR #404, opened 2019-03-05
Fix typo templaceSpec->templateSpec

ikewat:
Was unable to using custom template:
`expo init App --template expo-template-reason`
due to the typo.
detail:
```
ikewat λ EXPO_DEBUG=true expo init App --template expo-template-reason
templaceSpec is not defined
ReferenceError: templaceSpec is not defined
    at _callee$ (/expo-cli@2.11.6/src/commands/init.js:81:9)
...
...
...
```

After the fix, this script works fine.


fson:
Thanks for fixing this!

PR #407, opened 2019-03-06
Added web build command

EvanBacon:
# Why

Running `expo build:web` will generate static files.

# How

* Moved `WebpackDeepScopeAnalysisPlugin` to common
* Removed vendoring
* Removed custom `optimization` in favor of webpack defaults 
* Removed terser


packages/xdl/src/Project.js
```diff
@@ -1892,6 +1894,38 @@ async function stopWebpackServerAsync(projectRoot) {
   });
 }
 
+export async function bundleWebpackAsync(projectRoot, packagerOpts) {
+  let { exp } = await ProjectUtils.readConfigJsonAsync(projectRoot);
+  if (!exp.platforms.includes('web')) {
```
fson: Perhaps this should throw and error that explains how to enable `web` in app.json? Presumably people want to build for web if they run `expo build:web`, so it would be better to show an error than just not do anything.

packages/xdl/src/Project.js
```diff
@@ -1892,6 +1894,38 @@ async function stopWebpackServerAsync(projectRoot) {
...
+  }
+  let config = webpackConfig({
+    projectRoot,
+    development: packagerOpts.dev,
+    production: !packagerOpts.dev,
+  });
+  let compiler = webpack(config);
+
+  process.env.BABEL_ENV = 'production';
+  process.env.NODE_ENV = 'production';
```
fson: I guess these should also be set according to the `packagerOpts.dev` option like the webpack config above?

packages/xdl/src/Project.js
```diff
@@ -1892,6 +1894,38 @@ async function stopWebpackServerAsync(projectRoot) {
...
+  await new Promise((resolve, reject) =>
+    compiler.run((error, stats) => {
+      // TODO: Bacon: account for CI
+      if (error) {
+        // TODO: Bacon: Clean up error messages
+        return reject(error);
+      }
+
+      const statsPath = projectRoot + '/web-build-stats.json';
+      return bfj
```
fson: We use `@expo/json-file` for basically all JSON reading/writing in this repo, so we could avoid the `bfj` dependency and use `@expo/json-file`:
```js
import JsonFile from `@expo/json-file`;
  // ...
  let stats = await new Promise((resolve, reject) =>
    compiler.run((error, stats) => {
      // TODO: Bacon: account for CI
      if (error) {
        // TODO: Bacon: Clean up error messages
        return reject(error);
      }
      resolve(stats);
    })
  );
  await JsonFile.write(path.join(projectRoot, 'web-build-stats.json'), stats.toJson());
```

packages/expo-cli/src/commands/build/WebBuilder.js
```diff
@@ -0,0 +1,22 @@
...
+import { PLATFORMS } from './constants';
+
+const { WEB } = PLATFORMS;
+
+export default class WebBuilder extends BaseBuilder {
+  async run() {
+    // Check the status of any current builds
+    // await this.checkForBuildInProgress();
+
+    await Project.bundleWebpackAsync(this.projectDir, { dev: false });
```
fson: Should `--dev` be a command line flag for the build command?

packages/xdl/src/Web.js
```diff
@@ -2,14 +2,44 @@ import opn from 'opn';
...
     return { success: false, error: e };
   }
 }
+
+export function getWebSetupLogs() {
+  const appJsonRules = chalk.white(
+    `
+  ${chalk.whiteBright.bold(`app.json`)}
+  {
+    "packages": [
```
fson: I think this should be `"platforms": ["android", "ios", "web"]` (this is what we check in https://github.com/expo/expo-cli/blob/9137987338f5a4dbc372bbfba6f2653df2d7a3d5/packages/expo-cli/src/commands/start/TerminalUI.js#L146).

packages/expo-cli/src/commands/start/TerminalUI.js
```diff
@@ -142,9 +142,15 @@ export const startAsync = async projectDir => {
         return;
       }
       case 'w': {
+        const { exp } = await ProjectUtils.readConfigJsonAsync(projectDir);
+        const isWebConfigured = exp.platforms.includes('web');
+        if (!isWebConfigured) {
+          log(Web.getWebSetupLogs());
```
fson: Could you move this check and logging the error message to [`Web.openProjectAsync`](https://github.com/expo/expo-cli/blob/9137987338f5a4dbc372bbfba6f2653df2d7a3d5/packages/xdl/src/Web.js#L7)? Then the check would automatically also run when people run `expo start --web` ([here](https://github.com/expo/expo-cli/blob/9137987338f5a4dbc372bbfba6f2653df2d7a3d5/packages/expo-cli/src/urlOpts.js#L89-L91)).

packages/xdl/src/Project.js
```diff
@@ -1892,6 +1892,40 @@ async function stopWebpackServerAsync(projectRoot) {
...
+  await new Promise((resolve, reject) =>
+    compiler.run(async (error, stats) => {
+      // TODO: Bacon: account for CI
+      if (error) {
+        // TODO: Bacon: Clean up error messages
+        return reject(error);
+      }
+
+      const { stats: statsPath = 'web-build-stats.json' } = packagerOpts;
+      await JsonFile.writeAsync(path.join(projectRoot, statsPath), stats.toJson());
```
fson: I think we get an uncaught promise rejection if this method throws an error (because there's no `await` for this callback function anywhere). You could avoid this by calling `resolve(stats)` here and having a separate `await` after the `await new Promise(...)` call like this https://github.com/expo/expo-cli/pull/407#discussion_r263828107.

packages/xdl/src/Web.js
fson: The fields is `platforms`, not `packages`.
```suggestion
    "platforms": [
```

packages/expo-cli/src/commands/start.js
```diff
@@ -27,6 +28,17 @@ async function action(projectDir, options) {
...
     startOpts.reset = true;
   }
+  if (options.https) {
+    startOpts.https = true;
+  }
+  if (options.host) {
+    if (typeof options.host === 'string') {
+      startOpts.host = options.host;
+    } else {
+      startOpts.host = ip.address();
```
fson: We also accept `EXPO_PACKAGER_HOSTNAME` and `REACT_NATIVE_PACKAGER_HOSTNAME`: https://github.com/expo/expo-cli/blob/6dc22d9cf7796340293f8844515b0fa538147572/packages/xdl/src/UrlUtils.js#L238-L247.

In order to not make things any more complicated, could we just reuse that logic in `constructWebAppUrlAsync` and not have a `--host` flag at all? (You can extract that bit out to a separate function.)

The `--host` flag might be a bit confusing because it only applies to the webpack dev server and not other URLs created by `expo start`. And I would rather not have three or four ways to set the hostname.

packages/xdl/src/Web.js
```diff
@@ -1,15 +1,62 @@
...
+  const isWebConfigured = exp.platforms.includes('web');
+  return isWebConfigured;
+}
+
+function getWebSetupLogs() {
+  const appJsonRules = chalk.white(
+    `
+  ${chalk.whiteBright.bold(`app.json`)}
+  {
+    "packages": [
```
fson:
```suggestion
    "platforms": [
```
(The correct field is `platforms`, not `packages`.)

packages/expo-cli/src/commands/start.js
EvanBacon: we should do this in another PR. I'll move the current IP related changes.

PR #409, opened 2019-03-08
Updated webpack-config

EvanBacon:
Moved from #407 

* Removed web polyfills
* Udpated index.html style
* Use primaryColor for msapplication navbutton-color
* Moved WebpackDeepScopeAnalysisPlugin to common
* Removed vendoring
* Added support for `.otf` fonts
* Removed custom optimization in favor of webpack defaults
* Removed terser


packages/webpack-config/webpack/webpack.common.js
```diff
@@ -106,24 +107,7 @@ module.exports = function(env) {
...
-
-  function useWebModule(modulePathToHiJack, redirectPath, initialRoot = 'expo/build/web/') {
-    return new webpack.NormalModuleReplacementPlugin(
-      new RegExp(modulePathToHiJack),
-      getWebModule(initialRoot, redirectPath)
-    );
-  }
-
-  const publicPath = '/';
+  const publicPath = ''.replace(/\/$/, '');
```
fson: This is just an empty string `''`, is that on purpose?

EvanBacon: We may be adding `publicUrl` for export right? If so then this will prevent `//` in the index.html. This looks a little strange out of context because I split it into 2 PRs.

PR #415, opened 2019-03-11
[xdl] Allow deprecating universal modules

sjchmiela:


packages/xdl/src/modules/Modules.js
```diff
@@ -39,19 +44,23 @@ export function getVersionableModulesForPlatform(platform: Platform): Array<Modu
...
+    moduleConfig =>
+      !!moduleConfig.libName && doesVersionSatisfy(sdkVersion, moduleConfig.sdkVersions)
+  );
 }
 
-export function getPublishableModules(): Array<ModuleConfig> {
-  return expoSdkUniversalModulesConfigs.filter(moduleConfig => !!moduleConfig.libName);
+function doesVersionSatisfy(version: string, versionRequirement: string): boolean {
+  if (version === 'UNVERSIONED') {
+    return semver.satisfies(HIGHEST_KNOWN_VERSION, versionRequirement);
```
tsapeta: where is `HIGHEST_KNOWN_VERSION` defined?

sjchmiela: whoops, I removed it when refactoring 😅

what would you say to removing handling `UNVERSIONED` altogether?

tsapeta: iirc this was required for sth to work, but I don't remember for what 😅 maybe you can try 😂

sjchmiela: ah, for building ios shell apps, nvm

PR #425, opened 2019-03-15
Remove "SMS" from link-sending descriptions

ide:
This removes SMS from the UI for sending project development links. The server endpoint still behaves the same, but the UI now promotes email over SMS. Email is more reliable, much cheaper, and more customizable. We'll make a similar change on web as well.

## Test plan

Ran the CLI on a test project and verified that both the CLI and web UI showed just "email" without "SMS" for sending links.

Web:
Screen Shot 2019-03-15 at 12 23 58 PM

CLI:
Screen Shot 2019-03-15 at 12 24 13 PM

### DIFF
--- a/packages/dev-tools/TESTING.md
+++ b/packages/dev-tools/TESTING.md
@@ -14,8 +14,8 @@
   - ✅ Should show a toast that says simulator is opening.
   - ✅ Should replace the toast with success toast when finished.
   - ✅ Should replace the toast with an error toast, if opening failed.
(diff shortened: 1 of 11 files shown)


ide:
I missed a lot -- thanks for finding all of these.

ide:
Fixed all of those (searched for "phone" and "mobile" and looked through each result). Also changed "e-mail address" to "email address" to be consistent and more modern (https://www.grammarly.com/blog/spelling-e-mail-email/).

PR #426, opened 2019-03-15
Added unimodules to config

EvanBacon:
* Moved aliases from babel-preset-expo
* Added optional override for root webpack.config.js
* Added optional override for root webpack.dev.js
* Added optional override for root webpack.prod.js
* Fixed absolute location function
* Moved babel config to method
* Bump 0.1.3 -> 0.1.4

### DIFF
--- a/packages/webpack-config/webpack/createBabelConfig.js
+++ b/packages/webpack-config/webpack/createBabelConfig.js
@@ -1,0 +1,38 @@
+// Only compile files from the react ecosystem.
+const includeModulesThatContainPaths = [
+  'node_modules/react-native',
+  'node_modules/react-navigation',
+  'node_modules/expo',
+  'node_modules/unimodules',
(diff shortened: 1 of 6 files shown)


packages/webpack-config/webpack/createBabelConfig.js
```diff
@@ -0,0 +1,38 @@
...
+  'node_modules/unimodules',
+  'node_modules/@react',
+  'node_modules/@expo',
+  'node_modules/@unimodules',
+];
+
+module.exports = function(babelRoot, { pathsToInclude = [], ...options } = {}) {
+  const modules = [...includeModulesThatContainPaths, ...pathsToInclude];
+  return {
+    test: /\.jsx?$/,
```
sstur: It appears from this that we are not including/processing TypeScript files. Is that correct, or am I missing something?

PR #429, opened 2019-03-17
RFC: Support per project account via .expo-env.json

mfazekas:
See https://expo.canny.io/feature-requests/p/per-project-expo-account

You can use separate accounts used to publish different projects by placing a `.expo-env.json` file with a "env" value. Projects with different "env" value uses different credentials. With this you can avoid  `expo-cli logout` and `expo-cli login` when working on multiple projects with different expo accounts.

Example .expo-env.json:
```
{
  "env" : "foo
}
```
this results in expo using `~/.expo/foo-state.json` instead of the default `~/.expo/state.json` to store credentials.

This PR is a prototype, no docs, no tests, comments are welcome.

### DIFF
--- a/packages/expo-cli/src/exp.ts
+++ b/packages/expo-cli/src/exp.ts
@@ -369,7 +369,7 @@ function runAsync(programName: string) {
       process.exit(0);
     });
 
-    program.on('command:*', subCommand => {
+    program.on('command:*', (subCommand: string) => {
       log.warn(
         `"${subCommand}" is not an ${programName} command. See "${programName} --help" for the full list of commands.`
       );
--- a/packages/xdl/src/UserSettings.ts
+++ b/packages/xdl/src/UserSettings.ts
@@ -25,9 +25,41 @@ export type UserData = {
 
 const SETTINGS_FILE_NAME = 'state.json';
 
+const ENV_FILE_NAME = '.expo-env.json';
+
+function findExpoEnvFile() {
+  let dir = path.resolve(process.cwd());
+  while (dir != '/') {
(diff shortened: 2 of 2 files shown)


mfazekas:
@fson thanks for the comment, i'll rework my PR.

- I'll read `owner` from app.json
- should the `owner` just affect puslisher?  I assume the current state of PR affect much more - anything that's affected by UserSettings, not just publishes.

fson:
cc @tcdavis who worked on adding the owner field.

brentvatne:
there's info on how the `owner` field impacts the expo-cli commands here: https://expo.io/settings/services-guide

that said, it could still be useful to have multiple accounts -- maybe you need to switch accounts for a project to be on an account where you are part of the owner's team.

i think this is a fairly reasonable thing to add but i'm curious to learn more first. can someone investigate what other popular / comparable cli tools do to handle switching accounts between projects and list some of the behaviors here? i just want to make sure we consider alternative approaches before settling on this one.

tcdavis:
Generally you can think of the owner field as telling expo-cli to run all operations as if you were signed in as the account listed.  Though there's still a few commands around credential management and webhooks that don't support owner yet.

As for other ways to handle it, the only options I'm aware of are:
- having the account to be used specified as project configuration: amplify uses this approach and the owner field discussed here
- using the currently authed user and providing a command to switch: the original approach here or now-cli and netlify-cli

EvanBacon:
Closing due to inactivity, please feel free to reopen in the future. Also I wonder if this functionality can be achieved using the newer `app.config.js` 🤔

PR #434, opened 2019-03-18
Use `react-dev-utils` to bundle Webpack

EvanBacon:
Reusing the webpack formatting in react-dev-utils to cleanup the webpack output.

* Reduced webpack spam
* Fixed order of output logs


packages/xdl/src/Project.js
```diff
@@ -1865,34 +1865,179 @@ function getWebpackInstance() {
...
+import clearConsole from 'react-dev-utils/clearConsole';
+import { prepareUrls } from 'react-dev-utils/WebpackDevServerUtils';
+
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+('use strict');
```
fson: Did these imports, copyright header and `'use strict'` directive in the middle of the file come from some accidental copy paste?

packages/xdl/src/Project.js
```diff
@@ -1865,34 +1865,179 @@ function getWebpackInstance() {
...
+
+  // "invalid" event fires when you have changed a file, and Webpack is
+  // recompiling a bundle. WebpackDevServer takes care to pause serving the
+  // bundle, so if you refresh, it'll wait instead of serving the old one.
+  // "invalid" is short for "bundle invalidated", it doesn't imply any errors.
+  compiler.hooks.invalid.tap('invalid', () => {
+    if (isInteractive) {
+      clearConsole();
+    }
+    console.log(chalk.white('Compiling...'));
```
fson: Can you route these logs through the log functions in `ProjectUtils` (`ProjectUtils.logInfo` etc) so that they work correctly with Expo Dev Tools, debug logging etc.?

packages/xdl/src/Project.js
EvanBacon: oops, I'll move this to another file since it's getting kinda bulky.

packages/xdl/src/createCompiler.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+function log(projectRoot, message) {
+  logToConsole(message);
+  ProjectUtils.logDebug(projectRoot, CONSOLE_TAG, message);
+}
+
+function logWarning(projectRoot, message) {
+  logToConsole(message);
```
fson:
```suggestion
```
This line is not needed. `ProjectUtils.logWarning` will log the message to console as well, so now we're logging everything twice.

packages/xdl/src/createCompiler.js
```diff
@@ -0,0 +1,157 @@
...
+function logToConsole(message) {
+  // Prevent double messaging
+  if (!process.env.DEBUG) {
+    console.log(message);
+  }
+}
+
+function log(projectRoot, message) {
+  logToConsole(message);
+  ProjectUtils.logDebug(projectRoot, CONSOLE_TAG, message);
```
fson:
```suggestion
  ProjectUtils.logInfo(projectRoot, CONSOLE_TAG, message);
```

packages/xdl/src/createCompiler.js
EvanBacon: The format seems to be broken when I use chalk with `ProjectUtils.logWarning` or `ProjectUtils.logInfo`. I can't reset the style after using underline, and newlines seem to be trimmed.

packages/xdl/src/Project.js
```diff
@@ -39,6 +42,7 @@ import * as Android from './Android';
 import Api from './Api';
 import ApiV2 from './ApiV2';
 import Config from './Config';
+import createCompiler from './createCompiler';
```
ide: Maybe call the function & file createWebpackCompiler so it makes more sense in this context where everything else is not necessarily about webpack

EvanBacon:
For context I'm stuck trying to get the two different logging types to match.

## using `console.log`

Screen Shot 2019-04-10 at 9 32 31 PM

## using `ProjectUtils.logInfo`

Screen Shot 2019-04-10 at 9 36 38 PM

packages/xdl/src/createWebpackCompiler.js
```diff
@@ -0,0 +1,168 @@
...
+function printInstructions(projectRoot, appName, urls, useYarn) {
+  const _log = message => log(projectRoot, message);
+
+  _log(` `);
+  _log(`You can now view ${chalk.bold(appName)} in the browser.`);
+  _log(` `);
+  if (urls.lanUrlForTerminal) {
+    _log(`  ${chalk.bold('Local:')}            ${urls.localUrlForTerminal}`);
+    _log(`  ${chalk.bold('On Your Network:')}  ${urls.lanUrlForTerminal}`);
+    _log(` `);
```
fson: These logs look really sparse in Dev Tools because each line is a separate message:
Screenshot 2019-04-11 at 11 37 54

Perhaps you can do something like call `_log( <a long message for the whole thing with newlines in it> )` instead to make it look nicer. Especially logging a `' '` (single space) looks bad in Dev Tools, you can replace that by adding newlines to the message.

fson: Also there are no webpack build progress bars/messages visible in Dev Tools right now.

packages/xdl/src/createWebpackCompiler.js
```diff
@@ -0,0 +1,168 @@
...
+    // New line after the bundle analyzer finishes
+    log(projectRoot, ` `);
+
+    const isSuccessful = !messages.errors.length && !messages.warnings.length;
+    if (isSuccessful) {
+      log(projectRoot, chalk.bold.cyan(`Compiled successfully!`));
+      printPreviewNotice(projectRoot);
+    }
+    if (isSuccessful && (!nonInteractive || isFirstCompile)) {
+      printInstructions(projectRoot, appName, urls, useYarn);
```
fson: Right now this message pops up every time I edit a file. I wonder if that's the best way to do this. Especially since the message is rather long, it's hard to see the "Compiled successfully!" message (the most relevant thing at that moment) from all the other stuff we print:
Screenshot 2019-04-11 at 12 02 34

fson:
Screenshot 2019-04-11 at 12 11 48

Looks like the second message is redundant, when the progress bar is already at 100% and the build time is shown next to it?

packages/xdl/src/createWebpackCompiler.js
```diff
@@ -0,0 +1,168 @@
...
+      // Teach some ESLint tricks.
+      logWarning(
+        projectRoot,
+        '\nSearch for the ' +
+          chalk.underline(chalk.yellow('keywords')) +
+          ' to learn more about each warning.'
+      );
+      logWarning(
+        projectRoot,
+        'To ignore, add ' + chalk.cyan('// eslint-disable-next-line') + ' to the line before.\n'
```
fson: These instructions seem wrong: I don't think we run the code through ESLint, so `// eslint-disable-next-line` should not have any effect. Also not sure what "keywords" refers to.

Screenshot 2019-04-11 at 16 00 13
Got these warnings when running in `native-component-list`.

packages/xdl/src/createWebpackCompiler.js
EvanBacon: I tried to keep this looking like CRA, but may need to change it

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
+    throw new XDLError(ErrorCode.WEB_NOT_CONFIGURED, errors.join('\n'));
+  }
+}
+
+async function promptToInstallReactNativeWeb(dependencies, command) {
+  // clearConsole();
+  const question = {
+    type: 'confirm',
+    name: 'should',
+    message: `It appears you don't have all of the required packages installed. \n  Would you like to install: ${chalk.white.underline(
```
fson: Screenshot 2019-04-12 at 17 58 02

Please avoid using `chalk.white` as it is invisible/illegible in terminals using a light background color.

fson:
```suggestion
    message: `It appears you don't have all of the required packages installed. \n  Would you like to install: ${chalk.underline(
```

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
 
-export async function ensureWebSupportAsync(projectRoot) {
+async function promptToAddWebPlatform() {
+  // clearConsole();
+  const question = {
+    type: 'confirm',
+    name: 'should',
+    message: `It appears you don't explicitly define "${chalk.underline(
+      `web`
+    )}" as one of the supported "${chalk.underline(`platforms`)}" in your project's ${chalk.white(
```
fson:
```suggestion
    )}" as one of the supported "${chalk.underline(`platforms`)}" in your project's ${chalk.bold(
```

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
+  };
+  const answer = await inquirer.prompt(question);
+  return !!answer.should;
 }
 
 function getWebSetupLogs() {
-  const appJsonRules = chalk.white(
-    `
-  ${chalk.whiteBright.bold(`app.json`)}
+  const appJsonRules = chalk.white(`
```
fson: Please remove white color here as well.

packages/xdl/src/createWebpackCompiler.js
```diff
@@ -0,0 +1,166 @@
...
+  // "invalid" event fires when you have changed a file, and Webpack is
+  // recompiling a bundle. WebpackDevServer takes care to pause serving the
+  // bundle, so if you refresh, it'll wait instead of serving the old one.
+  // "invalid" is short for "bundle invalidated", it doesn't imply any errors.
+  compiler.hooks.invalid.tap('invalid', () => {
+    if (!nonInteractive) {
+      clearConsole();
+    }
+    console.log('');
+    log(projectRoot, chalk.white('Compiling...'));
```
fson:
```suggestion
    log(projectRoot, 'Compiling...');
```

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
+  }
+  if (currentPlatforms.includes('web')) {
+    return;
+  }
+
+  await ConfigUtils.writeConfigJsonAsync(projectRoot, { platforms: [...currentPlatforms, 'web'] });
+}
+
+function isUsingYarn(projectRoot) {
+  const yarnLockPath = path.resolve(projectRoot, 'yarn.lock');
```
fson: This false detected `expo/apps/native-component-list` as not using yarn. Can we try something like this?
```js
const workspaceRoot = findWorkspaceRoot(projectRoot);
if (workspaceRoot) {
  return fs.existsSync(path.join(workspaceRoot, 'yarn.lock'));
} else {
  return fs.existsSync(path.join(projectRoot, 'yarn.lock'));
}
```

packages/webpack-config/webpack/webpack.common.js
```diff
@@ -30,40 +44,64 @@ const imageLoaderConfiguration = {
...
+// that fall through the other loaders.
+const fallbackLoaderConfiguration = {
+  loader: require.resolve('file-loader'),
+  // Exclude `js` files to keep "css" loader working as it injects
+  // its runtime that would otherwise be processed through "file" loader.
+  // Also exclude `html` and `json` extensions so they get processed
+  // by webpacks internal loaders.
+
+  // Excludes: js, jsx, ts, tsx, html, json
+  exclude: [/\.[jt]sx?$/, /\.html$/, /\.json$/],
```
fson:
```suggestion
  exclude: [/\.jsx?$/, /\.tsx?$/, /\.html$/, /\.json$/],
```

packages/xdl/src/Web.js
```diff
@@ -70,6 +58,60 @@ export async function hasWebSupportAsync(projectRoot) {
...
+  return `${name}@${version}`;
+}
+
+async function getMissingReactNativeWebPackagesAsync(projectRoot, appPackage) {
+  const { dependencies = {} } = appPackage;
+  const requiredPackages = [
+    { name: 'react', version: '*' },
+    { name: 'react-dom', version: '*' },
+    { name: 'react-native-web', version: '^0.11.2' },
+    { name: 'expo', version: '^33.0.0-alpha.web.1' },
```
fson: Note that this depends on the SDK version used. If write the versions here, we'll have to release a new version of Expo CLI whenever there's a new version that isn't covered by the semver range and users will also have to update to it to use that SDK version.

I think you could add `react-dom` and `react-native-web` to the version endpoint (https://exp.host/--/api/v2/versions), which already includes `react` and `expo` versions and is versioned by SDK version. You can use `Versions.sdkVersionsAsync()` in XDL to fetch them.

fson: Also this code probably should go in the `Doctor` module in XDL, which handles checks like this.

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
+  return path.resolve(workspaceRoot || projectRoot, 'node_modules');
+}
+
+function areModulesInstalled(projectRoot) {
+  const modulesPath = getModulesPath(projectRoot);
+  return fs.existsSync(modulesPath);
+}
+
+async function attemptToGetPackageVersionAtPathAsync(modulesPath, moduleName) {
+  const possibleModulePath = path.resolve(modulesPath, moduleName, 'package.json');
```
fson: You can use ``ConfigUtils.resolveModule(`${moduleName}/package.json`, projectRoot, exp)``, which takes in account custom paths for node_modules (can be overridden in app.json) and resolves the packages using the Node module resolution algorithm (like `require.resolve`) so it should also work in cases like a project that's using Yarn workspaces etc.

packages/xdl/src/Web.js
```diff
@@ -79,39 +121,158 @@ export async function onlySupportsWebAsync(projectRoot) {
...
+  const question = {
+    type: 'confirm',
+    name: 'should',
+    message: `It appears you don't have all of the required packages installed. \n  Would you like to install: ${chalk.white.underline(
+      dependencies.join(', ')
+    )}?`,
+    default: true,
+  };
+  const answer = await inquirer.prompt(question);
+  return !!answer.should;
```
fson: `should` should be a boolean here, no need to coerce it?

packages/xdl/src/createWebpackCompiler.js
```diff
@@ -0,0 +1,166 @@
...
+function logError(projectRoot, message) {
+  ProjectUtils.logError(projectRoot, CONSOLE_TAG, message);
+}
+
+function printInstructions(projectRoot, appName, urls, showInDevtools) {
+  const _log = message => log(projectRoot, message, showInDevtools);
+
+  console.log('');
+  _log(`You can now view ${chalk.bold(appName)} in the browser.`);
+  console.log('');
```
fson: I think these console.logs should be removed. Add newlines inside the `_log` message as necessary.

packages/xdl/src/logs/PackagerLogsStream.js
```diff
@@ -177,14 +177,21 @@ export default class PackagerLogsStream {
...
-          if (chunk.tag !== 'metro' && chunk.tag !== 'expo') {
-            return;
-          } else if (this._getCurrentOpenProjectId() !== projectId) {
+          if (this._getCurrentOpenProjectId() !== projectId) {
             // TODO: We should be confident that we are properly unsubscribing
             // from the stream rather than doing a defensive check like this.
             return;
           }
 
+          if (chunk.tag === 'webpack') {
```
fson: This change can be removed after rebasing the branch, because we no longer call `_formatMsg` on the logs.

packages/xdl/src/Web.js
EvanBacon: Can I use `ConfigUtils.resolveModule(``, projectRoot, exp);` to determine if the `node_modules` have been installed?

packages/xdl/src/createWebpackCompiler.js
EvanBacon: but this will add new lines to the devtools. Doing it this way only adds newlines to the terminal

packages/xdl/src/logs/PackagerLogsStream.js
EvanBacon: This just makes nothing log out

packages/xdl/src/createWebpackCompiler.js
fson: My point is that everything should go in a single log message with some line breaks in it. Only then it will look the same in Dev Tools and the terminal.

You can do it like this:
```js
const message = `You can now view ${chalk.bold(appName)} in the browser.\n\n`;
if (urls.lanUrlForTerminal) {
  message += `  ${chalk.bold('Local:')}            ${urls.localUrlForTerminal}\n`;
  message += `  ${chalk.bold('On Your Network:')}  ${urls.lanUrlForTerminal}\n`
} else {
  message += `  ${urls.localUrlForTerminal}\n`;
}
```
Then you also won't need to log those empty strings to the console.

PR #436, opened 2019-03-19
PWA Manifest

EvanBacon:
# Why

This will enable webpack to generate a `manifest.json` using the `app.json` from either `expo-cli` or `react-native-cli`. 

# How

* Added `@expo/webpack-pwa-manifest-plugin` (forked from [webpack-pwa-manifest](https://www.npmjs.com/package/webpack-pwa-manifest))
  * Originally this was just to add webpack 4 support, but ended up moving all of the `app.json` parsing and validation into it as well.
* Deleted the default icons and `manifest.json` in favor of `app.json` values.
* Added splash screen generation.
  * Uses `splash.image` with fallback to `icon` (parity with expo native apps)
  * `splash.backgroundColor` Defines the splash screen background color
  * `supportsTablet: true`: Adds tablet images
  * `orientation`: Will determine whether or not to generate landscape / portrait screens.
* Added iOS icon generation.
  * Uses `ios.icon` with fallback to `icon`.
  * TODO: Prevent generation of certain icons when `supportsTablet: false`
* Uses `icon` on Android / other platforms
* Added noScript
* Unified language
* Added better support for react-native-cli projects in `@expo/webpack-config`
* Added open graph, twitter, msapp meta tags.
* Fixed viewport on iPhoneX
* Added `X-UA-Compatible`
* Added minifyHTML option to `app.json`


packages/webpack-pwa-manifest-plugin/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "application"
+  ],
+  "scripts": {
+    "start": "yarn run build && yarn run watch",
+    "build": "gulp clean && gulp build",
+    "watch": "gulp watch",
+    "prepare": "yarn build",
+    "pkg": "pkg ."
+  },
+  "author": "Arthur A. Bergamaschi <arthurbergmz@gmail.com>",
```
fson: I think this field can be removed?

packages/webpack-pwa-manifest-plugin/yarn.lock
```diff
@@ -0,0 +1,4969 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
fson: I think because we use workspaces, there should only be one yarn.lock file, at the root of the repo. Can you remove this an re-run `yarn`?

packages/webpack-config/webpack/webpack.common.js
```diff
@@ -43,14 +46,28 @@ const mediaLoaderConfiguration = {
...
   const babelConfig = createBabelConfig(locations.root);
-  const indexHTML = createIndexHTMLFromAppJSON(locations);
-  const clientEnv = createClientEnvironment(locations);
 
-  const nativeAppManifest = require(locations.appJson);
-  const appManifest = nativeAppManifest.expo || nativeAppManifest;
+  const appJSON = require(locations.appJson);
+  if (!appJSON) {
+    throw new Error('app.json could not be found at: ' + locations.appJson);
+  }
```
fson: Expo CLI reads app.json configs using this function: https://github.com/expo/expo-cli/blob/a9ce6ead81f850c9bc5c421fe7a41ac30d9224f2/packages/xdl/src/project/ProjectUtils.js#L212 It handles things like showing errors with code snippets if there's a syntax error in the JSON file (instead of the unhelpful `SyntaxError: app.json: Unexpected token in JSON at position 113` from `require()`). It also sets a couple of default values and supports JSON5 as an alternative.

I think it would make sense to use the same code to load `app.json` configuration everywhere. We can extract the config reading code from `Expo.ProjectUtils` to a small package, e.g. `@expo/config` so the webpack doesn't need to have a (circular) dependency on XDL.

EvanBacon: https://github.com/expo/expo-cli/pull/452

packages/webpack-pwa-manifest-plugin/src/createMetatagsFromConfig.js
```diff
@@ -1,5 +1,50 @@
...
+      return pascalToKebab(input);
+    case '_':
+      return pascalToSnake(input);
+  }
+  return input;
+}
+
+function pascalToSnake(input) {
+  return input
+    .replace(/(?:^|\.?)([A-Z])/g, function(x, y) {
```
ide: pick better variables than x y

PR #437, opened 2019-03-19
Move root `webpack.config.js` check to XDL

EvanBacon:
Now you can create a `webpack.config.js` in your root project which reexports `@expo/webpack-config` without creating a stack overflow.

### DIFF
--- a/packages/webpack-config/webpack.config.js
+++ b/packages/webpack-config/webpack.config.js
@@ -1,7 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const developmentConfig = require('./webpack/webpack.dev');
-const productionConfig = require('./webpack/webpack.prod');
 
 function invokePossibleFunction(objectOrMethod, ...args) {
   if (typeof objectOrMethod === 'function') {
--- a/packages/webpack-config/webpack.config.js
+++ b/packages/webpack-config/webpack.config.js
@@ -1,29 +1,10 @@
-const fs = require('fs');
-const path = require('path');
-
-function invokePossibleFunction(objectOrMethod, ...args) {
-  if (typeof objectOrMethod === 'function') {
-    return objectOrMethod(...args);
-  } else {
-    return objectOrMethod;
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Web.js
```diff
@@ -1,9 +1,31 @@
...
+  if (typeof objectOrMethod === 'function') {
+    return objectOrMethod(...args);
+  } else {
+    return objectOrMethod;
+  }
+}
+
+export function invokeWebpackConfig(env, argv) {
+  // Check if the project has a webpack.config.js in the root.
+  const projectWebpackConfig = path.resolve(process.cwd(), 'webpack.config.js');
```
fson: This should look for the config in the project root folder, `env.projectRoot`, not current working folder. You can run `expo start <full-path-to-project-root.` from anywhere and it should work.

packages/webpack-config/webpack.config.js
```diff
@@ -11,23 +9,16 @@ function invokePossibleFunction(objectOrMethod, ...args) {
   }
 }
 
-module.exports = function(env = {}, argv) {
-  // Check if the project has a webpack.config.js in the root.
-  const projectWebpackConfig = path.resolve(process.cwd(), 'webpack.config.js');
-  if (fs.existsSync(projectWebpackConfig)) {
-    const webpackConfig = require(projectWebpackConfig);
-    return invokePossibleFunction(webpackConfig, env, argv);
+function getConfigFromRootIfPossible(filename) {
```
fson: Do we still need this here? Since you can have `webpack.config.js` people can structure their config anyway they want, if they override it. That would keep the `webpack.dev.js` and `webpack.prod.js` filenames as an implementation detail too, so we can also change the structure internally, if needed.

EvanBacon: this makes sense. I think originally the idea was to make it easier to edit the prod output, but I think the logic is getting somewhat crippling.

PR #452, opened 2019-03-21
@expo/config

EvanBacon:
Created a module for interacting with the expo config (`app.json` )


packages/config/.babelrc
```diff
@@ -0,0 +1,16 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "node": "7"
```
fson: Right now we target Node v6.0.0 and up: https://github.com/expo/expo-cli/blob/7da18891ed56a2c2b323ec0a7ea72d96bf5b7e9f/packages/expo-cli/bin/expo.js#L10-L22

However, 6.x LTS end of life is on April 30th, so we could then drop it and support 8.x and up.

packages/config/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
fson: You can remove this file and the package will use the default `.eslintrc.js` from the root folder.

packages/config/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 650 Industries
```
fson:
```suggestion
Copyright (c) 2015-present 650 Industries, Inc. (aka Expo)
```

packages/config/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "babel-preset-stage-1": "^6.24.1",
+    "concurrently": "^3.5.0",
+    "eslint": "^4.7.2",
+    "eslint-config-universe": "^1.0.7",
+    "flow-bin": "^0.55.0",
+    "flow-copy-source": "^1.2.1",
+    "jest": "^23.4.1",
+    "mkdirp": "^0.5.1",
+    "prettier": "^1.7.4",
+    "rimraf": "^2.6.2"
```
fson: Looks like most of these `devDependencies` are not used and should be removed.

packages/expo-cli/src/exp.js
```diff
@@ -30,6 +30,7 @@ import {
   ProjectUtils,
   User as UserManager,
 } from 'xdl';
+import * as ConfigUtils from '@epxo/config';
```
fson:
```suggestion
import * as ConfigUtils from '@expo/config';
```

packages/xdl/src/project/ProjectUtils.js
```diff
@@ -126,203 +118,64 @@ export function attachLoggerStream(projectRoot: string, stream: any) {
...
-  "expo": {
-    (JSON contents from exp.json)
-  }
-}`
-    );
+  try {
+    return await ConfigUtils.findConfigFileAsync(projectRoot);
+  } catch (error) {
+    if (error instanceof ConfigUtils.ConfigError) {
+      logWarning(projectRoot, 'expo', error.message);
```
fson: Not sure if this error handling works as intended: when you catch the error and log a warning, this function returns `undefined` and then the code crashes soon after with `Cannot read property 'platforms' of null`. (When running `expo start` in a project with `exp.json` and no `app.json`)

However, `app.json` has been the default for more than two years now and now we've also had this deprecation warning live for almost two months, so I'd say let's drop the `catch` block here and let it be a hard error if you're still using `exp.json`.

packages/xdl/src/project/ProjectUtils.js
```diff
@@ -122,203 +118,60 @@ export function attachLoggerStream(projectRoot: string, stream: any) {
...
-Example app.json:
-{
-  "expo": {
-    (JSON contents from exp.json)
-  }
-}`
-    );
+  try {
+    return await ConfigUtils.findConfigFileAsync(projectRoot);
+  } catch (error) {
```
fson: If you’re only going to rethrow, just remove the try-catch altogether?

PR #466, opened 2019-03-25
Remove legacy config functions from xdl

fson:
These functions have moved to @expo/config


fson:
@EvanBacon could you sanity check?

PR #472, opened 2019-03-27
Add `expo install` command

fson:
`expo install` makes it easier to install compatible unimodule packages in managed Expo projects by automatically choosing the correct version to use based on the Expo SDK version.

* For unimodules with native code, `expo install` always installs the version the `expo` package depends on (and whose native code is included in the Expo Client). For JS-only unimodules and other npm packages it passes the request to npm/Yarn without any version clobbering. 🤗 
  Screenshot 2019-03-27 at 22 44 36
* It uses npm if `package-lock.json` exists, Yarn if `yarn.lock` exists. (Can be overridden with `--npm`/`--yarn` flags.) 👇 
  Screenshot 2019-03-27 at 22 51 34
* It works even if `node_modules` has been deleted. ☺️ 
  Screenshot 2019-03-27 at 22 45 24
* It refuses to install unimodules that are not compatible with the current SDK version. 🙀 
  Screenshot 2019-03-27 at 22 45 59
* **BONUS** Did something seem unusual in the screenshots above? Oh yeah, no peer dependency warnings. `expo install` filters those pesky things out from both npm and Yarn output! 🙌 

*Note*: the old `expo install:android` and `expo install:ios` have nothing to with this. I have deprecated them and moved the functionality they provide to `expo client:install:android` and `expo client:install:ios`, respectively. (I know, these commands have so many colons in them. But they're rarely used and I'm not going to change the whole naming scheme just yet. But I will.)


packages/expo-cli/src/commands/install.js
```diff
@@ -1,28 +1,129 @@
...
+async function listBundledNativeModulesAsync(projectRoot, exp) {
+  const packageJsonPath = ConfigUtils.resolveModule('expo/package.json', projectRoot, exp);
+  if (!packageJsonPath) {
+    throw new CommandError(
+      'NO_EXPO',
+      "Package 'expo' not found in node_modules. Please make sure this is a managed Expo project."
+    );
+  }
+  const packageJson = await JsonFile.readAsync(packageJsonPath);
+  const sdkVersion = packageJson.version;
```
brentvatne: is this reliable? version in package.json could be 32.0.3 for example. why not use app.json?

packages/expo-cli/src/commands/install.js
```diff
@@ -1,28 +1,129 @@
...
+    const spec = npmPackageArg(arg);
+    const { name } = spec;
+    if (
+      ['tag', 'version', 'range'].includes(spec.type) &&
+      name &&
+      (compatible[name] || incompatible[name])
+    ) {
+      if (compatible[name]) {
+        // Unimodule packages from npm registry are modified to use the bundled version.
+        const version = dependencies[name];
```
brentvatne: i'm probably reading this wrong, it looks to me like we're depending on the unimodule being in the dependencies of the local copy of the expo package in order to determine the version that should be used. won't they all be removed from dependencies on the expo package?

packages/expo-cli/src/commands/install.js
fson: The compatibility data only deals in major versions and we're using semver, so I think this should be reliable. See:
https://github.com/expo/expo-cli/blob/f96bd374d5415f192dfd0a7ee4bd7958954101a2/packages/xdl/src/modules/config.js#L44-L55

I went back and forth between using `app.json` version and using `package.json` version, but I think if use the word "SDK version" to mean the native version and not the `expo` SDK JS version (and I think we do), then reading the version from `app.json` is more correct. I'll change the code back to use `app.json`.

fson:
I've added support for vendored modules: Screenshot 2019-03-28 at 14 02 45

packages/xdl/src/modules/config.js
```diff
@@ -551,6 +551,20 @@ const universalModules = [
...
+  { libName: '@react-native-community/netinfo', sdkVersions: '>=33.0.0', isNativeModule: true },
+  { libName: 'lottie-react-native', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-branch', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-gesture-handler', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-maps', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-reanimated', sdkVersions: '>=28.0.0', isNativeModule: true },
+  { libName: 'react-native-screens', sdkVersions: '>=30.0.0', isNativeModule: true },
+  { libName: 'react-native-svg', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-view-shot', sdkVersions: '>=26.0.0', isNativeModule: true },
+  { libName: 'react-native-webview', sdkVersions: '>=33.0.0', isNativeModule: true },
```
fson: @brentvatne could you verify that this list is correct? I compiled it manually by comparing the dependencies of the `expo` package and the gulp tasks like [this one](https://github.com/expo/expo/blob/5f93ff3156234e0bf8010abf42256cdd581e9796/tools/gulpfile.js#L121-L134) in the expo/expo repo.

There are `updateVendoredNativeModule` tasks for `amazon-cognito-identity-js`, `react-tipsi-stripe`, `react-native-admob` and `react-native-fbads` too, but I didn't find them in the SDK reference and couldn't figure out if they should be installable in managed apps or not.

brentvatne: good point - i think no need to add those ones.

ultimately i don't think this should live in xdl because it requires that we publish a new version of xdl and expo-cli any time we want to change a version. it'd probably be best if it lived at some remote endpoint, like native.directory or some endpoint on www.

fson: > ultimately i don't think this should live in xdl because it requires that we publish a new version of xdl and expo-cli any time we want to change a version.
I 100% agree this data shouldn't live in XDL, I'm only adding it here to support older SDK versions which won't have this data available in the `expo` package.

My idea is:
* For SDK <33, use the existing list of Unimodules in XDL combined with version numbers from `expo/package.json`.
* For SDK >=33, use a single file in the `expo` package that includes names of the supported packages and their versions. It's a finite set, because the only packages that are included are those whose native code is baked in the SDK and Expo Client.

I'd like to see how far we can get with just a flat file alongside the code that's distributed for each SDK version, without relying on a special endpoint that we'd host for this functionality. I think this has a few nice properties:
- Versioning is simple. The data is versioned in git together with the SDK that it describes and it's easy to find the history of changes and compare versions. 
- Making changes to it is not different from contributing any other changes to the SDK. You can commit / send a PR. Anyone who has permissions to publish `expo` to npm can make a release without a need to have separate write permissions to our internal API (or S3) or deployment permissions to www.
- The data is available offline in managed Expo projects, without any special caching considerations on our part. (Together with npm/Yarn cache it's possible to install packages you've previously downloaded without having an internet connection.)
- To use this feature it's inevitable that you must be able to connect to the npm registry. Some people need to use a proxy for this. Not relying on an additional web service (when it's not necessary) makes it available to more people.

brentvatne: ok that makes sense, let's see how far we can go with this approach then!

packages/expo-cli/src/PackageManager.js
```diff
@@ -0,0 +1,97 @@
...
+    return 'Yarn';
+  }
+  async installAsync() {
+    await this._runAsync(['install']);
+  }
+  async addAsync(...names) {
+    await this._runAsync(['add', ...names]);
+  }
+
+  // Private
```
alvaropaco: could remove this comment.

packages/expo-cli/src/commands/install.js
```diff
@@ -1,28 +1,125 @@
...
+  const versionedPackages = packages.map(arg => {
+    const spec = npmPackageArg(arg);
+    const { name } = spec;
+    if (
+      ['tag', 'version', 'range'].includes(spec.type) &&
+      name &&
+      (compatible[name] || incompatible[name])
+    ) {
+      if (compatible[name]) {
+        // Unimodule packages from npm registry are modified to use the bundled version.
```
alvaropaco: Could remove this comment.

packages/expo-cli/src/commands/install.js
```diff
@@ -1,28 +1,125 @@
...
+        throw new CommandError(
+          'INCOMPATIBLE_NATIVE_MODULE',
+          `'${name}' is incompatible with Expo SDK ${exp.sdkVersion}.\nSupported SDK versions: ${
+            incompatible[name].sdkVersions
+          }`
+        );
       }
-    }, true);
+    } else {
+      // Other packages are passed through unmodified.
```
alvaropaco: Could remove this comment.

PR #475, opened 2019-03-30
[expo-cli] update client:ios copy to be more verbose

esamelson:
Feedback from https://exponent-internal.slack.com/archives/CFSP4GXL6/p1553565327070500, added somewhat more verbose copy to this command to explain the steps of this flow in a bit more detail to users. Also noticed this command didn't have a description, so I added one.

### DIFF
--- a/packages/expo-cli/src/commands/client/index.js
+++ b/packages/expo-cli/src/commands/client/index.js
@@ -89,9 +89,6 @@ export default program => {
         log(chalk.green(`${result.registrationUrl}`));
         log.newLine();
         log(
-          "Nothing will be saved on your device, but this step is necessary to get your device's unique identifier (UDID)."
-        );
-        log(
           "After you register your device, we'll start building your client, and you'll receive an email when it's ready to install."
         );
       } else {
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/client/index.js
```diff
@@ -77,16 +83,25 @@ export default program => {
...
-              result.registrationUrl
-            }`
-          )
+          'Open the following link on your iOS device (or scan the QR code) and follow the instructions to install the development profile:'
         );
         log.newLine();
+        log(chalk.green(`${result.registrationUrl}`));
+        log.newLine();
+        log(
+          "Nothing will be saved on your device, but this step is necessary to get your device's unique identifier (UDID)."
```
fson: This is the only message I’m not sure about. Why would saving something to the device be a concern / something we need to assure is not going to happen?

esamelson: sorry for letting this rot 😞 I think you're right, this message isn't necessary. I've gone ahead and removed it.

fson:
Thank you!

PR #487, opened 2019-04-03
add check for websupport

### LINKED ISSUE #460: 'Webpack is not running' Issue on expo-cli
## Description

I get 'webpack is not running' message in red color in terminal when I run expo build  commands. There may be other commands which I have not yet found.

## Expected Behavior

Previous versions, as far I recall, did not show these messages. I started getting them after i updated to latest version. 

## Observed Behavior
No error though, just that red message warning. 

Also specify:

Using Expo-cli latest version at the moment.

## Reproducible Demo

When I run commands such as expo build:ios or expo build:android.
Or after the build is finished and I press ctrl + c, webpack is not running message appears.

cruzach:
This closes #460.

Problem is that `stopWebpackServerAsync` gets called (whether or not the webpack server is running) when `stopinternalasync` is called, and results in the logged message shown in the issue (through calling `getWebpackInstance`). I thought the simplest fix would be to check for web support before that call.
This way-> `getWebpackInstance` [here](https://github.com/expo/expo-cli/blob/229fc54e20a812aaa0b0422b035dc49ec56695bf/packages/xdl/src/Project.js#L1832 ) still works as originally intended if it is used somewhere else.
This all ties up with checking for websupport before the `startWebpackServerAsync` call because it results in more consistency and clarity (i.e: why should startwebpack function run and NOT start the webpack?)

Let me know if there's something more for me to add/do here. I know the docs ask for a test plan but I'm not 100% clear on how exactly to do that.

### DIFF
--- a/packages/xdl/src/Project.js
+++ b/packages/xdl/src/Project.js
@@ -1841,10 +1841,6 @@ async function startWebpackServerAsync(projectRoot, options, verbose) {
     ProjectUtils.logError(projectRoot, 'expo', 'Webpack is already running.');
     return;
   }
-  const hasWebSupport = await Web.hasWebSupportAsync(projectRoot);
-  if (!hasWebSupport) {
-    return;
-  }
   let { dev, https } = await ProjectSettings.readAsync(projectRoot);
   let config = Web.invokeWebpackConfig({ projectRoot, development: dev, production: !dev, https });
   let webpackServerPort = await _getFreePortAsync(19000);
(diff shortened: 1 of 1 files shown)


fkanout:
Is it fixed? I'm still getting the error message `Webpack is not running.` 😞 
expo version `2.14.0`

cruzach:
It's merged, but a publish hasn't been done yet to the `npm` package. Once that's done and you update your version of the `expo-cli`, you shouldn't get that message anymore 👍 

-Charlie

fkanout:
@cruzach, thanks for your reply.
Just to be sure, this fix is not only about hiding this message, right?
because the real problem is when I publish the App, I can't get it from my Expo client, getting the error message `there was a problem running the requested app`.

cruzach:
@fkanout, no this fix is **only** about not showing this error message when you are not building for Web anyways.

Your issue must be different. If you need help or have questions, I recommend taking a look at [the Expo Forums](https://forums.expo.io/) and seeing if someone else has had a similar problem and how they solved it.
If you think what you're experiencing is a bug that needs fixing on our end, then I suggest making an issue here on Github

Best of luck 👍 
-Charlie

PR #497, opened 2019-04-09
[web] Add webpack support for TypeScript files

sstur:
This allows webpack to load (and babel to process) *.ts and *.tsx files.

This brings the Expo Web behavior inline with the normal Expo behavior.


EvanBacon:
thanks for this!

PR #499, opened 2019-04-09
[web] Removed QR when metro isn't fired up

EvanBacon:
Trying to create a very simple path to starting web so things like code sandbox can use it easily.

### DIFF
--- a/packages/expo-cli/src/commands/start.js
+++ b/packages/expo-cli/src/commands/start.js
@@ -83,7 +83,7 @@ async function action(projectDir, options) {
       log.newLine();
       urlOpts.printQRCode(url);
     }
-    log(`Your app is running at ${chalk.underline(url)}`);
+    log(`Your native app is running at ${chalk.underline(url)}`);
   }
 
   log.nested(chalk.green('Logs for your project will appear below. Press Ctrl+C to exit.'));
--- a/packages/expo-cli/src/commands/start.js
+++ b/packages/expo-cli/src/commands/start.js
@@ -65,6 +65,7 @@ async function action(projectDir, options) {
   const startOpts = await parseStartOptionsAsync(projectDir, options);
 
   await Project.startAsync(rootPath, startOpts);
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/start.js
```diff
@@ -72,11 +78,13 @@ async function action(projectDir, options) {
   if (!nonInteractive && !exp.isDetachexped) {
     await TerminalUI.startAsync(projectDir, startOpts);
   } else {
```
ide:
```suggestion
  } else if (!options.webOnly) {
```

packages/expo-cli/src/commands/start.js
```diff
@@ -72,11 +78,13 @@ async function action(projectDir, options) {
...
-    if (!exp.isDetached) {
-      log.newLine();
-      urlOpts.printQRCode(url);
+    if (!options.webOnly) {
```
ide: When web is enabled, do we print the URL of the site?

EvanBacon: we open the devtools with the link to open it. May be nice to have both though.

ide: I think it'd be useful to show the URL in the CLI too for people who don't have the web UI of Expo CLI enabled (they just use CLI) and also so we clarify which URLs are for which platforms, now that we have multiple URLs.

packages/expo-cli/src/commands/start.js
```diff
@@ -71,7 +78,7 @@ async function action(projectDir, options) {
 
   if (!nonInteractive && !exp.isDetachexped) {
     await TerminalUI.startAsync(projectDir, startOpts);
-  } else {
+  } else if (!options.webOnly) {
```
ide: Maybe change:
```
log(`Your app is running at ${chalk.underline(url)}`)
```
to 
```
log(`Your native app is running at ${chalk.underline(url)}`);
```

PR #504, opened 2019-04-11
Select and send a push key in the adhoc build request

fson:
We previously didn't send the push key because it was not handled by the server end point in the beginning.


packages/expo-cli/src/commands/client/index.js
```diff
@@ -32,7 +33,7 @@ export default program => {
 
       const distributionCert = await selectDistributionCert(context);
       // TODO(fson): Select and save a push key.
```
ide: remove todo?

fson:
```suggestion
```

PR #505, opened 2019-04-11
Fix 'Opening iOS Sim' indefinite hang

cruzach:
Problem booting simulator device if simulator is open but no devices booted.
To repro, have the simulator open, then close the simulator window (⌘W). The simulator doesn't quit, but has no windows. Then try to open an app on iOS simulator with Expo CLI: it hangs indefinitely, waiting for simulator to start.

Device ID is for iPhoneX, wasn't sure if that's an issue. There didn't seem to be any `simctl` args for default sim. If that poses a problem, might be easier to just call `_quitSimulatorAsync` and return false.

### DIFF
--- a/packages/xdl/src/Simulator.js
+++ b/packages/xdl/src/Simulator.js
@@ -180,7 +180,7 @@ async function _getDefaultSimulatorDeviceUDIDAsync() {
     const { stdout: defaultDeviceUDID } = await spawnAsync('defaults', ['read', 'com.apple.iphonesimulator', 'CurrentDeviceUDID']);
     return defaultDeviceUDID.trim();
   } catch (e) {
-    return false;
+    return null;
   }
 }
 
--- a/packages/xdl/src/Simulator.js
+++ b/packages/xdl/src/Simulator.js
@@ -153,6 +153,17 @@ export async function _isSimulatorRunningAsync() {
   }
 
   let bootedDevice = await _bootedSimulatorDeviceAsync();
+  
+  if (!bootedDevice) {
+    Logger.global.info(`Booting device in iOS simulator...`);
+    try {
+      return await _xcrunAsync(['simctl', 'boot', '5E560ABA-4019-4A8E-A1BB-F01C15533F0F']);
+    } catch (e) {
+      Logger.global.error(`There was a problem booting a device in iOS Simulator. Quit Simulator, and try again.`);
+      throw e;
+    }
+  }
+  
   return !!bootedDevice;
 }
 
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Simulator.js
```diff
@@ -153,6 +153,17 @@ export async function _isSimulatorRunningAsync() {
   }
 
   let bootedDevice = await _bootedSimulatorDeviceAsync();
+  
+  if (!bootedDevice) {
+    Logger.global.info(`Booting device in iOS simulator...`);
+    try {
+      return await _xcrunAsync(['simctl', 'boot', '5E560ABA-4019-4A8E-A1BB-F01C15533F0F']);
```
fson: Unfortunately this doesn't work for everyone, because the UDIDs of the devices aren't the same. For example, on my computer the UDID of the iPhone X simulator is `4BDC42E4-...` and there are no simulator devices with your UDID.

Looks like you can get the active / previously used device with this command:
```
defaults read com.apple.iphonesimulator CurrentDeviceUDID
```
However, I don't know if this will always be set. (What if simulator was never used before? etc). So perhaps have a fallback in case it's not set, e.g. find an iPhone that has `isAvailable: true` from the output of `xcrun simctl list --json`.

packages/xdl/src/Simulator.js
```diff
@@ -153,9 +153,37 @@ export async function _isSimulatorRunningAsync() {
...
+  } catch (e) {
+    Logger.global.error(`There was a problem booting a device in iOS Simulator. Quit Simulator, and try again.`);
+    throw e;
+  }
+}
+
+async function _getDefaultSimulatorDeviceUDIDAsync() {
+  try {
+    const { stdout: defaultDeviceUDID } = await spawnAsync('defaults', ['read', 'com.apple.iphonesimulator', 'CurrentDeviceUDID']);
+    return defaultDeviceUDID.replace(/\n$/, '');
```
fson:
```suggestion
    return defaultDeviceUDID.trim();
```

packages/xdl/src/Simulator.js
```diff
@@ -153,9 +153,37 @@ export async function _isSimulatorRunningAsync() {
...
+  } catch (e) {
+    return false;
```
fson:
```suggestion
    return null;
```

packages/xdl/src/Simulator.js
```diff
@@ -165,6 +193,21 @@ async function _waitForSimulatorRunningAsync() {
...
+  for (let runtime in simulatorDeviceInfo.devices) {
+    let devices = simulatorDeviceInfo.devices[runtime];
+    for (let i = 0; i < devices.length; i++) {
+      let device = devices[i];
+      if (device.isAvailable && device.name.includes('iPhone')) {
+        return device;
+      }
+    }
+  }
+  console.warn('No iPhone devices available in Simulator.')
```
fson:
```suggestion
  Logger.global.warn('No iPhone devices available in Simulator.')
```
(All logging in XDL goes through `Logger`. This also makes it work in Expo Dev Tools and not only in terminal.)

packages/xdl/src/Simulator.js
fson: This is to make the type of this function simpler (nullable `string` vs `string | boolean`).

PR #507, opened 2019-04-12
Add check and warning for nextDeprecatedVersion

AdamJNavarro:
Fetches the lowest supported version and if the user is building with said version, it will warn them to upgrade.


packages/xdl/src/Versions.js
```diff
@@ -83,6 +83,16 @@ export async function newestSdkVersionAsync() {
   return result;
 }
 
+export async function nextDeprecatedSdkVersionAsync() {
+  const sdkVersions = await sdkVersionsAsync();
+  const supportedVersions =  _.pickBy(sdkVersions, v => v.isDeprecated === undefined);
```
fson:
```suggestion
  const supportedVersions =  _.pickBy(sdkVersions, v => !v.isDeprecated);
```
(Should work also if `isDeprecated` is `false`.)

packages/expo-cli/src/commands/build/BaseBuilder.js
```diff
@@ -89,6 +89,13 @@ export default class BaseBuilder {
       log.error(`'expo build:${this.platform()}' is not supported for detached projects.`);
       process.exit(1);
     }
+    
+    // Warn user if building a project using the next deprecated SDK version
+    let nextDeprecatedVersion = await Versions.nextDeprecatedSdkVersionAsync();
+    if (this.manifest.sdkVersion.substring(0, 2) === nextDeprecatedVersion.toString()) {
```
fson: Please use `server.major(this.manifest.sdkVersion)` to parse the major version (it doesn't assume the major version only has 2 digits).

packages/xdl/src/Versions.js
```diff
@@ -83,6 +83,16 @@ export async function newestSdkVersionAsync() {
   return result;
 }
 
+export async function nextDeprecatedSdkVersionAsync() {
```
fson: This function doesn't return the full semver version, but the major version part of it. Maybe make it return the whole semver version (preferred) or change the function name to mention it returns a major version (e.g. `oldestSupportedMajorVersionAsync`).

packages/expo-cli/src/commands/build/BaseBuilder.js
```diff
@@ -89,6 +89,13 @@ export default class BaseBuilder {
...
+      let {version} = await Versions.newestSdkVersionAsync();
+      log.warn(`\nSDK${nextDeprecatedVersion} will be DEPRECATED soon! We recommend upgrading versions, ideally to the latest (SDK${version.substring(0, 2)}), so you can continue to build new binaries of your app and develop in the Expo Client.\n`);
```
fson: Maybe use `${chalk.bold('deprecated')}` instead of all caps (screaming)?

packages/xdl/src/Versions.js
fson: You can use `semver.compare` as a compare function passed to `Array.prototype.sort` to sort by semver order.

PR #516, opened 2019-04-14
Parallelizing PWA icon generation

EvanBacon:
* Removed iOS splash screen prop from web PWA splash screen since the backgroundColor is used across both iOS and Android
* Changed default `apple-touch-icon` from 1024 to 180
* Removed all icon sizes `96, 128, 192, 256, 384, 512` in favor of browser optimized `192, 512`
* Removed unused template `apple-touch-icon` & `apple-touch-startup-image ` in favor of manifest plugin generated copies
* Fixed splash screen resize mode
* Fixed splash screen contain mode
* Parallelizing all icon generation for faster results
* Fixed splash screens meta tags not being generated

### DIFF
--- a/packages/config/src/Config.js
+++ b/packages/config/src/Config.js
@@ -443,7 +443,7 @@ function inferWebHomescreenIcons(config: Object = {}, getAbsolutePath: Function,
 }
 
 function inferWebStartupImages(config: Object = {}, getAbsolutePath: Function, options: Object) {
-  const { icon, web = {}, ios = {}, splash = {}, primaryColor } = config;
+  const { icon, web = {}, splash = {}, primaryColor } = config;
   if (Array.isArray(web.startupImages)) {
     return web.startupImages;
   }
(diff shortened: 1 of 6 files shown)


packages/config/src/Config.js
```diff
@@ -448,8 +450,18 @@ function inferWebStartupImages(config: Object = {}, getAbsolutePath: Function, o
 
   let splashImageSource;
   if (webSplash.image || iOSSplash.image || splash.image) {
+    const resizeMode =
+      webSplash.resizeMode || iOSSplash.resizeMode || splash.resizeMode || 'contain';
+    const backgroundColor =
+      webSplash.backgroundColor ||
+      iOSSplash.backgroundColor ||
+      splash.backgroundColor ||
```
fson: Is this logic correct? Given this configuration:
{
  splash: {
    backgroundColor: 'black'
  },
  ios: {
    splash: {
      backgroundColor: 'white'
    }
  }
}
```
I'd expect web to use the default `splash` configuration because platform specific `web` configuration is not specified. However, it seems it will use the iOS specific configuration instead.

I think this code should not use the `ios` splash configuration at all.

EvanBacon: I read an outdated document about iOS being the only platform to use the background_color value. Looks like Google does too: https://developers.google.com/web/tools/lighthouse/audits/custom-splash-screen

PR #520, opened 2019-04-16
Add assetPlugins arg to React Native server start for SDK33+

brentvatne:
Start the React Native server with the assetPlugins CLI flag from https://github.com/react-native-community/cli/pull/318 and stop using a query parameter in the URL, as support for this regressed in Metro recentlyish and it's not an ideal way to inject this plugin anyhow.

### DIFF
--- a/packages/xdl/src/UrlUtils.js
+++ b/packages/xdl/src/UrlUtils.js
@@ -133,15 +133,20 @@ export async function constructBundleQueryParamsAsync(projectRoot: string, opts:
 
   let { exp } = await ProjectUtils.readConfigJsonAsync(projectRoot);
 
-  // Use an absolute path here so that we can not worry about symlinks/relative requires
-  let pluginModule = ConfigUtils.resolveModule('expo/tools/hashAssetFiles', projectRoot, exp);
-  queryParams += `&assetPlugin=${encodeURIComponent(pluginModule)}`;
-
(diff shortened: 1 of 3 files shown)


packages/xdl/src/UrlUtils.js
```diff
@@ -133,15 +133,20 @@ export async function constructBundleQueryParamsAsync(projectRoot: string, opts:
 
   let { exp } = await ProjectUtils.readConfigJsonAsync(projectRoot);
 
-  // Use an absolute path here so that we can not worry about symlinks/relative requires
-  let pluginModule = ConfigUtils.resolveModule('expo/tools/hashAssetFiles', projectRoot, exp);
-  queryParams += `&assetPlugin=${encodeURIComponent(pluginModule)}`;
-
-  // Only sdk-10.1.0+ supports the assetPlugin parameter. We use only the
-  // major version in the sdkVersion field, so check for 11.0.0 to be sure.
+  // SDK11 to SDK32 require us to inject hashAssetFiles through the params, but this is not
```
ide: SDK32 actually doesn't support the plugin URL parameter either, and we supply it by a manual patch to RN#sdk-32, so this check can look for 31 and below.

brentvatne: I left it in for SDK32 because we still pass the param there even though it's unused and I didn't want to change any behavior for SDK32 - I guess this comment is not so accurate though.

Issue #439, opened 2019-03-19
Add flag to disable web server


sjchmiela:
Not always do people with `app.json#platforms` including `web` want to develop their apps on all 3 platforms at the same time.

It could be beneficial to add some sort of `--disable-web` flag or something similar to be able to prevent Webpack from starting when running `expo start`.

fson:
Instead of adding a flag, perhaps we can make it so that there’s no harm whatsoever from webpack running in the background even if you don’t currently need it? We would not compile for the web (or at least not surface any web specific errors) until a web page is requested from the dev server. This would be analogous to how Metro works for Android/iOS and only serves bundles when requested.

The experience of running `expo start` without ant flags should be as good as possible and when you want to switch platforms, you then don’t have to restart the server with different flags, only press `a`, `i` or `w` to open the app in that platform.

sjchmiela:
Even better! 🎉

PR #523, opened 2019-04-17
Remove `Web` from Default Platforms

cruzach:
If `platforms` key is missing in `app.json`, default is to provide it with all three (ios, android, & web). This results in the warning message - 

>Error: Problem validating fields in app.json. See https://docs.expo.io/versions/v32.0.0/workflow/configuration/
 • Field: platforms - should NOT have more than 2 items.

And also starts the webpack server. Proposing that by default just include ios and android in `platforms`

cc @brentvatne

### DIFF
--- a/packages/config/src/Config.js
+++ b/packages/config/src/Config.js
@@ -558,7 +558,7 @@ export async function readConfigJsonAsync(
   }
 
   if (exp && !exp.platforms) {
-    exp.platforms = ['android', 'ios', 'web'];
+    exp.platforms = ['android', 'ios'];
   }
 
   if (exp.nodeModulesPath) {


brentvatne:
i think this is good! i believe web should remain explicitly opt-in due to the overhead of running webpack

fson:
I agree this should be the default for now. Many existing projects have not been made with web support in mind, so web should be opt-in for existing projects.

To make web support the default in *new* projects, starting in SDK 33 we could add explicit `"platforms": ["android", "ios", "web"]` to `app.json` in the project templates, so new projects will have web enabled by default. But this should *not* make the developer experience any worse (see https://github.com/expo/expo-cli/issues/439#issuecomment-474979318). I believe this is not true yet, but we're getting there.

brentvatne:
i think we should still leave web out of default in new projects until we actually have some people shipping apps with it and have kicked the tires more extensively

PR #524, opened 2019-04-17
Fix Logging of Expo Web server

cruzach:
Right now, `logURL()` for web is called when web isn't supported, so the message 
>Expo Web is running at null

is displayed in logs.

Change was to check for web support before calling `logURL()`

Tested these changes locally


fson:
Thanks, looks good to me.

Could you also change the `logURL` function so that it doesn't `running at null`, if it for some reason gets called when webpack is not running to avoid this kind of error in the future?

cruzach:
Sure, I’ll submit a PR for that today

fson:
Actually never mind, the `logURL` function will be removed in https://github.com/expo/expo-cli/pull/513, so there's nothing more that needs to be done right now.

PR #533, opened 2019-04-19
.web.tsx has higher priority than .tsx

petrbela:
`.web.tsx` was missing from the extensions priority list, causing `.tsx` files to precede `.web.tsx` files. The same for `.web.jsx`.

### DIFF
--- a/packages/webpack-config/webpack/webpack.common.js
+++ b/packages/webpack-config/webpack/webpack.common.js
@@ -299,7 +299,7 @@ module.exports = function(env = {}, argv) {
         new ModuleScopePlugin(locations.template.folder, [locations.packageJson]),
       ],
       symlinks: false,
-      extensions: ['.web.ts', 'web.tsx', '.ts', '.tsx', '.web.js', '.js', '.jsx', '.json'],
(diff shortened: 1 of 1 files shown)


packages/webpack-config/webpack/webpack.common.js
```diff
@@ -299,7 +299,7 @@ module.exports = function(env = {}, argv) {
         new ModuleScopePlugin(locations.template.folder, [locations.packageJson]),
       ],
       symlinks: false,
-      extensions: ['.web.ts', '.ts', '.tsx', '.web.js', '.js', '.jsx', '.json'],
+      extensions: ['.web.ts', 'web.tsx', '.ts', '.tsx', '.web.js', 'web.jsx', '.js', '.jsx', '.json'],
```
fson: Looks like it should be `'.web.tsx'` and not `'web.tsx'`?

petrbela: Yeah, you're right. I made sure to double check it and still missed it 😂

petrbela: Fixed by #538

PR #534, opened 2019-04-19
[web][xdl] Auto add web platform field

EvanBacon:
Prompt the user to add the "web" platform if it's missing at the time the user invokes web functionality.

* Moved validation to `Doctor`.
* Check process for interaction support (For use in CI).
* Prompt the user to add web if the user tries to open web, start webpack, or bundle webpack.
* Throw XDL Error if the user rejects the prompt
* Mutate the `app.json` with the new field.

Screen Shot 2019-04-19 at 1 42 49 PM

Moved from #513


packages/xdl/src/project/Doctor.js
```diff
@@ -432,6 +436,77 @@ export async function validateWithNetworkAsync(projectRoot: string): Promise<num
...
+  let inputExp = exp;
+  if (!exp) {
+    inputExp = (await ProjectUtils.readConfigJsonAsync(projectRoot)).exp;
+  }
+  const { platforms } = inputExp;
+  if (!Array.isArray(platforms)) {
+    return false;
+  }
+
+  const isWebConfigured = platforms.includes('all') || platforms.includes('web');
```
fson: Is `all` actually valid value in the platforms array? Not sure if we want to add this to the schema.

EvanBacon: I saw it somewhere, if it's deprecated I can remove it.

fson: Only `android`, `ios` and `web` values are allowed: https://github.com/expo/universe/blob/94351697e532afedcd51b672f91b3e07716f9b25/server/www/xdl-schemas/UNVERSIONED-schema.json#L44

```suggestion
  const isWebConfigured = platforms.includes('web');
```

PR #547, opened 2019-04-24
Fix Booting when Running Bug

### LINKED ISSUE #536: Unable to boot device in current state: Booted
## Description

I'm occasionally getting this error when starting the iOS simulator: "Unable to boot device in current state: Booted"

Screenshot 2019-04-22 at 19 09 31

## Expected Behavior

Simulator should start without errors.

## Observed Behavior

The `xcrun simctl boot` command fails because when it's called, the simulator is already running:
```
Error running `xcrun simctl boot 4BDC42E4-0213-4850-926E-72D5BB81CACA`: An error was encountered processing the command (domain=com.apple.CoreSimulator.SimError, code=164):
Unable to boot device in current state: Booted
```
This seems to only happen when Simulator.app is initially not running.

I think what might be happening is:
- Simulator.app is not running
- We run `open -a Simulator.app`. This starts **and** boots the simulator.
- While the simulator is still booting, we check the status (in `_waitForSimulatorRunningAsync`).
- The status is not *yet* booted, so we decide to try and boot the simulator.
- In the meantime the simulator has booted so the command fails.

We should probably not try to manually boot the simulator if Simulator.app is initially not running. Instead we could only boot it using `xcrun simctl boot` if it's initially started but not booted.

## Environment

**Tested on master** (4ca219ebb2f7bf46cef2b8ed7b6b2f908b678306)

```
  Expo CLI 2.15.3 environment info:
    System:
      OS: macOS 10.14.4
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 11.13.0 - /usr/local/bin/node
      Yarn: 1.15.2 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5199772
      Xcode: 10.2/10E125 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^33.0.0-alpha.web.1 => 33.0.0-alpha.web.1 
      react-native: 0.57.1 => 0.57.1 
    npmGlobalPackages:
      expo-cli: 2.15.2

```
Phone/emulator/simulator platform, model and version: iPhone X (simulator)

cruzach:
Closes #536 

Follows the logic suggested in the issue, but also needed to create a new waiting function because without it, IF the user quits Simulator app and immediately tries to open their app in iOS simulator after, it won't wait for the device to boot, only for simulator to be running, resulting in the CLI trying to install Expo Client before a device is even booted. Just wanted to explain the need for that new function.

Tested locally, and didn't identify any issues.


packages/xdl/src/Simulator.js
```diff
@@ -141,6 +141,14 @@ export async function _openSimulatorAsync() {
     Logger.global.info('Opening iOS simulator');
     await spawnAsync('open', ['-a', 'Simulator']);
     await _waitForSimulatorRunningAsync();
+    await _waitForDeviceToBoot();
+  }
```
fson: Please run `yarn prettier --write packages/xdl/src/Simulator.js` to fix the formatting of this file. I recommend installing a Prettier plugin in your editor so it formats code automatically.

packages/xdl/src/Simulator.js
```diff
@@ -214,6 +216,13 @@ async function _listSimulatorDevicesAsync() {
   return info;
 }
 
+async function _waitForDeviceToBoot() {
+  let bootedDevice = await _bootedSimulatorDeviceAsync();
+  while (!bootedDevice) {
+    bootedDevice = await _bootedSimulatorDeviceAsync();
+  }
```
fson: A `do-while` loop would be nice here.

I think you might also want to add a small wait between the tries (e.g. `await delayAsync(100)`) in this loop so we don't continuously fire the `xcrun simctl list devices` command.

Also what happens if the device never boots? Perhaps add a timeout/counter to make sure we won't be stuck in this loop indefinitely.

cruzach: added a timeout, set for 10 seconds. When testing, it never took longer than 5, but let me know if you think it should be set to longer/shorter 👍

packages/xdl/src/Simulator.js
```diff
@@ -219,6 +219,21 @@ async function _listSimulatorDevicesAsync() {
...
+  let bootedDevice;
+  const start = Date.now();
+  do {
+    await delayAsync(100);
+    bootedDevice = await _bootedSimulatorDeviceAsync();
+    if (Date.now() - start > 10000) {
+      Logger.global.error(
+        `iOS Simulator device failed to boot. Try opening Simulator first, then running your app.`
+      );
+      throw 'Timed out waiting for iOS Simulator device to boot.';
```
fson:
```suggestion
      throw new Error('Timed out waiting for iOS Simulator device to boot.');
```
It's best to never throw strings, only errors.

cruzach:
@fson attempted to clean up the merge conflicts by updating the `@cruzach/boot-sim-fix`, although now it's a little ugly with "79 Commits" and "Files Changed: 145" for a small change

Would you rather I close this and just open a new PR with the relevant changes to keep things clean?

fson:
Closing this PR in favor of #883.

Issue #545, opened 2019-04-24
After upgrade expo-cli version 2.15.4  run expo web


khalisafkari:
Starting project at /run/media/khalis/DATA/westmanga/westmanga2
Expo DevTools is running at http://localhost:19002
Opening DevTools in the browser... (press shift-d to disable)
Error: Problem validating fields in app.json. See https://docs.expo.io/versions/v32.0.0/workflow/configuration/
 • Field: platforms - should NOT have more than 2 items.
Starting Metro Bundler on port 19003.
Cannot find module './validators/Presets'
Set EXPO_DEBUG=true in your env to view the stack trace.

cruzach:
Hi @khalisafkari,

Can you share your `app.json`? (Specifically the "platforms" key)

-Charlie

khalisafkari:
`"platforms":["android","ios","web"],`

byCedric:
I ran into this issue too when creating the [Expo Github Action](https://github.com/expo/expo-github-action). After some digging around I found the cause, which is kind of a "dooh" moment for me 😅 

Whenever the Expo CLI is started, in any way (e.g. `build` or `start`) it tries to validate the schema right? The JSON Schema is fetched based on the SDK version used. Currently, the latest Expo SDK should be `32`. _Buuuuuut_, I don't think Expo SDK 32 "supports" the web platform right?

You can see this IRL for yourself too, just [use this link to fetch the JSON Schema for Expo SDK 32](https://expo.io/--/xdl-schema/32.0.0) and search for the `platforms` rules. 😄 

I think this can only be fixed during the release of SDK 33, by adding the `web` as platform enum.

z0al:
@khalisafkari what OS do you use? I had the same problem on Debian/Linux and fixed by #554

z0al:
> I think this can only be fixed during the release of SDK 33, by adding the web as platform enum.

@byCedric  Actually, it works just fine even with that error (at least for me). The problem is what I commented about above :point_up: .

Waiting for @khalisafkari to verify the fix.

byCedric:
Ah wait! Maybe it's [because of this (highlighted part)](https://blog.expo.io/expo-cli-and-sdk-web-support-beta-d0c588221375#d5a4)?

> it will not be until the next SDK version that Expo Web projects are compatible with managed Android and iOS projects since this web beta is not part of an official Expo SDK version.

I actually tried to run an existing Android/iOS app with Web, and it resulted in that error. 😄

[...]

PR #548, opened 2019-04-26
Fix validator filename casing for sensitive file systems

brookemitchell:
We noticed a linux only error occurring in our docker builds `Cannot find module './validators/Presets'` and think this is due to case sensitivity from the webpack-pwa-manifest's imports.

To reproduce run `expo build:web` from a linux machine/docker image

I've renamed the files to correctly match these lines  in `webpack-pwa-manifest-plugin/index.js:10-11`.
```js 
import validateColors from './validators/Colors';
import validatePresets from './validators/Presets';
```

Thanks for looking this over


coop182:
Looks like this is at least part of the problem with #545 too.

z0al:
I can confirm. It's actually the problem.

PR #551, opened 2019-04-29
Added `--no-pwa` flag

EvanBacon:
with `expo build:web --no-pwa` you can prevent the bundling of PWA features which can save up to a minute of generation time.


rickysullivan:
> Prevent webpack from generating the manifest.json

Yet `manifest.json` still exists when supplying `--no-pwa`.

Should https://github.com/expo/expo-cli/blob/master/packages/webpack-config/src/webpack.config.ts#L217 be prefixed with `env.pwa && ` to prevent `WebpackPWAManifestPlugin` being added?

I'm building my own `manifest.json` and would like to prevent one from being generated.

PR #555, opened 2019-04-29
Add image optimization functionality

sergeichestakov:
## Overview
This PR is the first step towards asset optimization for Expo projects. It allows for compression (via quality reduction using the [Sharp](https://github.com/lovell/sharp) library) of JPGs and PNGs in order to decrease the spacial footprint of an app. 

This is most relevant for published apps so we will offer two methods of optimization prior to publishing: integration with `expo publish` on the CLI side and a toggle in the Publishing Section of the dev-tools on the web side.

## Summary of Changes
- Adds an `optimize` command to expo-cli with an optional `--save` flag to preserve a backup of the originals. If the `--save` flag is not passed in, the user will be prompted and asked if they want to save a copy of the originals. Copies are saved by appending a `.expo` extension such that `image.png` becomes `image.expo.png`. The original images are overwritten with the compressed versions.
- Adds a toggle to the Project Manager Publishing Section (the screen that pops up when the user intends to publish their app) of the dev-tools that lets the user decide whether or not they want to optimize the assets in their app. It is on by default.
- Adds an `optimizeAsync` function along with some helper functions under `Project.js` in xdl.
- Adds a GraphQL mutation so that the front end of dev-tools can call the `optimizeAsync` function in xdl.
- Updates the `expo publish` command to check for whether or not the user has already optimized their assets. If not, it will prompt the user and ask them whether or not they want to optimize prior to publishing.
- When this command is run, it will create a `.expo-shared` folder in the project's root directory and an `assets.json` within that folder if they don't already exist. This is intended to be committed by the user in order to share the state of asset optimization between collaborators so that if one person optimizes an image, the others will not be able to optimize the same images when they pull changes. `assets.json` contains the hash of the contents of each image in a project. This guarantees that if `expo optimize` was run on a certain image, it will not be rerun on that image unless it was modified. Additionally, every time `expo optimize` is run, it will go through and remove any entries that no longer exist (i.e. assets that were modified or removed) so that `assets.json` only contains relevant entries.


packages/expo-cli/src/commands/publish.js
```diff
@@ -27,6 +30,18 @@ export async function action(projectDir: string, options: Options = {}) {
     );
     process.exit(1);
   }
+  const hasOptimized = fs.existsSync(projectDir + '/.expo-shared/assets.json');
```
esamelson: use `path.join` rather than string concatenation here

packages/xdl/src/Project.js
```diff
@@ -2139,6 +2142,163 @@ export async function getUrlAsync(projectRoot: string, options: Object = {}) {
...
+  assetBundlePatterns.forEach(pattern => {
+    files.push(...glob.sync(pattern, options));
+  });
+  return files.map(file => `${projectDir}/${file}`.replace('//', '/'));
+};
+
+/*
+ * Read the contents of assets.json under .expo-shared folder. Create the file/directory if they don't exist.
+ */
+const readAssetJsonAsync = async projectDir => {
```
esamelson: i would suggest moving this (and possibly some of these other helper functions) to `ProjectUtils` or perhaps even a new `AssetUtils` module, since this file is already gigantic

packages/expo-cli/src/commands/publish.js
esamelson: this `.expo-shared` folder will start showing up for people and I wonder if we should add some explicit instructions in the command (maybe only the first time it's run) that people should commit this folder -- what do you think @sergeichestakov ?

sergeichestakov:
For reference, here is the only visual change on the dev-tools side. Notice the `Optimize Assets` toggle right under the description:

image

I also went ahead and tested this across multiple devices (e.g. two users sharing one repo and pushing/pulling changes related to the assets and running the optimization commands) as well as on my Linux machine (Ubuntu 19.04) and everything seems to be working fine.

packages/xdl/src/Project.js
sergeichestakov: Yeah, I was thinking about that too but wasn't sure where to put them. `AssetUtils` sounds good!

packages/expo-cli/src/commands/publish.js
sergeichestakov: I pushed a commit addressing that issue 1 min before you made this comment 😄 Let me know what you think about the message in b2a0d49

esamelson: 😄 👌 looks good!

esamelson: you read my mind

sergeichestakov:
> implementation looks pretty clean and slick 🚀 i have a few comments in addition to these:
> 
>     * what if you want to optimize only some specific images, but not others? i think we should also be allowed to pass in a glob pattern of files to include/exclude as an optional CLI argument.
> 
>     * do you think there should be a way to toggle `--save` in the dev-tools UI?
> 
>     * what do you think about adding some text to `expo start` as well, that informs people of `expo optimize` if they have not yet run it? I think adding a prompt might be too disruptive to getting started quickly but I wonder if it would be worthwhile to add a warning or some text. (up for discussion though)

Thanks for the feedback! In response to your last points, 
- Added the options to include/exclude certain glob patterns in 4ce0321
- I don't think we should add an option for `--save` in devtools. While that would be a trivial change to make, I think that we should discourage saving the original assets as much as possible because the optimized versions should be indistinguishable and saving them would just clutter the user's assets directory with twice as many files. More importantly, if the user saves all of their assets immediately before publishing then they're uploading the original assets as well as the optimized ones which is counter to our original goal of saving space on our servers and end user's devices. My thinking is that if they don't like the optimized images after they publish, they can just restore them from git and then republish.
- It might be good to add a prompt during `expo start` but I was thinking that this is only really important when publishing so adding it to the publishing workflow should be sufficient. While a prompt during `expo start` would help with discoverability of the command, I want to avoid cluttering the output with stuff that isn't directly relevant, especially given how often people run `expo start`.

Let me know what you think.

esamelson:
Cool, your change looks great pending adding clarification of the `include` and `exclude` options being relative to project root. Agree with your rationale about `--save`.

The only thing I want to add re: `expo start` is that one situation we'd like to try to avoid, if possible, is for people to publish their app (and optimize their assets along the way) and then see a significant downgrade in quality in an asset or two -- i.e. see that their app looks different in production than in development. However, I agree that cluttering the output is not great, and if you think this is sufficiently addressed by all of the output of `optimize` that makes it clear what is going on, I trust your call.

packages/expo-cli/src/commands/publish.js
```diff
@@ -27,6 +31,18 @@ export async function action(projectDir: string, options: Options = {}) {
     );
     process.exit(1);
   }
+  const hasOptimized = fs.existsSync(path.join(projectDir, '/.expo-shared/assets.json'));
+  if (!hasOptimized) {
+    log.warn('It seems your assets have not been optimized yet.');
+    const { allowOptimization } = await prompt({
+      type: 'confirm',
+      name: 'allowOptimization',
+      message: 'Do you want to do this now?',
```
fson:
```suggestion
      message: 'Do you want to optimize assets now?',
```

sergeichestakov:
Great, thanks for the suggestions! I updated the prompt to say `Do you want to save a backup of each file?` Is that better?

I agree that `.orig` is clearer than `.expo` but I think images should be saved as `icon.png` -> `icon.orig.png` rather than `icon.png.orig` because otherwise the OS doesn't know how to interpret/open the backed up images. This makes it a pain for the user if they want to compare the originals with the compressed version since they will most likely be unable to open files that end with `.orig`. Let me know if that makes sense.

Also updated the CLI to not prompt the user to backup assets if there are none to optimize in 
60deb84.

ricardovf:
Please, can you make an option to disable the question everytime if i do not want to optimize?

How to make this default?
"It seems your assets have not been optimized yet.
? Do you want to optimize assets now? No"

Thanks!

esamelson:
Hey @ricardovf - could you explain why you don't want to optimize your images? We currently host all your assets for free on our CDN (which is not free for us) and we'd like to be hosting optimized images as much as possible, but would like to hear in more detail what isn't working for you about this workflow.

(Note that you can also exclude certain assets from being optimized: https://docs.expo.io/versions/v32.0.0/guides/assets/#optimization)

ricardovf:
Hi @esamelson sure! The problem is that i noticed that a gradient image was not very good after the optimisation run. If you like i can post an example.

esamelson:
Hey @ricardovf - you can get around this by running `expo optimize --quality=<some number > 60>` and seeing if you can get better results, or if not, `expo optimize --exclude="<path/to/gradient.png>"`. Let us know if that doesn't work for you!

pckz:
I optimized my images but I want to go back and restore originals. How can I do that?
Thanks

sergeichestakov:
If you didn't save a backup the best way to restore images is through your git history. Just checkout the file at the commit before optimization.

pckz:
@sergeichestakov and if I save the backup images, exists a parameter to restore? (example expo optimize --rollback)

sergeichestakov:
No, we don't currently support that option

sergeichestakov:
Fixed in https://github.com/expo/expo-cli/pull/647 @ricardovf

PR #562, opened 2019-05-01
[xdl] Provide an appropriate warning message for SDK 33 on iOS devices

brentvatne:
This is important because when we release SDK 33 a compatible client will not yet be available on the iOS App Store and users will need to do an ad-hoc build to run it on their devices. I need to follow this PR up with a page at https://expo.io/ad-hoc that re-directs to the appropriate doc when it is ready.

<hr />

Screen Shot 2019-04-30 at 18 50 00

### DIFF
--- a/packages/xdl/src/Project.js
+++ b/packages/xdl/src/Project.js
@@ -1672,6 +1672,44 @@ function shouldExposeEnvironmentVariableInManifest(key) {
   return key.startsWith('REACT_NATIVE_') || key.startsWith('EXPO_');
 }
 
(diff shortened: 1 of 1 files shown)


brentvatne:
It'd be good to land this soon and ship it so that more folks are already on the latest version of expo-cli when we release SDK 33, and can benefit from this warning.

fson:
Looks like this message won't be necessary for SDK 33, should we close the PR for now @brentvatne?

brentvatne:
yup sounds good

Issue #124, opened 2018-10-30
Reuse existing Chrome tab for dev tools


fson:
When Chrome is set as the default browser, we should reuse the existing tab instead of opening a new tab. On macOS we could do something like this: https://github.com/facebook/create-react-app/blob/d671b8fb0430d11ebc6f399274bd5f2cfbb1de65/packages/react-dev-utils/openChrome.applescript, but we should also first check if Chrome is the default browser. To detect if Chrome is the default browser, we can do `defaults read ~/Library/Preferences/com.apple.LaunchServices/com.apple.launchservices.secure` and find the handler for `LSHandlerURLScheme = http`.

imflavio:
Any update on this feature?

methodbox:
I strongly support this!

I just opened my browser after working for a few hours on an app last night to find my pinned tab (btw you should be sure this also re-uses the pinned tab) plus another 6 tabs opened. Those are just the ones I missed when I put it to sleep last night.

Every time I push a commit I get a new tab. It's pretty annoying.

moshfeu:
Just like the debugger window.
Another sub task, if somehow the process exited, just close the tab.

Thanks!

byCedric:
I would be up to get some work started on this issue! Yesterday I had 29 tabs open 😅... 

To get started I think some guidance is required. E.g. where should the script(s) for this functionality be placed? I noticed [Vue CLI ripped the code from CRA](https://github.com/vuejs/vue-cli/commit/985654943238e80916677fd8487460fdfacefb35) and placed it in a similar spot. I don't think Expo has the same structure right? (a specific `expo-dev-utils` folder). Should it be placed in the `xdl`, `dev-tools`, `expo-cli` or should it be in a new folder named something like `open-dev-tools`?

For now, I think I can start on creating a dedicated script to find the best approach to open the browser (e.g. platform/browser check). After that, I can try to make the default browser check happening?

PS. I think this feature is also a good opportunity to upgrade `opn` to the latest `open` version (it was renamed), but that [requires dropping `node@6` support](https://github.com/expo/expo-cli/issues/457#issuecomment-487370630).

byCedric:
This is now fixed by #565, right? Thanks Evan ❤️

PR #565, opened 2019-05-01
Reuse browser tabs

EvanBacon:
Currently, when using web and devtools you get a new tab each time. This exacerbates the annoying feeling of needing to restart the bundler. Using `react-dev-utils/openBrowser` wrapper for the `opn` module we can attempt to reuse tabs when possible.

### DIFF
--- a/packages/dev-tools/server/dev-server.js
+++ b/packages/dev-tools/server/dev-server.js
@@ -3,7 +3,7 @@ import { Project } from 'xdl';
 import express from 'express';
 import http from 'http';
 import next from 'next';
-import opn from 'opn';
+import openBrowser from 'react-dev-utils/openBrowser';
 
 import { startGraphQLServer } from './DevToolsServer';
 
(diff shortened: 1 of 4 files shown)


fson:
Nice! People have asked about this before (https://github.com/expo/expo-cli/issues/124). It will make some Chrome users very happy. (Safari always reuses the same tab by default, so it already works the same way.)

A side effect of this approach is that you get a scary looking prompt the first time the CLI tries to open something in Chrome:
Screenshot 2019-05-03 at 1 26 53

EvanBacon:
:/ Good catch, that is unfortunate. AFAICT there don't seem to be any complaints with CRA doing this.

Related: [`react-dev-utils/openBrowser`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js)

qwoodmansee:
Any chance there is a way to allow these permissions after the fact? I didn't realize this would be necessary and I denied the permissions very early on in the process!

ide:
Maybe try: Settings > Security & Privacy > Privacy > Automation > Terminal

Screen Shot 2019-07-11 at 3 33 48 PM

newfylox:
Any chance this could work with Ubuntu and Chrome?

PR #566, opened 2019-05-02
Config upgrades for codesandbox

EvanBacon:
* Possible to generate a temp expo config in the `.expo` folder when one isn't present. This will also log out: "Config not found. Using virtual app.json, it's recommended that you create an app.json in the root of your project."
* Async webpack.config for reading a newly generated expo config.

# Test Plan

In a project without an `app.json` use `node start.js`

```js
// /start.js
const fs = require("fs");
const { Webpack } = require("xdl");

Webpack.startAsync(
  fs.realpathSync(process.cwd()),
  { nonInteractive: process.stdout.isTTY },
  true
);
```

which should generate `.expo/app.json`, and start the web project.

Once published, the following sandbox should work: https://codesandbox.io/s/rr8kl1lkpn

### DIFF
--- a/packages/config/src/Config.js
+++ b/packages/config/src/Config.js
@@ -232,7 +232,7 @@ export async function getConfigForPWA(
   getAbsolutePath: Function,
   options: Object
 ) {
-  const { exp } = await readConfigJsonAsync(projectRoot, { isConfigOptional: true });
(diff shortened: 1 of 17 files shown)


EvanBacon:
Stale

Issue #107, opened 2018-10-22
Eject doesn't create index.js file anymore


Aximem:
## Description

Eject command doesn't create index.js file anymore.

## Expected Behavior

expo init Test
cd Test
expo start
expo eject

## Observed Behavior

No index.js file are created (like it was in previous version), we get a : 

>  Cannot find entry file index.js

And we have to create, manually, index.js file.

## Environment

Environment:
  OS: macOS 10.14
  Node: 8.12.0
  Yarn: 1.3.2
  npm: 6.4.1
  Watchman: 4.9.0
  Xcode: Xcode 9.4.1 Build version 9F2000
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  expo: Not Found
  react: 16.3.1 => 16.3.1
  react-native: https://github.com/expo/react-native/archive/sdk-30.0.0.tar.gz => 0.55.4

Diagnostics report:
  https://exp-xde-diagnostics.s3.amazonaws.com/aximem4363-f21198ce-aaec-4ede-9b5e-5f6c9d620287.tar.gz

Phone/emulator/simulator iOS or Android

j3r1ch0-2007:
I understand that there limited development resources to work on issues but this was opened in October. I just encountered the same issue.

bmwertman:
I'm just gettin started with React Native and ran into this issue. I ejected using the "React Native" option following this [Medium post](https://codeburst.io/how-to-avoid-wasting-time-building-a-mobile-app-and-make-a-release-with-single-click-part-1-19c598d21392) on getting started.

sergeichestakov:
Sorry for the delay everyone. This issue was resolved in #567.

PR #567, opened 2019-05-03
Fix problems with building/running an ejected Expo app

sergeichestakov:
This PR addresses several errors related to building and running an ejected Expo app described below:

1. `react-native run-android` and `react-native run-ios` were both throwing the following error:
```
$ react-native run-ios
internal/modules/cjs/loader.js:651
   throw err;
   ^

Error: Cannot find module '@babel/runtime/helpers/interopRequireDefault'
   at Function.Module._resolveFilename (internal/modules/cjs/loader.js:649:15)
   at Function.Module._load (internal/modules/cjs/loader.js:575:25)
   at Module.require (internal/modules/cjs/loader.js:705:19)
   at require (internal/modules/cjs/helpers.js:14:16)
   at Object.<anonymous> (/Users/sergeichestakov/Documents/test-eject/node_modules/react-native/local-cli/cliEntry.js:11:41)
   at Module._compile (internal/modules/cjs/loader.js:799:30)
   at Module._compile (/Users/sergeichestakov/Documents/test-eject/node_modules/pirates/lib/index.js:99:24)
   at Module._extensions..js (internal/modules/cjs/loader.js:810:10)
   at Object.newLoader [as .js] (/Users/sergeichestakov/Documents/test-eject/node_modules/pirates/lib/index.js:104:7)
   at Module.load (internal/modules/cjs/loader.js:666:32)
error Command failed with exit code 1.
```
This is fixed by adding `@babel/runtime` to the dependencies in `package.json` for ejected apps.

2. The following error was being thrown on Android only:
```
Could not find the "expo" package in your project when configuring the packager
JS server already running.
Building and installing the app on the device (cd android && ./gradlew installDebug)...

> Task :app:compileDebugJavaWithJavac FAILED
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:5: error: cannot find symbol
import com.facebook.react.ReactApplication;
                        ^
 symbol:   class ReactApplication
 location: package com.facebook.react
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:6: error: cannot find symbol
import com.facebook.react.ReactNativeHost;
                        ^
 symbol:   class ReactNativeHost
 location: package com.facebook.react
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:14: error: cannot find symbol
public class MainApplication extends Application implements ReactApplication {
                                                           ^
 symbol: class ReactApplication
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:16: error: cannot find symbol
 private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
               ^
 symbol:   class ReactNativeHost
 location: class MainApplication
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:36: error: cannot find symbol
 public ReactNativeHost getReactNativeHost() {
        ^
 symbol:   class ReactNativeHost
 location: class MainApplication
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:16: error: cannot find symbol
 private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
                                                      ^
 symbol:   class ReactNativeHost
 location: class MainApplication
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainApplication.java:35: error: method does not override or implement a method from a supertype
 @Override
 ^
/Users/sergeichestakov/Documents/test-eject/android/app/src/main/java/com/testeject/MainActivity.java:5: error: MainActivity is not abstract and does not override abstract method getPackages() in ReactActivity
public class MainActivity extends ReactActivity {
      ^
8 errors


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```
This is fixed by setting an explicit version of React Native in `package.json` (e.g. `0.57.1` rather than  `https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz` as is default).

3. I removed a check for sdk version that was preventing ejected apps >= sdk22.0.0 from generating an `index.js` which meant `App.js` would not register and the app would therefore throw a 404 as non-expo React Native apps require an `index.js` as an entry point.

A comment in the code erroneously stated:
```
// Starting from react-native 0.49.x (SDK 22), react-native eject template includes this out of the box.
```
when in fact that is not the case as described [here](https://github.com/facebook/react-native/releases/tag/v0.49.0). RN 0.49.x is simply the release that switched to having a single, shared point of entry (`index.js`) rather than one for each platform.

### DIFF
--- a/packages/expo-cli/src/commands/eject/Eject.js
+++ b/packages/expo-cli/src/commands/eject/Eject.js
@@ -231,7 +231,9 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
     const versions = await Versions.versionsAsync();
     const reactNativeVersion = versions['sdkVersions'][sdkVersion]['facebookReactNativeVersion'];
 
-    pkgJson.dependencies['@babel/runtime'] = 'latest';
+    if (semver.satisfies(sdkVersion, '31 - 32')) {
+      pkgJson.dependencies['@babel/runtime'] = '^7.0.0';
+    }
     pkgJson.dependencies['react-native'] = reactNativeVersion;
 
     if (pkgJson.jest !== undefined) {
--- a/packages/expo-cli/src/commands/eject/Eject.js
+++ b/packages/expo-cli/src/commands/eject/Eject.js
@@ -259,16 +259,14 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
 
     log(chalk.green('Your package.json is up to date!'));
 
-    // Starting from react-native 0.49.x (SDK 22), react-native eject template includes this out of the box.
-    if (!Versions.gteSdkVersion(exp, '22.0.0')) {
-      log(`Adding entry point...`);
-      const lolThatsSomeComplexCode = `import { AppRegistry } from 'react-native';
+    log(`Adding entry point...`);
+    const indexjs = `import { AppRegistry } from 'react-native';
 import App from './App';
+
 AppRegistry.registerComponent('${appJson.name}', () => App);
 `;
-      await fse.writeFile(path.resolve('index.js'), lolThatsSomeComplexCode);
-      log(chalk.green('Added new entry points!'));
-    }
+    await fse.writeFile(path.resolve('index.js'), indexjs);
+    log(chalk.green('Added new entry points!'));
 
     log(`
 Note that using \`${npmOrYarn} start\` will now require you to run Xcode and/or
--- a/packages/expo-cli/src/commands/eject/Eject.js
+++ b/packages/expo-cli/src/commands/eject/Eject.js
@@ -4,6 +4,7 @@ import chalk from 'chalk';
 import fse from 'fs-extra';
 import matchRequire from 'match-require';
 import path from 'path';
+import semver from 'semver';
 import spawn from 'cross-spawn';
 import spawnAsync from '@expo/spawn-async';
 import { ProjectUtils, Detach, Versions } from 'xdl';
--- a/packages/expo-cli/src/commands/eject/Eject.js
+++ b/packages/expo-cli/src/commands/eject/Eject.js
@@ -227,6 +227,13 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
     pkgJson.scripts.ios = 'react-native run-ios';
     pkgJson.scripts.android = 'react-native run-android';
 
+    const { sdkVersion } = exp;
+    const versions = await Versions.versionsAsync();
+    const reactNativeVersion = versions['sdkVersions'][sdkVersion]['facebookReactNativeVersion'];
+
+    pkgJson.dependencies['@babel/runtime'] = 'latest';
+    pkgJson.dependencies['react-native'] = reactNativeVersion;
+
     if (pkgJson.jest !== undefined) {
       newDevDependencies.push('jest');
 


packages/expo-cli/src/commands/eject/Eject.js
```diff
@@ -227,6 +227,13 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
     pkgJson.scripts.ios = 'react-native run-ios';
     pkgJson.scripts.android = 'react-native run-android';
 
+    const { sdkVersion } = exp;
+    const versions = await Versions.versionsAsync();
+    const reactNativeVersion = versions['sdkVersions'][sdkVersion]['facebookReactNativeVersion'];
+
+    pkgJson.dependencies['@babel/runtime'] = 'latest';
```
fson: This is risky, because it always pulls in the latest version of the package, so any new release of `@babel/runtime` might break it.

As far as I know, this issue also only affected react-native 0.57.1 ([0.57.2 added `@babel/runtime` in RN's dependencies](https://github.com/facebook/react-native/blob/4a440f40189ea54bd64f7a03bd05bf26619ad320/package.json#L152)).

Because of this, I think it would be best to use a specific version of `@babel/runtime` and limit the fix to only those affected versions (only SDK 31 and SDK 32 used the 0.57.1 fork, SDK 33 will use 0.59):
```suggestion
    if (semver.satisfies(sdkVersion, '31 - 32')) {
      pkgJson.dependencies['@babel/runtime'] = '^7.0.0';
    }
```

fson:
This fixes https://github.com/expo/expo-cli/issues/107.

packages/expo-cli/src/commands/eject/Eject.js
sergeichestakov: Good call.

packages/expo-cli/src/commands/eject/Eject.js
```diff
@@ -227,6 +228,15 @@ from \`babel-preset-expo\` to \`babel-preset-react-native-stage-0/decorator-supp
     pkgJson.scripts.ios = 'react-native run-ios';
     pkgJson.scripts.android = 'react-native run-android';
 
+    const { sdkVersion } = exp;
+    const versions = await Versions.versionsAsync();
+    const reactNativeVersion = versions['sdkVersions'][sdkVersion]['facebookReactNativeVersion'];
+
+    if (semver.satisfies(sdkVersion, '31 - 32')) {
```
esamelson: @sergeichestakov is this only needed for SDKs 31 and 32 (and explicitly NOT for 33 and above)? If not would prefer something like `semver.gte(sdkVersion, '31')` to be more forward-compatible.

esamelson:
(ignore my deleted comment, just saw Ville's! 😅 )

sergeichestakov:
> (ignore my deleted comment, just saw Ville's! 😅 )

Not a bad suggestion though. Made the check more robust in 688dacd

PR #569, opened 2019-05-03
[expo-cli] order options from most desirable to least

quinlanj:
For the Adhoc build workflow, the most desirable options (in my opinion) is
1/ use existing certificate we have on our servers (least effort)
2/ create new one
3/ upload cert from user's filesystem (most effort)

in the terminal workflow, we should display the options from most desirable to least.


brentvatne:
this is nice :) for distribution certs -- could we also present options for the ones that already exist on apple developer portal? my understand is that it is possible to download those. this would be helpful for people who have not built anything with expo before, but have used react-native or built ios apps some other way before, and want to get a client easily

PR #571, opened 2019-05-04
webpack-pwa-manifest-plugin tests

EvanBacon:
Added tests to ensure that the PWA manifest and assets are generated correctly.


EvanBacon:
Currently the failing results are too cryptic. The tests are also pretty fragile.

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
...
+                ? previous + 1
+                : previous;
+            }, 0);
+            if (result === testContentLength) {
+              console.log(chalk.bgGreen.black(`Test "${name}" passed.`));
+              console.timeEnd(name);
+              run(next.shift(), next);
+            } else {
+              console.log(chalk.bgRed.black('There are files missing or with different content.'));
+              console.log(chalk.bgRed.black(`Test "${name}" failedon file ${outputContent}.`));
```
ide:
```suggestion
              console.log(chalk.bgRed.black(`Test "${name}" failed on file ${outputContent}.`));
```

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
...
+        const outputPathLength = outputPath.length;
+
+        // Get the file contents for the directory we just generated
+        ls(outputPath, (err, outputTree) => {
+          if (err) throw err;
+
+          const outputContent = outputTree.filter(i => /.*\..+/.test(i));
+          const outputContentLength = outputContent.length;
+          if (testContentLength === outputContentLength) {
+            const result = outputContent.reduce((previous, current) => {
```
ide: `reduce` obfuscates the intent more than alternatives -- either write out the for-loop or run `filter` and get the `.length`

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
...
+              console.log(chalk.bgRed.black(`Test "${name}" failedon file ${outputContent}.`));
+              console.timeEnd(name);
+              assert(result === testContentLength);
+            }
+          } else {
+            console.log(chalk.bgRed.black(`Expected ${testContentLength} file(s).`));
+            console.log(chalk.bgRed.black(`Found ${outputContentLength} file(s).`));
+            console.log(chalk.bgRed.black(`Test "${name}" failed.`));
+            console.timeEnd(name);
+            assert(testContentLength === outputContentLength);
```
ide: this assertion is going to fail, we've already printed out the error, just do nothing here instead?

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
...
+            }, 0);
+            if (result === testContentLength) {
+              console.log(chalk.bgGreen.black(`Test "${name}" passed.`));
+              console.timeEnd(name);
+              run(next.shift(), next);
+            } else {
+              console.log(chalk.bgRed.black('There are files missing or with different content.'));
+              console.log(chalk.bgRed.black(`Test "${name}" failedon file ${outputContent}.`));
+              console.timeEnd(name);
+              assert(result === testContentLength);
```
ide: see note below about assertions

packages/webpack-pwa-manifest-plugin/tests/index.js
```diff
@@ -0,0 +1,5 @@
+const runTest = require('./runTest');
+const tests = ['basic'];
+
+console.log('Running tests...');
+runTest(tests.shift(), tests);
```
ide: make it so the caller doesn't need to do this `shift()` stuff:

```js
const runTests = require('./runTests');
runTests(tests);
```

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
...
+const path = require('path');
+const ls = require('list-directory-contents');
+const fs = require('fs');
+const assert = require('assert');
+const chalk = require('chalk');
+
+function run(name, next) {
+  console.time(name);
+  if (!name) {
+    return console.log('End.');
```
ide: the intent is not to return whatever console.log returns (in practice, undefined)
```suggestion
   console.log('End.');
   return;
```

packages/webpack-pwa-manifest-plugin/tests/index.js
```diff
@@ -0,0 +1,5 @@
+const runTest = require('./runTest');
+const tests = ['basic'];
+
+console.log('Running tests...');
```
ide: Move this into the test runner function, since it's responsible for printing out the rest of the test status output

packages/webpack-pwa-manifest-plugin/tests/runTest.js
```diff
@@ -0,0 +1,67 @@
+const webpack = require('webpack');
+const path = require('path');
+const ls = require('list-directory-contents');
+const fs = require('fs');
+const assert = require('assert');
+const chalk = require('chalk');
+
+function run(name, next) {
```
ide: seems like it would be easier to make this an async function, also that way it could return a value if needed. (would need to util.promisify(webpack) probably)

yarn.lock
```diff
@@ -1102,6 +1102,16 @@
     mv "~2"
     safe-json-stringify "~1"
 
+"@expo/config@^1.0.3":
```
ide: Not sure this should be here -- yarn workspaces should link everything

PR #574, opened 2019-05-06
[web] Fixed various bugs related to building PWAs

EvanBacon:
* Set a non-zooming viewport 
* Notify user which mode they're running
* Fix undefined icons type
* Added override for lib require 
* Removed extra apple parsing
* Filter out startup images to prevent android PWA bundling
* Update `createMetatagsFromConfig.js`
* Removed public path for manifest
* Ensure plugins are in the correct order
* Removed extra process


EvanBacon:
This currently adds a lot of extra build time for reloads.

PR #587, opened 2019-05-08
[expo-cli] autoselect creds if they exist for adhoc builds

quinlanj:
Most users probably don't want to be bothered with manually selecting credentials unless absolutely neccessary. If we find valid credentials, we should autoselect it and just expose a confirmation (Y/n).

### DIFF
--- a/packages/expo-cli/src/commands/client/selectDistributionCert.js
+++ b/packages/expo-cli/src/commands/client/selectDistributionCert.js
@@ -13,6 +13,21 @@ export default selectDistributionCert;
 async function selectDistributionCert(context, options = {}) {
   const certificates = context.username ? await chooseUnrevokedDistributionCert(context) : [];
   const choices = [...certificates];
+
+  if (certificates.length > 0 && !options.disableAutoSelectExisting) {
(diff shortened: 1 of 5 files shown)


quinlanj:
TODO: maybe just autoselect w/o prompting if we find valid creds?

fson:
I agree we should just use the previous credentials (if valid) without any extra prompts.

packages/expo-cli/src/commands/client/selectDistributionCert.js
```diff
@@ -13,6 +13,14 @@ export default selectDistributionCert;
 async function selectDistributionCert(context, options = {}) {
   const certificates = context.username ? await chooseUnrevokedDistributionCert(context) : [];
   const choices = [...certificates];
+
+  // autoselect creds if we find valid ones
+  if (certificates.length > 0 && !options.disableAutoSelectExisting) {
+    const autoselectedCertificate = certificates[0];
```
fson: We take the first cert from the array here — how is this array sorted?

I believe we should prefer the one that shows as ”used in @username/home` (or whatever the slug off the client is).

PR #592, opened 2019-05-08
feat: add a `--changelog` flag to the `expo upload:ios` command

mackbrowne:
Add `changelog` option to IOS_OPTIONS so I can add test notes for my QA team when uploading to TestFlight through circleCI

### DIFF
--- a/packages/expo-cli/src/commands/upload.js
+++ b/packages/expo-cli/src/commands/upload.js
@@ -33,6 +33,7 @@ export default program => {
     'sku',
     'language',
     'publicUrl',
+    'changelog'
   ];
   const iosCommand = program.command('upload:ios [projectDir]').alias('ui');
   setCommonOptions(iosCommand, '.ipa');


fson:
Thanks, this seems like a useful addition. Could you add a section "Test plan" to the PR description, where you describe how you tested that this feature works correctly?

packages/expo-cli/src/commands/upload.js
```diff
@@ -33,6 +33,7 @@ export default program => {
     'sku',
     'language',
     'publicUrl',
+    'changelog'
```
fson: I believe that for the `changelog` to be passed correctly to the App Store API, you'll also need to add it to to the [options of the upload command](https://github.com/expo/expo-cli/blob/a741eb76fdd1f62eb6194bec97c9c3135be12c3c/packages/expo-cli/src/commands/upload.js#L41-L69)
and [IOSUploader.js](https://github.com/expo/expo-cli/blob/a741eb76fdd1f62eb6194bec97c9c3135be12c3c/packages/expo-cli/src/commands/upload/IOSUploader.js#L174-L188). Please test the command to make sure it works.

mackbrowne:
@fson thanks for the reply!  

I'll add a test plan and address your comments as well.  One question though.  What's the easiest way to test this locally?  will `npm i -g mackbrowne/expo-cli#master` work?

fson:
The easiest way to test it locally is documented here: https://github.com/expo/expo-cli/blob/master/CONTRIBUTING.md#setting-up-the-repository-for-development

After setting up the local repo according to the instructions in CONTRIBUTING.md, you can run the local Expo CLI with `./packages/expo-cli/bin/expo.js`

mackbrowne:
Thanks @fson working on this now, will report back within a few days

brentvatne:
hey there, it looks like there hasn't been movement on this for a while. i'll close the issue but if anyone wants to pick it up this would be a welcome feature!

PR #602, opened 2019-05-10
feat: added commit message template

cruzach:
New template follows the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) format. Also added a step in the `Setting up the repository for development` section of the `Contributing` readme to configure this template for your local repo, but maybe we can make this part of the `bootstrap` script so there aren't 2 separate scripts to run to set up the repo for development?

This is the first step towards an automated changelog. 
Second step- We already use `lerna publish` command in the `publish` script, adding the [--conventional-commits flag](https://github.com/lerna/lerna/tree/master/commands/version#--conventional-commits) should automatically up our versions & generate/update changelog files (NOTE: I'm not familiar with Lerna so feel free to correct me if I'm wrong about that implementation)

### DIFF
--- a/.github/.COMMIT_TEMPLATE
+++ b/.github/.COMMIT_TEMPLATE
@@ -1,0 +1,21 @@
+# 🚨🚨ATTENTION🚨🚨
+# Please follow the commit message format below 📝
+#
+# <type>: <description>
+# [optional body]
+# [optional footer]
+# 
+# types: "fix" = patches a bug in the codebase
+# 	 "feat" = introduces new feature to codebase
+#	 "improvement" = improvement without adding a feature or fixing a bug
+#	 various others: "chore" "docs" "style" "refactor" "perf" "test"
+# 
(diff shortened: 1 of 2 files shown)


.github/.COMMIT_TEMPLATE
```diff
@@ -0,0 +1,21 @@
...
+# 
+# Include "BREAKING CHANGE:" at the beginning of the body or footer if your commit introduces a breaking API change
+
+
+
+
+
+
+# Example commit messages @ https://www.conventionalcommits.org/en/v1.0.0-beta.4/#examples
+# Thank you for cooperating, this allows us to auto-generate fantastic a fantastic changelog :)
```
sjchmiela:
```suggestion
# Thank you for cooperating, this allows us to auto-generate a fantastic changelog :)
```
😅

cruzach: 😟Oops! Thanks for catching that

PR #626, opened 2019-05-17
feat: allow webpack alias to be specified in app.json

sstur:
This allows us to customize the webpack module resolution using `alias` from within our app.json.

Example app.json:
```json
{
  "expo": {
    "name": "Test Expo Web",
    "slug": "testme-web",
    "privacy": "public",
    "sdkVersion": "32.0.0",
    "platforms": ["ios", "android", "web"],
    "version": "1.0.0",
....
    "web": {
      "build": {
        "alias": {
          "react-thing$": "@me/custom-react-thing"
        }
      }
    }
  }
}
```

Side note: I assume I need to update JSON Schema somewhere to reflect this new property, but I am not sure where. Can you point me in the right direction?

### DIFF
--- a/packages/webpack-config/webpack/webpack.common.js
+++ b/packages/webpack-config/webpack/webpack.common.js
@@ -421,7 +421,7 @@ module.exports = function(env = {}, argv) {
       ],
     },
     resolve: {
-      alias: DEFAULT_ALIAS,
+      alias: { ...DEFAULT_ALIAS, ...config.web.build.alias },
       extensions: [
         '.web.ts',
         '.web.tsx',


sstur:
Hey @EvanBacon. Anything I can add to this PR to help move it forward?

It would be super helpful to get it in the next release since we are limited at the moment what webpack config options we can specify in app.json.

Also, is there a JSON Schema somewhere that should document all the webpack related configuration? If not, you want me to make one?

fson:
Note that you can also create a `webpack.config.js` file in the root folder of your app and Expo CLI will use it instead of the default configuration. You can import the base configuration from the [`@expo/webpack-config` package](https://github.com/expo/expo-cli/tree/a7996867b42c2ce2c9699cc249b11be85a92d4b5/packages/webpack-config) and override any parts that you want (for example, by making use of the `webpack-merge` package).

I think creating a custom `webpack.config.js` is be a more sustainable way to configure less frequently used or more advanced webpack options than `app.json`. Even if we extended `app.json` to support every webpack option imaginable, it would probably not be as well documented as the webpack configuration itself, always lag behind a bit when new options get added.

@EvanBacon I think we should make sure the way to override options by creating a `webpack.config.js` is documented, since it's the way people can use more advanced webpack configuration in their Expo web apps.

fson:
@sstur as far as the JSON Schema is concerned, Expo CLI currently downloads it from our API (e.g. http://exp.host/--/xdl-schema/32.0.0) and the schema file is stored in the (private) server repo. I think it should be moved to one of our open source repos as quickly as possible to allow anyone to contribute.

sstur:
Thanks @fson, that is helpful to know that I can use a custom `webpack.config.js` which unblocks me on multiple fronts.

Regarding the change proposed in this PR, if you do get a final decision on if it could be closed or not just let me know.

Regarding the xdl schema, moving that to a public facing repo makes sense if that happens I'm happy to contribute. Thanks!

brentvatne:
made an issue about moving the schemas to the oss repo: https://github.com/expo/expo/issues/6449

i agree that using a custom webpack config (extending the default expo one) is a good approach for such things. appreciate the pr nonetheless. @EvanBacon if you see this and think differently please chime in. i'll close the pr for now 😍

PR #632, opened 2019-05-18
[xdl] Update Podfile generator for CocoaPods 1.6 support for SDK 33

ide:
CocoaPods 1.6 changed the Ruby API so we need to make some changes in the emitted Ruby code within the generated Podfiles. This is not backwards-compatible with CocoaPods 1.5. So for SDK 33 we'll require CocoaPods 1.6, and for SDK 32 we'll require CocoaPods 1.5 (and want to keep 1.5 compatibility so that people can keep the version of CocoaPods they already have for SDK 32 projects).

This commit changes the way we get the `native_target` object to use an approach recommended by one of the CocoaPods maintainers. It also changes the way we set up the framework search paths; namely, we want to fall back on `$(inherited)`, which gets set to look for all of the installed CocoaPods' frameworks. This is so that we can use `@import` syntax in Objective-C, which is necessary for FBAudienceNetwork and GoogleSignIn, since their headers are no longer installed to `Pods/Headers/Public/`. Lastly, it adjusts the indentation a little to make the generated Podfile look better.

Tested by linking xdl into `expo/tools-public` and then running `generate-files-ios.sh` in the expo repo, along with some other changes in the expo repo so that the newly generated Podfile works and that the Obj-C code compiles. Re-installed the Pods and verified that the iOS client compiles and runs.

### DIFF
--- a/packages/xdl/src/detach/IosPodsTools.js
+++ b/packages/xdl/src/detach/IosPodsTools.js
@@ -247,23 +247,34 @@ async function _concatTemplateFilesInDirectoryAsync(directory, filterFn) {
 }
 
 function _renderDetachedPostinstall(sdkVersion, isServiceContext) {
+  const sdkMajorVersion = parseSdkMajorVersion(sdkVersion);
+  const podNameExpression = sdkMajorVersion < 33 ? 'target.pod_name' : 'pod_name';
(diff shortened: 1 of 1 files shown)


ide:
(Merging this now since commits in the Expo repo depend on these changes, will publish them as a prerelease version to npm.)

PR #636, opened 2019-05-20
Rename xdl to @expo/xdl (scoped package)

fson:
npm currently has issues with 3 character package names (https://status.npmjs.org/incidents/qg46dsfk1vt2) and scoped packages also have more permission options available (e.g. `npm access` command only works with scoped packages).


fson:
I need to figure out why the test is failing.

PR #642, opened 2019-05-22
Convert @expo/json-file to TypeScript

fson:


packages/babel-preset-cli/index.js
```diff
@@ -0,0 +1,14 @@
+module.exports = () => ({
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: '8.0.0',
```
ide: Node 8.9.0 is when LTS started, and notably 8.7.0 is when native async functions were added. So by supporting LTS and newer, we can use native async functions (and the latest Node 12 supports experimental async stack traces when debugging).

```suggestion
          node: '8.9.0',
```

packages/babel-preset-cli/index.js
```diff
@@ -0,0 +1,14 @@
...
+        },
+      },
+    ],
+    '@babel/preset-typescript',
+  ],
+  plugins: ['@babel/plugin-proposal-class-properties', '@babel/plugin-proposal-object-rest-spread'],
```
ide: Node 8.6.0+ supports object rest & spread properties. Also class properties are part of TypeScript -- does the TS preset handle them?

packages/json-file/src/JsonFile.ts
```diff
@@ -1,33 +1,34 @@
...
-  default?: JSONObject,
-  json5?: boolean,
-  space?: number,
+const writeFileAtomicAsync: (
+  filename: string,
+  data: string | Buffer,
+  options: writeFileAtomic.Options
+) => void = promisify(writeFileAtomic);
+
+type JSONT = { [key: string]: any };
```
ide: Could be stricter with this:

```typescript
type JSONValue = boolean | number | string | null | JSONArray | JSONObject;
interface JSONArray extends Array<JSONValue> {}
interface JSONObject {
  [key: string]: JSONValue;
}
```

packages/json-file/src/JsonFile.ts
```diff
@@ -43,7 +44,7 @@ const DEFAULT_OPTIONS: Options<*> = {
  * and object with string keys and either objects or primitive types as values.
  * @type {[type]}
  */
-export default class JsonFile<JSONObject: JSONT> {
+export default class JsonFile<JSONObject extends JSONT> {
```
ide: In TS and other languages, the convention for generic type parameter names is to use one or two characters usually:
```suggestion
export default class JsonFile<J extends JSONT> {
```

(maybe it is not actually a good convention, but when reading through the rest of this file it makes it more evident which types are generics and which are concrete types)

.circleci/config.yml
```diff
@@ -19,8 +19,14 @@ jobs:
...
       - run:
-          working_directory: ~/expo-dev-tools/packages/dev-tools
+          name: 'dev-tools tests'
+          working_directory: ~/expo-cli/packages/dev-tools
           command: yarn test
       - run:
-          working_directory: ~/expo-dev-tools/packages/expo-codemod
+          name: 'json-file tests'
+          working_directory: ~/expo-cli/packages/json-file
+          command: yarn test
```
ide: do you think it would be useful to add `yarn build` to check for type errors too? (the babel preset just converts TS but doesn't do typechecking)

packages/json-file/src/JsonFile.ts
fson: What do you think about prefixing generic types with `T`, for example `TJSONObject`? This makes it really easy to distinguish them from concrete types, yet allows us to use descriptive names.

.circleci/config.yml
fson: This already happens as a part of the [bootstrap script](https://github.com/expo/expo-cli/blob/4079500ae3edcec5b153013aec2ee3841b399173/package.json#L7-L8) because all our packages include build as the `prepare` script which we run post bootstrap.

packages/babel-preset-cli/index.js
fson: Good catch, removed the `@babel/plugin-proposal-object-rest-spread` transform.

`@babel/plugin-proposal-class-properties` is still needed – they're not transformed when only using `@babel/preset-typescript`.

fson:
@ide thanks for a thoughtful review!

packages/json-file/src/JsonFile.ts
ide: I think that's fine, too.

Issue #457, opened 2019-03-22
Drop Node 6.x support after April 30th 2019


fson:
Expo CLI supports [Current Release, Active LTS and Maintenance LTS versions of Node.js.](https://github.com/nodejs/Release#release-schedule)

Node 6.x is coming to end-of-life on 30th of April 2019. After this it will no longer be supported by Expo CLI. After April, we will continue to support Expo CLI running on following Node.js versions:
- Current Release (12.x, to be released in April)
- Active LTS (10.x "Dubnium")
- Maintenance LTS (8.x "Carbon")

byCedric:
Yes, please! 😄 

I see that `opn@5.4.0` is used in the `expo-cli` package. Sindre actually renamed it now to `open`, because the original `open` package is deprecated:

> Note: The original open package was recently deprecated in favor of this package, and we got the name, so this package is now named open instead of opn. If you're upgrading from the original open package (open@0.0.5 or lower), keep in mind that the API is different.

If you want to update, you'll have to switch to `open@6.x.x`. Which has dropped support for Node 6 already, as [you can see here](https://github.com/sindresorhus/open/commit/5c525b5f4f2472e35ace19d6df746bb355b8060a).

fson:
This was done in #643! We now support Node v8, v10 and v12.

PR #643, opened 2019-05-22
Drop Node v6 support

fson:
See https://github.com/expo/expo-cli/issues/457.

### DIFF
--- a/packages/expo-cli/bin/expo.js
+++ b/packages/expo-cli/bin/expo.js
@@ -1,22 +1,13 @@
 #!/usr/bin/env node
-'use strict';
-
 var getenv = require('getenv');
-// validate that used node version is supported
 var semver = require('semver');
-var ver = process.versions.node;
-ver = ver.split('-')[0]; // explode and truncate tag from version
+var version = process.versions.node;
 
-if (semver.satisfies(ver, '>=6.0.0')) {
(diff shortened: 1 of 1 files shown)


hugo-marques-m3:
I understand why this was done but why was this released under a patch version?

fson:
Thanks for the question. You are right that this should have been a major version bump.

The reason why this sometimes happens is that right now we don’t have a good system in place to track which changes are breaking in this repo. We have many people in our team who contribute to this repository and every team member is allowed to publish new versions. This allows bug fixes to go out quickly, without having to wait for another team member to cut the release. On the other hand, it’s hard to keep track of all the changes, when the best you can do is look at all the commits since the previous version and try to figure out if they are breaking changes. It’s really time consuming (when you are just trying to get that bug fix to users) and it’s easy to miss a breaking change.

That said, we have already identified this problem and there’s work in progress (see https://github.com/expo/expo-cli/pull/602) to adopt [Conventional Commits](https://www.conventionalcommits.org) as the commit message format. With some additional tooling, this will allow us to more easily keep track of breaking changes, automatically bump major versions when there are some and also to keep a changelog of changes (which currently is missing from the releases).

hugo-marques-m3:
Thanks for the reply and I'm glad there's already work in progress to prevent this from happening again. Keep up the good work :) 👍

PR #649, opened 2019-05-24
feat: add ExpoInAppPurchases module to xdl config

sergeichestakov:
As per step one in [this guide](https://github.com/expo/expo/blob/master/guides/Creating%20Universal%20Modules.md#integrate-with-the-native-projects), it seems I have to add the IAP module to the xdl module config in order for it to show up as a dependency in my Podfile and be usable within Expo apps via the Expo client.

Not sure if these instructions are fully up to date (some of the Android ones were a little outdated), but figured this was still a necessary prerequisite. Also, if you guys have any suggestions for how I can get started developing IAP on iOS quickly please let me know. But as of now, it seems this is required for the module to even be exposed when importing the native module as such:
```
import { NativeModulesProxy } from '@unimodules/core';
export default NativeModulesProxy.ExpoInAppPurchases; // this is null on iOS
```

### DIFF
--- a/packages/xdl/src/modules/config.js
+++ b/packages/xdl/src/modules/config.js
@@ -328,6 +328,19 @@ const expoSdkUniversalModules = [
     sdkVersions: '>=33.0.0',
   },
   {
+    podName: 'EXInAppPurchases',
+    libName: 'expo-in-app-purchases',
+    sdkVersions: '>=33.0.0',
+    config: {
+      ios: {
+        includeInExpoClient: false,
+      },
+      android: {
+        includeInExpoClient: false,
+      },
+    },
+  },
+  {
     libName: 'expo-intent-launcher',
     sdkVersions: '>=33.0.0',
     config: {


sergeichestakov:
Additionally, when I run `./generate-files-ios.js` in `tools-public`, I get the following output:
```
generate-dynamic-macros is falling back to `template-files/keys.json`
Downloading DEV published kernel manifest...
Generating local kernel manifest from project root /Users/sergeichestakov/Documents/expo/home and url exp://10.1.10.42:80...
Unable to generate manifest from /Users/sergeichestakov/Documents/expo/home: Failed to fetch manifest from www
Using manifest: {"android":{"package":"host.exp.exponent"},"dependencies":["@expo/react-native-action-sheet","@expo/react-native-fade-in-image","@expo/react-native-touchable-native-feedback-safe","apollo-boost","apollo-cache-inmemory","dedent","es6-error","expo","expo-analytics-amplitude","expo-asset","expo-barcode-scanner","expo-camera","expo-constants","expo-font","expo-linear-gradient","expo-location","expo-permissions","expo-task-manager","expo-web-browser","graphql","graphql-tag","immutable","lodash","prop-types","querystring","react","react-apollo","react-native","react-native-infinite-scroll-view","react-native-maps","react-navigation","react-navigation-material-bottom-tabs","react-redux","redux","redux-thunk","sha1","url"],"description":"","extra":{"amplitudeApiKey":"081e5ec53f869b440b225d5e40ec73f9"},"icon":"https://s3.amazonaws.com/exp-brand-assets/ExponentEmptyManifest_192.png","iconUrl":"https://s3.amazonaws.com/exp-brand-assets/ExponentEmptyManifest_192.png","ios":{"bundleIdentifier":"host.exp.exponent","supportsTablet":true},"locales":{},"name":"expo-home","orientation":"portrait","packagerOpts":{"config":"metro.config.js"},"platforms":["ios","android"],"primaryColor":"#cccccc","privacy":"unlisted","scheme":"exp","sdkVersion":"33.0.0","slug":"expo-home-dev-b6e367a03cf168a5e3017ab494233dc4e1f065f7","updates":{"checkAutomatically":"ON_LOAD","fallbackToCacheTimeout":0},"version":"33.0.0","id":"@expo-home-dev/expo-home-dev-b6e367a03cf168a5e3017ab494233dc4e1f065f7","revisionId":"33.0.0-r.01jqp3B3JO","publishedTime":"2019-05-20T01:01:50.085Z","commitTime":"2019-05-20T01:01:50.130Z","bundleUrl":"https://d1wp6m56sqw74a.cloudfront.net/%40expo-home-dev%2Fexpo-home-dev-b6e367a03cf168a5e3017ab494233dc4e1f065f7%2F33.0.0%2Fd0ff0c5e95b38eb9c5e525fc5299c1c7-33.0.0-ios.js","releaseChannel":"default","hostUri":"expo.io/@expo-home-dev/expo-home-dev-b6e367a03cf168a5e3017ab494233dc4e1f065f7"}
```

Not sure why I'm getting `Failed to fetch manifest from www` and where the url `exp://10.1.10.42:80` comes from. Do I need to be serving something locally when running this command? I couldn't find any instructions detailing that.

sjchmiela:
If I remember correctly, this list is kinda deprecated in favor of detecting unimodules by `react-native-unimodules`. This list would be only be used for `tools/publish-packages.js`. If `NativeModules.ExpoInAppPurchases` is `null` on iOS, I would debug with breakpoints if the constructor of the module is run when starting the app (this would also require the module to be installed with CocoaPods in the project, so autodetected by the scripts).

The `Unable to generate manifest from <HOME PATH>` is ok. When we generate dynamic macros, the script first tries to download the manifest of Home from your running `expo start` instance (in case you would like Expo Client to run it instead of published dev Home). For this, it tries to infer the URL from the configuration (`.expo`) of `home/`. Since the packager is not running, XDL's `UrlUtils.constructManifestUrlAsync` falls back to [`${IP}`](https://github.com/expo/expo-cli/blob/fe979915ba4f8a7f39db911ee285c1899fb8f777/packages/xdl/src/UrlUtils.js#L257)[`:80`](https://github.com/expo/expo-cli/blob/fe979915ba4f8a7f39db911ee285c1899fb8f777/packages/xdl/src/UrlUtils.js#L298-L303), under which no packager is running, so it throws an error. Then, as a fallback, `dev-home-config.json` file is used to get the URL of the Home app that should be used when running Expo Client (unless `EXBuildConstants.DEV_KERNEL_SOURCE === "LOCAL"`, when the script will fail since you've explicitly decided you would like to use a local instance of Home app and still it couldn't be found).

sergeichestakov:
Nevermind, I got it to work by running `pod install` in the `ios/` directory and making sure I was exporting the module correctly. Does this mean this step is unnecessary? Won't I have to do this eventually if I wish to publish this as an npm package? 

I believe that's the end goal right? So that people will run `expo install expo-in-app-purchases` (or via yarn/npm).

sjchmiela:
Yes yes, it's still required, just wanted to explain 1. the error message shown when generating macros, 2. why this list is/was needed 🙂

sergeichestakov:
Ok great! Is this good to merge then or do I need to add anything?

packages/xdl/src/modules/config.js
```diff
@@ -327,6 +327,19 @@ const expoSdkUniversalModules = [
...
+  {
+    podName: 'EXInAppPurchases',
+    libName: 'expo-in-app-purchases',
+    sdkVersions: '>=33.0.0',
+    config: {
+      ios: {
+        includeInExpoClient: false,
+      },
+      android: {
+        includeInExpoClient: false,
```
sjchmiela: If I remember correctly, we add `expo-payments-stripe` to Expo Client on Android. Do we not include it due to some external restriction, or module restriction, or do we want to have feature parity between Clients?

sergeichestakov: Yeah it seems we do. I know that we won't be able to include it on iOS because it is not likely to get approved for the app store. I had assume it would be the case on Android but perhaps not. Perhaps @ide can chime in?

ide: IAPs require configuration on Google Play / App Store Connect. Since we are the owners of `host.exp.Exponent`, only we can configure IAPs for the Expo client -- it's not useful for developers to test their own IAPs.

PR #653, opened 2019-05-28
BREAKING CHANGE: convert @expo/config to TypeScript and remove support for exp.json

fson:
- Use TypeScript.
- Remove support for supplying configuration in `exp.json` or the `exp` property in `package.json`. These configuration formats have been deprecated for a long time and using them has resulted in an error that explained how to upgrade to `app.json` (which has been the default for more than two years).

Note: for the return type of the `app.json` config object we currently resort to `any`. An improvement over this could possibly be achieved by generating TypeScript types from the JSON Schema declarations we have for `app.json`. However, it might also require us to perform runtime validation of `app.json` any time we read it, to make it possible to narrow the types down.


packages/config/src/Config.ts
```diff
@@ -38,15 +35,15 @@ const DEFAULT_PREFER_RELATED_APPLICATIONS = true;
 export async function addPlatform(
   projectRoot: string,
   platform: 'ios' | 'android' | 'web'
-): Promise<{ exp: ?Object, pkg: ?Object, rootConfig: ?Object }> {
-  const { exp } = await readConfigJsonAsync(projectRoot);
+): Promise<{ exp: any; pkg: any; rootConfig: any }> {
```
ide: Instead of "any", how about "object" or perhaps an aliased type (`type ExpoConfig = object;`)?

packages/config/src/Config.ts
```diff
@@ -108,49 +105,26 @@ function _findConfigPath(projectRoot: string) {
...
     configPath = await _findConfigPathAsync(projectRoot);
   }
-  return getConfigPaths(configPath);
+  return { configPath, configName: APP_JSON_FILE_NAME, configNamespace: 'expo' };
 }
 
 export function findConfigFile(
   projectRoot: string
-): Promise<{ configPath: string, configName: string, configNamespace: ?string }> {
+): { configPath: string; configName: string; configNamespace: string | null } {
```
ide:
```suggestion
): { configPath: string; configName: string; configNamespace: 'expo' } {
```

packages/config/src/Config.ts
```diff
@@ -228,11 +193,16 @@ function sanitizePublicPath(publicPath) {
   return publicPath + '/';
 }
 
-export function getConfigForPWA(projectRoot: string, getAbsolutePath: Function, options: Object) {
+export function getConfigForPWA(
+  projectRoot: string,
+  getAbsolutePath: (...pathComponents: string[]) => string,
+  options: Object
```
ide:
```suggestion
  options: object
```

packages/config/src/Config.ts
```diff
@@ -473,7 +453,11 @@ function inferWebStartupImages(config: Object = {}, getAbsolutePath: Function, o
   return startupImages;
 }
 
-export function ensurePWAConfig(appJSON: Object, getAbsolutePath: Function, options: Object) {
+export function ensurePWAConfig(
+  appJSON: Object,
```
ide:
```suggestion
  appJSON: object,
```

packages/config/src/Config.ts
```diff
@@ -473,7 +453,11 @@ function inferWebStartupImages(config: Object = {}, getAbsolutePath: Function, o
...
+  getAbsolutePath: ((...pathComponents: string[]) => string) | undefined,
+  options: Object
```
ide:
```suggestion
  options: object
```

ide:
Left some small comments that should be easy to fix up.

PR #667, opened 2019-06-03
improvement: apple-mobile-web-app-capable meta tag

pixtron:
The meta tag should be `apple-mobile-web-app-capable` and not `mobile-web-app-capable`

### DIFF
--- a/packages/webpack-pwa-manifest-plugin/src/createMetatagsFromConfig.js
+++ b/packages/webpack-pwa-manifest-plugin/src/createMetatagsFromConfig.js
@@ -41,7 +41,7 @@ export default function createMetatagsFromConfig(config) {
     // Disable automatic phone number detection.
     'format-detection': apple.formatDetection,
     'apple-touch-fullscreen': apple.touchFullscreen,
-    'mobile-web-app-capable': apple.mobileWebAppCapable,
+    'apple-mobile-web-app-capable': apple.mobileWebAppCapable,
     'apple-mobile-web-app-status-bar-style': apple.barStyle,
     'apple-mobile-web-app-title': web.shortName,
   };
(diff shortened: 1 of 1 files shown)


packages/webpack-pwa-manifest-plugin/src/createMetatagsFromConfig.js
```diff
@@ -41,7 +41,7 @@ export default function createMetatagsFromConfig(config) {
     // Disable automatic phone number detection.
     'format-detection': apple.formatDetection,
     'apple-touch-fullscreen': apple.touchFullscreen,
-    'mobile-web-app-capable': apple.mobileWebAppCapable,
+    'apple-mobile-web-app-capable': apple.mobileWebAppCapable,
```
EvanBacon:
```suggestion
    // [Android Chrome] To enable multi-resolution icons of size [196x196]
    'mobile-web-app-capable': apple.mobileWebAppCapable,
```

EvanBacon: This should probably be moved out of the `apple` config and into some Android specific field.

ide: (This is in addition to keeping apple-mobile-web-app-capable, which we want in the apple config; we want both.)

pixtron: I wasn't aware that `mobile-web-app-capable` was introduced by chrome/android. I added it again in a new `android` field, if the field is undefined, it will fall back to `apple.mobileWebAppCapable`, so the change should not break anything.

The updated PR changes the following things:

1. Add the `apple-` prefix for iOS devices.
2. Move `mobile-web-app-capable` to own android specific field.

EvanBacon:
This form of meta-tag generation seems really static now, I think it'd be better if we pulled all of this and just made a guide on using react-helmet or something with Expo web.

PR #678, opened 2019-06-07
[xdl] Support Android App Bundles

wkozyra95:


packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1074,50 +1134,16 @@ async function buildShellAppAsync(context: StandaloneContext, sdkVersion: string
...
-          'prodKernel',
-          'release',
-          'app-prodKernel-release.apk'
-        );
-      }
-      await fs.copy(apkPath, 'shell.apk');
+      await fs.copy(outputPath, shellFile);
       // -c means "only verify"
-      await spawnAsync(`zipalign`, ['-c', '-v', '4', 'shell.apk'], {
+      await spawnAsync(`zipalign`, ['-c', '-v', '4', shellFile], {
```
sjchmiela: Does this also apply to AAB?

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1040,24 +1041,83 @@ export async function runShellAppModificationsAsync(
...
+  buildType: string
+) {
   let shellPath = shellPathForContext(context);
+  const ext = buildType === 'app-bundle' ? 'aab' : 'apk';
 
-  if (context.build.android) {
-    let androidBuildConfiguration = context.build.android;
+  const isRelease = !!context.build.android;
+  // concat on those strings is not very readable, but only alternative here is huge if statement
+  const debugOrRelease = isRelease ? 'Release' : 'Debug';
```
sjchmiela: This could be called `buildType` https://developer.android.com/studio/build/build-variants

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1040,24 +1041,83 @@ export async function runShellAppModificationsAsync(
...
+  const devOrProd = isRelease ? 'Prod' : 'Dev';
```
sjchmiela: And this `legacyFlavor`

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1040,24 +1041,83 @@ export async function runShellAppModificationsAsync(
...
+  const debugOrReleaseL = isRelease ? 'release' : 'debug';
```
sjchmiela: and we could uppercase them on demand in the code

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1040,24 +1041,83 @@ export async function runShellAppModificationsAsync(
...
+
+  let gradleBuildCommand;
+  let outputPath;
+  if (buildType === 'app-bundle') {
+    if (ExponentTools.parseSdkMajorVersion(sdkVersion) >= 33) {
+      gradleBuildCommand = `bundle${debugOrRelease}`;
+      outputPath = path.join(outputDirPath, debugOrReleaseL, `app.aab`);
+    } else if (ExponentTools.parseSdkMajorVersion(sdkVersion) >= 32) {
+      gradleBuildCommand = `bundle${devOrProd}Kernel${debugOrRelease}`;
+      outputPath = path.join(outputDirPath, `${devOrProdL}Kernel${debugOrRelease}`, `app.aab`);
```
sjchmiela: So we do support SDK32?

packages/xdl/src/detach/AndroidShellApp.js
sjchmiela: Then `buildType` would have to become productType? resultType?

packages/xdl/src/detach/AndroidShellApp.js
wkozyra95: it's quite unreadable as it is, additional logic at this point will make this worse

packages/xdl/src/detach/AndroidShellApp.js
wkozyra95: aab is also a zip file

packages/xdl/src/detach/AndroidShellApp.js
wkozyra95: yes 32 and 33 for app bundles (this has nothing to do width 64bit support)
older sdk should also work, but there were some problems with the build and I don't think it's worth investigating given that it will need to be deprecated when 64 bit builds will be required.

packages/xdl/src/detach/AndroidShellApp.js
wkozyra95: the only clean way I see here is to write all the combination but after adding bundle support this number doubled

packages/expo-cli/src/commands/build/index.js
```diff
@@ -86,8 +86,14 @@ export default (program: any) => {
     .option('--keystore-path <app.jks>', 'Path to your Keystore.')
     .option('--keystore-alias <alias>', 'Keystore Alias')
     .option('--public-url <url>', 'The URL of an externally hosted manifest (for self-hosted apps)')
+    .option(
+      '-t --type <build>',
+      'Type of build: [app-bundle|apk].',
+      /^(app-bundle|apk)$/i,
+      'app-bundle'
```
sjchmiela: Is this the new default?

wkozyra95: yes

wkozyra95: old cli will build apk by default

PR #686, opened 2019-06-09
Account for webpack config being async

EvanBacon:
Custom configs could be async, this'll account for that possibility

### DIFF
--- a/packages/xdl/src/Web.js
+++ b/packages/xdl/src/Web.js
@@ -15,16 +15,16 @@ function invokePossibleFunction(objectOrMethod, ...args) {
   }
 }
 
-export function invokeWebpackConfig(env, argv) {
+export async function invokeWebpackConfigAsync(env, argv) {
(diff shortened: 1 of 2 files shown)


packages/xdl/src/Web.js
```diff
@@ -15,16 +15,16 @@ function invokePossibleFunction(objectOrMethod, ...args) {
...
 }
 
-export function invokeWebpackConfig(env, argv) {
+export async function invokeWebpackConfigAsync(env, argv) {
   // Check if the project has a webpack.config.js in the root.
   const projectWebpackConfig = path.resolve(env.projectRoot, 'webpack.config.js');
   if (fs.existsSync(projectWebpackConfig)) {
     const webpackConfig = require(projectWebpackConfig);
-    return invokePossibleFunction(webpackConfig, env, argv);
+    return await invokePossibleFunction(webpackConfig, env, argv);
```
EvanBacon: wrong

PR #693, opened 2019-06-10
[xdl] exclude expo-face-detector and expo-payments from being used on ejected apps by default

tsapeta:
In `use_unimodules` method, `expo-face-detector` and `expo-payments` should be excluded by default when ejecting an app.

### DIFF
--- a/packages/xdl/src/detach/IosPodsTools.js
+++ b/packages/xdl/src/detach/IosPodsTools.js
@@ -381,7 +381,7 @@ require_relative '../node_modules/react-native-unimodules/cocoapods.rb'
 use_unimodules!(
   exclude: [
     'expo-face-detector',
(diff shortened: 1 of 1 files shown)


packages/xdl/src/detach/IosPodsTools.js
```diff
@@ -378,7 +378,12 @@ function _renderUnversionedUniversalModulesDependencies(universalModules, sdkVer
       `
 # Install unimodules
 require_relative '../node_modules/react-native-unimodules/cocoapods.rb'
-use_unimodules!`,
+use_unimodules!(
+  exclude: [
+    'expo-face-detector',
+    'expo-payments',
```
sjchmiela:
```suggestion
    'expo-payments-stripe',
```

PR #714, opened 2019-06-12
Replace Jimp with sharp

EvanBacon:
Reduce PWA image generation (`expo build:web`) time by ~60 seconds


sergeichestakov:
>
>@EvanBacon: *opens PR at 8pm, asks 3 people for review*
>
>One hour later: *well I guess since no one's responding it's good to merge* 

Seems legit

PR #720, opened 2019-06-12
[xdl] Fix iOS shell app when using unimodules autolinking

tsapeta:
# Why

It turned out shell app for SDK33 was not using autolinking because it used an old version of xdl. Unfortunately, I found some issues with using them there - mostly related to the fact that shell app workspace doesn't look like a common project workspace.

# How

- Replaced `shellAppWorkspaces/ios/default` with `shellAppWorkspaces/default/ios`
  I suppose `default` was going to be some kind of a flavor, but having it inside `ios` folder makes the difference in paths when compared to a common projects (paths were different in shell app and when ejecting). 
- Added `UNIVERSAL_MODULES_PATH` to podfile substitutions so we know where to look for unimodules.
- Temporarily duplicated installing `packagesToInstallWhenEjecting` when creating iOS workspace. This is needed to install react-native-unimodules. I used word _temporarily_ because I'd like to refactor creating a shell app on iOS. Its process is overcomplicated imho. 

# Test Plan

Already tested on expo's CI - I've successfully built iOS shell app with these changes.

### DIFF
--- a/packages/xdl/src/detach/Detach.js
+++ b/packages/xdl/src/detach/Detach.js
@@ -381,7 +381,7 @@ async function prepareDetachedServiceContextIosAsync(projectDir: string, args: a
   // path generated inside IosShellApp. When we support more than one path, this needs to
   // be smarter.
   const expoRootDir = path.join(projectDir, '..', '..');
-  const workspaceSourcePath = path.join(projectDir, 'default');
+  const workspaceSourcePath = path.join(projectDir, 'ios');
   const buildFlags = StandaloneBuildFlags.createIos('Release', { workspaceSourcePath });
   const context = StandaloneContext.createServiceContext(
     expoRootDir,
--- a/packages/xdl/src/detach/IosShellApp.js
+++ b/packages/xdl/src/detach/IosShellApp.js
@@ -164,7 +164,7 @@ async function _createStandaloneContextAsync(args) {
   if (args.workspacePath) {
     workspaceSourcePath = args.workspacePath;
   } else {
-    workspaceSourcePath = path.join(expoSourcePath, '..', 'shellAppWorkspaces', 'ios', 'default');
+    workspaceSourcePath = path.join(expoSourcePath, '..', 'shellAppWorkspaces', 'default', 'ios');
   }
   let { privateConfigFile, privateConfigData } = args;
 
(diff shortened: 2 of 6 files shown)


packages/xdl/src/detach/IosShellApp.js
```diff
@@ -164,7 +164,7 @@ async function _createStandaloneContextAsync(args) {
   if (args.workspacePath) {
     workspaceSourcePath = args.workspacePath;
   } else {
-    workspaceSourcePath = path.join(expoSourcePath, '..', 'shellAppWorkspaces', 'ios', 'default');
+    workspaceSourcePath = path.join(expoSourcePath, '..', 'shellAppWorkspaces', 'default', 'ios');
```
sjchmiela: I hope nothing else depends on this path…

packages/xdl/src/detach/IosWorkspace.js
```diff
@@ -175,26 +176,27 @@ async function _renderPodfileFromTemplateAsync(
...
-      ...module,
-      path: path.join(expoDependenciesPath, module.libName, module.subdirectory),
-    }));
   } else if (context.type === 'service') {
     reactNativeDependencyPath = path.join(
       expoRootTemplateDirectory,
       'react-native-lab',
       'react-native'
     );
+    modulesPath = path.join(expoRootTemplateDirectory, 'packages');
```
sjchmiela: 😢

packages/xdl/src/detach/IosWorkspace.js
```diff
@@ -309,6 +326,25 @@ async function createDetachedAsync(context: StandaloneContext) {
   }
 }
 
+async function _getPackagesToInstallWhenEjecting(sdkVersion) {
+  const versions = await Versions.versionsAsync();
+  return versions.sdkVersions[sdkVersion].packagesToInstallWhenEjecting;
+}
+
+async function _installRequiredPackagesAsync(projectRoot, sdkVersion) {
```
sjchmiela: `// copied from …` so we remember this code is duplicated?

packages/xdl/src/detach/Detach.js
```diff
@@ -381,7 +381,7 @@ async function prepareDetachedServiceContextIosAsync(projectDir: string, args: a
   // path generated inside IosShellApp. When we support more than one path, this needs to
   // be smarter.
   const expoRootDir = path.join(projectDir, '..', '..');
-  const workspaceSourcePath = path.join(projectDir, 'default');
+  const workspaceSourcePath = path.join(projectDir, 'ios');
```
sjchmiela: Don't we have to do some `if (version >= 33)` here and later on?

tsapeta: I believe we don't, this can run only in a service context (CI), that means it's used only when creating a shell app, which are always built with unversioned SDK 🙂

PR #722, opened 2019-06-12
Implement Auto-Lock on Inactive & Closed Issues

cruzach:
Once an issue has been closed, and is inactive for a certain period of time (in this case- 60 days, but we can change/update that), it should be locked. This results in less people commenting "I experience this bug, too" after the issue has been closed, and instead they will open up a new issue with more information and a repro

Main things to get feedback on-

- how long before we lock issues? I've chosen 60 days but that's pretty long
- adding the label outdated right now upon locking, any objections? Also see lock comment

Note: I've created a different config file for `expo/expo` repo and `expo/expo-cli` repo, but can always just extend one to the other. I just thought it might be nice to be able to customize each

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,37 @@
+# Configuration for Lock Threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 60
+
+# Skip issues and pull requests created before a given timestamp. Timestamp must
+# follow ISO 8601 (`YYYY-MM-DD`). Set to `false` to disable
+skipCreatedBefore: false
+
+# Issues and pull requests with these labels will be ignored. Set to `[]` to disable
+exemptLabels: []
+
+# Label to add before locking, such as `outdated`. Set to `false` to disable
+lockLabel: outdated
+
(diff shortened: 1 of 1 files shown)


.github/lock.yml
```diff
@@ -0,0 +1,37 @@
...
+# issues:
+#   exemptLabels:
+#     - help-wanted
+#   lockLabel: outdated
+
+# pulls:
+#   daysUntilLock: 30
+
+# Repository to extend settings from
+# _extends: repo
```
ide: What do you think about extending from `expo/expo` and using all of the same settings?

cruzach: Fine with me. If they need to be fine tuned, we can always change it later. I'll make the change

.github/lock.yml
```diff
@@ -0,0 +1,34 @@
...
+# issues:
+#   exemptLabels:
+#     - help-wanted
+#   lockLabel: outdated
+
+# pulls:
+#   daysUntilLock: 30
+
+# Repository to extend settings from
+_extends: expo
```
ide: Since we're extending from `expo` -- delete the rest of the fields unless they're intentional overrides

Qwerty33Iam (commented):
avatar

PR #729, opened 2019-06-13
Added expo-device package

EvanBacon:


packages/xdl/src/modules/config.ts
```diff
@@ -309,6 +309,11 @@ const expoUniversalModules = [
     libName: 'expo-gl',
     sdkVersions: '>=29.0.0',
   },
+  {
+    podName: 'EXDeviceInfo',
+    libName: 'expo-device-info',
```
ide: Just name it `expo-device`

EvanBacon: CC: @ninjaachibi @vivianzzhu91

ninjaachibi: > Just name it `expo-device`

got em

tsapeta:
@ninjaachibi, just fyi: when we're adding new modules to xdl config, the corresponding PR on `expo/expo` repo needs to be merged first. Right now some scripts are failing because they use this config and try to find `expo-device` in `packages` folder but it isn't there yet.

ninjaachibi:
Gotcha, thanks @tsapeta!

PR #744, opened 2019-06-19
[expo-cli][traveling-fastlane] enable push notifs from apple portal

quinlanj:
# why
- in order to make push notifications work for adhoc builds, you need to go into the Apple Development Portal into the `Identifiers` tab, and check the `Push Notifications` box in order to get push notifications working.
- this is very frustrating and not very intuitive at all, and we should enable it for Adhoc builds automatically for users. Maybe we should consider doing it for all standalone builds too (afaik we dont currently do this?)

### DIFF
--- a/packages/traveling-fastlane/actions/ensure_app_exists.rb
+++ b/packages/traveling-fastlane/actions/ensure_app_exists.rb
@@ -22,7 +22,7 @@ def ensure_app_exists()
   if app == nil
     {created: true, app: Spaceship::Portal.app.create!(bundle_id: $bundleId, name: $name)}
   else
-    {created: false , app: app}
+    {created: false, app: app}
   end
 end
 
--- a/packages/traveling-fastlane/actions/ensure_app_exists.rb
+++ b/packages/traveling-fastlane/actions/ensure_app_exists.rb
@@ -20,7 +20,7 @@ ENV['FASTLANE_TEAM_ID'] = $teamId
 def ensure_app_exists()
   app = Spaceship::Portal.app.find $bundleId
   if app == nil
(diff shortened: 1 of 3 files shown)


packages/traveling-fastlane/actions/ensure_app_exists.rb
```diff
@@ -1,18 +1,35 @@
 require 'spaceship'
 require 'json'
 require 'base64'
+require 'optparse'
 require_relative 'funcs'
 
+options = {}
+OptionParser.new do |opts|
```
ide: Thanks for setting this up

packages/traveling-fastlane/actions/ensure_app_exists.rb
```diff
@@ -1,18 +1,35 @@
...
+
 $appleId, $password, $teamId, $bundleId, $name = ARGV
 ENV['FASTLANE_TEAM_ID'] = $teamId
 
 def ensure_app_exists()
   app = Spaceship::Portal.app.find $bundleId
   if app == nil
-    Spaceship::Portal.app.create!(bundle_id: $bundleId, name: $name)
-    true
+    {created: true , app: Spaceship::Portal.app.create!(bundle_id: $bundleId, name: $name)}
```
ide:
```suggestion
    {created: true, app: Spaceship::Portal.app.create!(bundle_id: $bundleId, name: $name)}
```

packages/traveling-fastlane/actions/ensure_app_exists.rb
```diff
@@ -1,18 +1,35 @@
...
   else
-    false
+    {created: false , app: app}
```
ide:
```suggestion
    {created: false, app: app}
```

PR #747, opened 2019-06-20
[expo-cli] adhoc builds: grab available app.json, improve disable services info text

quinlanj:
# Why

- Google maps doesnt work on adhoc build clients
- Normally when a user makes a standalone build, they declare `ios.config.googleMapsApiKey` in their `app.json`. Regular Expo clients have the Expo googleMapsApiKey, but we are not going to use this since we are going to constrain it to the expo client bundleIdentifier soon.

# How
- when a user runs `expo client:ios`, we try to find a valid `app.json` in their cwd or from a passed in `--config` flag. 
- we thread this through to turtle so it knows to use the custom ios configuration 

# Related PRs
https://github.com/expo/expo-cli/pull/747
https://github.com/expo/expo/pull/4674
https://github.com/expo/universe/pull/3589

# Test Plan

- custom adhoc build now works with Google Maps

# Deploy

- [ ] Run postgres migrations in prod

### DIFF
--- a/packages/expo-cli/src/commands/client/index.js
+++ b/packages/expo-cli/src/commands/client/index.js
@@ -41,7 +41,7 @@ export default program => {
       };
 
       // get custom project manifest if it exists
-      // Note: this is some random user's project, NOT the expo client manifest
+      // Note: this is the current developer's project, NOT the Expo client's manifest
       const spinner = ora(`Finding custom configuration for the Expo client...`).start();
       const appJsonPath = options.config || path.join(projectDir, 'app.json');
       const appJsonExists = await ConfigUtils.fileExistsAsync(appJsonPath);
--- a/packages/expo-cli/src/commands/client/index.js
+++ b/packages/expo-cli/src/commands/client/index.js
@@ -1,5 +1,9 @@
 import chalk from 'chalk';
+import idx from 'idx';
+import ora from 'ora';
+import path from 'path';
 import CliTable from 'cli-table';
+import * as ConfigUtils from '@expo/config';
 import { Android, Simulator, User, Credentials } from '@expo/xdl';
 
 import CommandError from '../../CommandError';
@@ -19,7 +23,7 @@ const { IOS } = PLATFORMS;
 
 export default program => {
   program
-    .command('client:ios')
+    .command('client:ios [project-dir]')
     .option(
       '--apple-id <login>',
       'Apple ID username (please also set the Apple ID password as EXPO_APPLE_PASSWORD environment variable).'
@@ -27,7 +31,31 @@ export default program => {
     .description(
       'Build a custom version of the Expo Client for iOS using your own Apple credentials and install it on your mobile device using Safari.'
     )
-    .asyncAction(async options => {
+    .asyncActionProjectDir(async (projectDir, options) => {
+      const servicesDisabled = {
+        push_notifications:
+          'not yet available until API tokens are supported for the Push Notification system',
+      };
+
+      // get custom project manifest if it exists
+      // Note: this is some random user's project, NOT the expo client manifest
+      const spinner = ora(`Finding a valid app.json..`).start();
+      const appJsonPath = options.config || path.join(projectDir, 'app.json');
+      const appJsonExists = await ConfigUtils.fileExistsAsync(appJsonPath);
+      const { exp } = appJsonExists ? await ConfigUtils.readConfigJsonAsync(projectDir) : {};
+
+      if (exp) {
+        spinner.succeed(`Found app.json at ${appJsonPath}`);
+      } else {
+        spinner.warn(`Unable to find app.json.`);
+      }
+      if (!idx(exp, _ => _.ios.config.googleMapsApiKey)) {
+        const disabledReason = exp
+          ? 'ios.config.googleMapsApiKey does not exist in app.json'
+          : 'no app.json could be found';
+        servicesDisabled.google_maps = disabledReason;
+      }
+
       const authData = await appleApi.authenticate(options);
       const user = await User.getCurrentUserAsync();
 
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/client/index.js
```diff
@@ -19,15 +23,39 @@ const { IOS } = PLATFORMS;
...
+      const appJsonPath = options.config || path.join(projectDir, 'app.json');
+      const appJsonExists = await ConfigUtils.fileExistsAsync(appJsonPath);
+      const { exp } = appJsonExists ? await ConfigUtils.readConfigJsonAsync(projectDir) : {};
+
+      if (exp) {
+        spinner.succeed(`Found app.json at ${appJsonPath}`);
+      } else {
+        spinner.warn(`Unable to find app.json.`);
+      }
+      if (!idx(exp, _ => _.ios.config.googleMapsApiKey)) {
```
dsokal:
```suggestion
      if (!has(exp, _ => _.ios.config.googleMapsApiKey)) {
```

Also, remember to import `has` from lodash above: `import has from 'lodash/has';`

packages/expo-cli/src/commands/client/index.js
```diff
@@ -19,15 +23,39 @@ const { IOS } = PLATFORMS;
...
+      // Note: this is some random user's project, NOT the expo client manifest
+      const spinner = ora(`Finding a valid app.json..`).start();
+      const appJsonPath = options.config || path.join(projectDir, 'app.json');
+      const appJsonExists = await ConfigUtils.fileExistsAsync(appJsonPath);
+      const { exp } = appJsonExists ? await ConfigUtils.readConfigJsonAsync(projectDir) : {};
+
+      if (exp) {
+        spinner.succeed(`Found app.json at ${appJsonPath}`);
+      } else {
+        spinner.warn(`Unable to find app.json.`);
```
dsokal: Hmmm, I'm afraid this message may be misleading. I know that you updated Expo docs - https://github.com/expo/expo/pull/4674/files#diff-220abbc3ea5d5ad390fa92a6a9372a28R153 - but I still feel that users will be confused about which `app.json` we were unable to find. Maybe we should write here that if a user wishes to use google maps then he should create an `app.json` file for his ad-hoc build and fill that in with his `googleMapsApiKey`.

packages/expo-cli/src/commands/client/index.js
```diff
@@ -19,15 +23,39 @@ const { IOS } = PLATFORMS;
...
       '--apple-id <login>',
       'Apple ID username (please also set the Apple ID password as EXPO_APPLE_PASSWORD environment variable).'
     )
     .description(
       'Build a custom version of the Expo Client for iOS using your own Apple credentials and install it on your mobile device using Safari.'
     )
-    .asyncAction(async options => {
+    .asyncActionProjectDir(async (projectDir, options) => {
+      const servicesDisabled = {
+        push_notifications:
```
dsokal:
```suggestion
        pushNotifications:
```

packages/expo-cli/src/commands/client/index.js
```diff
@@ -19,15 +23,39 @@ const { IOS } = PLATFORMS;
...
+      const appJsonPath = options.config || path.join(projectDir, 'app.json');
+      const appJsonExists = await ConfigUtils.fileExistsAsync(appJsonPath);
+      const { exp } = appJsonExists ? await ConfigUtils.readConfigJsonAsync(projectDir) : {};
+
+      if (exp) {
+        spinner.succeed(`Found custom configuration for the Expo client at ${appJsonPath}`);
+      } else {
+        spinner.warn(`Unable to find custom configuration for the Expo client.`);
+      }
+      if (!_.has(exp, _ => _.ios.config.googleMapsApiKey)) {
```
dsokal:
```suggestion
      if (!_.has(exp, 'ios.config.googleMapsApiKey')) {
```

dsokal: Screenshot 2019-06-26 at 09 47 26
;)

fson:
```suggestion
      if (!idx(exp, _ => _.ios.config.googleMapsApiKey)) {
```
(I think this can be better typechecked by TypeScript compiler than the stringy `_.has` syntax.)

packages/expo-cli/src/commands/client/index.js
```diff
@@ -19,15 +23,39 @@ const { IOS } = PLATFORMS;
...
-    .asyncAction(async options => {
+    .asyncActionProjectDir(async (projectDir, options) => {
+      const servicesDisabled = {
+        push_notifications:
+          'not yet available until API tokens are supported for the Push Notification system',
+      };
+
+      // get custom project manifest if it exists
+      // Note: this is some random user's project, NOT the expo client manifest
+      const spinner = ora(`Finding a valid app.json..`).start();
```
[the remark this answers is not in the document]
quinlanj: i think u are commented on an outdated diff. i recently changed this to `Finding custom configuration for the Expo client...`

packages/expo-cli/src/commands/client/index.js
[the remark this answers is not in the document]
quinlanj: i've changed this to also search for `ios.config.googleApiKey`, but it doesnt look like we use `ios.config.googleApiKey` for any other configuration anywhere :/ Or did you mean we were eventually wanting to consolidate all the keys under a new `ios.config.googleApiKey` field?

packages/expo-cli/src/commands/client/index.js
quinlanj: @fson @dsokal we should decide if we want to prefer `idx` or `_.has`. I'm just going to keep `_.has` for now because that's what I've been changing it to across different pr's for consistency sake.

packages/expo-cli/src/commands/client/index.js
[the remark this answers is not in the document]
quinlanj: ok so from the slack discussion we are going to remove the check for `ios.config.googleApiKey` for now and look into falling back to a google services file if we cant find `googleMapsApiKey`

packages/expo-cli/src/commands/client/index.js
dsokal: In my opinion, the choice is obvious. It's better to have one npm package and not two.

quinlanj:
Addressing the unresolved comments for this PR here:

## Unifying google api keys
As outlined in https://github.com/expo/expo/issues/4736 , when we unify our google api keys, the following work will need to be done on adhoc builds:
- (minor) `expo client:ios` should detect if any google services are disabled by looking at googleMapsApiKey, googleApiKey then a googleServicesFile so the user can be notified. Currently it only looks for the presence of googleMapsApiKey. 
- (major) will need to import googleServicesFile into adhoc client in order for MLKit to work

Unifying google api keys shouldn't be a blocker for this PR, and any additional changes should be put in a separate PR.

## `idx` vs lodash `has/get`
Tradeoff here is that `idx` has a more robust check while using lodash `has/get` avoids using Yet Another Package. Are we ok just going with lodash here (i'm fine w either)?

ide:
- Google API Keys: let's do the cleanup in another PR. This PR doesn't make anything worse.
- idx vs. has: idx is probably better since there's no mini interpreted language but I don't think it matters a lot since both are crummy compared to a native operator that we might get next year: https://github.com/tc39/proposal-nullish-coalescing.

PR #761, opened 2019-06-25
feat(webpack-config): add abiliity to disable deep scope

EvanBacon:
Some libraries have too many side-effects to enable any kind of tree-shaking (like three.js). In this case we want to completely disable tree-shaking so you can push to prod.

Enable only in production mode:
```js
const createExpoWebpackConfig = require('@expo/webpack-config');
const { getMode } = require('@expo/webpack-config/utils');

module.exports = async function(env, argv) {
  const mode = getMode(env);
  const isProd = mode === 'production';

  const config = await createExpoWebpackConfig(
    {
      ...env,
      removeUnusedImportExports: isProd,
    },
    argv);
  return config;
};
```

### DIFF
--- a/packages/webpack-config/webpack/webpack.common.js
+++ b/packages/webpack-config/webpack/webpack.common.js
@@ -147,6 +147,14 @@ module.exports = async function(env = {}, argv) {
   const isDev = mode === 'development';
   const isProd = mode === 'production';
 
+  // Enables deep scope analysis in production mode.
+  // Remove unused import/exports
+  // override: `env.deepScopeAnalysis`
+  const deepScopeAnalysisEnabled = overrideWithPropertyOrConfig(
+    env.removeUnusedImportExports,
+    isProd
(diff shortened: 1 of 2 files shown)


packages/webpack-config/webpack/webpack.common.js
```diff
@@ -158,6 +158,14 @@ module.exports = async function(env = {}, argv) {
   const { noJavaScriptMessage } = config.web.dangerous;
   const noJSComponent = createNoJSComponent(noJavaScriptMessage);
 
+  if (
+    (isProd && config.web.build.deepScopeAnalysis !== false) ||
+    (isDev && config.web.build.deepScopeAnalysis != null)
```
fson: So for `--dev` builds, deep scope analysis is enabled even if you configure `
{ expo: web: { build: { deepScopeAnalysis: false } } } }`? That seems confusing.

PR #765, opened 2019-06-25
refactor(xdl): convert tools/ImageUtils.js and its dependencies in detach folder to TypeScript

fson:


packages/xdl/src/detach/Logger.ts
```diff
@@ -1,31 +1,53 @@
...
-    levels.forEach(level => {
-      this[level] = function(...args) {
-        this.logLine(level, ...args);
-      };
-    });
-  },
-  configure(loggerObj) {
+  }
+
+  configure(loggerObj: Console | bunyan) {
```
ide: add `void` return types to these methods

PR #769, opened 2019-06-25
chore: test on node 12.x

nicknovitski:
We should be able to support this and I've been worried about a few reports of problems.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,33 +1,59 @@
 version: 2.1
 
 executors:
-  node:
+  node-10:
     docker:
       - image: circleci/node:10
+  node-12:
+    docker:
+      - image: circleci/node:12
 workflows:
   version: 2
   build-and-test:
     jobs:
-      - build
+      - build:
+          name: build12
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,33 +1,59 @@
...
+            - build12
+      - test-json-file:
+          e: node-10
+          requires:
+            - build10
+      - test-expo-codemod:
+          e: node-12
           requires:
-            - build
+            - build12
```
fson: TIL CircleCI doesn't support build matrixes, bummer. 😕

Anyway, it's important that we do this, even if it introduces some accidental complexity to the CI configuration 👍

.circleci/config.yml
```diff
@@ -1,33 +1,59 @@
...
       - test-expo-codemod:
+          e: node-10
           requires:
-            - build
+            - build10
 jobs:
   build:
-    executor: node
+    parameters:
+      e:
```
fson:
```suggestion
      executor:
```

PR #775, opened 2019-06-26
feat(xdl): set deployment target for all pods

sjchmiela:
Having default deployment targets raises warnings in Xcode. Fixes a TODO left by Ben.


sjchmiela:
It looks like setting the deployment target inhibits warnings for the pod, we'll need to filter out unversioned development pods then.

sjchmiela:
Ok, it wasn't the case—cleaning and rebuilding shows all the warnings again. Ready for review.

PR #781, opened 2019-06-27
feat(expo-cli): add support for logging in with a session file

quinlanj:
Add support for logged in users to generate a session file so they can give it to others. Other people can then run `login` with a session flag and the session file path to log in as that user.

# why
so you dont need to give others your username and password to log into your account

# usage
```
  Usage: login|signin [options]

  Login with your Expo account

  Options:

    -u, --username [string]        Username
    -p, --password [string]        Password
    -g, --generate-session [path]  Generate user session file
    -s, --session <path>           Login with user session file
    -h, --help                     output usage information
```

# related prs
https://github.com/expo/universe/pull/3597


fson:
What are the intended use cases for this feature?

Sharing your authentication session, even if you're not handing out your password, is not something that should be done without serious consideration. Giving access to your session means handing out full read-write permissions to your user account. The person you give the session will be able to:
- push OTA updates to apps on your account
- delete your projects
- manage your subscriptions
- edit your settings
- change your password
- delete your account
- etc.

For many purposes this is quite broad. For this reason, many apps implement more granular authorization features and things like roles. This is also one of the reasons why things like organizations and teams are implemented: they allow teams to collaborate on shared projects in a more controlled way.

Because sharing your authentication session with other people gives full access to your account, also the session file, once generated, should kept safe. Sending it through any unencrypted channel can compromise your whole user account. The current implementation doesn't include any guidance on using this feature or warnings about the security implications.

Another possible concern is how making a shared user account (in contrast to separate user accounts and teams) an "official" workflow by adding this feature to the CLI might affect things like pricing paid services, etc., but I don't want to focus too much on that right now.

So the reason I'm asking about intended use cases is to figure out, if there's a more secure way to enable those use cases than crude "session hijacking". These alternatives might include things like:
- using/extending the team features @tcdavis has been working on. Great for things like:
  * allowing everyone in a team to publish the app they're working on
  * only allowing billing roles manage credit card info etc
  * only adding admin roles add more people to the team
- API tokens. Scoped access tokens that have a granular list of permissions granted to them. You can create/delete them on your settings page. Great for things like:
  * programmatic access to your account, e.g. scripts
  * continuous integration (CI)
  * integration with 3rd party tools

The best alternative depends on what problem we're trying to solve (we know some people share their passwords with other people, but why are they doing it?) so let's start with answering that question and then think about the best solution to that problem.

ide:
We talked this through in person and it seems like teams and API tokens meet the needs, but we'll need to do a lot of work to support API tokens for all the operations that accounts can do. We're first starting with just push notifications -- and for the shared-login use case that Quin detailed, we'll use teams.

quinlanj:
The original use-case for all this came from users who wanted to be able to generate a token/secret, give it to other coworkers out-of-band so they could use it to login to the account in order to perform `publish`/`build` operations from expo-cli.

@fson raised some really good points about the drawbacks of an all-powerful 'login-token'. Technically, anyone with the login token can change the account password or even delete the account itself. We came to the conclusion irl that tc's teams feature would let people do `publish`/`build` operations on the same project, so will be closing this PR :)

@fson raised some really good points about the drawbacks of an all-powerful 'login-token'. Technically, anyone with the login token can change the account password or even delete the account itself. We came to the conclusion irl that tc's teams feature would let people do `publish`/`build` operations on the same project, so will be closing this PR :)

PR #803, opened 2019-07-02
BREAKING CHANGE: make supported Node.js versions stricter

fson:
We support Node.js maintenance LTS, active LTS and current release versions.
Previously the supported version range was defined as 8.x.x || >=10.0.0,
but this included the 11.x.x versions that reached end-of-life on June 1st 2019.

Update the supported versions to include the exact ranges we support and also
make the error message clearly tell which versions are supported.

Note: the version ranges purposefully use the normal forms such as '>=8.9.0 <9.0.0'
instead of '^8.9.0' to make them easy to understand even without knowledge of the
intricacies of semver.

### DIFF
--- a/packages/expo-cli/bin/expo.js
+++ b/packages/expo-cli/bin/expo.js
@@ -1,13 +1,37 @@
 #!/usr/bin/env node
+var chalk = require('chalk');
 var getenv = require('getenv');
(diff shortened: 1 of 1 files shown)


fson:
Screenshot 2019-07-02 at 17 02 56

PR #804, opened 2019-07-02
refactor(xdl): convert more modules to TypeScript

fson:
* Android
* Api
* Exp
* Extract
* Session
* ThirdParty
* UrlUtils


packages/xdl/src/UserSettings.ts
```diff
@@ -50,7 +51,7 @@ function userSettingsJsonFile(): JsonFile<UserSettingsData> {
 function recentExpsJsonFile() {
   // TODO(ville): Add array support to JsonFile.
   // @ts-ignore JsonFile doesn't officially support arrays, only objects
```
ide: Is this comment and the one above still relevant?

packages/xdl/src/Api.ts
```diff
@@ -237,8 +252,14 @@ export default class ApiClient {
     return await ApiClient._schemaCaches[sdkVersion].getAsync();
   }
 
-  static async downloadAsync(url, outputPath, options = {}, progressFunction, retryFunction) {
-    if (options.extract) {
+  static async downloadAsync(
+    url: string,
+    outputPath: string,
+    { extract = false } = {},
```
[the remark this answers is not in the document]
fson: TypeScript infers `boolean` type from the default value, so annotation is not necessary.

packages/xdl/src/UserSettings.ts
fson: Yes, still valid. Type safe support for top level arrays requires some work because some object specific operations don’t make sense (e.g. set a string key, merge). But `read` and `write` happen to work and this code makes use of that.

PR #812, opened 2019-07-04
[xdl] Remove Expo client's API keys from AndroidManifest when preparing a shell app

tsapeta:
It turned out we don't remove Expo client's API keys from `AndroidManifest.xml` when corresponding keys are not provided by the user. With these changes, `xdl` will always delete Expo client's API keys from `AndroidManifest.xml`, especially when they are not provided by the user.

Fixes https://github.com/expo/expo/issues/4776


ide:
We need to double check that standalone apps actually use the user's Google Maps key -- we got a report that it always uses Expo's key.

PR #818, opened 2019-07-05
refactor(expo-cli): typescript support

wkozyra95:


packages/xdl/src/xdl.ts
```diff
@@ -20,8 +19,7 @@ export { Api };
 import ApiV2 from './ApiV2';
 export { ApiV2 };
 
-// @ts-ignore untyped module yet to be converted to TypeScript
-import * as AssetUtils from './AssetUtils.js';
+const AssetUtils: any = require('./AssetUtils.js');
```
wkozyra95: needs to be done this way because ts-ignore comments don't propagate to expo-cli and are interpreted as errors there

packages/expo-cli/src/exp.ts
```diff
@@ -144,7 +145,7 @@ Command.prototype.asyncActionProjectDir = function(asyncFn, skipProjectValidatio
       }
     };
 
-    const logStackTrace = (chunk, logFn, nestedLogFn) => {
+    const logStackTrace = (chunk: any, logFn: (...args: any[]) => void, nestedLogFn: (...args: any[]) => void) => {
```
fson:
```suggestion
    const logStackTrace = (chunk: LogRecord, logFn: (...args: any[]) => void, nestedLogFn: (...args: any[]) => void) => {
```

packages/xdl/src/logs/PackagerLogsStream.ts
```diff
@@ -396,7 +393,7 @@ export default class PackagerLogsStream {
       if (bundleComplete) {
         if (this._onFinishBuildBundle) {
           const error = msg.type === 'bundle_build_failed' ? 'Build failed' : null;
-          this._onFinishBuildBundle(error, this._bundleBuildStart, bundleBuildEnd, progressChunk);
+          this._onFinishBuildBundle(error, this._bundleBuildStart || new Date(0), new Date(), progressChunk);
```
fson: Not sure about `new Date(0)`... it's 1st of January 1970. This might cause some funny errors if we ever calculate the duration from those dates and this happens. "Build took 49 years ..." etc.

packages/xdl/src/xdl.ts
fson: I just merged #813, which adds types to AssetUtils so you should be able to remove this particular one.

PR #819, opened 2019-07-05
refactor(xdl): convert Web, Webpack and createWebpackCompiler to TS

fson:


packages/xdl/src/Web.ts
```diff
@@ -53,21 +54,45 @@ function applyEnvironmentVariables(config) {
...
     // Prevent minimizing when running in debug mode.
-    config.optimization.minimize = false;
+    set(config, ['optimization', 'minimize'], false);
   }
 
   return config;
 }
 
-export async function invokeWebpackConfigAsync(env, argv) {
+type WebEnvironment = {
```
EvanBacon: Move to top?

packages/xdl/src/Web.ts
```diff
@@ -53,21 +54,45 @@ function applyEnvironmentVariables(config) {
...
+    ) => Promise<webpack.Configuration> | webpack.Configuration);
+
+export async function invokeWebpackConfigAsync(
+  env: WebEnvironment,
+  argv?: string[]
+): Promise<webpack.Configuration> {
   // Check if the project has a webpack.config.js in the root.
   const projectWebpackConfig = path.resolve(env.projectRoot, 'webpack.config.js');
-  let config;
+  let config: webpack.Configuration;
```
EvanBacon: https://github.com/expo/expo-cli/blob/c162e9de9885a456c7c3a30ca7b315c9f6d525aa/packages/webpack-config/src/types.ts#L2-L10

EvanBacon: `WebpackConfiguration | DevConfiguration`

fson:
@EvanBacon could you check this PR once more? I have now rebased this branch, fixed all the conflicts and TypeScript errors. However, I'm not 100% sure I got all the types right. In particular, I was a bit confused by the multiple conversions done to options passed to `bundleAsync`:
- `transformCLIOptions` converts CLI options to `BundlingOptions`
- `getWebpackConfigEnvFromBundlingOptionsAsync` converts `BundlingOptions` to `WebEnvironment`
Maybe we don't need the intermediate step? Could we refactor this to create the "env" object directly from CLI options? Or maybe we could parse CLI options on the Expo CLI side so things like parsing boolean options don't leak into XDL to complicate the webpack interface code.

EvanBacon:
@fson should be fine to compressed. Would be good to have unit tests to ensure the expected transforms are passed.

PR #823, opened 2019-07-06
[web] Remove deep scope until further notice

EvanBacon:
Something has broken in the prod bundling. This will get expo web back to working.
Also added better tools for debugging production bundles when using `EXPO_WEB_DEBUG=true`.


LouisJS:
Need this fix to be released ! 😁

PR #831, opened 2019-07-09
Fixed TerminalUI and GraphQLSchema

EvanBacon:


packages/dev-tools/server/graphql/GraphQLSchema.js
```diff
@@ -14,7 +14,7 @@ import {
   ProjectUtils,
   UrlUtils,
   UserSettings,
-  User,
+  UserManager,
```
ide: sort above UserSettings

PR #833, opened 2019-07-11
refactor(web): Make Webpack more readable

EvanBacon:
- Removed unused props
- Surface more functionality
- Reuse config logic across starting and bundling webpack
- Deprecated unused verbose prop


EvanBacon:
It'd be nice to add tests to this module as well

PR #844, opened 2019-07-15
[web][push] Add supports for push notification in Expo for Web

hesyifei:
- In CLI, added supports for uploading (`push:web:upload`), generating (`push:web:generate`), showing (`push:web:show`), and clearing (`push:web:clear`) VAPID key pair and VAPID subject.
- In service worker, added supports for displaying push notifications as well as sending push notifications payload to the client (the webpage) after clicking the notification.

TODO:
- Some options in `GenerateSW` are not transformed into `serviceWorker.js`.

### DIFF
--- a/packages/expo-cli/src/commands/push-creds.js
+++ b/packages/expo-cli/src/commands/push-creds.js
@@ -75,15 +75,15 @@ export default (program: any) => {
       log('All done!');
     }, true);
 
+  const vapidSubjectDescription =
+    'URL or `mailto:` URL which provides a point of contact in case the push service needs to contact the message sender.';
+
   program
     .command('push:web:upload [project-dir]')
-    .description('Uploads a VAPID key for web push notifications.')
+    .description('Uploads VAPID key pair and VAPID subject for web push notifications.')
     .option('--vapid-pubkey [vapid-public-key]', 'URL-safe base64-encoded VAPID public key.')
     .option('--vapid-pvtkey [vapid-private-key]', 'URL-safe base64-encoded VAPID private key.')
-    .option(
-      '--vapid-subject [vapid-subject]',
-      'URL or `mailto:` URL which provides a point of contact in case the push service needs to contact the message sender.'
-    )
+    .option('--vapid-subject [vapid-subject]', vapidSubjectDescription)
     .asyncActionProjectDir(async (projectDir, options) => {
       if (!options.vapidPubkey || !options.vapidPvtkey || !options.vapidSubject) {
         throw new Error(
@@ -111,6 +111,40 @@ export default (program: any) => {
       });
 
       log('All done!');
+    }, true);
+
+  program
+    .command('push:web:generate [project-dir]')
+    .description('Generates VAPID key pair for web push notifications.')
+    .option('--vapid-subject [vapid-subject]', vapidSubjectDescription)
+    .asyncActionProjectDir(async (projectDir, options) => {
+      if (!options.vapidSubject) {
+        throw new Error('Must specify --vapid-subject.');
+      }
+
+      log('Reading project configuration...');
+
+      const {
+        args: { remotePackageName },
+      } = await Exp.getPublishInfoAsync(projectDir);
+
+      log('Logging in...');
+
+      let user = await UserManager.getCurrentUserAsync();
+      let apiClient = ApiV2.clientForUser(user);
+
+      log("Generating and setting VAPID keys on Expo's servers...");
+
+      let results = await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
+        vapidSubject: options.vapidSubject,
+      });
+
+      log('All done!');
+      log('Your VAPID public key is: ' + results.vapidPublicKey);
+      log('Your VAPID private key is: ' + results.vapidPrivateKey);
+      log(
+        "Don't forget to also add the generated public key to your `app.json` file! Learn more here: https://docs.expo.io/versions/latest/guides/using-vapid/"
+      );
     }, true);
 
   program
--- a/packages/expo-cli/src/commands/push-creds.js
+++ b/packages/expo-cli/src/commands/push-creds.js
@@ -102,7 +102,7 @@ export default (program: any) => {
       let user = await UserManager.getCurrentUserAsync();
       let apiClient = ApiV2.clientForUser(user);
 
-      log("Setting VAPID key on Expo's servers...");
+      log("Setting VAPID keys on Expo's servers...");
 
       await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
         vapidPublicKey: options.vapidPubkey,
--- a/packages/expo-cli/src/commands/push-creds.js
+++ b/packages/expo-cli/src/commands/push-creds.js
@@ -112,4 +112,54 @@ export default (program: any) => {
 
       log('All done!');
     }, true);
+
+  program
+    .command('push:web:show [project-dir]')
+    .description(
+      'Prints the VAPID public key, the VAPID private key, and the VAPID subject currently in use for web notifications for this project.'
+    )
+    .asyncActionProjectDir(async (projectDir, options) => {
+      const {
+        args: { remotePackageName },
+      } = await Exp.getPublishInfoAsync(projectDir);
+      let user = await UserManager.getCurrentUserAsync();
+      let apiClient = ApiV2.clientForUser(user);
+
+      let result = await apiClient.getAsync(`credentials/push/web/${remotePackageName}`);
+
+      if (
+        result.status === 'ok' &&
+        result.vapidPublicKey &&
+        result.vapidPrivateKey &&
+        result.vapidSubject
+      ) {
+        log(JSON.stringify(result));
+      } else if (result.status === 'error') {
+        throw new Error('Server returned an error: ' + result.error);
+      } else {
+        throw new Error('Server returned an invalid result!');
+      }
+    }, true);
+
+  program
+    .command('push:web:clear [project-dir]')
+    .description(
+      'Deletes previously uploaded VAPID public key, VAPID private key, and VAPID subject.'
+    )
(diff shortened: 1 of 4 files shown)


jaulz:
Great work. Instead of calling it `custom-service-worker.js` it might be better to call the workbox's service worker `workbox.js` and this one `service-worker.js` wouldn't it?
Also we would need to make `Notifications.addListener` etc. web compatible and send a message to the client once a notification was received. I see that you were already working on something like that or where does `notificationIcon` come from? https://github.com/expo/expo-cli/blob/be3788fc152133be8d72941cf52937d09240c28e/packages/webpack-config/web-default/custom-service-worker.js#L7-L11

hesyifei:
@jaulz Thanks for the comment!

I have initially named the file `service-worker.js` and let `WorkboxPlugin.GenerateSW` generate `workbox.js`. However, the problem with that is in development mode, `noopServiceWorkerMiddleware()` is called, which always override the `service-worker.js` file. Then it would not be possible to test push notifications in development mode. If we instead use `custom-service-worker.js` and then import `service-worker.js`, we can keep `noopServiceWorkerMiddleware()` while still being able to test push notifications in development mode.

https://github.com/expo/expo-cli/blob/4a69b374b512f9a771d4560dd125b175131e0280/packages/webpack-config/webpack/createDevServerConfigAsync.js#L70-L75

https://github.com/facebook/create-react-app/blob/100e12d22f35a75120fe79d9ef89646717ccdfe8/packages/react-dev-utils/noopServiceWorkerMiddleware.js#L11-L15

---

`Notifications.addListener` is also supported, but it is from the `expo` package as a side-effect, because we need to use the functions that is provided in that package and we cannot access those function directly from service worker.

https://github.com/expo/expo/blob/cb874e50ada1f6b41051948453a7ee4e007600ed/packages/expo/src/Notifications/ExponentNotifications.fx.web.ts#L1-L9

---

`notificationIcon` also comes from `expo` package during `_subscribeUserToPushAsync()` which is called by `getExponentPushTokenAsync()`:

https://github.com/expo/expo/blob/cb874e50ada1f6b41051948453a7ee4e007600ed/packages/expo/src/Notifications/ExponentNotifications.web.ts#L187-L192

Thanks! :)

packages/webpack-config/web-default/custom-service-worker.js
```diff
@@ -0,0 +1,78 @@
+/* global self */
+
+/**
+ * Store notification icon string in service worker.
+ * Ref: https://stackoverflow.com/a/35729334/2603230
+ */
+self.addEventListener('message', function(event) {
```
esamelson:
```suggestion
self.addEventListener('message', event => {
```
for consistency with your other listeners

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+
+      log('Reading project configuration...');
+
+      const {
+        args: { remotePackageName },
+      } = await Exp.getPublishInfoAsync(projectDir);
+
+      log('Logging in...');
+
+      let user = await UserManager.getCurrentUserAsync();
```
EvanBacon:
```suggestion
      const user = await UserManager.getCurrentUserAsync();
```
Same for all

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+      let apiClient = ApiV2.clientForUser(user);
+
+      log("Setting VAPID keys on Expo's servers...");
```
EvanBacon:
```suggestion
      log('Setting VAPID keys on Expo's servers...');
```

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+      let apiClient = ApiV2.clientForUser(user);
+
+      log("Generating and setting VAPID keys on Expo's servers...");
+
+      let results = await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
+        vapidSubject: options.vapidSubject,
+      });
+
+      log('All done!');
+      log('Your VAPID public key is: ' + results.vapidPublicKey);
```
EvanBacon:
```suggestion
      log(`Your VAPID public key is: ${results.vapidPublicKey}`);
```

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+      log('Your VAPID private key is: ' + results.vapidPrivateKey);
+      log(
+        "Don't forget to also add the generated public key to your `app.json` file! Learn more here: https://docs.expo.io/versions/latest/guides/using-vapid/"
```
EvanBacon:
```suggestion
        'Don't forget to also add the generated public key to your `app.json` file! Learn more here: https://docs.expo.io/versions/latest/guides/using-vapid/'
```

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+      const {
+        args: { remotePackageName },
+      } = await Exp.getPublishInfoAsync(projectDir);
+
+      log('Logging in...');
+
+      let user = await UserManager.getCurrentUserAsync();
+      let apiClient = ApiV2.clientForUser(user);
+
+      log("Generating and setting VAPID keys on Expo's servers...");
```
EvanBacon:
```suggestion
      log('Generating and setting VAPID keys on Expo's servers...');
```

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+
+      let results = await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
```
EvanBacon:
```suggestion
      const results = await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
```

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+
+      if (
+        result.status === 'ok' &&
+        result.vapidPublicKey &&
+        result.vapidPrivateKey &&
+        result.vapidSubject
+      ) {
+        log(JSON.stringify(result));
+      } else if (result.status === 'error') {
+        throw new Error('Server returned an error: ' + result.error);
```
EvanBacon:
```suggestion
        throw new Error(`Server returned an error: ${result.error}`);
```

packages/webpack-config/web-default/index.html
```diff
@@ -57,7 +57,7 @@
     <script>
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', function() {
-          navigator.serviceWorker.register('/service-worker.js');
+          navigator.serviceWorker.register('/custom-service-worker.js');
```
EvanBacon:
```suggestion
          navigator.serviceWorker.register('/expo-service-worker.js');
```

packages/expo-cli/src/commands/push-creds.js
hesyifei: There's `Expo's servers` in there, the `'` forces us to use double quotes instead of single quotes.

packages/expo-cli/src/commands/push-creds.js
```diff
@@ -72,6 +72,128 @@ export default (program: any) => {
...
+
+      const results = await apiClient.putAsync(`credentials/push/web/${remotePackageName}`, {
+        vapidSubject: options.vapidSubject,
+      });
+
+      log('All done!');
+      log(`Your VAPID public key is: ${results.vapidPublicKey}`);
+      log(`Your VAPID private key is: ${results.vapidPrivateKey}`);
+      log(
+        "Don't forget to also add the generated public key to your `app.json` file! Learn more here: https://docs.expo.io/versions/latest/guides/using-vapid/"
```
esamelson: one thought -- it might be nice to just automatically modify the app.json file to have this key. we do this in a few other places, e.g. in `eject` we add the bundleIdentifier/packageName after prompting for it, and it's always nice to save the developer little busywork steps where we can.

jaulz:
Nice to see such good progress! What just came to my mind: if we make the expo service worker as the main service worker, then workbox might not be able to pick up updates because the expo service worker file will rarely change. Could you please check that? I think the workbox webpack plug in also has an inject functionality which could help here: https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#injectmanifest_plugin

hesyifei:
@jaulz, unfortunately, with `InjectManifest`, it has the same problem (`noopServiceWorkerMiddleware()` overrides the `/service-worker.js`) as above, plus it removes many useful options that user could pass in `GenerateSW`.

What you mentioned did use to be a concern (https://stackoverflow.com/q/38120092/2603230, https://blog.pushpad.xyz/2016/07/service-worker-importscripts-never-updates-scripts/). However, it seems to addressed https://developers.google.com/web/updates/2018/06/fresher-sw here. We can set `updateViaCache: 'none',` during `navigator.serviceWorker.register('expo-service-worker.js')` such that "the HTTP cache will not be consulted when making requests for either the top-level /expo-service-worker.js or for any imported scripted."

This does not seem to create any additional problem because even by default

> the HTTP cache will never be consulted when checking for updates to the /service-worker.js script, but will be consulted when fetching any imported scripts (path/to/import.js, in our example). This is the default, and it matches the behavior starting in Chrome 68.

So we are just making sure the browser will also not cache the workbox-generated `/service-worker.js`.

packages/expo-cli/src/commands/push-creds.js
hesyifei: I've added this as well as warnings. :)

PR #849, opened 2019-07-18
[web] fix: allow custom default port

jaulz:
This fix will allow you to use a custom port for the webpack dev server which is especially handy if you need a specific port in development (e.g. because of CORS).

### DIFF
--- a/packages/xdl/src/Webpack.js
+++ b/packages/xdl/src/Webpack.js
@@ -32,9 +32,9 @@ export function getServer(projectRoot: string) {
   return webpackDevServerInstance;
 }
 
-async function choosePortAsync(): Promise<number | null> {
+async function choosePortAsync(port:number): Promise<number | null> {
   try {
-    return await choosePort(HOST, DEFAULT_PORT);
+    return await choosePort(HOST, port);
   } catch (error) {
     throw new XDLError('NO_PORT_FOUND', 'No available port found: ' + error.message);
   }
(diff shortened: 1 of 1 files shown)


EvanBacon:
```js
const port = parseInt(process.env.PORT, 10) || 19006;
```
Seems more standard. Then we can support other tools like https://github.com/c9/core/blob/d8917f33fd7d32141d895fb7de530cfee2a416cf/configs/standalone.js#L15. 

Also not sure is `process.env.IP` or `process.env.HOST` is better for the host value.

packages/xdl/src/Webpack.js
```diff
@@ -18,8 +18,8 @@ import ip from './ip';
 
 import type { User as ExpUser } from './User'; //eslint-disable-line
 
-const HOST = '0.0.0.0';
-const DEFAULT_PORT = 19006;
+const HOST = process.env.IP || '0.0.0.0';
```
EvanBacon: If you add this functionality you'll need to account for the URL getter https://github.com/expo/expo-cli/blob/f5d61b6fcc87cadaefc578ee770185fcc69fabff/packages/xdl/src/Webpack.js#L117-L125. This may need to read from the running instance of Webpack to ensure the correct IP is being used. Currently it's hardcoded.

jaulz:
Good point! I added the PORT environment variable but still respect the port defined in the webpack dev  server config. In this case we will be flexible for any future requirements.

Furthermore I added the `PUBLIC_HOST` constant and adjusted the URL getter correspondingly.

PR #853, opened 2019-07-18
[web] fix `expo customize:web` on Windows

FLGMwt:
Before, `expo customize:web` failed on Windows with the following error:

```
ENOENT: no such file or directory, scandir ''
```

This was due to some pathing logic using `/` as a path delimiter, whereas Windows uses `\` .

This swaps out the `substring` for `path.dirname` which works properly on all platforms.

Tested on Windows and Linux (Ubuntu WSL).

|Before|After|
|---|---|
|image|image|


EvanBacon:
@estebanjb I don't understand why? https://github.com/expo/expo-cli/blob/3aca69c94398330538346996836f6816a710404b/packages/expo-cli/src/commands/customize.js#L6

Issue #850, opened 2019-07-18
[web] expo build fails with remote icons


jaulz:
## Description

Building a web application fails when icons in `app.json` are URLs.

## Expected Behavior

Remote icons should work the same as for iOS or Android. 

## Observed Behavior

The error occurs here:
https://github.com/expo/expo-cli/blob/4a69b374b512f9a771d4560dd125b175131e0280/packages/webpack-pwa-manifest-plugin/src/icons.js#L95

https://github.com/expo/expo-cli/blob/4a69b374b512f9a771d4560dd125b175131e0280/packages/image-utils/src/sharp.ts#L68-L71

`sharp-cli` tries to open the URL as a local file (e.g. /Users/jaulz/.../https://example.com/icon.png).

## Environment

```
npmPackages:
 expo: ^33.0.0 => 33.0.7
npmGlobalPackages: 
  expo-cli: 2.122
```

## Reproducible Demo

n/a

jaulz:
I could implement a fix for this but the question is in which package you would like it to be. In `image-util` or in `webpack-pwa-manifest-plugin`? The fix would be to download the image to a temporary folder if it's an URL and pass that one on.

PR #856, opened 2019-07-19
[web] fix: allow URLs for splash screens and icons

jaulz:
This PR will allow users to specify URLs for splash screens and icons. The images will be downloaded first and will then run through the regular processing.


jaulz:
Will fix https://github.com/expo/expo-cli/issues/850

packages/webpack-config/webpack/utils/getPathsAsync.js
```diff
@@ -31,6 +31,11 @@ module.exports = async function getPaths({ locations, projectRoot }) {
   const inputProjectRoot = projectRoot || appDirectory;
 
   function absolute(...pathComponents) {
+    // Simple check if we are dealing with an URL
+    if (pathComponents && pathComponents.length === 1 && pathComponents[0].startsWith('http')) {
```
EvanBacon: We should wait for #861 cuz this may break somethings.

jaulz: @EvanBacon could we consider this PR now after #861 was merged?

Issue #859, opened 2019-07-19
webpack-config has unlisted dependency on eslint / eslint-config-universe


FLGMwt:
## Description

`yarn web` fails on a new project with expo-cli@3 seemingly because recent versions of webpack-config rely on eslint and eslint-config-universe being available. When webpack-config is used by the CLI (instead of by the project directly), this would seem to require global installs of both packages.

## Expected Behavior

I expected `yarn web` to successfully build the starter app and that I wouldn't have to add any unlisted global dependencies.

## Observed Behavior

```
> yarn web
yarn run v1.17.3
$ expo start --web
Starting project at C:\git\black-screen-test
Expo DevTools is running at http://localhost:19002
Opening DevTools in the browser... (press shift-d to disable)
Warning: This project contains unoptimized assets. Please run `expo optimize` in your project directory.
Starting Metro Bundler on port 19001.
Tunnel ready.
Cannot find module 'eslint'
```

I expect users would encounter this issue if they did any of the following:

1. Created a new project with `expo init` using cli@3+
2. Had an existing project **without** a `webpack.config.js` and updated to cli@3+
3. Had an existing project **with** a `webpack.config.js` and a dependency on `@expo/webpack-config` and updated `@expo/webpack-config` to a recent version with the eslint loader.

## Environment

```
expo diagnostics

  Expo CLI 3.0.2 environment info:
    System:
      OS: Windows 10
    Binaries:
      Yarn: 1.17.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.4.0.0 AI-183.6156.11.34.5692245
```

## Reproducible Demo

Assuming:
- expo-cli@3.0.2 is globally installed
- eslint is NOT globally installed
- eslint-config-universe is NOT globally installed

```
git clone git@github.com:FLGMwt/expo-cli-3-eslint-issue.git
cd expo-cli-3-eslint-issue
yarn install
yarn web
```

## Workaround

You can work around this issue by running `expo customize:web`, selecting webpack, then overriding the webpack-config version to one before the eslit config was added:

```
yarn add -D @expo/webpack-config@0.5.19
```

Alternatively, run the yarn / npm command above and create a `webpack.config.js` file in the root of your project with the following content:

```
const createExpoWebpackConfigAsync = require('@expo/webpack-config');

module.exports = async function(env, argv) {
[...]

PR #860, opened 2019-07-19
feat(webpack-config): Made ESLint experimental

EvanBacon:
- Added `eslint` to webpack-config
- Added `env.dangerouslyUseLinter`
- Fixed `index.html` template replacement error by reverting path


packages/webpack-config/package.json
```diff
@@ -31,7 +31,12 @@
     "copy-webpack-plugin": "^5.0.0",
     "css-loader": "^2.1.1",
     "deep-diff": "^1.0.2",
+    "eslint": "^6.0.1",
     "eslint-loader": "^2.2.1",
+    "eslint-plugin-babel": "^5.3.0",
+    "eslint-plugin-flowtype": "^3.12.1",
+    "eslint-plugin-import": "^2.18.1",
+    "eslint-plugin-prettier": "^3.1.0",
```
FLGMwt: Do we also need eslint-config-universe added here since that's the config we use? usage: https://github.com/expo/expo-cli/pull/860/files#diff-cc24e2c78b9d805246070e6c56fc13a6R451

Killer turnaround time on this >: D

packages/webpack-config/package.json
```diff
@@ -31,7 +31,12 @@
     "copy-webpack-plugin": "^5.0.0",
     "css-loader": "^2.1.1",
     "deep-diff": "^1.0.2",
+    "eslint": "^6.0.1",
     "eslint-loader": "^2.2.1",
+    "eslint-plugin-babel": "^5.3.0",
```
fson:
```suggestion
    "eslint-plugin-babel": "5.3.0",
```
Perhaps we should pin these versions? I think ESLint & plugins break very easily when new versions of them get published that can require upgrading other plugins or ESLint itself, or require some minor changes to the configuration. If we pin the version we can upgrade them ourself and test before publishing.

packages/webpack-config/webpack/webpack.common.js
```diff
@@ -436,7 +437,7 @@ module.exports = async function(env = {}, argv) {
         // First, run the linter.
         // It's important to do this before Babel processes the JS.
         // This will throw required errors and warnings during the build.
-        {
+        env.dangerouslyUseLinter && {
```
fson: This uses `eslint-config-universe` configuration by default – is it possible to override the configuration using an `.eslintrc` file and have you tested/documented that?

EvanBacon: will add this in the future, need to disable now so people stop getting #859

packages/webpack-config/package.json
EvanBacon: Good catch

EvanBacon:
fixes https://github.com/expo/expo/issues/5022

PR #862, opened 2019-07-20
fix(web): respect minify option

jaulz:
Currently, the minify option is not respected when you want to start Expo like this:
```
expo start --no-minify --no-dev --web
```


jaulz:
Not sure if I got everything since I am not 100% familiar with the code base and it's difficult catch everything 😊

packages/webpack-config/webpack/webpack.prod.js
```diff
@@ -30,7 +30,7 @@ module.exports = async function(env = {}, argv) {
         locations.absolute(info.absoluteResourcePath).replace(/\\/g, '/'),
     },
     optimization: {
-      minimize: true,
+      minimize: env.minify,
```
EvanBacon: Ensure the default value is `true` if it's undefined.

EvanBacon: This isn't minifying in dev mode.

jaulz:
@EvanBacon okay, I just added your feedback. Wouldn't that also mean that we could drop most of the stuff from `applyEnvironmentVariables`?

PR #863, opened 2019-07-22
fix(xdl): default splashScreen mode being wrongly assumed to be 'native' in splashScreen images check

bbarthec:


packages/xdl/src/Android.ts
```diff
@@ -364,7 +364,7 @@ export async function checkSplashScreenImages(projectDir: string): Promise<void>
   }
 
   const splashScreenMode =
-    _.get(exp, 'android.splash.resizeMode') || _.get(exp, 'splash.resizeMode');
+    _.get(exp, 'android.splash.resizeMode') || _.get(exp, 'splash.resizeMode') || 'contain';
```
fson:
```suggestion
    _.get(exp, 'android.splash.resizeMode') || _.get(exp, 'splash.resizeMode', 'contain'); 
```

PR #885, opened 2019-07-29
feat(webpack-config): add CspHtmlWebpackPlugin

jaulz:
This PR will add https://github.com/slackhq/csp-html-webpack-plugin to the webpack configuration. Unfortunately, I faced some issues with stylesheets so I excluded it explicitly in the config.


brentvatne:
hi @jaulz! i'm going to close this for now because i'm not convinced that it needs to be added. if you want to reply explain the purpose for it and why you think it belongs here in detail we can definitely reopen and reconsider. thanks :)

PR #889, opened 2019-07-30
fix: problem loading the requested app

fson:
Fix "ValidationError: "https" is not allowed" thrown when trying to load the manifest and change the manifest handler so that it logs all errors to make it possible to catch this kind of problems in the future.

Simulator Screen Shot - iPhone Xʀ - 2019-07-30 at 19 32 03
Screenshot 2019-07-30 at 19 32 32

### DIFF
--- a/packages/xdl/src/Project.js
+++ b/packages/xdl/src/Project.js
@@ -1792,7 +1792,7 @@ export async function startExpoServerAsync(projectRoot: string) {
         developerTool: Config.developerTool,
       });
     } catch (e) {
-      ProjectUtils.logDebug(projectRoot, 'expo', `Error in manifestHandler: ${e} ${e.stack}`);
+      ProjectUtils.logError(projectRoot, 'expo', e.stack);
       // 5xx = Server Error HTTP code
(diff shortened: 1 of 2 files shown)


packages/xdl/src/Project.js
```diff
@@ -1792,7 +1792,7 @@ export async function startExpoServerAsync(projectRoot: string) {
         developerTool: Config.developerTool,
       });
     } catch (e) {
-      ProjectUtils.logDebug(projectRoot, 'expo', `Error in manifestHandler: ${e} ${e.stack}`);
+      ProjectUtils.logError(projectRoot, 'expo', e.stack);
```
ide: It seems like it'd be helpful to know what part of Expo CLI had a problem when it prints an error to the console -- does it make sense to log some more information here?

fson: I removed the prefix because the log output was something like this:
```
Error in manifest handler: XDLError: Error: ...
```
The "what part of Expo CLI" information should be clear from the included stack trace.

PR #893, opened 2019-07-31
fix: detect git even if Xcode is not installed

wodin:
Don't guess whether git is installed by checking if Xcode is installed. Just try to use git and warn if it fails.

### Test plan

I've tested with `git` in the path, with `git` not in the path and with a non-executable `git` in the path.
If git is in the path and executable then it is used as expected.
If git is not in the path, the expected warning is printed.

Closes issue #905

### DIFF
--- a/packages/xdl/src/Exp.ts
+++ b/packages/xdl/src/Exp.ts
@@ -161,11 +161,6 @@ export async function extractTemplateAppAsync(
 }
 
 async function initGitRepoAsync(root: string) {
-  if (process.platform === 'darwin' && !Binaries.isXcodeInstalled()) {
-    Logger.global.warn(`Unable to initialize git repo. \`git\` not installed.`);
-    return;
-  }
-
   // let's see if we're in a git tree
   let insideGit = true;
   try {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Exp.ts
```diff
@@ -161,7 +162,7 @@ export async function extractTemplateAppAsync(
 }
 
 async function initGitRepoAsync(root: string) {
-  if (process.platform === 'darwin' && !Binaries.isXcodeInstalled()) {
+  if (process.platform === 'darwin' && !hasbin.sync('git')) {
```
fson: Thanks for fixing this issue!

I think we could remove this check and just try to call `git rev-parse` (see lines below). If that throws an error where `e.code` is `ENOENT` that means git is not installed and we can print a warning.

wodin: Sounds good to me. I was wondering why we should check explicitly on darwin but not on e.g. Linux.

wodin: I've updated the change. Please let me know if the way I'm checking for the ENOENT exception is OK.

fson:
Looks good to me! 🚢

PR #896, opened 2019-07-31
Added E2E tests for webpack-config

EvanBacon:
# Why

We can catch if a project actually builds and runs correctly with webpack-config.

# How

Added `jest-puppeteer` to run E2E tests on a basic project built with Expo.
- Divided unit tests from E2E tests in webpack-config

# Test Plan

- CI should build a puppeteer env
- CI should run E2E tests for building, and starting projects


packages/webpack-config/jest-puppeteer.config.js
```diff
@@ -0,0 +1,37 @@
...
+      headless: true,
+    };
+
+const config = {
+  start: {
+    url: 'https://localhost:19006',
+    launch,
+    server: {
+      command: `../expo-cli/bin/expo.js start tests/basic/ --web-only --non-interactive --https`,
+      port: 19006,
```
ide: Since jest-dev-server will wait until a server launches on this port, it'd be preferable to explicitly specify the port to Expo CLI so that we are guaranteed to get the same port every time.

packages/webpack-config/jest-puppeteer.config.js
```diff
@@ -0,0 +1,37 @@
...
+    launch,
+    server: {
+      command: `node jest/build-project.js tests/basic/ && serve tests/basic/web-build`,
+      // The default serve-cli port
+      port: 5000,
+      launchTimeout: 30000,
+      debug: true,
+    },
+  },
+}[process.env.EXPO_E2E_COMMAND];
```
ide: Add an assertion that this env var is defined and so is the config object:

```js
assert(process.env.EXPO_E2E_COMMAND, `EXPO_E2E_COMMAND must be defined`);
assert(config, `"${process.env.EXPO_E2E_COMMAND}" is not a valid E2E test`);
```

packages/webpack-config/jest-puppeteer.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+assert(process.env.EXPO_E2E_COMMAND, `EXPO_E2E_COMMAND must be defined`);
+assert(
+  config,
+  `"${process.env.EXPO_E2E_COMMAND}" is not a valid E2E test. Expected one of ${Object.keys(
+    config
+  ).join(', ')}`
+);
+process.env.WEB_PORT = config.server.port;
```
ide:
```suggestion
// Tell Expo CLI to use the same port on which the test runner expects there to be a server
process.env.WEB_PORT = config.server.port;
```

PR #897, opened 2019-07-31
Remove deprecated `@babel/polyfill`

EvanBacon:


ide:
What's the recommended way for a developer to add their own polyfills? Should they write a Babel config that extends babel-preset-expo and adds a polyfill plugin?

EvanBacon:
Close in favor of #1056

PR #917, opened 2019-08-05
fix: slow CLI startup time performance regression

fson:
When changing the build configurations to add support for TypeScript, the lazy loading of modules that was previously done using the `babel-transform-inline-imports-commonjs` package was removed from `@expo/xdl` and `expo-cli`:
- https://github.com/expo/expo-cli/commit/b45e9dbf724bc077b7fbaf5041961028c4ddf470#diff-77b4343b244f7776c55c7c643d13efe6
- https://github.com/expo/expo-cli/commit/e3c7af725987e17bac74d0e806c67b0691ca626b#diff-b80134f5b17e5b157842c17b91ec70f4

This affected startup time of the CLI negatively, causing every command to take 3-5 seconds to start.

This commit restores lazy loading of modules, now using the `lazy` option of `@babel/plugin-transform-modules-commonjs`.

Thanks to @ninjaachibi for his investigation and profiling that revealed the source of the slowdown was `require`! 🙌

### DIFF
--- a/packages/babel-preset-cli/index.js
+++ b/packages/babel-preset-cli/index.js
@@ -6,9 +6,13 @@ module.exports = () => ({
         targets: {
           node: '8.9.0',
         },
+        modules: false,
       },
     ],
     '@babel/preset-typescript',
   ],
-  plugins: ['@babel/plugin-proposal-class-properties'],
+  plugins: [
+    '@babel/plugin-proposal-class-properties',
+    ['@babel/plugin-transform-modules-commonjs', { lazy: source => true }],
+  ],
 });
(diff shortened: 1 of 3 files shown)


packages/babel-preset-cli/index.js
```diff
@@ -6,9 +6,13 @@ module.exports = () => ({
...
         },
+        modules: false,
       },
     ],
     '@babel/preset-typescript',
   ],
-  plugins: ['@babel/plugin-proposal-class-properties'],
+  plugins: [
+    '@babel/plugin-proposal-class-properties',
+    ['@babel/plugin-transform-modules-commonjs', { lazy: source => true }],
```
ide: would just `lazy: true` work?

fson: According to [`@babel/plugin-transform-modules-commonjs` documentation](https://babeljs.io/docs/en/next/babel-plugin-transform-modules-commonjs.html#lazy), when set to `true` it will not make local imports lazy.

PR #921, opened 2019-08-07
[webpack-config] Simplify the module include method

EvanBacon:


packages/webpack-config/webpack/loaders/createBabelLoaderAsync.js
```diff
@@ -75,18 +67,18 @@ module.exports = async function({
...
-          return inputPath;
+          return true;
         }
       }
       // Is inside the project and is not one of designated modules
-      if (inputPath.includes(ensuredProjectRoot)) {
+      if (inputPath.match(ensuredProjectRoot)) {
         for (const excluded of excludedRootPaths) {
-          if (inputPath.includes(excluded)) {
+          if (inputPath.match(excluded)) {
```
ide: this doesn't look like it should have changed

EvanBacon: matching against regex now

packages/webpack-config/webpack/loaders/createBabelLoaderAsync.js
```diff
@@ -75,18 +67,18 @@ module.exports = async function({
 
     include(inputPath) {
       for (const possibleModule of modules) {
-        if (inputPath.includes(possibleModule)) {
+        if (inputPath.match(possibleModule)) {
```
ide: `modules` has some strings and some regexes in it. it's probably going to be confusing that strings are converted to regexes automatically. I think the config will need to document more clearly how the `include` option is supposed to work:

```
module.exports = async function({
  /**
   * The webpack mode: `"production" | "development"`
   */
  mode,
  babelProjectRoot,
  include = [],
  verbose,
  ...options
} = {}) {
  const ensuredProjectRoot = await ensureRootAsync(babelProjectRoot);
  const modules = [...includeModulesThatContainPaths, ...include];
```

PR #926, opened 2019-08-09
feat(web): add support for Next.js with Expo for Web

hesyifei:
# Why

With Next.js, we can support server-side rendering and many other features Next.js provided.

# How

- With `_createNextJsConf` function, it grabs the user's `next.config.js` and add Expo related webpack config to it from `@expo/webpack-config/webpack/webpack.config.unimodules`.
- With `getProjectUseNextJsAsync`, it checks if the user set `web.useNextJs` to true in `app.json`.
- With `startNextJsAsync`, it uses `_createNextJsConf`'s config to start a `express` server where `next` handles every request.
  - In production mode, it will `bundleNextJsAsync` first before starting (see Next.js docs).
  - It means that in the production mode, reload is not supported and every time the user has to restart (`expo start --no-dev` again) to see any changes.
- With `bundleNextJsAsync`, it uses `_createNextJsConf`'s config to build the app with `next/dist/build` function.
- During both `start` and `build:web`, `.expo/_document.js` is generated (or overridden). If `pages/_document.js` does not exist, it will also be generated (but will not be overridden), which `import` then `export` `.expo/_document.js`.
- Zero-config building/bundling with Now.sh should be supported (haven't figure out how to test it yet) given that the user add `expo-cli` as a `devDependencies` and set `scripts.build` to `expo build:web` in `package.json`.
- See also https://github.com/expo/expo/pull/5275 and https://github.com/expo/universe/pull/3645.

# Test Plan

- Set `web.useNextJs` to `true` in `app.json`.
- Run `yarn add next` in the app directory.
- `yarn link` related packages with https://github.com/expo/expo/pull/5275.
- Add `_document.js` like this: https://github.com/zeit/next.js/blob/master/examples/with-react-native-web/pages/_document.js
- `[path]/expo-cli/node_modules/.bin/expo-cli start --web-only`


EvanBacon:
Can you add tests to `webpack-config` which ensures that these can start and build as expected. https://github.com/expo/expo-cli/blob/master/packages/webpack-config/jest-puppeteer.config.js

packages/xdl/src/Webpack.js
```diff
@@ -204,6 +239,12 @@ export async function getProjectNameAsync(projectRoot: string): Promise<string>
   return webName;
 }
 
+export async function getProjectUseNextJsAsync(projectRoot: string): Promise<boolean> {
+  const { exp } = await ProjectUtils.readConfigJsonAsync(projectRoot);
+  const { useNextJs } = exp.web || {};
```
EvanBacon: Let's change this to `exp.web.use: "nextjs"`

packages/xdl/src/Webpack.js
```diff
@@ -280,7 +322,13 @@ async function createWebpackConfigAsync(
 
   setMode(env.mode);
 
-  const config = await Web.invokeWebpackConfigAsync(env);
+  let config;
+  if (unimodules) {
+    const webpackConfigUnimodules = require('@expo/webpack-config/webpack/webpack.config.unimodules');
```
EvanBacon: Fix up `@expo/webpack-config/withUnimodules` (add async) and use that instead.

packages/xdl/src/Webpack.js
```diff
@@ -336,3 +384,115 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+      }
+      onFinished();
+    });
+  });
+}
+
+async function bundleNextJsAsync({ projectRoot, expoConfig }) {
+  let nextBuild;
+  try {
+    nextBuild = require(path.join(projectRoot, 'node_modules', 'next', 'dist', 'build')).default;
```
EvanBacon: Try to use `@expo/config`s `resolveModule` method instead

packages/xdl/src/Webpack.js
```diff
@@ -336,3 +384,115 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+
+function _createNextJsConf({ projectRoot, expoConfig }) {
+  let userNextConfigJs = {};
+  const userNextConfigJsPath = path.join(projectRoot, 'next.config.js');
+  if (fs.existsSync(userNextConfigJsPath)) {
+    userNextConfigJs = require(userNextConfigJsPath);
+  }
+
+  // `include` function is from https://github.com/expo/expo-cli/blob/3933f3d6ba65bffec2738ece71b62f2c284bd6e4/packages/webpack-config/webpack/loaders/createBabelLoaderAsync.js#L76-L96
+  const includeFunc = expoConfig.module.rules[1].include;
```
EvanBacon: Create a util method which attempts to find the loader by name and throws if it cannot be found.

packages/xdl/src/Webpack.ts
```diff
@@ -93,30 +105,51 @@ export async function startAsync(
     !!process.stdout.isTTY
   );
 
-  const server: WebpackDevServer = await new Promise(resolve => {
-    // Create a webpack compiler that is configured with custom messages.
-    const compiler = createWebpackCompiler({
+  let server: DevServer;
+  if (usingNextJs) {
+    server = await new Promise(resolve => {
```
fson:
```suggestion
    server = await startNextJsAsync({
```
Async functions should usually be called with `await` so that the code waits for the async function to execute. This also ensures errors get handled instead of becoming unhandled promise rejections.

Please also remove the `onFinished` callback and instead make the `startNextJsAsync` function wait until starting the server has finished. Async functions and callbacks are both programming techniques for asynchronous code. In this codebase we strongly prefer async functions (promises) over callbacks. You can use the `await new Promise(..)` construct inside `startNextJsAsync` to convert from callbacks to promises.

packages/xdl/src/Webpack.ts
```diff
@@ -296,7 +341,8 @@ function transformCLIOptions(options: CLIWebOptions): BundlingOptions {
 
 async function createWebpackConfigAsync(
   projectRoot: string,
-  options: CLIWebOptions = {}
+  options: CLIWebOptions = {},
+  unimodules: boolean = false
```
fson: It's usually best to avoid positional boolean arguments because they can lead in code that's harder to read because it's not possible to guess what each argument means from a function call like `createWebpackConfigAsync(projectRoot, options, true, false)`.

You can add a `unimodules` property to `options` instead.

packages/xdl/src/Webpack.ts
```diff
@@ -363,3 +415,133 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
     ...(options.webpackEnv || {}),
   };
 }
+
+async function startNextJsAsync({ projectRoot, port, dev, expoConfig, onFinished }) {
+  let next;
+  try {
+    const { exp } = await ProjectUtils.readConfigJsonAsync(projectRoot);
```
fson:
```suggestion
    const { exp } = await ConfigUtils.readConfigJsonAsync(projectRoot);
```
(`ProjectUtils.readConfigJsonAsync` will be deprecated soon.)

packages/xdl/src/Webpack.ts
```diff
@@ -363,3 +415,133 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+    next = require(ConfigUtils.resolveModule('next', projectRoot, exp));
+  } catch {
+    throw new XDLError('NEXTJS_NOT_INSTALLED', 'Next.js is not installed in your app.');
```
fson: We should include instructions for fixing this or a link to the relevant page in the documentation.

packages/xdl/src/Webpack.ts
```diff
@@ -363,3 +415,133 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+    dev,
+    dir: projectRoot,
+    conf: _createNextJsConf({
+      projectRoot,
+      expoConfig,
+    }),
+  });
+  const handle = app.getRequestHandler();
+
+  app.prepare().then(() => {
```
fson:
```suggestion
  await app.prepare();
```

packages/xdl/src/Webpack.ts
```diff
@@ -363,3 +415,133 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+    throw new XDLError(
+      'NEXTJS_NOT_INSTALLED',
+      'Next.js (or its build component) is not installed in your app.'
+    );
+  }
+
+  await nextBuild(projectRoot, _createNextJsConf({ projectRoot, expoConfig }));
+}
+
+function _createNextJsConf({ projectRoot, expoConfig }) {
```
fson:
```suggestion
function _createNextJsConfig({ projectRoot, expoConfig }) {
```
(In general, we avoid abbreviations in code, but `config` is so well established as an abbreviation of "configuration" that it's OK. `conf` is not. 😄)

packages/xdl/src/Webpack.ts
```diff
@@ -363,3 +415,133 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
...
+  if (fs.existsSync(userNextConfigJsPath)) {
+    userNextConfigJs = require(userNextConfigJsPath);
+  }
+
+  // `include` function is from https://github.com/expo/expo-cli/blob/3933f3d6ba65bffec2738ece71b62f2c284bd6e4/packages/webpack-config/webpack/loaders/createBabelLoaderAsync.js#L76-L96
+  const expoBabelLoader = _findBabelLoader(expoConfig.module.rules);
+  if (!expoBabelLoader) {
+    throw new XDLError(
+      'BABEL_LOADER_NOT_FOUND',
+      'Cannot find `babel-loader` generated by `@expo/webpack-config/withUnimodules`.'
```
fson: Include instructions how to fix this?

packages/xdl/src/Webpack.ts
```diff
@@ -501,7 +504,7 @@ function _createNextJsConfig({
 }: {
   projectRoot: string;
   expoWebpackConfig: Web.WebpackConfiguration;
-}) {
+}): any {
```
EvanBacon: fix type

5ervant:
Hi there! @hesyifei @EvanBacon @fson 

I'm about to implement Next.js on my new Expo app to support server-side rendering by following this new guide: [Using Next.js with Expo for Web](https://docs.expo.io/versions/latest/guides/using-nextjs/)

But the big question is: **Will Next.js pages with `react-navigation` also work on an Android or iOS device or you need to create a specific screen for a native device?**

Any reply would be appreciated !!! :slight_smile:

fson:
@5ervant please post your questions to https://forums.expo.io, it's more likely that you'll get an answer there! 🙂

PR #929, opened 2019-08-12
fix(expo-cli): use production mode in build:web

jaulz:
Currently, when running `build:web` the option `dev` is `undefined` by default. If for any reason `dev` is `true` in `.expo/settings.json` or the command is executed for the first time the project will subsequently be built in dev mode.


EvanBacon:
We've fully removed the ability to build in dev mode. This functionality has been replaced by  `process.env.EXPO_WEB_DEBUG=true`

jaulz:
Okay, thanks for the info!

PR #930, opened 2019-08-12
fix(xdl) - fix build command for xcode 11

wkozyra95:


packages/xdl/src/detach/IosShellApp.js
```diff
@@ -117,7 +117,7 @@ async function _buildAsync(
       `${projectName}.app`
     );
   } else if (type === 'archive') {
-    buildCmd += ` -sdk iphoneos -destination generic/platform=iOS archive -archivePath ${buildDest}/${projectName}.xcarchive CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO | xcpretty`;
+    buildCmd += ` -sdk iphoneos -destination generic/platform=iOS archive -archivePath ${buildDest}/${projectName}.xcarchive CODE_SIGNING_ALLOWED=NO | xcpretty`;
```
sjchmiela: How would this work on Xcode < 11? Would it?

sjchmiela: For future reference — this is used only to build a shell app. This functionality isn't used by other users/clients, so only our servers depend on it. It should be safe to merge as long as we only include this change on Xcode 10.3+ servers. (**It's probably a breaking change.**)

wkozyra95: I tested those changes on 10.1 and shellapp build worked correctly

PR #934, opened 2019-08-13
[xdl] Make simulator builds and Android APKs SDK-specific

tsapeta:
Instead of saving iOS simulator builds and Android APKs urls to the global config, we would like to store them at SDK-specific configuration - it is required for https://github.com/expo/expo/pull/5331.
I've also exported a few more methods and made it possible internally to install an app from any url and not only the latest one.

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -29,7 +29,7 @@ export function isPlatformSupported(): boolean {
   );
 }
 
-async function _getAdbOutputAsync(args: string[]): Promise<string> {
+export async function _getAdbOutputAsync(args: string[]): Promise<string> {
   await Binaries.addToPathAsync('adb');
 
   try {
@@ -111,7 +111,7 @@ function _apkCacheDirectory() {
   return dir;
 }
 
-async function _downloadApkAsync() {
+export async function _downloadApkAsync(url?: string) {
   let versions = await Versions.versionsAsync();
   let apkPath = path.join(_apkCacheDirectory(), `Exponent-${versions.androidVersion}.apk`);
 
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Android.ts
```diff
@@ -28,7 +28,7 @@ export function isPlatformSupported(): boolean {
   );
 }
 
-async function _getAdbOutputAsync(args: string[]): Promise<string> {
+export async function _getAdbOutputAsync(args: string[]): Promise<string> {
```
sjchmiela: Do we want to export `_privateFunctions`?

packages/xdl/src/tools/UpdateVersions.ts
```diff
@@ -71,8 +77,37 @@ export async function updateAndroidApk(s3Client: any, pathToApp: string, appVers
...
+async function updateClientUrlAndVersionAsync(
+  sdkVersion: string | undefined,
+  platform: 'ios' | 'android',
+  clientUrl: string,
+  appVersion: string
+) {
+  // Unfortunately it needs to be of `any` type to be able to change dynamic keys in the object.
+  const versions: any = await Versions.versionsAsync();
+
+  // Create new SDK version config if not there yet.
```
sjchmiela:
```suggestion
  // Create new SDK version config if it's not there yet.
```

packages/xdl/src/tools/UpdateVersions.ts
```diff
@@ -71,8 +77,37 @@ export async function updateAndroidApk(s3Client: any, pathToApp: string, appVers
...
+  if (sdkVersion && !versions.sdkVersions[sdkVersion]) {
+    versions.sdkVersions[sdkVersion] = {};
+  }
+
+  // For compatibility reason we have to maintain that global config, but only when we're updating the most recent SDK.
```
sjchmiela:
```suggestion
  // For compatibility reasons we have to maintain that global config, but only when we're updating the most recent SDK.
```

PR #938, opened 2019-08-15
[webpack-config] Implement public path

EvanBacon:
# Why

To make this work -> https://docs.expo.io/versions/latest/distribution/publishing-websites/#github-pages

# How

Added the ability to publish to a publicPath based on the `homepage` field of the `package.json`.

# Test Plan

1. `expo init`
2. Add URL
```diff
// package.json
+   "homepage": "http://evanbacon.github.io/gh-pages-expo-web-deployment-test",
```
3. `expo build:web && yarn gh-pages -d web-build`


cruzach:
Deploying to Github pages worked, but if you have `homepage` set in package.json there are issues with the other deployment methods (Now & Netlify)

Is it possible to still have the public path url work for all the methods, even if `homepage` is set for gh pages? Or should we just make it clear to not have that set if you're building for not gh pages

EvanBacon:
@cruzach setting `process.env.WEB_PUBLIC_PATH="gh-pages-expo-web-deployment-test"` should work for that. Could we work this into the docs somehow? Also we need to update the docs to reflect the correct build path:
```diff
- build
+ web-build
```

cruzach:
Yeah that worked, I'll update the docs

PR #939, opened 2019-08-16
[webpack-config] Added lighthouse audits to CI

EvanBacon:
Run a lighthouse audit on the example Expo project that is built in production mode.

### DIFF
--- a/packages/webpack-config/tests/__tests__/basic-test.prod.js
+++ b/packages/webpack-config/tests/__tests__/basic-test.prod.js
@@ -1,0 +1,60 @@
+/* global browser */
+import lighthouse from 'lighthouse';
+import url from 'url';
+import config from '../../jest-puppeteer.config';
+
+describe(`Lighthouse`, () => {
+  async function getLighthouseResultsAsync(lhr, property) {
+    const result = lhr.categories[property] || lhr.audits[property];
+
+    const score = await result.score;
+
+    // A null value is returned when the test wasn't run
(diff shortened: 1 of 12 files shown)


EvanBacon:
Would probably make sense to implement a hook that published and tested on a staging URL. Puppeteer scores are swayed by the serve-cli

packages/webpack-config/tests/__tests__/basic-test.prod.js
```diff
@@ -0,0 +1,59 @@
...
+  for (const lighthouseTest of tests) {
+    const { minValue = 1.0 } = lighthouseTest;
+    it(
+      minValue >= 1.0
+        ? `should have a perfect ${lighthouseTest.name} score`
+        : `should have ${lighthouseTest.name} score of ${minValue * 100} or greater`,
+      async () => {
+        const score = await getLighthouseResultsAsync(lhr, lighthouseTest.field);
+        expect(score).toBeGreaterThanOrEqual(minValue);
+        expect(score).toMatchSnapshot();
```
ide: keep an eye on this in case it gets annoying

ide:
> Would probably make sense to implement a hook that published and tested on a staging URL. > Puppeteer scores are swayed by the serve-cli

What specifically do you mean by this? We probably do want a local server so that CI is shielded from external I/O (specifically networks and uptime).

EvanBacon:
stale - we should use Lighthouse CI now

PR #940, opened 2019-08-16
[webpack-config] Pass the CI env down to projects

EvanBacon:
Projects should be able to act a certain way when running in CI

### DIFF
--- a/packages/webpack-config/webpack/createClientEnvironment.js
+++ b/packages/webpack-config/webpack/createClientEnvironment.js
@@ -5,7 +5,7 @@ module.exports = function createClientEnvironment(mode, publicPath, nativeAppMan
   const environment = mode || 'development';
   const __DEV__ = environment !== 'production';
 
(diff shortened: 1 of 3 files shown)


packages/webpack-config/webpack/createClientEnvironment.js
```diff
@@ -5,7 +5,7 @@ module.exports = function createClientEnvironment(mode, publicPath, nativeAppMan
   const environment = mode || 'development';
   const __DEV__ = environment !== 'production';
 
-  const ENV_VAR_REGEX = /^(EXPO_|REACT_NATIVE_)/i;
+  const ENV_VAR_REGEX = /^(EXPO_|REACT_NATIVE_|CI)/i;
```
ide: change this to be more restrictive -- if someone has a variable like CI_SECRET_TOKEN we don't want that passed through I think

```suggestion
  const ENV_VAR_REGEX = /^(EXPO_|REACT_NATIVE_|CI$)/i;
```

PR #944, opened 2019-08-20
fix(webpack-config): read correct values from config

jaulz:
Currently, for example the `serviceWorker` is not correctly read from `app.json` because it's always overridden by the default value `{}`. This fix returns the property from `app.json` if it's truthy and an object.


jaulz:
@EvanBacon could you please have a look at this PR? I think it's still valid.

EvanBacon:
I'd like to just delete and or drastically reduce the values in `app.json` due to type safety and cross-framework compatibility. If you'd still like to land this then please add tests to `packages/webpack-config/src/utils/__tests__/config-test.js`

jaulz:
Sure, will add them beginning of next week!

jaulz:
@EvanBacon done 😊

PR #945, opened 2019-08-20
Convert @expo/schemer, xdl.Project and xdl.ExpSchema to TypeScript

fson:


packages/xdl/src/Project.ts
```diff
@@ -76,16 +78,85 @@ const treekillAsync = promisify(treekill);
...
+      manifestString: string;
+      signedManifest: string;
+    };
 
 let _cachedSignedManifest: CachedSignedManifest = {
   manifestString: null,
   signedManifest: null,
 };
 
+type Asset =
```
ide: One thing that might be confusing is that there are multiple types named `Asset` in the code. Perhaps more descriptive names like AssetInfo would help disambiguate them.

packages/xdl/src/Project.ts
```diff
@@ -1220,14 +1385,14 @@ async function uploadAssetsAsync(projectRoot, assets) {
...
+    current?: boolean;
+    mode?: string;
+    platform?: 'android' | 'ios' | 'all';
+    expIds?: Array<string>;
+    type?: string;
+    releaseChannel?: string;
+    bundleIdentifier?: string;
+    publicUrl?: string;
   } = {}
 ) {
```
ide: add return types, especially for exported functions

packages/xdl/src/Project.ts
```diff
@@ -1847,15 +2019,15 @@ export async function stopExpoServerAsync(projectRoot: string) {
 
 async function _connectToNgrokAsync(
   projectRoot: string,
-  args: mixed,
-  hostnameAsync: Function,
-  ngrokPid: ?number,
+  args: ngrok.NgrokOptions,
+  hostnameAsync: () => Promise<string>,
+  ngrokPid: number | null | undefined,
```
ide: Might be worth looking at the caller and seeing if we can consistently pass in null or undefined but not both.

fson: The value comes from `ProjectSettings.readPackagerInfoAsync` which has this type:
```ts
type PackagerInfo = {
  expoServerPort?: number | null;
  packagerPort?: number | null;
  packagerPid?: number | null;
  expoServerNgrokUrl?: string | null;
  packagerNgrokUrl?: string | null;
  ngrokPid?: number | null;
  devToolsPort?: number | null;
  webpackServerPort?: number | null;
};
```
Initially all of these properties are non-existent (`undefined`) and then they are set to numbers/string as these services get started. Then when the services are stopped the ports/PIDs are explicitly set to `null` to mark the service as stopped. We treat `undefined` and `null` the same.

I agree setting the PIDs/ports to `undefined` (=removing the field from JSON) would be more ideal and clean. But we should also keep it consistent so that if we set `ngrokPid` to `undefined`, we also treat `packagerPid` or `expoServerPort` the same way.

Is it worth it to go over every call that sets any of these values to `null` and change it to `undefined`?

ide: Let's just accept `null | undefined` for now. In this case, pid = 0 is reserved anyway so if we just do a truthiness check for the pid it will always be non-zero if set.

packages/xdl/src/Project.ts
fson: [We already have `AssetInfo`](https://github.com/expo/expo-cli/pull/945/files#diff-8d46d7b06777595a524a77ee67736074R45) 😄 

I've renamed the `Asset` in schemer to  a more descriptive name `AssetField`.

packages/xdl/src/ThirdParty.ts
```diff
@@ -4,7 +4,7 @@ import XDLError from './XDLError';
 
 export async function getManifest(
   publicUrl: string,
-  opts: { platform?: 'android' | 'ios' | 'all' } = {}
+  { platform = 'all' }: { platform?: 'android' | 'ios' | 'all' } = {}
```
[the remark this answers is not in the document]
fson: This results in an error:
> Property 'platform' is missing in type '{}' but required in type '{ platform: "android" | "ios" | "all"; }'.ts(2741)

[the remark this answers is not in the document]
fson: Any ideas how to work around this?

packages/schemer/src/Error.ts
```diff
@@ -19,8 +14,20 @@ export class ValidationError extends ExtendableError {
...
-  constructor({ errorCode, fieldPath, message, data, meta }: Object) {
+  meta: any;
+  constructor({
+    errorCode,
+    fieldPath,
+    message,
+    data,
+    meta,
+  }: {
+    errorCode: keyof typeof ErrorCodes;
```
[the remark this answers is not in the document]
fson: I've added the `ErrorCode` union type, I prefer it to using enums.

However, I've kept the `ErrorCodes` map around for backwards compatibility.

Please let me know if you think it should be removed (which requires a major version bump for Schemer).

packages/xdl/src/project/ExpSchema.ts
```diff
@@ -1,34 +1,33 @@
...
- */
-
 import fs from 'fs';
 import path from 'path';
 import Schemer from '@expo/schemer';
 import Api from '../Api';
-import * as ProjectUtils from './ProjectUtils';
+import * as ConfigUtils from '@expo/config';
+
+export type Schema = any;
```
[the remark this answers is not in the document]
fson: `object` is not very useful here because it doesn't allow accessing any properties. I think it's the correct type only in cases where any object will suffice (e.g. the type of the argument to `Object.keys` builtin method).

I used `any` as a shortcut because the type definitions for packages that `@expo/schemer` uses to validate/traverse JSON Schema didn't include type definitions for the schema JSON. But maybe there's a separate package that we can install to get the full TypeScript type for this format.

packages/schemer/src/index.ts
```diff
@@ -243,17 +239,9 @@ export default class Schemer {
...
-    meta,
-  }: {
-    fieldPath: string,
-    data: string,
-    meta: Meta,
-  }) {
+  async _validateAssetAsync({ fieldPath, data, meta }: Asset) {
     if (meta && meta.asset && data) {
-      if (meta.asset.contentTypePattern && meta.asset.contentTypePattern.startsWith('^image')) {
+      if (meta.contentTypePattern && meta.contentTypePattern.startsWith('^image')) {
```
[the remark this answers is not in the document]
fson: Yeah, I fixed this because the original code seemed to be wrong in an obvious way that was caught by the TypeScript types (yay!). (The underlying data structure looks like this: https://github.com/expo/universe/blob/68816802d45e7941ea781c80a3004a8dd7e27f7a/server/www/xdl-schemas/UNVERSIONED-schema.json#L72-L74).

The code was introduced by a recent commit https://github.com/expo/expo-cli/commit/7f1860da9cdda9128b715724182c91d30e30b698.

However, I didn't test my fix at runtime yet. I can test it and extract the change to a separate PR.

packages/xdl/src/Project.ts
fson: good call

packages/xdl/src/project/ExpSchema.ts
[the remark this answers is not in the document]
ide: `object` in TS is different than in Flow (which is the fairly useless type). In TS it is basically `{ [key: string]: any }`.

packages/xdl/src/AssetUtils.ts
```diff
@@ -33,14 +33,24 @@ export const optimizeImageAsync = async (inputPath: string, quality: number) =>
...
+
+export type OptimizationOptions = {
+  quality: number;
+  include?: string;
+  exclude?: string;
+  save?: boolean;
 };
 
-type Options = { quality: number; include: string; exclude: string };
+export type AssetInfo = { [hash: string]: boolean };
```
ide: I'm thinking it might be better to call this `ProjectAssetInfo` or something that explains that this contains information about the assets in the project because `AssetInfo` could also describe the info for a single asset (like suggested in one of my previous comments :P)

packages/schemer/src/Error.ts
```diff
@@ -30,6 +37,8 @@ export class ValidationError extends ExtendableError {
   }
 }
 
+export type ErrorCode = keyof typeof ErrorCodes;
```
ide: IMO an enum named `ErrorCode` would be nicer to use (and works well without TypeScript too since you'd write `ErrorCode.SOME_ERROR`) but I don't think it's a big deal either way.

packages/xdl/src/project/ExpSchema.ts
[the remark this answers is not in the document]
fson: If I change the the type to `object`, I get 4 errors in this file:
```
@expo/xdl: src/project/ExpSchema.ts(31,14): error TS2339: Property 'meta' does not exist on type 'object'.
@expo/xdl: src/project/ExpSchema.ts(31,27): error TS2339: Property 'meta' does not exist on type 'object'.
@expo/xdl: src/project/ExpSchema.ts(34,29): error TS2339: Property 'properties' does not exist on type 'object'.
@expo/xdl: src/project/ExpSchema.ts(78,3): error TS2741: Property 'schema' is missing in type '{}' but required in type '{ schema: object; }'.
```
Maybe you are thinking of some other type or maybe this depends on `tsconfig` settings?

packages/xdl/src/AssetUtils.ts
fson: Renamed to `AssetOptimizationState`.

packages/schemer/src/Error.ts
fson: I personally prefer union types, but I think they're both equally nice to use.

When used with TypeScript, autocompletion will be provided for both enums and string literals.

When used with JavaScript, there's also no difference, since neither approach will protect from typos when there's no type checker to detect them.

PR #946, opened 2019-08-20
Changes Axios Request error handling

AdamJNavarro:
Wrapped the Axios Request in a try/catch and used some de-structuring to get a clear error message to be spit out.

You'll need to fire up this branch of `www` locally before testing. https://github.com/expo/universe/pull/3652

### DIFF
--- a/packages/xdl/src/Api.ts
+++ b/packages/xdl/src/Api.ts
@@ -101,33 +101,42 @@ async function _callMethodAsync(
       options = { ...options, ...requestOptions };
     }
   }
-  let response = await axios.request(options);
-  if (!response) {
-    throw new Error('Unexpected error: Request failed.');
-  }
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Api.ts
```diff
@@ -101,33 +101,42 @@ async function _callMethodAsync(
...
-    responseObj = responseBody;
-  }
-  if (responseObj.err) {
-    let err = new ApiError(
-      responseObj.code || 'API_ERROR',
-      'API Response Error: ' + responseObj.err
+  } catch (error) {
+    const { response } = error;
+    const { ...errorObject } = response;
+    throw new Error(
```
fson: I think we should throw an `ApiError`, because we handle them differently from `Error` in that we don't show a stack trace.

Issue #933, opened 2019-08-12
Installing Expo client on iOS simulator not installing


glasje:
### Environment
run expo start --ios

### Steps to Reproduce

when I execute the project with expo start --ios, the application is lifted but it does not install the expo in the simulator, I already tried with models iphone 6,7,8, x, xr, xs and still there is no change. The same happens if I run the app with react-native run-ios does not install the app.
version S.O. : mojave 10.14.3
version expo-cli: 3.0.9
version xcode: Version 10.3 (10G8)
version simulator: Version 10.3 (SimulatorApp-880.5 CoreSimulator-587.35)
### Reproducible Demo
![image](https://user-images.githubusercontent.com/34929785/62892161-91d30780-bd15-11e9-966a-53966fd28a01.png)

glasje:
@cruzach  It happens to me with blank projects, with developed app and we install expo cli in the simulator.
I did tests on a co-worker's mac and exactly the same thing happens

cruzach:
What do you mean by "we installed expo cli in the simulator"? Can you try running `expo client:install:ios`

glasje:
I ran the indicated command and threw this after execution.
![image](https://user-images.githubusercontent.com/34929785/63056499-32602d80-beb6-11e9-84dc-14b6c557de6a.png)
 
I need to solve this issue urgently. :(

cruzach:
Could you try reinstalling [`watchman`](https://docs.expo.io/versions/latest/introduction/installation/#watchman)? Also, go ahead and share your `app.json` here

glasje:
if you reinstall watchman but the same error came out
watchman version : 4.9.0
this is app.json :
{
  "expo": {
    "name": "prueba",
    "slug": "AwesomeProject",
    "privacy": "public",
    "sdkVersion": "34.0.0",
    "platforms": [
      "ios",
      "android",
      "web"
    ],
    "version": "1.0.0",
    "orientation": "portrait",
    "icon": "./assets/images/icon.png",
    "splash": {
      "image": "./assets/images/splash.png",
      "resizeMode": "contain",
      "backgroundColor": "#ffffff"
    },
    "updates": {
      "fallbackToCacheTimeout": 0
    },
    "assetBundlePatterns": [
      "**/*"
    ],
    "ios": {
      "supportsTablet": true
    }
  }
}
-------------------------------
this is package.json
{
  "main": "node_modules/expo/AppEntry.js",
  "scripts": {
    "start": "expo start",
[...]

PR #957, opened 2019-08-23
fix(xdl): Check if Client is installed via application directory [iOS 13/Xcode 11 compatibility]

cruzach:
Revealed in  [933](https://github.com/expo/expo-cli/issues/933#issuecomment-523872521) that iOS 13 / Xcode 11 Command Line Tools resulted in a forever loop after installing the Expo Client onto the simulator. 

The loop we get stuck in was [`_waitForExpoAppInstalledOnCurrentBootedSimulatorAsync`](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/Simulator.ts#L275). `_isExpoAppInstalledOnCurrentBootedSimulatorAsync` would continually return `falsy` because there was no longer `AppID/Library/Caches/Snapshots` directory, so the `glob` match would fail. 

This PR makes it so we search for the app in the App Store Application Directory, instead of the Data Directory

**new commit** I realized that the previous glob search would also match any other app that happened to be installed on the sim with a bundle name starting with "Exponent". Good news is we already handle that in `_expoVersionOnCurrentBootedSimulatorAsync` with some regex testing 😄


ahansson89:
Can this get merged in? iOS 13 and XCode 11 is out now. Thanks!

mosnicholas:
+1 -- we're seeing issues and are unable to load our projects on iOS 13 and XCode 11 either.

kitingChris:
Problem still exists and a real blocker! When will this be merged?

jordan-acosta:
For anyone who ends up here, the newly-released Expo SDK 35 supports iOS 13 and XCode 11: https://blog.expo.io/expo-sdk-35-is-now-available-beee0dfafbf4

cruzach:
@fson this was fixed in [this commit](https://github.com/expo/expo-cli/commit/5f209e26ba47d978ca7006d7b528db93930ae9e3#diff-71db691b019e067851e82c9e8a3a6bd6), not sure if you want to keep that as it is or merge this one?

fson:
Let's ship this?

cruzach:
Wfm, resolving conflicts now

Issue #822, opened 2019-07-05
v3.0.0 build:android fails with property error


brunogirin:
# Required format for issues

## Description

Running the `build:android` command with expo-cli v3.0.0, the command fails with an error. It works with expo-cli v2.21.2

## Expected Behavior

Ran `expo build:android`
The command should succeed with no error.

## Observed Behavior

The command fails if the expo-cli version is 3.0.0 but works if it is a 2.x version (I tried with 2.20.11 and 2.21.2).

```
$ expo build:android
[16:18:49] Cannot read property 'readConfigJsonAsync' of undefined
[16:18:49] TypeError: Cannot read property 'readConfigJsonAsync' of undefined
    at Object.checkSplashScreenImages (/@expo/xdl@56.0.0/src/Android.ts:359:37)
    at AndroidBuilder._callee$ (/expo-cli@3.0.0/src/commands/build/AndroidBuilder.js:27:19)
    at tryCatch (/usr/local/share/.config/yarn/global/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:62:40)
    at Generator.invoke [as _invoke] (/usr/local/share/.config/yarn/global/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:296:22)
    at Generator.prototype.(anonymous function) [as next] (/usr/local/share/.config/yarn/global/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:114:21)
    at step (/usr/local/share/.config/yarn/global/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /usr/local/share/.config/yarn/global/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Environment

Please run `expo diagnostics` in the project folder and copy the output here:

```
  Expo CLI 3.0.0 environment info:
    System:
      OS: Linux 4.19 Debian GNU/Linux 9 (stretch) 9 (stretch)
      Shell: 4.4.12 - /bin/bash
    Binaries:
      Node: 10.16.0 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
    npmPackages:
      expo: ^32.0.0 => 32.0.6 
      react: 16.5.0 => 16.5.0 
[...]

PR #960, opened 2019-08-26
fix(expo-cli): correctly import @expo/config

z0al:
## The problem

```sh
$ expo customize:web
Cannot read property 'readConfigJsonAsync' of undefined
TypeError: Cannot read property 'readConfigJsonAsync' of undefined
    at action (/expo-cli@3.0.10/src/commands/customize.ts:126:3)
    at Command.action (/expo-cli@3.0.10/src/exp.ts:82:13)
```

## Changes

I looked for every import to `@expo/config` and fixed the import, however, only one file needed a fix.

Related to #822 (Which I think this should be closed)

### DIFF
--- a/packages/expo-cli/src/commands/customize.ts
+++ b/packages/expo-cli/src/commands/customize.ts
@@ -1,6 +1,6 @@
 import { Command } from 'commander';
 import spawnAsync from '@expo/spawn-async';
-import ConfigUtils from '@expo/config';
+import * as ConfigUtils from '@expo/config';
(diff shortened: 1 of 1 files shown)


fson:
Thank you!

chayancbnits:
After install expo 3.0.11-alpha.1 ,then expo build:web

z0al:
@chayancbnits I don't think this PR is related to your issue, but in any case, please open a new issue with more context since this PR is closed.

PR #972, opened 2019-08-29
feat(web): add supports for push notifications for Next.js with Expo for Web

hesyifei:
(See also https://github.com/expo/expo-cli/pull/844 and https://github.com/expo/expo-cli/pull/926.)

We copy and serve `/expo-service-worker.js` (and `/service-worker.js`) with Express when using Next.js. This means that `expo start --dev` and `expo start --no-dev` will both support the service worker without any additional configurations.

For services like Now.sh, configurations like https://github.com/hanford/next-offline#now-20 are needed. See docs from https://github.com/expo/expo/pull/5501.

Problem:
- `/static/expo-service-worker.js` will always get created even if the user is not using web Notifications.

### DIFF
--- a/packages/xdl/src/Webpack.ts
+++ b/packages/xdl/src/Webpack.ts
@@ -14,7 +14,7 @@ import createWebpackCompiler, {
   printInstructions,
   printSuccessMessages,
 } from './createWebpackCompiler';
-import ip from './ip'; 
+import ip from './ip';
 // @ts-ignore
 import * as Doctor from './project/Doctor';
 import * as ProjectUtils from './project/ProjectUtils';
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Webpack.ts
```diff
@@ -539,6 +539,22 @@ async function startNextJsAsync({
   await app.prepare();
 
   const server = express();
+
+  server.get('/expo-service-worker.js', (req, res) => {
```
EvanBacon: Will this work for locally testing prod builds? Some people use serve-cli to test before deploying.

hesyifei: It will work with `expo start --no-dev --web`. However, I don't think it will work with `expo build:web` and then `yarn next start` (since then it's not using our Express server).

It will work with `now` with the `now.json` configuration in the docs.

PR #984, opened 2019-09-03
test(expo-cli): add e2e (end-to-end) tests for start and init

fson:
This PR adds end-to-end tests for the `expo start` and `expo init` commands.

### DIFF
--- a/packages/expo-cli/e2e/__tests__/init-test.ts
+++ b/packages/expo-cli/e2e/__tests__/init-test.ts
@@ -20,7 +20,7 @@ test('init', async () => {
   jest.setTimeout(60000);
   const cwd = temporary.directory();
   const { stdout } = await runAsync(
-    ['init', 'hello-world', '--template', 'blank', '--name', 'Hello', '--yarn'],
+    ['init', 'hello-world', '--template', 'blank', '--name', 'Hello'],
     { cwd }
   );
   expect(stdout).toMatch(`Your project is ready at ${cwd}`);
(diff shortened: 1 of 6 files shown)


packages/expo-cli/e2e/__tests__/init-test.ts
```diff
@@ -0,0 +1,30 @@
+import path from 'path';
+
+import JsonFile from '@expo/json-file';
+import temporary from 'tempy';
+
+import { runAsync, tryRunAsync } from '../TestUtils';
+
```
ide: might want to set up beforeEach and afterEach blocks to create a temporary directory and set process.env.PWD to the temp dir, and delete the temp dir when done so that the test cleans up after itself

packages/expo-cli/e2e/__tests__/start-test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { runAsync, EXPO_CLI } from '../TestUtils';
+
+const projectDir = path.join(temporary.directory(), 'my-app');
+
+beforeAll(async () => {
+  jest.setTimeout(60000);
+  await runAsync(['init', projectDir, '--template', 'blank', '--name', 'My App']);
+});
+
```
ide: delete the temp dir in afterAll() (this is more for our own computers, not CI)

packages/expo-cli/e2e/__tests__/init-test.ts
fson: I considered this, when writing the test, but decided against cleaning up the temporary directory for a few reasons:
- the test sets `$PWD` to a temporary folder inside `/tmp` so the operating system will automatically clean it up afterwards, on reboot
- cleaning up would therefore add unnecessary complexity to the test suite
- if there's a failure, it's a lot easier to debug, if the folder is still there after the test finishes.

PR #991, opened 2019-09-06
[expo-cli] permissions commands

EvanBacon:
- added: `expo permissions:ios`
  - Managed workflow
    - [x] This prompts the user to update the permission dialogs for their iOS app. This will also enable users to view what the auto generated Expo permission dialogs will look like in their app.
    - [x] Add reminder that permission dialogs will only be available in Standalone, and native builds.
  - Bare, Detached, (anything with an iOS project)
    - [x] Will prompt you to update, add, remove, and sync permissions from the iOS `Info.plist`
    - [x] Sync changes to the `app.json`
- added: `expo permissions:android`
  - Managed workflow
    - [x] Prompts users to include or exclude the Android permissions
    - [x] Clearly demonstrates that selecting _no_ permissions means enabling all permissions
    - [x] Select/Deselect All
  - Non-Managed Native Android
    - [x] Reads/Writes/Syncs from the Android manifest
    - [x] Sync changes to the `app.json`

## iOS

Screen Shot 2019-09-06 at 12 44 23 AM

## Android 

Screen Shot 2019-09-06 at 12 44 44 AM


EvanBacon:
It'd be cool to apply this to approach to deep linking too

EvanBacon:
HOLD for #1036

ide:
Might be good to introduce just `expo permissions` and then add `--platforms {android,ios,web}` so that by default we get text for all of the permissions universally, while letting people still focus on just one platform if they want to.

EvanBacon:
But the permission flows are completely different. Android doesn't provide a description like iOS does so when is a checklist and the other is a table.

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
+import { ExpoConfig, readConfigJsonAsync, writeConfigJsonAsync } from '@expo/config';
+import { IosPlist, IosWorkspace } from '@expo/xdl';
+import StandaloneContext from '@expo/xdl/build/detach/StandaloneContext';
+import * as Manifest from '@expo/android-manifest';
+import chalk from 'chalk';
+import { Command } from 'commander';
+// @ts-ignore
```
ide: explain what you are ignoring

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+import { Form, MultiSelect } from 'enquirer';
+import fs from 'fs-extra';
+import path from 'path';
+
+// @ts-ignore
```
ide: same

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+    },
+  });
+}
+
+function getPermissionDescription(
+  appName: string,
+  { ios = {} }: ExpoConfig,
+  permission: string
+): string {
+  // @ts-ignore
```
ide: me too

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+  const { infoPlist = {} } = ios;
+
+  if (permission in infoPlist && typeof infoPlist[permission] === 'string') {
```
ide:
```suggestion
  if (typeof infoPlist[permission] === 'string') {
```

same thing?

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+  const prompt = new MultiSelect({
+    header() {
+      const descriptions: AnyPermissions = {
+        // TODO(Bacon): Add descriptions of what each permission is used for
+      };
+      return descriptions[Object.keys(Manifest.UnimodulePermissions)[this.index]];
+    },
+    initial: permissions,
+    hint:
+      '(Use <space> to select, <return> to submit, <a> to toggle, <i> to inverse the selection)',
```
ide:
```suggestion
      '(Use <space> to select, <return> to submit, <a> to toggle, <i> to invert the selection)',
```

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+        chalk.magenta(
+          `${CHEVRON} Expo will default to using all permissions in your project by deleting the ${chalk.underline`expo.android.permissions`} array.`
+        )
+      );
+    }
+  }
+
+  await writeConfigJsonAsync(projectDir, {
+    android: {
+      ...(exp.android || {}),
```
ide:
```suggestion
      ...exp.android,
```

we all good

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+async function promptForPermissionDescriptionsAsync(
+  choices: IOSPermissionChoice[],
+  hasNativeConfig: boolean,
+  currentDescriptions: AnyPermissions
+): Promise<AnyPermissions> {
+  console.log('');
+
+  const keys = Object.keys(DefaultiOSPermissionNames);
+  const prompt = new Form({
+    name: hasNativeConfig ? 'Native project' : 'Universal Project',
```
ide:
```suggestion
    name: hasNativeConfig ? 'Native project' : 'Universal project',
```

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+        `${CHEVRON} Update the description for the permission "${chalk.bold(permission)}"`
+      );
+    },
+    choices,
+  });
+
+  try {
+    const answer: AnyPermissions = await prompt.run();
+    // Trimming allows users to pass in " " to delete a permission
+    return Object.keys(answer).reduce(
```
ide: simpler:
```
for (const key of Object.keys(answer)) {
  answer[key] = answer[key].trim();
}
```

or https://lodash.com/docs/2.4.2#mapValues. mapValues seems particularly useful to remove all the reduce() calls here

packages/expo-cli/src/commands/permissions.ts
```diff
@@ -0,0 +1,416 @@
...
+  } else {
+    console.log(
+      `${CHEVRON} ${chalk.bold`Remember:`} ${chalk.reset
+        .dim`Permission messages are a build-time configuration. Your selection will only be available in Standalone, or native builds. You will continue to see the predefined permission dialogs when using your app in the Expo client.`}`
+    );
+  }
+}
+
+// Find the location of the native iOS info.plist if it exists
+// TODO(Bacon): Search better
```
ide: explain what you mean by this - someone reading this comment is not going to know what to do

packages/expo-cli/src/commands/permissions.ts
EvanBacon: nah this is nullable

PR #994, opened 2019-09-08
Convert @expo/webpack-config to TypeScript

EvanBacon:


fson:
Let's add the generated files (compiled sources and source maps) to `.gitignore`.

PR #996, opened 2019-09-09
[Actions] Created a workflow to move unit tests off of CircleCI

EvanBacon:
CircleCI is not as free as Github Actions. 

I've opted to manually create a test for each package just to keep the migration consistent. The script has the ability to automatically find packages in the future.


.github/workflows/packages.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: yarn install --frozen-lockfile
+      - run: yarn bootstrap
+        env:
+          CI: true
+      - name: Test @expo/dev-tools
+        run: cd packages/dev-tools && ../../scripts/module-scripts-check-packages.sh
```
ide:
```suggestion
        run: cd packages/dev-tools && "$GITHUB_WORKSPACE/scripts/module-scripts-check-packages.sh"
```

scripts/module-scripts-check-packages.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright 2019-present 650 Industries. All rights reserved.
+
+# exit when any command fails
+set -e
```
ide:
```suggestion
set -euo pipefail
```

scripts/module-scripts-check-packages.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright 2019-present 650 Industries. All rights reserved.
+
+# exit when any command fails
```
ide:
```suggestion
```

EvanBacon:
`yarn` and `yarn bootstrap` both seem to take twice as long as Circle CI, we should try and get those to at least be the same duration or faster

ide:
My guess is it's the cache. If you want to email GitHub support about it or tweet at them, go ahead. I think it'd be a pretty meaningful addition to Actions.

byCedric:
> My guess is it's the cache. If you want to email GitHub support about it or tweet at them, go ahead. I think it'd be a pretty meaningful addition to Actions.

I also pinged them related to the Expo GH Action. If they reply with information relevant to you guys too, I'll comment it here. 😄 In my opinion, this is almost a required feature for most large projects.

byCedric:
> > My guess is it's the cache. If you want to email GitHub support about it or tweet at them, go ahead. I think it'd be a pretty meaningful addition to Actions.
> 
> I also pinged them related to the Expo GH Action. If they reply with information relevant to you guys too, I'll comment it here. 😄 In my opinion, this is almost a required feature for most large projects.

TL;DR; They are working on it and hoping to release this feature before the end of the year.

brentvatne:
discussed with @ide - there are a few blockers for merging this so we're going to close it for now.

- github action per-user pricing model is tricky for us 🤑 
- no artifact caching yet, so it's slow
- we can decide what to do when caching is available

byCedric:
Caching is [sort-of here](https://github.com/actions/cache), but you can't download the cache from outside GitHub I think.

PR #1006, opened 2019-09-19
Fix CI caching parameters

EvanBacon:
# Why

https://circleci.com/gh/expo/expo-cli/7454?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

- error computing cache key: template: cacheKey:1: function "parameters" not defined

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,12 +4,18 @@ executors:
   node-10:
     docker:
       - image: circleci/node:10
+    environment:
+      EXECUTOR_NAME: node-10
   node-12:
     docker:
       - image: circleci/node:12
+    environment:
+      EXECUTOR_NAME: node-12
   puppeteer:
     docker:
       - image: circleci/node:latest-browsers
+    environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -79,7 +85,7 @@ jobs:
       - restore_cache:
           name: Restore Yarn Package Cache
           keys:
-            - yarn-{{ parameters.e }}-{{ checksum "yarn.lock" }}
+            - yarn-{{ $EXECUTOR_NAME }}-{{ checksum "yarn.lock" }}
```
ide: AFAIK you can't use arbitrary environment variables in cache keys. You can use `{{ .Environment.variableName }}` but only for environment variables that CircleCI defines.

ide: As a workaround you could `echo "$EXECUTOR_NAME" > /tmp/.executor_name` and then do `{{ checksum "/tmp/.executor_name" }}`.

PR #1008, opened 2019-09-19
fix(dev-tools): add authentication to the WebSocket API

fson:
This patch adds four measures to limit access to the local WebSocket
API of Dev Tools.

- The WebSocket API endpoint URL now includes a secret token.
  Requests to the API without the ticket are unauthorized, for example:
  ws://localhost:19002/M2vbMDtkD-hVJGCegdw3BytRFV5JbQXPM2pLbYwgQDo/graphql
- WebSocket connections without a correct `Origin` header are rejected.
- After a connection is established, the client needs to inclue
  `clientAuthenticationToken`, another secret token.
- The HTTP server of Dev Tools is only listening connections to `localhost`.

The secret tokens are obtained from a new HTTP endpoint `/dev-tools-info`.

Thanks to Spencer von der Ohe for reporting this issue.

### DIFF
--- a/packages/dev-tools/pages/index.js
+++ b/packages/dev-tools/pages/index.js
@@ -2,6 +2,7 @@ import React from 'react';
 import gql from 'graphql-tag';
 import { ApolloProvider, Query } from 'react-apollo';
 import { SubscriptionClient } from 'subscriptions-transport-ws';
+import pTimeout from 'p-timeout';
 
 import * as State from 'app/common/state';
 import createApolloClient from 'app/common/createApolloClient';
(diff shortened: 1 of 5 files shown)


packages/dev-tools/pages/index.js
```diff
@@ -459,27 +460,37 @@ export default class IndexPage extends React.Component {
...
       disconnected: false,
+      client: null,
     };
     this.unsubscribers = [];
     if (process.browser) {
-      this.subscriptionClient = new SubscriptionClient(`ws://${window.location.host}/graphql`, {
-        reconnect: true,
-      });
-      this.client = createApolloClient(this.subscriptionClient);
+      this.connect().catch(error => this.setState({ disconnected: true }));
```
ide: I think we should be doing this work in `componentDidMount`, not the constructor.

fson: You're right, fixed now.

abhaykumar01234:
How can i open the dev tools now? How do i use this 
{
webSocketGraphQLUrl: "ws://192.168.43.235:19002/abMPg_W5NmomhbzAMeNs7fSiWAhW4gFUTEn0idm0WY4/graphql",
clientAuthenticationToken: "4IfHxm6meiwQhlO9JrWdVgrZ5nT2RPA8MAMD-E7xXPE"
}

PR #1020, opened 2019-09-23
Prevent prod builds from using dev flag

EvanBacon:
- Dev flag doesn't provide enough control for debugging a prod build.
- Persistent dev flag causes a lot of confusion

### DIFF
--- a/packages/xdl/src/Webpack.ts
+++ b/packages/xdl/src/Webpack.ts
@@ -466,6 +466,10 @@ async function getWebpackConfigEnvFromBundlingOptionsAsync(
   projectRoot: string,
   options: BundlingOptions
 ): Promise<Web.WebEnvironment> {
+  // Bacon: Prevent dev flag from being used in production
+  if (options.mode === 'production') {
+    options.dev = false;
+  }
   let { dev, https } = await applyOptionsToProjectSettingsAsync(projectRoot, options);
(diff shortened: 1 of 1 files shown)


kopax:
How can we have multiple prod environment ?

My configuration look like follow:

```jsx
import Constants from 'expo-constants';
export const { version } = Constants.manifest;

export const locales = [
  'en',
  'fr',
];

/* eslint-disable global-require, consistent-return, arrow-body-style */
export const getEnvVars = (env = Constants.manifest.releaseChannel) => {
  let config = null;
  // What is __DEV__ ?
  // This variable is set to true when react-native is running in Dev mode.
  // __DEV__ is true when run locally, but false when published.
  if (__DEV__) { // eslint-disable-line no-undef
    config = require('./env/default');
  } else if (env === 'staging') {
    config = require('./env/staging');
  } else if (env === 'preprod') {
    config = require('./env/preprod');
  } else if (env === 'prod') {
    config = require('./env/prod');
  } else {
    config = require('./env/prod');
  }

  if (!['prod', 'test'].includes(config.name)) {
    // eslint-disable-next-line no-console
    console.log('Loading', config.name, 'config');
  }
  return config;
};
```

I am not able to build the web for `staging` and `preprod`, any idea how I can do?

Issue #1025, opened 2019-09-24
Cannot find `babel-loader` generated by `@expo/webpack-config/withUnimodules`


uber5001:
## Description

Following the guide for [expo with nextjs](https://docs.expo.io/versions/latest/guides/using-nextjs/) results in an error when in powershell on windows.

## Expected Behavior

Following the guide for expo with nextjs should work on windows with powershell without errors

Installing `next-server` shouldn't be needed, or should be mentioned in the docs

## Observed Behavior

I need to install `next-server` (see #1015), but that's not mentioned in the docs. I'm now getting an error `Cannot find 'babel-loader'`.

```
$ expo start --web
Starting project at C:\Users\uber5\Desktop\expotest3
Expo DevTools is running at http://localhost:19002
Press d to open DevTools now, or shift-d to always open it automatically.
Starting Metro Bundler on port 19001.
Tunnel ready.
 web  Starting Next.js on port 19006 in development mode.
Cannot find `babel-loader` generated by `@expo/webpack-config/withUnimodules`. It is likely an Expo issue. Please create a new issue at https://github.com/expo/expo-cli.
Error: Cannot find `babel-loader` generated by `@expo/webpack-config/withUnimodules`. It is likely an Expo issue. Please create a new issue at https://github.com/expo/expo-cli.
    at _createNextJsConfig (C:\@expo\xdl@56.2.3\src\Webpack.ts:643:11)
    at startNextJsAsync (C:\@expo\xdl@56.2.3\src\Webpack.ts:531:11)
error Command failed with exit code 1.
```

## Environment

```
  Expo CLI 3.1.1 environment info:
    System:
      OS: Windows 10
    Binaries:
      Yarn: 1.17.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.11.3 - C:\Program Files\nodejs\npm.CMD
```

## Reproducible Demo

```
npm i -g windows-build-tools #from admin powershell, make sure it installs python
npm i -g expo-cli
```
```
expo init #pick managed blank, name it expotest, use yarn to install dependencies
cd expotest
yarn add react react-dom next next-server # including next-server as per #1015
# add pages/index.js and edit app.json with expo.web.use = "nextjs"
yarn add -D @babel/plugin-proposal-decorators
yarn web
# Cannot find `babel-loader` generated by `@expo/webpack-config/withUnimodules`. It is likely an Expo issue. Please create a new issue at https://github.com/expo/expo-cli.
```

PR #1029, opened 2019-09-25
#1025 support windows paths in _findBabelLoader

uber5001:


davidkcyip:
Hi Evan, was wondering when would you be releasing this fix?

uber5001:
@davidkcyip Looks like it was just released

PR #1031, opened 2019-09-25
refactor(expo-cli): make importing commands non-dynamic

fson:
Previously Expo CLI searched every file and directory in the `commands`
folder and imported commands dynamically.

Removing this dynamic importing requires us to list all commands in
`commands/index.ts`, but it is necessary for tools that use static
analysis, such as [`ncc`](https://github.com/zeit/ncc) to to work.


fson:
I ran some tests and there's no significant difference. The imports are still lazy due to the Babel transports we use.
In theory this change should make things

PR #1034, opened 2019-09-26
[expo-cli] First pass at expo upgrade command

brentvatne:
Just putting it up here, still some more work to do, feel free to review though

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -1,0 +1,217 @@
+import { Command } from 'commander';
+import { Versions } from '@expo/xdl';
+import JsonFile from '@expo/json-file';
+import * as ConfigUtils from '@expo/config';
+import chalk from 'chalk';
+import * as PackageManager from '../PackageManager';
+import CommandError from '../CommandError';
+import prompt from '../prompt';
+import log from '../log';
+import path from 'path';
+import fs from 'fs';
+
+type Options = {
+  npm?: boolean;
+  yarn?: boolean;
+};
+
+function maybeFormatSdkVersion(sdkVersionString: string | null) {
+  if (!sdkVersionString) {
+    return sdkVersionString;
+  } else if (sdkVersionString.includes('.')) {
+    return sdkVersionString;
+  }
+
+  return `${sdkVersionString}.0.0`;
+}
+
+type DependencyList = { [key: string]: string };
+
+/**
+ * Produce a list of dependencies used by the project that need to be updated
+ */
+async function getUpdatedDependenciesAsync(
+  projectRoot: string,
+  targetSdkVersion: { expoReactNativeTag: string; facebookReactVersion?: string } | null
+): Promise<DependencyList> {
+  let result: DependencyList = {};
+
+  // Get the supported react-native version.
+  // * We can't update the react-native version if the version data isn't published
+  if (targetSdkVersion) {
+    result['react-native'] = `https://github.com/expo/react-native/archive/${
+      targetSdkVersion.expoReactNativeTag
+    }.tar.gz`;
+
+    // React version apparently is optional in SDK version data
+    if (targetSdkVersion.facebookReactVersion) {
+      result['react'] = targetSdkVersion.facebookReactVersion;
+    }
+
+    // - TODO: how do we know what version of react-native-web, react-dom to use?
+    // - TODO: we need to track supported devDependencies versions somewhere programmatic too
+  } else {
+    console.log(`Supported React Native version unknown, please update it manually.`);
+  }
+
+  // Get the updated version for any bundled modules
+  let { exp, pkg } = await ConfigUtils.readConfigJsonAsync(projectRoot);
+  let bundledNativeModules = (await JsonFile.readAsync(
+    ConfigUtils.resolveModule('expo/bundledNativeModules.json', projectRoot, exp)
+  )) as DependencyList;
+
+  let { dependencies } = pkg;
+  Object.keys(bundledNativeModules).forEach(name => {
+    if (dependencies[name]) {
+      result[name] = bundledNativeModules[name];
+    }
+  });
+
+  return result;
+}
+
+async function upgradeAsync(requestedSdkVersion: string | null, options: Options) {
+  let { projectRoot, workflow } = await findProjectRootAsync(process.cwd());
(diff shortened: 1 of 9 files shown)


packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+    log.error('No sdkVersion field is present in app.json, cannot upgrade project.');
+    return;
+  }
+
+  let currentSdkVersionString = exp.sdkVersion;
+  let sdkVersions = await Versions.sdkVersionsAsync();
+  let latestSdkVersion = await Versions.newestSdkVersionAsync();
+  let latestSdkVersionString = latestSdkVersion.version;
+  let targetSdkVersionString =
+    maybeFormatSdkVersion(requestedSdkVersion) || latestSdkVersion.version;
```
fson:
```suggestion
    semver.valid(semver.coerce(requestedSdkVersion)) || latestSdkVersion.version;
```

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    // React version apparently is optional in SDK version data
+    if (targetSdkVersion.facebookReactVersion) {
+      result['react'] = targetSdkVersion.facebookReactVersion;
+    }
+
+    // - TODO: how do we know what version of react-native-web, react-dom to use?
+    // - TODO: we need to track supported devDependencies versions somewhere programmatic too
+  } else {
+    console.log(`Supported React Native version unknown, please update it manually.`);
```
fson:
```suggestion
    log.warn(`Supported React Native version unknown, please update it manually.`);
```

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+  });
+
+  return result;
+}
+
+async function upgradeAsync(requestedSdkVersion: string | null, options: Options) {
+  let { projectRoot, workflow } = await findProjectRootAsync(process.cwd());
+  let { exp } = await ConfigUtils.readConfigJsonAsync(projectRoot);
+
+  if (Versions.lteSdkVersion(exp, '32.0.0')) {
```
fson:
```suggestion
  if (!Versions.gteSdkVersion(exp, '33.0.0')) {
```
(I know we currently don't have versions like `33.0.1`, but we should probably still treat the version number like a semver.)

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+    log.error('The upgrade command is only available on SDK 33 and higher. Please refer to https://docs.expo.io/versions/latest/workflow/upgrading-expo-sdk-walkthrough/ for upgrade instructions.');
+    return;
+  }
+
+  // Can't upgrade if we don't have a SDK version
+  if (!exp.sdkVersion) {
```
fson: This would be a separate PR, but this reminded me that I think it would make sense to change config file handling so that we'd always have a version, even if it's not specified in `app.json` (like we discussed at some point, I think). We could infer the version from the installed `expo` package version.

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+  let sdkVersions = await Versions.sdkVersionsAsync();
+  let latestSdkVersion = await Versions.newestSdkVersionAsync();
+  let latestSdkVersionString = latestSdkVersion.version;
+  let targetSdkVersionString =
+    maybeFormatSdkVersion(requestedSdkVersion) || latestSdkVersion.version;
+  let targetSdkVersion = sdkVersions[targetSdkVersionString];
+
+  // TODO: figure out how this should work for bare workflow
+  if (workflow === 'bare') {
+    console.log('This command is currently only supported on the managed workflow.');
```
fson:
```suggestion
    log.error('This command is currently only supported on the managed workflow.');
```

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+    return;
+  }
+
+  // Bail out early if people are trying to update to the current version
+  if (targetSdkVersionString === currentSdkVersionString) {
+    console.log(
+      'You are already using the latest SDK version. Follow https://blog.expo.io for new release information.'
+    );
+    return;
```
fson: We should throw an error (e.g. `CommandError`) to make the return value of the command non-zero.

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+      targetSdkVersionString = selectedSdkVersionString;
+    }
+  } else if (!targetSdkVersion) {
+    // If they provide an apparently unsupported sdk version then let people try
+    // anyways, maybe we want to use this for testing alpha versions or
+    // something...
+    let answer = await prompt({
+      type: 'confirm',
+      name: 'attemptUnknownUpdate',
+      message: `You provided the target SDK version value of ${targetSdkVersionString} which does not seem to exist. But hey, I'm just a program, what do I know. Do you want to try to upgrade to it anyways?`,
```
fson: LOL

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -0,0 +1,217 @@
...
+    });
+
+    if (!answer.attemptUnknownUpdate) {
+      return;
+    }
+  }
+
+  let packageManager = PackageManager.createForProject(projectRoot, {
```
fson: 👍 👍

brentvatne:
linking this issue as people are having trouble upgrading https://github.com/kmagiera/react-native-gesture-handler/issues/742#issuecomment-533739788

packages/expo-cli/src/commands/upgrade.ts
brentvatne: i think in this case i'd expect it to exit with status 0 because, while it didn't upgrade, the command ran successfully. we already were on latest version so no need to change anything

brentvatne:
Screen Shot 2019-10-01 at 19 37 19

the main blocker on landing this is that https://github.com/expo/expo-cli/issues/1054 is particularly painful when using `expo upgrade` because we do `npm install react@16.8.3` which `npm` incorrectly interprets as `npm install react@^16.8.3`

PR #1037, opened 2019-09-27
fix: respect the owner field specified in app.json

tcdavis:
Fix some cases where expo-cli commands would ignore an owner key specified in app.json

### DIFF
--- a/packages/expo-cli/src/commands/build/AndroidBuilder.js
+++ b/packages/expo-cli/src/commands/build/AndroidBuilder.js
@@ -61,7 +61,7 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
   }
 
   async _clearCredentials() {
-    const username = this.user.username;
+    const username = this.manifest.owner || this.user.username;
     const experienceName = `@${username}/${this.manifest.slug}`;
 
     const credentialMetadata = {
@@ -110,7 +110,7 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
   }
 
   async collectAndValidateCredentials() {
-    const username = this.user.username;
+    const username = this.manifest.owner || this.user.username;
     const experienceName = `@${username}/${this.manifest.slug}`;
 
     const credentialMetadata = {
--- a/packages/expo-cli/src/commands/build/AndroidBuilder.js
+++ b/packages/expo-cli/src/commands/build/AndroidBuilder.js
@@ -61,14 +61,10 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
   }
 
   async _clearCredentials() {
-    const username = this.manifest.owner || this.user.username;
-    const experienceName = `@${username}/${this.manifest.slug}`;
-
-    const credentialMetadata = {
-      username,
-      experienceName,
-      platform: ANDROID,
-    };
+    const credentialMetadata = await Credentials.getCredentialMetadataAsync(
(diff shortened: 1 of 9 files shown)


packages/expo-cli/src/commands/build/AndroidBuilder.js
```diff
@@ -61,7 +61,7 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
   }
 
   async _clearCredentials() {
-    const username = this.user.username;
+    const username = this.manifest.owner || this.user.username;
```
fson: Looks like `IOSBuilder` has some similar code that assumes username is `this.user.username`: https://github.com/expo/expo-cli/blob/83a758a75892200b3bb517e389ea8093e16e23a2/packages/expo-cli/src/commands/build/ios/IOSBuilder.js#L54-L59

Can you check if it needs to be changed as well?

packages/xdl/src/Exp.ts
```diff
@@ -304,9 +304,12 @@ export async function getPublishInfoAsync(root: string): Promise<PublishInfo> {
     throw new Error('Attempted to login in offline mode. This is a bug.');
   }
 
-  const { username } = user;
+  let { username } = user;
 
   const { exp } = await ConfigUtils.readConfigJsonAsync(root);
+  if (exp.owner) {
+    username = exp.owner;
```
fson: As you mentioned before, `getPublishInfoAsync` gets called in a lot of places. Most of them seem harmless and don't even read the `username`. But it's also called [here](https://github.com/expo/expo-cli/blob/b12ba961c46e1bea014298c23dd33c66df7746a0/packages/xdl/src/Project.ts#L1965-L1971) to get the arguments passed to the `signManifest` API as-is. I'm not sure what happens when we now pass the `owner` (if there's one) in place of the username we passed before.

I think it might be safer to make a new function to implement the "`owner` or `username`" logic and leave `getPublishInfoAsync` unchanged. The new function could be added in the `Credentials` module and have a signature like this:
```ts
import { Platform } from '@expo/config';

async function getCredentialMetadataAsync(projectRoot: string, platform: Platform): CredentialMetadata;
```
where `CredentialMetadata` is already defined as: https://github.com/expo/expo-cli/blob/b12ba961c46e1bea014298c23dd33c66df7746a0/packages/xdl/src/credentials/Credentials.ts#L6-L11

Then you can use that function [here](https://github.com/expo/expo-cli/blob/83a758a75892200b3bb517e389ea8093e16e23a2/packages/expo-cli/src/commands/fetch/ios.js#L21), [here](https://github.com/expo/expo-cli/blob/83a758a75892200b3bb517e389ea8093e16e23a2/packages/expo-cli/src/commands/upload/IOSUploader.js#L125-L130), [here](https://github.com/expo/expo-cli/blob/b12ba961c46e1bea014298c23dd33c66df7746a0/packages/expo-cli/src/commands/build/AndroidBuilder.js#L67-L71) and [here](https://github.com/expo/expo-cli/blob/b12ba961c46e1bea014298c23dd33c66df7746a0/packages/expo-cli/src/commands/build/AndroidBuilder.js#L116-L120), instead of `getPublishInfoAsync`, which I'd rather see go away eventually and be split to more meaningful pieces, as this comment might suggest as well:
https://github.com/expo/expo-cli/blob/83a758a75892200b3bb517e389ea8093e16e23a2/packages/xdl/src/Exp.ts#L299

PR #1044, opened 2019-09-29
[expo-cli][init] Added --yes command to skip questions 

EvanBacon:
`npm init` has a command that lets you skip all the questions and create a basic project in the current directory. This adds that similar functionality to `expo-cli`.

### DIFF
--- a/packages/expo-cli/src/commands/init.ts
+++ b/packages/expo-cli/src/commands/init.ts
@@ -29,6 +29,7 @@ type Options = {
   androidPackage?: string;
   iosBundleIdentifier?: string;
   parent?: Command;
+  yes?: true;
 };
 
 const FEATURED_TEMPLATES = [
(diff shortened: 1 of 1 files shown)


ide:
Perhaps we should just use non-interactive mode to enable this behavior -- @fson what do you think?

fson:
```
expo init --yes
```
is currently synonymous to
```
expo init . --template blank --non-interactive --name "Crossy Road"
```
I think we could make `--name` non-mandatory managed apps. Like iOS `bundleIdentifier` or Android `package`, it's not necessary until you want to build a standalone app, at which point we can prompt it if it's not defined yet.
We could also have `--template` default to `blank` for non-interactive mode.

After these changes, you could run `expo init . --non-interactive` for the same effect. Not as fast to type, but also allows adding options as necessary, e.g. `expo init hello-world --non-interactive --template expo-template-blank-typescript`. Maybe `--yes` could then be added as an alias of `--non-interactive` for brevity?

PR #1046, opened 2019-09-29
feat: Allow custom webpack config & static dir paths

artalat:
Right now we can only use `webpack.config.js` and `web` directory from the project root. There is no way to use these files in other locations, e.g. in nested folders. This PR changes this behavior by allowing custom paths through `app.json`.

```json
{
  "expo": {
    ...

    "web": {
      "staticDir": "foo/static",
      "webpackConfigPath": "foo/webpack.config.js"
    }
  }
}
```

### DIFF
--- a/packages/webpack-config/src/utils/paths.ts
+++ b/packages/webpack-config/src/utils/paths.ts
@@ -55,7 +55,7 @@ export function getPossibleProjectRoot(): string {
 
 function parsePaths(
   projectRoot: string,
-  { exp: nativeAppManifest, pkg }: ConfigUtils.ExpoConfig
+  { exp: nativeAppManifest, pkg, web }: ConfigUtils.ExpoConfig
 ): FilePaths {
   const inputProjectRoot = projectRoot || getPossibleProjectRoot();
 
@@ -118,7 +118,8 @@ function parsePaths(
   const productionPath = absolute(config.web.build.output);
 
   function templatePath(filename: string = ''): string {
-    const overridePath = absolute('web', filename);
+    const staticDir = web.staticDir || 'web';
+    const overridePath = absolute(staticDir, filename);
     if (fs.existsSync(overridePath)) {
       return overridePath;
     }
--- a/packages/xdl/src/Web.ts
+++ b/packages/xdl/src/Web.ts
@@ -113,7 +113,9 @@ export async function invokeWebpackConfigAsync(
   argv?: string[]
 ): Promise<WebpackConfiguration> {
   // Check if the project has a webpack.config.js in the root.
-  const projectWebpackConfig = path.resolve(env.projectRoot, 'webpack.config.js');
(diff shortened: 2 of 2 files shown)


packages/webpack-config/src/utils/paths.ts
```diff
@@ -55,7 +55,7 @@ export function getPossibleProjectRoot(): string {
 
 function parsePaths(
   projectRoot: string,
-  { exp: nativeAppManifest, pkg }: ConfigUtils.ExpoConfig
+  { exp: nativeAppManifest, pkg, web }: ConfigUtils.ExpoConfig
```
EvanBacon: This type is wrong, and `web` doesn't exist on the object.

EvanBacon:
Perhaps just re-export your Webpack config from the root Webpack config. Maybe it would be useful if you provided your use case for this feature.

artalat:
We have a large project of a universal app. The project has multiple target platforms (i.e expo, electron, web(our own setup that we plan on deprecating in favor of expo) as well as storybook), but a single codebase. Each of these has its own webpack config files and static folders. So to organize we have a dedicated folder for each of these platforms:

```
/web
/expo
/electron
/storybook
```

I feel this isn't something unique to us, and a lot of people with multiple integrations in the same project would feel the need for this feature.

artalat:
> This PR is missing a lot of features and wouldn't work. A feature change of this size would require a test as well. Regardless I don't think it's a good idea to change the `webpack.config.js` path.

Would be happy to contribute if you can point in the right direction please.

EvanBacon:
Related https://github.com/expo/examples/pull/23

artalat:
@EvanBacon is this something we can work on? I don't think having this functionality is wrong, we're already doing the same with `app.json` file through the `--config` flag. 

This would help us in our project a lot.

brentvatne:
@EvanBacon - when you get a chance can you decide on how we should proceed with this?

brentvatne:
@artalat - sorry for the delay - if you can resolve conflicts we can land this!

artalat:
@brentvatne would love to do it, but I'm busy for next 2 weeks. Will try to look into it after that.

BTW. @EvanBacon mentioned it's missing some features and will not work. Any pointers I can get to make it work will be very helpful and much appreciated. Thanks.

EvanBacon:
I'm thinking maybe a better approach could be to use `webpack-cli` directly. Or we could surface the interface in expo/xdl rather than making this available in the `app.config` initially. 
`expo start:web` is a very thin wrapper around `webpack-cli` that enforces some defaults like `@expo/webpack-config`. 

I think it would be best to avoid adding bundler specific properties to the Expo config spec since Expo shouldn't really be coupled to a bundler. Also `expo/webpack-config` can be used for modified versions of React Native apps, so putting webpack config values in the web object wouldn't be the best place either.

EvanBacon:
Closing due to inactivity, feel free to open in the future.

PR #1058, opened 2019-10-02
Automatically exclude fonts from webpack-config

EvanBacon:
Revise the issue where fonts can sometimes be loaded twice when mixing with a third-party config like Storybook, Gatsby, Next, etc...

```diff
withUnimodules(
      otherConfig,
      { projectRoot: '...' },
-      { supportsFontLoading: false }
)
```


packages/webpack-config/src/utils/loaders.ts
```diff
@@ -0,0 +1,201 @@
...
+  loader: RuleSetUseItem;
+  loaderIndex: number;
+}
+
+export function findLoader(loaderName: string, rules: RuleSetRule[]): RuleSetRule | null {
+  for (const rule of rules) {
+    if (
+      rule.use &&
+      (rule.use as any).loader &&
+      ((rule.use as any).loader.includes(`/${loaderName}`) ||
```
hesyifei:
```suggestion
      ((rule.use as RuleSetLoader).loader!.includes(`/${loaderName}`) ||
```

PR #1060, opened 2019-10-03
Created @expo/next-adapter

EvanBacon:
This library will act as the main entry point for Next.js projects using Unimodules.

# Test Plan

Updated example to use this library as the entry point


hesyifei:
We should also remove `nextJsDocument` from `Webpack.ts` in `xdl` https://github.com/expo/expo-cli/blob/e85fefb8ee5e428388ec87cb6ffe12a0e19718f4/packages/xdl/src/Webpack.ts#L614.

Also what do you think would be the best way to handle the service worker file?

EvanBacon:
Might make sense too slowly deprecate it while we transition to using this new library. I have no preference on the service worker at this time.

fson:
I think this is a step in the right direction!

Going forward, I'd like to see more of the Next.js specific code in XDL extracted to a separate optional plugin rather than shipped with the core Expo CLI distribution (even when you don't need it). Here's what we still have in XDL:
https://github.com/expo/expo-cli/blob/519ed266ce3803d9076f1186112d76149dd5f52e/packages/xdl/src/Webpack.ts#L489-L692
This not would not only make the installation package of the CLI smaller, potentially (and more importantly) enable projects to upgrade Next.js independently of Expo CLI.

Nice work 👍

packages/next-document/README.md
```diff
@@ -0,0 +1,98 @@
...
+yarn add @expo/next-document
+```
+
+## ⚽️ Usage
+
+Re-export this component from the `pages/_document.js` file of your Next.js project.
+
+```js
+// pages/_document.js
+import Document from '@expo/next-document';
```
fson: Since the Next.js server that's currently hard-coded in XDL is strongly coupled with Next.js (even if `next` is not a direct dependency of XDL), maybe we should already plan to extract it out as well? Then a more appropriate name for this package might be something like `@expo/next-plugin`?
```suggestion
import { Document } from '@expo/next-plugin';
```

EvanBacon:
I think our service-worker can/should be included via the `next-offline` plugin

Issue #997, opened 2019-09-09
[Expo Web] Asset(s) exceed the recommended size limit (244 KiB).


LouisJS:
## Description

I have set a CI in Gitlab that build my app (expo build:web) and push it to my server.
Which have this consequence :

> Treating warnings as errors because process.env.CI = true.
> Most CI servers set it automatically.
> 

## Observed Behavior

As it said, warning are treated as error, and i cannot disable it, so i end up with

```
Failed to compile.

entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  app (2.26 MiB)
      static/js/runtime~app.efa64587.js
      static/js/2.dec8fd5e.chunk.js
      static/js/app.f5840f5e.chunk.js

ERROR: Job failed: exit code 1
```
## Investigation

The error leads me to this [github](https://github.com/webpack/webpack/issues/3486) issue related to WebPack configuration.
So i've tried to run `expo customize:web` to custom my webpack.config.js

But i'm facing the `Cannot read property 'readConfigJsonAsync' of undefined` error for now.

## Question

How can i get over the `expo customize:web` issue ?
What line should i change in the generated webpack to solve my issue ?

Best regards,
Louis L

PR #1071, opened 2019-10-04
[webpack-config] Disable performance warning in CI

EvanBacon:
fix https://github.com/expo/expo-cli/issues/997


LouisJS:
Thanks a lot ! 💪✨
The GOAT as always !

PR #1075, opened 2019-10-06
Added MJS support

EvanBacon:


davidbiller:
@EvanBacon 
Still get this error with expo-cli 3.5.0

.../node_modules/graphql/index.mjs 49:0-53:205
Can't reexport the named export 'BREAK' from non EcmaScript module (only default export is available)


should fix this right?

Issue #449, opened 2019-03-21
[web] Adjust service worker to use network caching policy by default


ide:
Currently, the service worker has a longer caching policy than is specified in HTTP response headers. When you redeploy a site and visit the page, you still see the old site for awhile or until you force-reload the page.

The service worker defaults should prioritize correctness and understandability -- when I deploy the site, if the server has the new code, I should see the new site on my next visit. When I change HTTP caching headers, they should take effect as if the service worker weren't there unless the developer explicitly takes action to opt into murkier caching.

danmaas:
I just ran into this issue when deploying an Expo project as a PWA with `expo build:web`.

My web server is configured to return a strict Cache-Control header on `index.html` and `service-worker.js` that tells the browser never to cache these files:

```
Cache-Control: max-age=0,no-cache,no-store,must-revalidate
```

Despite this, I often see an outdated version of index.html when visiting the app. I usually have to clear the browser cache and/or un-register the service worker in order to see updates. I'm not 100% sure, but it seems like the service worker is not obeying this header and returning a stale version of `index.html`.

As a work-around, I created a `webpack.config.js` that tweaks Expo's webpack configuration to adjust the Google Workbox parameters, telling Workbox never to cache these two files. Complete contents of `webpack.config.js` follows:

```
// partially override expo-cli's built-in webpack config to fix some
// errors in production builds.
// note: this only affects the expo web build, not native/mobile.

const expoConfig = require('@expo/webpack-config');

module.exports = async function(env, argv) {
    let conf = await expoConfig(env, argv);
    
    // adjust Google Workbox (service worker) config to avoid caching problems
    if(conf['plugins']) {
        conf['plugins'].forEach((plugin) => {
            // detect workbox plugin
            if(plugin['config'] && plugin['config']['swDest'] === 'service-worker.js') {
                // tell it never to cache index.html or service-worker.js
                plugin['config']['exclude'].push(/index.html/);
                plugin['config']['exclude'].push(/service-worker.js/);

                // (optional) tell it to start new service worker versions immediately, even if tabs
                // are still running the old one.
                plugin['config']['skipWaiting'] = true;
            }
        });
    }

    return conf;
};
```

nandorojo:
I'm still facing this issue with the latest expo sdk and CLI.

deathemperor:
for anyone encounter this issue, the solution can be found here https://www.npmjs.com/package/@expo/webpack-config#fully-disabling-the-service-worker.

[...]

PR #1082, opened 2019-10-08
Added offline support for expo web projects

EvanBacon:
Because this changes a lot of static files, we should release this as a major bump.
fix https://github.com/expo/expo-cli/issues/449


DrKylstein:
I'm still having issues like the ones marked as fixed by this change, on expo-cli 3.11.5. Is it necessary to enable this somehow?

PR #1085, opened 2019-10-08
fix(dev-tools): fix publish failing when "optimize assets" is selected

fson:
Not passing an options object caused an error in `Project.optimizeAsync`.

Changed the quality option to be non-mandatory so that the default needs to be defined in just one place and made the options object optional.


ide:
Are there some TS type hints we could easily specify (without making things too complicated) that would prevent this from regressing?

fson:
TypeScript compiler would have caught this issue, but the bug was in `@expo/dev-tools`, which is still in JS, so it wasn't caught.

PR #1090, opened 2019-10-09
[web] fixed screen height

oxyii:
ref [expo/expo#5081](https://github.com/expo/expo/issues/5081)

### DIFF
--- a/packages/webpack-config/web-default/index.html
+++ b/packages/webpack-config/web-default/index.html
@@ -19,8 +19,9 @@
         -webkit-overflow-scrolling: touch;
         margin: 0px;
         padding: 0px;
-        /* Allows content to fill the viewport and go beyond the bottom */
-        min-height: 100%;
+        /* DO NOT Allows content to fill the viewport and go beyond the bottom */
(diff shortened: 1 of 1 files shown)


packages/webpack-config/web-default/index.html
```diff
@@ -19,15 +19,18 @@
         -webkit-overflow-scrolling: touch;
         margin: 0px;
         padding: 0px;
-        /* Allows content to fill the viewport and go beyond the bottom */
-        min-height: 100%;
+        /* DO NOT Allows content to fill the viewport and go beyond the bottom */
```
EvanBacon: Fix grammer

EvanBacon:
This may cause major visual regressions, we should test against changes like this in the future.

packages/webpack-config/web-default/index.html
oxyii: @EvanBacon My english is bad. Please do it your-self.

FiberJW:
This causes a lot of damage, so I'm going to revert this. Sorry for the confusion :(

PR #1092, opened 2019-10-09
Auto use expo entry point in native projects

EvanBacon:
## Why

Add ability to run a project with the bare minimum #1101

## How

- [x] Fallback to `node_modules/expo/AppEntry.js` if the `main` field in the `package.json` is missing. https://github.com/expo/expo-cli/pull/1092/commits/6bb4324ac62d3c1c96faee88ae5f871bf793f083


packages/xdl/src/Exp.ts
```diff
@@ -37,6 +39,23 @@ export async function determineEntryPointAsync(root: string) {
     entryPoint = exp.entryPoint;
   }
 
+  const userDefinedEntryPointExists = await fs.pathExists(path.resolve(projectRoot, entryPoint));
+  if (!userDefinedEntryPointExists) {
+    entryPoint = ConfigUtils.resolveModule('expo/AppEntry', projectRoot, exp);
+    const expoEntryPointExists = await fs.pathExists(entryPoint);
+    // Remove project root from file path
+    entryPoint = entryPoint.split(projectRoot).pop();
```
fson:
```suggestion
    entryPoint = path.relative(projectRoot, entryPoint);
```

PR #1102, opened 2019-10-10
Add more customize options

EvanBacon:
## Why

Now that we're making it optional to use a lot of files it'd be nice to have a way to regenerate them.

## How

- Added jest, next, storybook


packages/expo-cli/src/commands/customize.ts
```diff
@@ -31,9 +31,38 @@ async function maybeWarnToCommitAsync(projectRoot: string) {
   }
 }
 
-const dependencyMap: { [key: string]: string[] } = {
-  'babel.config.js': ['babel-preset-expo'],
-  'webpack.config.js': ['@expo/webpack-config'],
+const dependencyMap: {
```
fson: Not certain how this works, so please correct me, but I wonder if this configuration should live in the same package as the templates itself. Right now if you edit the templates, you will also need to edit this file in `expo-cli`, which consumes the templates?

PR #1103, opened 2019-10-10
default sdkVersion to expo package version

EvanBacon:
If the `sdkVersion` is not defined in the `app.json` then we should attempt to read it from the installed `expo` version.

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -149,7 +149,7 @@ export function getConfigForPWA(
   getAbsolutePath: (...pathComponents: string[]) => string,
   options: object
 ) {
-  const { exp } = readConfigJson(projectRoot, true);
+  const { exp } = readConfigJson(projectRoot, true, true);
   return ensurePWAConfig(exp, getAbsolutePath, options);
 }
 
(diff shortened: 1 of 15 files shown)


packages/config/src/Config.ts
```diff
@@ -555,6 +563,14 @@ export async function readConfigJsonAsync(
     exp.version = pkg.version;
   }
 
+  if (exp && !exp.sdkVersion) {
```
ide: move this logic into a shared function

packages/config/src/Config.ts
```diff
@@ -501,6 +501,14 @@ export function readConfigJson(
...
+    try {
+      const packageJsonPath = resolveModule('expo/package.json', projectRoot, exp);
+      const expoPackageJson = require(packageJsonPath);
+      exp.sdkVersion = expoPackageJson.version;
```
ide: need to zero out the minor and patch versions. could cause issues if people have sdkVersion: 35.0.1 unintentionally. so for now i think zeroing out the version is safest.

packages/xdl/src/Env.ts
```diff
@@ -13,3 +13,7 @@ export function isStaging(): boolean {
 export function isLocal(): boolean {
   return getenv.boolish('EXPO_LOCAL', false);
 }
+
+export function skipManifestValidation(): boolean {
```
ide:
```suggestion
export function maySkipManifestValidation(): boolean {
```

packages/config/src/Config.ts
```diff
@@ -501,6 +501,14 @@ export function readConfigJson(
...
+    } catch (_) {}
```
fson: Maybe we can throw an error here, because that allows us to guarantee that the configuration object returned by this function *always* has `sdkVersion`. IMHO this is better than failing in a weird way later.

We can make the error describe how to fix this, i.e. run npm/Yarn to ensure `expo` is installed in `node_modules`.

PR #1111, opened 2019-10-11
feat(expo-cli): ask to close expo when upgrading

### LINKED ISSUE #1108: Warn if server is running when you run upgrade command and prompt to close it
The first step in "Upgrading Your App" in the [release notes](https://blog.expo.io/expo-sdk-35-is-now-available-beee0dfafbf4) is to "Close your Expo CLI server". We should get rid of this by instead checking if a server is running for the project and asking the user if they want us to close it. If they don't., then we stop the upgrade.

byCedric:
Fixes #1108

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -157,7 +157,7 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     });
 
     if (!answer.closeExpo) {
-      return;
+      throw new CommandError('EXPO_RUNNING_FOR_UPGRADE_COMMAND');
     }
 
     await Project.stopAsync(projectRoot);
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -1,5 +1,5 @@
 import { Command } from 'commander';
-import { Versions } from '@expo/xdl';
+import { Project, Versions } from '@expo/xdl';
 import JsonFile from '@expo/json-file';
 import * as ConfigUtils from '@expo/config';
 import chalk from 'chalk';
@@ -145,6 +145,23 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
       log.error('No sdkVersion field is present in app.json, cannot upgrade project.');
       throw new CommandError('SDK_VERSION_REQUIRED_FOR_UPGRADE_COMMAND_IN_MANAGED');
     }
+  }
+
+  // Can't upgrade if Expo is running
+  let status = await Project.currentStatus(projectRoot);
+  if (status === 'running') {
+    let answer = await prompt({
+      type: 'confirm',
+      name: 'closeExpo',
+      message: 'Expo needs to close before we can upgrade your dependencies. Do you want to close it?',
+    });
+
+    if (!answer.closeExpo) {
+      return;
+    }
+
+    await Project.stopAsync(projectRoot);
+    log.newLine();
   }
 
   let currentSdkVersionString = exp.sdkVersion;
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -150,18 +150,9 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
   // Can't upgrade if Expo is running
   let status = await Project.currentStatus(projectRoot);
   if (status === 'running') {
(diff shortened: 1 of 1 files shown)


byCedric:
Just a random side note, but it seems that Circle isn't running the tests for non-expo-members. I mean, I'm happy to contribute but not running those tests "kind of" makes me a bit nervous 😅

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -147,6 +147,23 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
...
+  let status = await Project.currentStatus(projectRoot);
+  if (status === 'running') {
+    let answer = await prompt({
+      type: 'confirm',
+      name: 'closeExpo',
+      message: 'Expo needs to close before we can upgrade your dependencies. Do you want to close it?',
+    });
+
+    if (!answer.closeExpo) {
+      return;
```
fson: In this case the exit status of the command should be non-zero. You can, for example, throw a `CommandError` to achieve this.

byCedric: Thanks! Will do in about 3-4 hours!

byCedric: I added it, not sure if it has the proper name though, maybe `BAD_CHOICE` is also a good one? I didn't add it to the [error codes](https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/CommandError.ts#L5) because the other two command errors are not listed there. But I'd be happy to add it if necessary.

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -147,6 +147,23 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     }
   }
 
+  // Can't upgrade if Expo is running
+  let status = await Project.currentStatus(projectRoot);
+  if (status === 'running') {
+    let answer = await prompt({
+      type: 'confirm',
+      name: 'closeExpo',
+      message: 'Expo needs to close before we can upgrade your dependencies. Do you want to close it?',
```
brentvatne:
```suggestion
      message: 'We found an existing expo-cli instance running for this project, it needs to be closed to continue.',
```

maybe something like this and then present two options:

```
- Close it for me and continue upgrade
- Cancel, I'm not ready yet
```

tbh i'm not sure how important this is though.. we could just close it for people and tell them that we closed it. it's not a destructive action, it costs nothing for them to close and reopen the packager.

byCedric: I don't think a lot of devs will even encounter this. Maybe just telling them that we closed it is sufficient and leaves some extra room for other questions it might encounter in the future 😄

brentvatne: sounds good

brentvatne: approved so we can ship it once you have made this change!

packages/expo-cli/src/commands/upgrade.ts
fson: The error code doesn't matter all that much in this case – it's mostly useful for cases where we want to catch a specific error, e.g. a common pattern is like this:
```
try {
  doStuffThatMightFail();
} catch (error) {
  if (error.code === 'BAD_CHOICE') {
    trySomethingElse();
  } else {
    throw error;
  }
}
```
Tl;dr: `BAD_CHOICE` is a good choice! 😆

byCedric: Thanks Ville! ❤️  I'll remember that next time 😄 

For now, I think this specific instance is superseded by Brent's suggestion to just tell instead of ask, right?

packages/expo-cli/src/commands/upgrade.ts
byCedric: Done! 😄

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -147,6 +147,15 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     }
   }
 
+  // Can't upgrade if Expo is running
+  let status = await Project.currentStatus(projectRoot);
+  console.log({ status });
```
brentvatne:
```suggestion
```

byCedric: My bad, i'll remove it right away

byCedric:
My bad, it's gone and won't be back!

PR #1112, opened 2019-10-11
Ask to upgrade simulator and android clients when upgrading

### LINKED ISSUE #1109: Prompt user to install new version of client rather than just tell them there is a new version
When you have an outdated version of the client app in your simulator or emulator we should give users the ability to install the updated version in one or two keystrokes rather than having to manually remove the app from the simulator.

In cases where we can't do anything (physical iOS device, Android device not connected to machine) we should recommend people get the new version from the store or maybe do nothing? Up to the discretion of the implementer.

This will allow us to get rid of half of this step from "Upgrade Your App" bit of the release notes:

> Update the Expo app on your phones from the App Store / Google Play. expo-cli will automatically update your apps in simulators if you delete the existing apps, or you can run expo client:install:ios and expo client:install:android.

will become

> Update the Expo app on your phones from the App Store / Google Play. expo-cli will automatically prompt you to update the app in simulators.

byCedric:
Fixes #1109

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -232,7 +232,7 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     let answer = await prompt({
       type: 'confirm',
       name: 'upgradeSimulator',
-      message: 'You might have to upgrade your iOS simulator, but before you can do that, you have to run the simulator. Do you want to upgrade it later?',
+      message: 'You might have to upgrade your iOS simulator. Before you can do that, you have to run the simulator. Do you want to upgrade it now?',
       default: false,
     });
 
@@ -251,7 +251,7 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     let answer = await prompt({
       type: 'confirm',
       name: 'upgradeAndroid',
-      message: 'You might have to upgrade your Android client, but before you can do that, you have to run the emulator, or a device plugged in. Do you want to upgrade it later?',
+      message: 'You might have to upgrade your Android client. Before you can do that, you have to run the emulator, or plug a device in. Do you want to upgrade it now?',
       default: false,
     });
 
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -227,8 +227,10 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     }
   }
 
+  const platforms = exp.platforms || [];
+
   // Check if we can, and probably should, upgrade the (ios) simulator
-  if (Simulator.isPlatformSupported()) {
+  if (Simulator.isPlatformSupported() && platforms.includes('ios')) {
     let answer = await prompt({
       type: 'confirm',
       name: 'upgradeSimulator',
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -219,6 +219,42 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
...
+      if (!result) {
+        log.error('The upgrade of your simulator didn\'t go as planned. You might have to reinstall it manually with expo client:install:ios.');
+      }
+    }
+
+    log.newLine();
+  }
+
+  // Check if we can, and probably should, upgrade the android client
+  if (Android.isPlatformSupported()) {
```
byCedric: Should we check on `Android.assertDeviceReadyAsync()` here too and just log "remember, your Android client might need some love too"? When I tried it errored out because no Android device was connected. It might be a bit too tempting to just say "yes" for a developer.

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -219,6 +219,42 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     }
   }
 
+  // Check if we can, and probably should, upgrade the (ios) simulator
+  if (Simulator.isPlatformSupported()) {
```
byCedric: I don't think there is a (public) method available to double-check if an upgrade is necessary right?

packages/expo-cli/src/commands/upgrade.ts
brentvatne: yeah this one is a bit tricky.. maybe we default for both of these to upgrading later? so the `yes` people can get through it quickly. we also might want to tell people that they need to plug in their android device or open their emulator

byCedric: Sorry for the delay, I set the default to false and add a bit more info when you could perform the step. It's still a prompt, mostly to make developers "more-aware" of this. Personally, I think it's the best compromise between forcing them do to something but also protect them from skipping everything 😄 

Screenshot 2019-11-19 at 00 22 58

packages/expo-cli/src/commands/upgrade.ts
brentvatne: not that i am aware of, cc @fson

brentvatne:
one other small thing - it may be worth us checking if the user has explicitly listed supported platforms in app.json and skipping upgrading emulators if the platform isn't listed.

eg: a brand new project includes this in app.json

```
"platforms": [
  "ios",
  "android",
  "web"
],
```

if we changed it to this:

```
"platforms": [
  "ios",
  "web"
],
```

and then ran the upgrade command, we should only be prompted to update the ios client.

packages/expo-cli/src/commands/upgrade.ts
byCedric: It's ok, might be a nice to have but not necessary now It think.

byCedric:
Done! 😄 It might be better if we add this in the simulator/android or even project XDL. But right now I didn't have to reuse it. I mean, if a developer wants to install another ios client while ios is disabled, I don't think we should "block" them right?

But if you want to make it resuable, I suggest adding this?
- `Simulator.isPlatformEnabled(project: ExpoConfig): boolean`
- `Android.isPlatformEnabled(project: ExpoConfig): boolan`

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -227,6 +227,46 @@ async function upgradeAsync(requestedSdkVersion: string | null, options: Options
     }
   }
 
+  const platforms = exp.platforms || [];
```
byCedric: The fallback would also be a great reason for moving this to `.isPlatformEnabled` I think, since it may or may not be `undefined`.

PR #1113, opened 2019-10-12
feat(expo-cli): use npm-check or yarn-check to interactively update 3rd party packages

christianjuth:
# Why
`expo update` currently offers no way to update 3rd party packages. This is an attempt at making it easier for the user to update 3rd party packages WITHOUT effecting Expo SDK packages since specific versions of those packages are required by the SDK. 

**TL;DR** This script should only update "safe to update" packages (aka any package that Expo SDK does not require a specific supported version of).

# How
I am using `npm-check` or `yarn-check` depending on the project configuration to interactively allow the user to update "safe to update" packages. Any package that is not safe to update (any package returned by the function `getUpdatedDependenciesAsync()` will be excluded from the interactive update.

# Test Plan
I built and installed `expo-cli` locally using `npm link`. As far as I can tell, everything is working, but this should probably be tested on multiple npm/node versions and debugged more thoroughly.

### DIFF
--- a/packages/expo-cli/package.json
+++ b/packages/expo-cli/package.json
@@ -97,6 +97,7 @@
     "klaw-sync": "6.0.0",
     "lodash": "4.17.15",
     "match-require": "2.1.0",
+    "npm-check": "^5.9.0",
     "npm-package-arg": "6.1.0",
     "open": "6.3.0",
     "ora": "3.4.0",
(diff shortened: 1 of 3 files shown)


christianjuth:
I think I should adopt a "measure twice cut once" mentality and not be so quick to hit commit. Sorry those last two commits were sloppy

brentvatne:
I think we do a lot here automatically so I'm going to close this for now. It's a good idea and perhaps we will revisit in the future!

PR #1118, opened 2019-10-14
[xdl] Add support for Facebook buildtime customizations

sjchmiela:
x-ref https://github.com/expo/expo/pull/5924

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -739,14 +739,17 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
   }
 
   // Add shell app scheme
-  if (scheme) {
+  const schemes = [scheme, manifest.facebookScheme].filter(e => e);
+  if (schemes.length > 0) {
     const searchLine = isDetached
       ? '<!-- ADD DETACH SCHEME HERE -->'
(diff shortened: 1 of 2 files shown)


packages/xdl/src/detach/IosNSBundle.js
```diff
@@ -311,6 +311,27 @@ async function _configureInfoPlistAsync(context) {
...
 
+    // if the developer has set a non-empty value or if Facebook SDK will want to autoinitialize itself
+    // (given the default is to autoinitialize and all the prerequisites are there)
+    if (config.facebookAutoInitEnabled !== undefined || config.facebookAppID) {
+      infoPlist.FacebookAutoInitEnabled = config.facebookAutoInitEnabled || false;
+    } else {
+      delete infoPlist['FacebookAutoInitEnabled'];
+    }
+
+    if (config.facebookAutoLogAppEventsEnabled !== undefined) {
```
ide: Just to be clear, if auto-initialization is disabled (set to false), then auto-logging app events does not do anything and the Facebook SDK will not collect data?

sjchmiela: With all the Facebook features enabled:

Zrzut ekranu 2019-10-15 o 14 22 51

With all the Facebook features enabled BUT auto init:

Zrzut ekranu 2019-10-15 o 14 22 09

PR #1121, opened 2019-10-15
[xdl] Removed unused ts-ignores

EvanBacon:
- restructure imports
- remove unused imports


packages/xdl/src/Env.ts
```diff
@@ -1,6 +1,5 @@
-import os from 'os';
-
 import getenv from 'getenv';
+import os from 'os';
```
fson: There's a reason I sort imports like this:
```ts
import os from 'os';

import getenv from 'getenv';
```
It's an old habit I picked up from PEP8 (the Python coding standard) many years ago. In [PEP8:](https://www.python.org/dev/peps/pep-0008/#imports)
> Imports should be grouped in the following order:
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
>
> You should put a blank line between each group of imports.

I like this because it makes it clear which modules are from Node and which are 3rd party dependencies and brings some organization to the list, if it's very long. On the other hand, it's not a big deal, I don't think anyone else is doing this and it might be hard to automate in the editors, so it's probably not applied consistently across the code base.

If you think it's better to just sort everything alphabetically, I guess that's fine. Just wanted to share some background 😆 

I think a more interesting question is, whatever the style we use for this, can we make Prettier (or Eslint etc) do this sorting business for us and never have to think about it again?

PR #1122, opened 2019-10-15
Convert upload command to TS

EvanBacon:


packages/expo-cli/src/commands/upload/IOSUploader.ts
```diff
@@ -110,16 +121,19 @@ export default class IOSUploader extends BaseUploader {
...
 
-    return teamId;
+    if (credential) {
+      return credential.teamId;
+    }
+    return;
   }
 
-  async _getAppleIdCredentials() {
+  async _getAppleIdCredentials(): Promise<IosPlatformOptions> {
```
fson: This function does not return a full `IosPlatformOptions` object, it seems, only the subset `{  appleId: string;  appleIdPassword: string }` of it.
```suggestion
  async _getAppleIdCredentials(): Promise<Pick<IosPlatformOptions, 'appleId', 'appleIdPassword'>> {
```

packages/expo-cli/src/commands/upload/IOSUploader.ts
```diff
@@ -110,16 +121,19 @@ export default class IOSUploader extends BaseUploader {
...
     const appleCredsKeys = ['appleId', 'appleIdPassword'];
-    const result = pick(this.options, appleCredsKeys);
+    const result: any = pick(this.options, appleCredsKeys);
```
fson: It's tricky that the type definitions of lodash make the return type `Partial< IosPlatformOptions>` here, which is too broad. Using `any` here is not ideal, it masked the error that showed the return type of this function was wrong.

Maybe an explicit cast would be slightly better?
```suggestion
    const result: any = pick(this.options, appleCredsKeys) as Pick< IosPlatformOptions, 'appleId' | 'appleIdPassword'>;
```
or maybe drop the `lodash.pick`, since  it gives a lot of TypeScript troubles, but little benefit in a simple case like this:
```suggestion
    const result = { appleId: this.options.appleId, appleIdPassword: this.options. appleIdPassword };
```
I think I actually prefer the latter.

packages/expo-cli/src/commands/upload/IOSUploader.ts
```diff
@@ -110,16 +121,22 @@ export default class IOSUploader extends BaseUploader {
...
   }
 
-  async _getAppleIdCredentials() {
+  async _getAppleIdCredentials(): Promise<Pick<IosPlatformOptions, 'appleId' | 'appleIdPassword'>> {
     const appleCredsKeys = ['appleId', 'appleIdPassword'];
-    const result = pick(this.options, appleCredsKeys);
+    const result: Partial<Pick<IosPlatformOptions, 'appleId' | 'appleIdPassword'>> = pick(
+      this.options,
+      appleCredsKeys
+    );
```
fson: I don't understand why this is `Partial`. In `IosPlatformOptions` both `appleId` and `appleIdPassword` are never `undefined`:
```
appleId: string;
appleIdPassword: string;
```
`lodash.pick` makes this whole thing really complicated. Also not sure why the `credsPresent` checks below are necessary.

```suggestion
    const result = { appleId: this.options.appleId, appleIdPassword: this.options.appleIdPassword };
```

fson: My guess is that the code was originally written to be more generic than it currently needs to be, but I'd simplify it because it's just two fields we are interested in. No need for `pick` or `intersection`.

EvanBacon: I incorrectly assumed that appleId and appleIdPassword were always defined.

PR #1123, opened 2019-10-15
hotfix for not a constructor function

FrenchBully:
#### Bug:
Running `expo start --web` fails to build

Error Msg: `WebpackPWAManifestPlugin is not a constructor`

Resolution:
instantiate function by calling `.default`

Steps to reproduce:
1. `yarn install`
2. `expo start --web`

### DIFF
--- a/packages/webpack-config/src/webpack.config.ts
+++ b/packages/webpack-config/src/webpack.config.ts
@@ -204,7 +204,7 @@ export default async function(env: Environment, argv: Arguments = {}): Promise<C
         ROOT_ID: rootId,
       }),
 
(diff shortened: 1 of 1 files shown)


EvanBacon:
I cannot repro, maybe your version of `expo-cli` is not updated?

EvanBacon:
Closing as I don't see this being an issue.

kevinulrich:
We are running into the exact same issue with an application that we have upgraded to SDK version 35 and have modified the webpack configuration according to the docs: https://docs.expo.io/versions/latest/guides/customizing-webpack/

We cannot reproduce the issue either with a new application so I suspect this happened somewhere when upgrading.

PR #1124, opened 2019-10-15
Convert Eject to TS

EvanBacon:


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -14,17 +12,21 @@ import * as PackageManager from '../../PackageManager';
...
-import prompt from '../../prompt';
+import prompt, { Question } from '../../prompt';
 import { loginOrRegisterIfLoggedOut } from '../../accounts';
 
 const EXPO_APP_ENTRY = 'node_modules/expo/AppEntry.js';
 
-async function warnIfDependenciesRequireAdditionalSetupAsync(projectRoot: string) {
+async function warnIfDependenciesRequireAdditionalSetupAsync(projectRoot: string): Promise<void> {
   const { exp, pkg: pkgJson } = await ProjectUtils.readConfigJsonAsync(projectRoot);
+  if (!exp || !pkgJson) {
```
fson: Instead of `ProjectUtils.readConfigJsonAsync`, please use `ConfigUtils.readConfigJsonAsync` (from `@expo/config`), which never returns `null`.

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -180,18 +202,20 @@ async function ejectToBareAsync(projectRoot, options) {
...
   try {
-    let tempDir = temporary.directory();
+    const tempDir = temporary.directory();
     await Exp.extractTemplateAppAsync(templateSpec, tempDir, appJson);
     fse.copySync(path.join(tempDir, 'ios'), path.join(projectRoot, 'ios'));
     fse.copySync(path.join(tempDir, 'android'), path.join(projectRoot, 'android'));
-    let packageJson = fse.readJsonSync(path.join(tempDir, 'package.json'));
-    defaultDependencies = packageJson.dependencies;
+    const { dependencies } = JsonFile.read(path.join(tempDir, 'package.json'), { json5: true });
+    if (isStringObject(dependencies)) {
```
fson: Not sure about this – if the template has e.g. `{"dependencies": {"lodash": 4.0, "react-native": "0.61.2"} }` (an error, because the versions should be strings), we just silently ignore the dependencies.

It seems to me that we should at least print a warning if we ignore the dependencies from the template, since this means that the template has an error that needs to be fixed.

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -180,18 +202,20 @@ async function ejectToBareAsync(projectRoot, options) {
...
    */
   try {
-    let tempDir = temporary.directory();
+    const tempDir = temporary.directory();
     await Exp.extractTemplateAppAsync(templateSpec, tempDir, appJson);
     fse.copySync(path.join(tempDir, 'ios'), path.join(projectRoot, 'ios'));
     fse.copySync(path.join(tempDir, 'android'), path.join(projectRoot, 'android'));
-    let packageJson = fse.readJsonSync(path.join(tempDir, 'package.json'));
-    defaultDependencies = packageJson.dependencies;
+    const { dependencies } = JsonFile.read(path.join(tempDir, 'package.json'), { json5: true });
```
fson:
```suggestion
    const { dependencies } = JsonFile.read(path.join(tempDir, 'package.json'));
```
`package.json` can not be JSON5 (neither npm nor Yarn accepts it).

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -213,10 +237,17 @@ async function ejectToBareAsync(projectRoot, options) {
...
+  const combinedDependencies = { ...defaultDependencies, ...pkgJson.dependencies };
+
+  for (const dependenciesKey of [
+    'react',
+    'react-dom',
+    'react-native-web',
+    'react-native-unimodules',
+    'react-native',
+  ]) {
+    combinedDependencies[dependenciesKey] = defaultDependencies[dependenciesKey];
```
fson: What happens if the `package.json` of the project has a dependency (e.g. `react-native-web`), but it doesn't exist in the template? It seems that it would become `"react-native-web": undefined` and be removed from dependencies.

Should we check that the key exists in `defaultDependencies` before copying it?

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -129,18 +140,45 @@ export async function ejectAsync(projectRoot: string, options) {
...
+  if (typeof value !== 'object') {
+    throw new Error(`Dependency map is invalid, expected object but got ${typeof value}`);
+  }
+
+  const outputMap: DependenciesMap = {};
+  if (!value) return outputMap;
+
+  for (const key of Object.keys(value)) {
+    if (key in value) {
+      if (isStringOrNumber(value)) {
```
fson: `number` is not a valid version number.

The JSON I gave in the earlier [comment](https://github.com/expo/expo-cli/pull/1124#discussion_r334874750):
```json
{"dependencies": {"lodash": 4.0, "react-native": "0.61.2"} }
```
was just an example of *invalid* package.json content that should throw an error. Sorry if the comment was unclear!

PR #1125, opened 2019-10-15
Convert IosNSBundle to TypeScript

EvanBacon:


packages/xdl/src/detach/IosNSBundle.ts
```diff
@@ -419,15 +418,17 @@ async function _configureShellPlistAsync(context) {
     if (parseSdkMajorVersion(config.sdkVersion) < 28) {
       shellPlist.isShell = true;
     }
-    shellPlist.manifestUrl = context.published.url;
-    shellPlist.releaseChannel = context.published.releaseChannel;
-    if (context.data.testEnvironment) {
+    if (context.published) {
```
fson: Looks like the type of `StandaloneContext.published` is wrong: it's always defined in both `StandaloneContextUser` and `StandaloneContextService`, so this null check should be unnecessary.

packages/xdl/src/detach/IosAssetArchive.ts
```diff
@@ -3,23 +3,20 @@ import path from 'path';
...
-  StandaloneContextDataService,
-} from './StandaloneContext';
+import { StandaloneContextUser, StandaloneContextService } from './StandaloneContext';
 
 /**
  *  Compile a .car file from the icons in a manifest.
  */
 async function buildAssetArchiveAsync(
-  context: StandaloneContext,
+  context: StandaloneContextUser | StandaloneContextService,
```
fson:
```suggestion
  context: AnyStandaloneContext,
```
Maybe?

PR #1148, opened 2019-10-23
[expo-cli] First draft of guide for installing simulator versions

byCedric:
> This is still a WIP, I'm creating this PR to "validate" if what I'm doing is correct.

#### Why this?
This is an idea from @brentvatne, born in #1109. I think this might help developers who are using older Expo versions.

#### What does it do?
I prototyped the idea for the simulator for now. But if we are sure we want this, I can create an Android version too ofc.

1. Checked if we are in an Expo project
  a. If so, _and_ it's the latest, prompt to install latest client.
  b. If so, but it's not latest _and_ we have an `iosClientUrl`, prompt to install this client version.
  c. If so, but it's not latest and we don't have an `iosClientUrl`, warn about it
2. When not in an Expo project, prompt to install the latest client.
3. When the dev denies all the recommendations, list the clients he/she can install.
4. Install selected client.

#### Things to think about
- Not all SDK versions have an `iosClientUrl`, like SDK `34`. That means we can't auto-install or detect these, so for now, I filtered it out.
- `iosVersion` and `iosClientVersion` are used, I think the latter is the one I should pick right?
- Maybe it's better to separate the "guides" from the other client commands. But I have no clue what the implication is to the command-loader.

#### Show me what you got

<details>
<summary>From latest Expo project (scenario 1.a)</summary>

</details>

<details>
<summary>From older Expo project, with `iosClientUrl` (scenario 1.b)</summary>

</details>

<details>
<summary>From older Expo project without `iosClientUrl` (scenario 1.c)</summary>

</details>

<details>
<summary>From outside Expo project (2)</summary>

</details>


brentvatne:
> Not all SDK versions have an `iosClientUrl`, like SDK 34. That means we can't auto-install or detect these, so for now, I filtered it out.

We should fix this and add `iosClientUrl` to SDK 34. I think it should be a required field. That said, it's good that we've added some handling for the case where this doesn't exist so this is more resilient.

> From older Expo project without `iosClientUrl` (scenario 1.c)

In this case we should probably not show a  client for an older SDK version because it's unlikely to work.

One other case to consider - it may be possible to not have any compatible clients available, then we can just ask people to download the latest.

byCedric:
Thanks for the input ❤️ I updated the code to do the following:

- Filter older SDKs, they probably won't work with their project
- Add a fallback message when no client was selected/is available, and bail before install
- Label latest SDK with "latest", they aren't asked if they want to upgrade to the latest one if they have a version with a compatible client

<details>
<summary>Filter older SDK scenario</summary>

</details>

<details>
<summary>"Updated" from older Expo project, with `iosClientUrl` (scenario 1.b)</summary>

</details>

I'm don't think I'm able to help with the API and filling the `iosClientUrl`. Best thing I could do is scrape together some links that might help 😄 

version | platform | SDK | source | link
--- | --- | --- | --- | ---
`2.12.2` | android | `31..32` | [SDK 35 blog](https://blog.expo.io/expo-sdk-35-is-now-available-beee0dfafbf4#47d1) | https://d1ahtucjixef4r.cloudfront.net/Exponent-2.12.2.apk
`2.12.3` | ios | `31..32` | [SDK 35 blog](https://blog.expo.io/expo-sdk-35-is-now-available-beee0dfafbf4#47d1) | https://dpq5q02fu5f55.cloudfront.net/Exponent-2.12.3.tar.gz
`2.10.6` | android | `26..30` | [SDK 33 blog](https://blog.expo.io/expo-sdk-v33-0-0-is-now-available-52d1c99dfe4c#dab0) | https://exp-android-apks.s3.amazonaws.com/Exponent-2.10.6.apk
`2.10.0` | ios | `26..30` | [SDK 33 blog](https://blog.expo.io/expo-sdk-v33-0-0-is-now-available-52d1c99dfe4c#dab0) | https://exp-ios-simulator-apps.s3.amazonaws.com/Exponent-2.10.0.tar.gz
`2.8.0`? | android | `30`? | [Eric's post](https://medium.com/@esamelson/youll-need-to-update-to-the-latest-version-of-the-expo-client-app-2-8-0-83b299d2a371) | -
`2.8.0`? | ios | `30`? | [Eric's post](https://medium.com/@esamelson/youll-need-to-update-to-the-latest-version-of-the-expo-client-app-2-8-0-83b299d2a371) | -

brentvatne:
this looks great! i'll update `iosClientUrl` myself when this lands :) can you rebase?

byCedric:
Done, I'll get started on the android counterpart tomorrow. Do you want me to add it in this PR or a new one @brentvatne?

brentvatne:
let's just add it to this PR :)

byCedric:
I just added the Android version, it's almost an exact copy of the Simulator's version. Unfortunately, I'm still a bit concerned about the API not returning the `xClientUrl` and `xClientVersion` properties.

As you can see here, the `androidClientUrl` is provided, but the `androidClientVersion` is missing for SDK 33.

Screenshot 2019-11-14 at 00 16 04

Please don't merge it before that's resolved 😄 If you want, I can perform a final test run to make sure all scenarios are properly working after that!

brentvatne:
@byCedric - perhaps to handle that situation we should leave out the client version string

```
> 33.0.0 - client version unknown
```

usually better to have a human message than display undefined in error cases

brentvatne:
we might want to extract the functions for these commands out so that we can call them at other times. for example in https://github.com/expo/expo-cli/pull/1112

byCedric:
Again, sorry for the delay! I've added the `client version unknown` fallback.

<details>
<summary>Fallback in action!</summary>

</details>

I'd love to make it more reusable, but they are basically all prompts. The only thing that might be valuable is a `Versions.androidClientSdkVersionsAsync(): SdkVersions` (same for ios ofc.). With that, we don't have to filter and other dark things to the original sdk array. And maybe a `Versions.isNewestSdk(sdk: SdkVersion)`?

We can also create something like a `commands/utils/ClientUtils.ts` or `UpgradeUtils.ts` where we can add some prompts and maybe make the existing client code a bit more readable. Does that sound like something you had in mind?

brentvatne:
@byCedric - yeah that would be good! let's do that in a followup PR :)

packages/expo-cli/src/commands/client/index.js
```diff
@@ -12,8 +12,14 @@ import log from '../../log';
 import prompt from '../../prompt';
 import urlOpts from '../../urlOpts';
 import { PLATFORMS } from '../build/constants';
+<<<<<<< HEAD
```
brentvatne: merge conflict

byCedric: my bad, it was already fixed in [dc250a5](https://github.com/expo/expo-cli/pull/1148/commits/dc250a5c70e985d8385bf5f202e45412f1d26ee2) but my server didn't push it to the right remote.

brentvatne:
this is 🔥

PR #1149, opened 2019-10-23
fix: device logging with React Native v0.61.x

fson:
As of React Native 0.61.x, Metro prints console logs from the
device to console.

Managed apps have a separate remote logging implementation
included in the Expo SDK, so we can just ignore these logs
(otherwise they would be printed twice).

Currently this has to be done by looking for log lines that
look like device logs, i.e. they start with " LOG ", " WARN "
or similar. This is less than ideal, but I plan submit a PR to
Metro to make it use the custom log reporter (if configured)
for these logs so LogReporter is able to filter what gets
printed to the console.

Additionally, the AppRegistry logging was changed recently
(https://github.com/facebook/react-native/commit/1f25d3c4ce605fc3bd2a4deefc754bd07cfbf1cf),
so the code detecting these logs had to be changed as well.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -1585,7 +1585,7 @@ function _logPackagerOutput(projectRoot: string, level: string, data: object) {
 
 function _isIgnorableMetroConsoleOutput(output: string) {
   // As of React Native 0.61.x, Metro prints console logs from the device to console, without
-  // passing them through the custom log rgeporter.
+  // passing them through the custom log reporter.
   //
   // Managed apps have a separate remote logging implementation included in the Expo SDK,
   // (see: _handleDeviceLogs), so we can just ignore these device logs from Metro.
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -1551,7 +1551,7 @@ async function _waitForRunningAsync(
   }
 }
 
-function _logPackagerOutput(projectRoot: string, level: string, data: Object) {
+function _logPackagerOutput(projectRoot: string, level: string, data: object) {
   let output = data.toString();
   if (!output) {
     return;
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Project.ts
```diff
@@ -1578,6 +1583,19 @@ function _logPackagerOutput(projectRoot: string, level: string, data: Object) {
   }
 }
 
+function _isIgnorableMetroConsoleOutput(output: string) {
+  // As of React Native 0.61.x, Metro prints console logs from the device to console, without
+  // passing them through the custom log rgeporter.
```
brentvatne:
```suggestion
  // passing them through the custom log reporter.
```

packages/xdl/src/Project.ts
```diff
@@ -1578,6 +1583,19 @@ function _logPackagerOutput(projectRoot: string, level: string, data: Object) {
...
+  //
+  // Managed apps have a separate remote logging implementation included in the Expo SDK,
+  // (see: _handleDeviceLogs), so we can just ignore these device logs from Metro.
+  // if (/^ () /)
+  //
+  // These logs originate from:
+  // https://github.com/facebook/metro/blob/e8181fb9db7db31adf7d1ed9ab840f54449ef238/packages/metro/src/lib/logToConsole.js#L50
+  return /^\s+(INFO|WARN|LOG|GROUP|DEBUG) /.test(output);
```
brentvatne: it's a bit unfortunate that we need to match specifically on these strings because if a user were to do `console.log('WARN: something')` or `console.log('INFO: blah')` it seems that we would filter out their log message

brentvatne:
> I plan submit a PR to Metro to make it use the custom log reporter (if configured) for these logs so LogReporter is able to filter what gets printed to the console.

:+1:

> Additionally, the AppRegistry logging was changed recently so the code detecting these logs had to be changed as well.

:+1:

packages/xdl/src/Project.ts
fson: This filtering only applies to the raw output from Metro, which normally is JSON, except in case of these new device console logs, which bypass the log reporter and are directly written to the console. https://github.com/facebook/metro/pull/465 will fix that and remove the need to match by a prefix in the future versions.

In any case, every message logged in the app itself will be printed, even if it has a prefix like “ WARN “, because the logs from the device/simulator are sent directly to Expo CLI by the SDK, and this code path doesn’t touch them at all.

PR #1159, opened 2019-10-30
fix(webpack-config): use custom register-service-worker.js if available

jonathonlui:
`expo customize:web` allows generating `web/register-service-worker.js` but  `webpack-config` only uses the one in web-default because the path is hardcoded.

This PR fixes this issue by using the `locations` object to like the other customizable files.


EvanBacon:
Thanks for this, there was a minor issue where the register-service-worker was being copied over twice, I fixed this though so it should be good now.

PR #1162, opened 2019-10-30
feat(expo-cli): enable development mode by default for 'start' command

cruzach:
Currently, `expo-cli` will read settings from the last time the project was run and use those. This can be confusing, and most likely if someone is starting their project via `expo start`, they want to run in development

`optsAsync` is used in 3 instances:
1. running `expo start`
2. running `expo url`
3. running `expo send`

I believe dev mode should be enabled for (1), and haven't changed anything for the other two commands. However, it might be worth the idea of enabling production mode by default for (2) and (3) since maybe those commands are used more for sharing projects rather than development


cruzach:
We could, but then `TerminalUI` would need to change how it checks whether or not the project is in dev (reading `.expo` when you hit `?` with the project running locally)

packages/expo-cli/src/urlOpts.ts
```diff
@@ -64,8 +64,14 @@ async function optsAsync(projectDir: string, options: any) {
   }
 
   let rawArgs = options.parent.rawArgs;
+
   if (hasBooleanArg(rawArgs, 'dev')) {
     opts.dev = getBooleanArg(rawArgs, 'dev');
+  } else {
+    if (rawArgs.includes('start') || rawArgs.includes('r')) {
```
fson: There's also `start:web` (alias `web`), which should behave in a similar way.

When it comes to the `url` and `send` commands, I think it's quite confusing that they even have the `--dev` and `--minify` options! (Even if they've existed for a long time.) Those commands are supposed to only work when you have the dev server running (started with `expo start`) and they just print/email the *manifest URL*. But I've just tested this and it turns `expo url --dev` or `expo send --minify` etc. clobber with the `dev`/`minify` settings of the project too. This won't make much sense, now that we start treating the setting as more non-persistent.

Could you remove the `--dev` and `--minify` flags from `url` and `send`? The other `urlOpts` (like `--lan` or `--tunnel`) still make sense for those commands (though I'm not sure if they should clobber the project settings either). This should allow you to get rid of these special cases for `start`/`r` and `start:web`/`web` as well, since any command that still has the `--dev` and `--minify` flags can then behave the same way.

cruzach: I removed the `http`/`https` flags as well, since if url and send are only meant for use when the server's running, then the protocol would already be configured, right?

packages/expo-cli/src/urlOpts.ts
```diff
@@ -64,8 +67,12 @@ async function optsAsync(projectDir: string, options: any) {
   }
 
   let rawArgs = options.parent.rawArgs;
+
   if (hasBooleanArg(rawArgs, 'dev')) {
     opts.dev = getBooleanArg(rawArgs, 'dev');
+  } else {
+    opts.dev = true;
+    opts.minify = false;
```
fson:
```suggestion
```
`opts.minify = false;` should be in the other `if-else` below that checks for the `--minify` flag.

cruzach:
If these options are now only going to be configurable via the `start` command, should they even by present in `.expo/settings.json` anymore?

fson:
I think they can still be in `.expo/settings.json`. You can set the initial value with the `--dev` and `--minify` options of the `start` command and after started, you can modify the settings for the currently running process in Dev Tools:
Screenshot 2019-11-14 at 17 09 51
or by pressing `p` in the terminal:
Screenshot 2019-11-14 at 17 10 58

`.expo/settings.json` is just how we store the current settings. However, the difference to how it worked before is that those settings should not be persisted between runs – each time you start `start` it should start with the settings you specify with command-line options (or the default settings if none given).

fson:
We *could* move these settings to `.expo/packager-info.json`, which is used to store transient packager settings such as port numbers and tunnel URLs – those have a similar lifespan. But it would not make a big difference, so we might as well keep them in `.expo/settings.json` I think.

cruzach:
Ah okay, I see what you mean. Made those changes now, and the options are handled in `start.js`

Only thing to note is that if someone runs `expo start --no-dev --minify` in one terminal, then in another decides they want to run it in development (or maybe run `expo start:web` in development) at the same time, the original project will also switch over to development. Is running two instances of a project, one in dev and one in prod something we want to support? If so, then we might have to rethink how we read/write these options

packages/expo-cli/src/commands/start.js
```diff
@@ -219,6 +258,12 @@ export default (program: any) => {
     )
     // TODO(anp) set a default for this dynamically based on whether we're inside a container?
     .option('--max-workers [num]', 'Maximum number of tasks to allow Metro to spawn.')
+    .option('--dev', 'Turns dev flag on')
```
brentvatne:
```suggestion
    .option('--dev', 'Turns the development mode flag on')
```

packages/expo-cli/src/commands/start.js
```diff
@@ -219,6 +258,12 @@ export default (program: any) => {
...
+    .option('--no-dev', 'Turns dev flag off')
```
brentvatne:
```suggestion
    .option('--no-dev', 'Turns the development mode flag off')
```

packages/expo-cli/src/commands/start.js
```diff
@@ -237,6 +282,12 @@ export default (program: any) => {
     .command('start:web [project-dir]')
     .alias('web')
     .description('Starts the Webpack dev server for web projects')
+    .option('--dev', 'Turns dev flag on')
```
brentvatne:
```suggestion
    .option('--dev', 'Turns the development mode flag on')
```

packages/expo-cli/src/commands/start.js
```diff
@@ -237,6 +282,12 @@ export default (program: any) => {
...
+    .option('--no-dev', 'Turns dev flag off')
```
brentvatne:
```suggestion
    .option('--no-dev', 'Turns the development mode flag off')
```

packages/expo-cli/src/commands/start.js
cruzach: Maybe clearer to say `turns development mode on/off`?

brentvatne: sounds good to me

PR #1181, opened 2019-11-05
Add post-publish hook for manifest and bundle for expo-ota.

mczernek:
We need to include manifest and bundle in both ios and android app archives, to access them in runtime.

However, I am not sure how this interfere with current updates available in standalone and detached apps. Maybe we should opt for new key in manifest, instead of already used "updates"?

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -291,9 +291,9 @@ async function _resolveManifestAssets(
 ) {
   try {
     // Asset fields that the user has set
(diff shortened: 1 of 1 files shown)


brentvatne:
rebased this and force pushed

brentvatne:
as far as i understand we are waiting on this pr for other parts of expo-ota to be ready, so i just added a label to indicate that. let me know if i'm wrong and if i am needed for anything

brentvatne:
@esamelson - we will probably need something like this in the future for expo-updates, but i'm going to close this for now since the expo-ota has been closed

PR #1185, opened 2019-11-08
enhancement: add jsx to sourceExts by default

### LINKED ISSUE #191: Renaming App.js to App.jsx fails when running `expo start`
## Description
* `expo start` does not resolve `import App from '../../App'` to `App.jsx`
* When using `expo start`, the root file set in `package.json`'s `main` property is ignored. It seems hard coded to `node_modules/expo/AppEntry.js`

## Expected Behavior

```
expo init jsx-file-test -t blank
cd jsx-file-test
mv App.js App.jsx
expo start
```
and then try visit the app on a device or simulator in order to trigger the JavaScript bundler to build.

Renaming `App.js` to `App.jsx` in order to meet our team coding standards, as files that contain JSX should be named `*.jsx` (specifically eslint, and `"rules": { "react/jsx-filename-extension": [1, {"extensions": [".js", ".jsx"] }]}`)

I expected that it would just work as normal when running `expo start`

## Observed Behavior

At first I got 

```
Unable to resolve "../../App" from "node_modules/expo/AppEntry.js"
```

when running `expo start`. The builder also got stuck at 0%.

## Attempted work around

I tried following the docs.expo.io instructions [What if I want to name my main app file something other than App.js?](https://docs.expo.io/versions/v31.0.0/sdk/register-root-component#what-if-i-want-to-name-my), but creating a main.js as described there, and pointing the `package.json` main property at that main.js file did not change the observed behaviour.

## Environment

```
  Expo CLI 2.4.0 environment info:
    System:
      OS: macOS High Sierra 10.13.6
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.2.0 - /usr/local/bin/node
      npm: 6.4.1 - /usr/local/bin/npm
    IDEs:
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^31.0.2 => 31.0.2
      react: 16.5.0 => 16.5.0
      react-native: https://github.com/expo/react-native/archive/sdk-31.0.0.tar.gz => 0.57.1
    npmGlobalPackages:
      expo-cli: 2.4.0

  Diagnostics report:
    https://exp-xde-diagnostics.s3.amazonaws.com/glutnix-b2d01851-bc4d-41b0-af66-ed9604386626.tar.gz
```

## Reproducible Demo

Not hard to reproduce if you follow the basic steps above.

fson:
Adding `jsx` as a source file extension is a reason why some people need to modify the `sourceExts` settings (see #191) 

We might as well add it to the default extensions. (I believe we already do this in the Webpack configuration, so this also makes the Metro config more consistent with web.)

Fixes #191.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -1722,7 +1722,18 @@ export async function startReactNativeServerAsync(
     port: packagerPort,
     customLogReporterPath,
     assetExts: ['ttf'],
-    sourceExts: ['expo.js', 'expo.ts', 'expo.tsx', 'expo.json', 'js', 'json', 'ts', 'tsx'],
+    sourceExts: [
+      'expo.js',
+      'expo.jsx',
+      'expo.ts',
+      'expo.tsx',
+      'expo.json',
+      'js',
+      'jsx',
+      'json',
+      'ts',
+      'tsx',
+    ],
   };
 
   if (options.nonPersistent && Versions.lteSdkVersion(exp, '32.0.0')) {


packages/xdl/src/Project.ts
```diff
@@ -1722,7 +1722,18 @@ export async function startReactNativeServerAsync(
     port: packagerPort,
     customLogReporterPath,
     assetExts: ['ttf'],
-    sourceExts: ['expo.js', 'expo.ts', 'expo.tsx', 'expo.json', 'js', 'json', 'ts', 'tsx'],
```
EvanBacon: Nice job, was planning on unifying these by merging the logic into this one function but couldn't find a way to emulate the ios/android platform extensions like we do on web.
```js
import { getManagedExtensions } from '@expo/config/build/paths';
const sourceExts  = getManagedExtensions([]);
/*
  "expo.ts",
  "expo.tsx",
  "expo.mjs",
  "expo.js",
  "expo.jsx",
  "ts",
  "tsx",
  "mjs",
  "js",
  "jsx",
  "json",
  "wasm",
*/
```

EvanBacon: I'm also thinking this order looks incorrect, will fix.

EvanBacon: Like this https://github.com/expo/expo-cli/pull/1187

EvanBacon:
https://github.com/expo/expo-cli/pull/1219

PR #1190, opened 2019-11-09
Ensure "Run on web browser" respects network setting

sstur:
The current behavior is that clicking "Run on web browser" will always use the computer's LAN address, even if the user selected "Local" for the network setting. This even happens when the option `--host localhost` was specified on the `expo start` command, which is probably not what the user expects.

Note: this is a client-side hack, but it seems less intrusive than messing with `packages/dev-tools/server/graphql/GraphQLSchema.js`.


sstur:
I'm closing this in favor of #1191 which fixes it at the source of truth.

PR #1191, opened 2019-11-09
Ensure webAppUrl respects network setting (LAN/Local)

sstur:
The current behavior of `constructWebAppUrlAsync` is to always use the LAN IP, even if the user selected "Local" for the network setting (either in the UI or the option `--host localhost` was specified). This is probably not what the user expects.

Note: This is a better solution than #1190 since it fixes the issue at the source, where the URL is generated.


fson:
Makes sense, thanks for fixing this! 🙌

PR #1192, opened 2019-11-10
Moved `expo-cli optimize` to `npx expo-optimize`

EvanBacon:
## Why 

- make `expo-cli` and `xdl` more lean

## How

- use `ncc` to compile optimize into a single 
- extern sharp to reduce npx size to 449kB (down from 25,311kB)
- created deprecation notice when using legacy command which redirects the command to `expo-optimize` and pipes the logs.

## Questions

- [x] does the naming make sense? Users might think `expo-optimize` is a unimodule...


packages/expo-optimize/README.md
```diff
@@ -0,0 +1,59 @@
+# optimze
+
+Compress the assets in your React project. This library uses the natively powered package [Sharp](https://sharp.pixelplumbing.com) behind the scenes to optimally reduce your project's assets without creating any visual regressions.
```
sergeichestakov: Thanks for adding these docs. Looks good!

packages/expo-optimize/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "expo-optimize": "./build/index.js"
+  },
+  "files": [
+    "build"
+  ],
+  "scripts": {
+    "prepare": "yarn run clean && yarn run build:prod",
+    "lint": "eslint .",
+    "watch": "yarn run build -w",
+    "build": "ncc build ./src/index.ts -o build/ --external @expo/image-utils",
```
fson: Can you explain how `ncc` helps here, if we leave `@expo/image-utils` as an external? Presumably `npx` will still have to download and install `@expo/image-utils` anyway?

packages/dev-tools/package.json
```diff
@@ -54,6 +54,7 @@
     "copy-to-clipboard": "^3.0.8",
     "emotion": "^8.0.5",
     "emotion-server": "^8.0.5",
+    "expo-optimize": "0.0.0",
```
fson: I don't think `expo-optimize` should be a dependency here.

packages/expo-cli/package.json
```diff
@@ -90,6 +90,7 @@
     "enquirer": "^2.3.2",
     "envinfo": "5.10.0",
     "es6-error": "3.2.0",
+    "expo-optimize": "0.0.0",
```
fson: Shouldn't we remove the dependency on `expo-optimize` from the default Expo CLI installation if we intend this to help with installation issues related to `@expo/image-utilts` / `sharp-cli`? Otherwise `npm i -g expo-cli` will still try to pull in Sharp.

EvanBacon: it's part of the deprecation. Running `expo optimize` will warn you to switch over for a little while.

packages/expo-optimize/package.json
EvanBacon: Because this package doesn't have an accessible API we don't need a types or source map file, `npx` will download the file significantly faster this way. It's using Webpack to compile the TS results down to one file, I was going to build my own implementation but this seemed to have everything I needed.

packages/expo-cli/src/commands/optimize.ts
```diff
@@ -18,31 +18,22 @@ export async function action(projectDir = './', options: Options = {}) {
...
-  }
-  return quality;
+  const args: string[] = [projectDir];
+
+  if (options.save) args.push('--save');
+  if (options.quality) args.push('--quality', options.quality);
+  if (options.include) args.push('--include', options.include);
+  if (options.exclude) args.push('--exclude', options.exclude);
+
+  const child = spawn(require.resolve('expo-optimize'), args);
```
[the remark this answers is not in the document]
EvanBacon: hmm I couldn't get that to work originally, will try again.

packages/expo-cli/package.json
fson: Sounds good to me

packages/expo-cli/src/commands/optimize.ts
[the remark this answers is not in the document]
EvanBacon: worked perfectly!

packages/dev-tools/package.json
EvanBacon: Will this still be able to resolve if we remove it? https://github.com/expo/expo-cli/pull/1192/files#diff-d087e14e7de95f68c0b8fac69087f279R571

fson: No it won't work unless it's a dependency. Could we spawn `npx` instead, or (maybe better) use `libnpx` (the programmatic API to npx) to run it?

EvanBacon: libnpx seems deprecated https://github.com/zkat/npx

EvanBacon: just using whichever globally installed npx module exists directly for now

PR #1194, opened 2019-11-10
[xdl] Deprecated ProjectUtils methods

EvanBacon:
- removed `readConfigJsonAsync`
- removed `readExpRcAsync`
- migrated `ProjectUtils` tests to `expo/config`
- updated changelog


packages/expo-cli/src/commands/start.js
```diff
@@ -185,7 +185,7 @@ async function configureProjectAsync(projectDir, options) {
 
   log(chalk.gray(`Starting project at ${projectDir}`));
 
-  const { exp, pkg } = await ProjectUtils.readConfigJsonAsync(projectDir, options.webOnly);
+  const { exp, pkg } = await ConfigUtils.readConfigJsonAsync(projectDir, options.webOnly);
   if (exp === null) {
```
fson: You can remove this `null` check now, because `readConfigJsonAsync` never returns `null`.

packages/expo-cli/src/commands/upload/BaseUploader.ts
```diff
@@ -40,7 +39,7 @@ export default class BaseUploader {
   }
 
   async _getProjectConfig(): Promise<void> {
-    const { exp } = await ProjectUtils.readConfigJsonAsync(this.projectDir);
+    const { exp } = await readConfigJsonAsync(this.projectDir);
     if (!exp) {
```
fson: You can remove this check.

packages/xdl/src/detach/Detach.js
```diff
@@ -360,7 +357,7 @@ async function _getIosExpoKitVersionThrowErrorAsync(iosProjectDirectory) {
 }
 
 async function prepareDetachedBuildIosAsync(projectDir, args) {
-  const { exp } = await ProjectUtils.readConfigJsonAsync(projectDir);
+  const { exp } = await readConfigJsonAsync(projectDir);
   if (exp) {
```
EvanBacon: @fson is this functionality now broken?

PR #1210, opened 2019-11-12
[config] export paths from alias

EvanBacon:
We should consider splitting this into a new package


EvanBacon:
https://github.com/expo/expo-cli/pull/1219

PR #1211, opened 2019-11-12
[webpack-config] Added package aliases for extra webpack features

EvanBacon:
```js
import {} from '@expo/webpack-config/loaders';
import {} from '@expo/webpack-config/addons';
import {} from '@expo/webpack-config/plugins';
import {} from '@expo/webpack-config/utils';
import {} from '@expo/webpack-config/env';
```

Added deprecated notice to `'@expo/webpack-config/withUnimodules'` in favor of `'@expo/webpack-config/addons'`.

- I'm not in love with the name addons but extensions seems confusing seeing as file extensions are a pretty heavy part of this.


packages/webpack-config/withUnimodules.js
```diff
@@ -1 +1,5 @@
-module.exports = require('./webpack/withUnimodules').default;
+module.exports = require('./addons').withUnimodules;
+
+console.warn(
+  '`@expo/webpack-config/withUnimodules` is deprecated, please use `import { withUnimodules } from @expo/webpack-config/addons` instead.'
+);
```
fson: 👍

Issue #1101, opened 2019-10-10
feat: Add ability to run a min project


EvanBacon:
## Why

Add ability to run a project with the bare minimum:

- `App.js`
```jsx
import React from 'react';
import { Text, View } from 'react-native';

export default function App() {
  return (
    <View>
      <Text>Open up App.js to start working on your app!</Text>
    </View>
  );
}
```
- `package.json`
```json
{
  "dependencies": {
    "expo": "^35.0.0",
    "react": "16.8.3",
    "react-dom": "16.8.3",
    "react-native": "https://github.com/expo/react-native/archive/sdk-35.0.0.tar.gz",
    "react-native-web": "^0.11.7"
  }
}
```

### Webpack

- [x] Use default babel preset when babel.config and .babelrc aren't present - #1100 
- [x] Fallback to `node_modules/expo/AppEntry.js` if the `main` field in the `package.json` is missing - #1100 
- [x] Don't require an `app.json`

### Metro

- [ ] Use default babel preset when babel.config and .babelrc aren't present 
  - https://github.com/facebook/metro/pull/460
- [x] Fallback to `node_modules/expo/AppEntry.js` if the `main` field in the `package.json` is missing. https://github.com/expo/expo-cli/pull/1092/commits/6bb4324ac62d3c1c96faee88ae5f871bf793f083
- [ ] Don't require an `app.json`

EvanBacon:
somewhat related https://github.com/expo/expo/issues/4156

EvanBacon:
Complete

PR #1213, opened 2019-11-13
Don't require an app.json config file to start a native project

EvanBacon:
## Why

- https://github.com/expo/expo-cli/issues/1101
- parity with web

## How

Disable config existence validation in the `readConfigJsonAsync` & `readConfigJson` methods. 
- This should only apply to starting a project. Publishing, deploying, and building should all require an `app.json` to exist (for now).

## Test Plan

Run `expod start --ios` in the `packages/webpack-config/tests/minimum` project.

## Future 

- prompting for bundleID, package name, slug, etc... writes just that single value to a config (and creates it if it doesn't exist)
- building, publishing, deploying without an `app.json`.
- ability to create an `app.config.js` with a method that provides the default config for users to mix features into.
- restructure docs to inform you to create an `app.json` before editing.
- remove `app.json` from template blank.


packages/config/src/Config.ts
```diff
@@ -489,11 +489,7 @@ function getRootPackageJsonPath(projectRoot: string, exp: ExpoConfig): string {
   return packageJsonPath;
 }
 
-export function readConfigJson(
-  projectRoot: string,
-  skipValidation: boolean = false,
-  skipNativeValidation: boolean = false
-): ProjectConfig {
+export function readConfigJson(projectRoot: string, skipValidation: boolean = true): ProjectConfig {
```
fson:
```suggestion
export function readConfigJson(projectRoot: string, { skipValidation = true }: { skipValidation?: boolean } = {}): ProjectConfig {
```
We should get rid of these positional boolean arguments.
```
readConfigJson(projectRoot, true);  // what does true mean here? Need to look at the definition.
readConfigJson(projectRoot, { skipValidation: true });  // obvious at a glance
```

fson: Also `skipValidation` sounds like doing validation is the default and you can optionally skip it, but the default now is `skipValidation = true`. Should it be `false` instead? Or rename the option to `requireExpoSDKVersion`?

packages/config/src/__tests__/Config-test.js
```diff
@@ -141,23 +141,25 @@ describe('readConfigJson', () => {
...
-      const { exp, pkg } = readConfigJson('/no-package', true, true);
+      const { exp, pkg } = readConfigJson('/no-package');
       expect(exp.name).toBe(pkg.name);
       expect(exp.description).toBe(pkg.description);
     });
 
     it(`will throw if the app.json is missing`, () => {
-      expect(() => readConfigJson('/no-config')).toThrow(/app.json must include a JSON object./);
+      expect(() => readConfigJson('/no-config', { requireLocalConfig: true })).toThrow(
+        /app.json must include a JSON object./
```
fson: I think this error message should include a bit more detail, maybe an example of a valid app.json and/or link to the relevant documentation.

EvanBacon:
I think we need a new approach to this PR. Currently if a project is missing an `app.json` then we consider it to be a different kind of project when building in turtle. Seems like we should slowly deprecate the method in favor of a differently named method so we can work through each instance individually.

PR #1216, opened 2019-11-13
fix performance hints in webpack

EvanBacon:


packages/webpack-config/src/webpack.config.ts
```diff
@@ -289,7 +289,9 @@ export default async function(
     },
     // Turn off performance processing because we utilize
     // our own (CRA) hints via the FileSizeReporter
-    performance: boolish('CI', false) ? false : undefined,
+
+    // TODO: Bacon: Remove this higher value
+    performance: boolish('CI', false) ? false : undefined, //{ maxAssetSize: 600000, maxEntrypointSize: 600000 },
```
fson: There are no changes to the previous code other than the added `TODO` comment?

EvanBacon: oops, uncommented the change 😅

PR #1219, opened 2019-11-13
[xdl] unify source extensions on native

EvanBacon:
- use the same extension order across web and native
- do we support MJS on native?

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -8,6 +8,8 @@ import {
   ExpoConfig,
   readConfigJsonAsync,
 } from '@expo/config';
+
+import { getManagedExtensions } from '@expo/config/paths';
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Project.ts
```diff
@@ -1716,7 +1718,8 @@ export async function startReactNativeServerAsync(
     port: packagerPort,
     customLogReporterPath,
     assetExts: ['ttf'],
-    sourceExts: ['expo.js', 'expo.ts', 'expo.tsx', 'expo.json', 'js', 'json', 'ts', 'tsx'],
+    // TODO: Bacon: Support MJS (Modern JS)
```
ide:
```suggestion
    // TODO: Bacon: Support .mjs (short-lived JS modules extension that some packages use)
```

EvanBacon:
fix #191

PR #1220, opened 2019-11-13
feat(dev-tools): copy url to clipboard feedback

jaysbytes:
Recently while using expo I asked myself why copy to clipboard on project link is not implemented.
Actually it was implemented, but there was no feedback to the user, so I decided to add it. 

EXPO CLI COPY


jaysbytes:
🔥

Issue #1207, opened 2019-11-12
Expo CLI upload doesn't respect owner field of app.json


chris-soar:
## Description

Expo CLI `expo upload:ios` doesn't use the owner field of app.json to determine where to download the latest build file from. So if the app is built (`expo build:ios`) with a different owner than the logged in user, the upload command is then unable to find the build on the Expo servers.

## Expected Behavior

The `expo upload:ios` command should retrieve the owner field from app.json (just like it currently retrieves the slug) and use the owner field to search for and download the latest build, if the logged in user has the required privileges. As it stands right now, it's necessary to manually download the build file and add the `--path` option.

## Observed Behavior

The Expo CLI attempted to search for the latest build of the app under the logged in user's builds. As we use delegated team accounts for building the app under the central account's ownership, the CLI wasn't able to find any builds. It seems there isn't a method of directly providing the owner field to the upload command either.
Looking at `_downloadLastestBuild()` in the upload command's `BaseUploader.ts`, it seems the command only retrieves the slug from the config and doesn't retrieve owner. By comparison the iOS build command seems to retrieve the owner and slug from the project's manifest/app.json.

The following error is observed, as Expo CLI doesn't find any builds for the current slug under the logged in user's builds.
```
Failed to upload the standalone app to the app store.
There are no builds on the Expo servers, please run 'expo build:ios' first
```

## Environment

```
Expo CLI 3.4.1 environment info:
    System:
      OS: macOS 10.15.1
      Shell: 5.7.1 - /usr/local/bin/zsh
    Binaries:
      Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
      Yarn: 1.19.1 - /usr/local/bin/yarn
      npm: 6.13.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Android Studio: 3.5 AI-191.8026.42.35.5900203
      Xcode: 11.2/11B52 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^35.0.0 => 35.0.1 
      react: 16.8.3 => 16.8.3 
      react-native: https://github.com/expo/react-native/archive/sdk-35.0.0.tar.gz => 0.59.8 
      react-navigation: ^4.0.7 => 4.0.10 
    npmGlobalPackages:
      expo-cli: 3.4.1
```

## Reproducible Demo

[...]

PR #1223, opened 2019-11-14
fix: pass owner name into API request

vincentvella:
Docs stated that the owner id would be respected (and is respected on build) but they are currently not being accounted for when searching for built apps
This PR was created to pass the owner name from the app.json context within the files through to the API request to be handled there.
Currently, when uploading an app that belongs to another team you'll be prompted with:
``` 
Failed to upload the standalone app to the app store.
There are no builds on the Expo servers, please run 'expo build:ios' first 
```
The build ran fine with the owner name so I'm assuming some backend work still needs to be taken care of for this to take effect.


vincentvella:
Referenced in #1207

PR #1235, opened 2019-11-16
[dev-tools] Add informative error page when root GQL query fails

wschurman:
why
===

Previously, when the root GraphQL query resulted in an error, the resulting interface in dev-tools would be a blank black page (null component). 

how
===

This PR adds a new component to handle that case which displays errors in a more informative way to users (the same way the CLI does in the shell).

test plan
===

1. Create blank test expo project via `expo init`
1. `cd` into `dev-tools` directory within this repo in a new terminal tab.
1. `yarn run dev <path-to-expo-project>`
1. Change the `expo` key to `blah` in `app.json` within the expo project.
1. See that the errors are displayed to the user rather than a blank page.

Also test the production build:
1. In this repo, `yarn start`
1. In expo project from above, run `<path-to-expo-cli-repo>/packages/expo-cli/bin/expo.js start`
1. Toggle between a valid and invalid `app.json` via same method as above, ensure errors are shown in the error case and not in the valid case.

Before:
Screen Shot 2019-11-15 at 5 01 09 PM
After:
Screen Shot 2019-11-15 at 5 00 23 PM

### DIFF
--- a/packages/dev-tools/components/IndexPageErrors.js
+++ b/packages/dev-tools/components/IndexPageErrors.js
@@ -1,0 +1,40 @@
+import React from 'react';
+
+import { css } from 'react-emotion';
+
+import * as Constants from 'app/common/constants';
+
+const STYLES_ERROR_CONTAINER = css`
+  background: ${Constants.colors.foreground};
+  position: relative;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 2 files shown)


fson:
Excellent, thanks for fixing this.

PR #1135, opened 2019-10-18
[expo-cli] Refresh generate-module command & rewrite to TS

bbarthec:
Adjustments for `expo generate-module` command.
Corresponding `expo-module-template` PR: https://github.com/expo/expo/pull/6022

PR #1246, opened 2019-11-19
[expo-cli] Convert generate-module command to TypeScript

wschurman:
why
===
Fairly straightforward conversion.

how
===
Rename to `.ts`, fix errors.

test plan
====
Run `~/expo/expo-cli/packages/expo-cli/bin/expo.js generate-module`, ensure prompts work and generated module has correct format.


packages/expo-cli/src/commands/generate-module/configureModule.ts
```diff
@@ -14,13 +16,13 @@ const asyncForEach = async (array, callback) => {
  * @param {string[]} extensions - array of extensions for files that would be renamed, must be provided with leading dot or empty for no extension, e.g. ['.html', '']
  * @param {{ from: string, to: string }[]} renamings - array of filenames and their replacers
  */
-const renameFilesWithExtensions = async (directoryPath, extensions, renamings) => {
+const renameFilesWithExtensions = async (directoryPath: string, extensions: string[], renamings: { from: string, to: string }[]) => {
```
brentvatne: it looks like we need to update the [precommit hook](https://github.com/expo/expo-cli/blob/a06ccd4abc5dfb76286cb6a97c54a91ded57c479/package.json#L18-L24) to run on ts files too, this should be formatted according to the root [.prettierrc](https://github.com/expo/expo-cli/blob/master/.prettierrc)

can you run prettier on these files manually? we can fix precommit hook in another pr

brentvatne:
looks like this is actually handled by https://github.com/expo/expo-cli/pull/1135 already 😱

PR #1253, opened 2019-11-21
[dev-tools] Option to choose devtools bind address

### LINKED ISSUE #1081: Securing WebSocket API prevents use of Expo DevTools
## Description

While using expo-cli from Docker container, access to Expo DevTools isn't anymore easily achieved. Previously Expo DevTools would listen 0.0.0.0:19002, which was easy to redirect in Docker. Now #1008 has increased security, but at same time forces Expo DevTools to bind only to 127.0.0.1:19002, preventing their use in this case, and also if wanting to have multiple computers accessing Expo DevTools.

## Expected Behavior

I would expect that this securing effort would be extended, so there would be option to choose which address Expo DevTools bind. There is already REACT_NATIVE_PACKAGER_HOSTNAME that is used to give correct address when expo-cli is ran inside Docker, there should also be something like EXPO_DEVTOOLS_LISTEN_ADDRESS, which could specify specific IP address, or even 0.0.0.0, while defaulting to 'localhost'.

## Observed Behavior

While expo-cli starts inside Docker just fine, and mobile phone can connect to it, trying to access port 19002 from outside Docker gives only error that no content is received. Inspecting what is happening inside Docker container, it is seen that ports 19000 and 19001 are listening ::, while 19002 is bound to 127.0.0.1

## Environment

Please run `expo diagnostics` in the project folder and copy the output here:
```
Expo CLI 3.2.3 environment info:
    System:
      OS: Linux 4.9 Debian GNU/Linux 9 (stretch) 9 (stretch)
      Shell: 4.4.12 - /bin/bash
    Binaries:
      Node: 12.10.0 - /usr/local/bin/node
      Yarn: 1.17.3 - /usr/local/bin/yarn
      npm: 6.10.3 - /usr/local/bin/npm
    npmPackages:
      expo: ^32.0.0 => 32.0.6
      react: 16.5.0 => 16.5.0
      react-native: https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz => 0.57.1
      react-navigation: ^3.0.9 => 3.4.1
    npmGlobalPackages:
      expo-cli: 3.2.3
```

kerbe:
### Why
As it was reported and discussed in #1081, recent changes break compatibility with Docker.

### How
Adds one new environment variable, `EXPO_DEVTOOLS_LISTEN_ADDRESS` that can be used to give different ip where devtools bind. If no environment variable is given, defaults to `localhost`, which is currently hardcoded value in codebase.

This fix takes use of already used environment variable `REACT_NATIVE_PACKAGER_HOSTNAME`, as it is already needed when using `expo-cli` in Docker container.

### Test plan
- Run expo-cli as usual in normal environment, and check that port 19002 listens only 127.0.0.1
- Run expo-cli with `EXPO_DEVTOOLS_LISTEN_ADDRESS=0.0.0.0 expo-cli start` and check that port 19002 listens all interfaces

Fixes #1081

### DIFF
--- a/packages/dev-tools/server/DevToolsServer.js
+++ b/packages/dev-tools/server/DevToolsServer.js
@@ -51,29 +51,7 @@ export async function createAuthenticationContextAsync({ graphqlHostname, port }
 export async function startAsync(projectDir) {
   const port = await freeportAsync(19002, { hostnames: [null, 'localhost'] });
   const server = express();
-
-  const devtoolsHost = () => {
-    let devtoolHost;
-    if (process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS) {
-      devtoolHost = process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS.trim();
-    } else {
-      devtoolHost = `localhost`;
-    }
-    return devtoolHost;
-  };
   const listenHostname = devtoolsHost();
-
-  const devtoolsGraphQLHost = () => {
-    let devtoolsGraphQLHost;
-    if (process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS && process.env.REACT_NATIVE_PACKAGER_HOSTNAME) {
-      devtoolsGraphQLHost = process.env.REACT_NATIVE_PACKAGER_HOSTNAME.trim();
-    } else if (process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS) {
-      devtoolsGraphQLHost = process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS;
-    } else {
-      devtoolsGraphQLHost = `localhost`;
-    }
-    return devtoolsGraphQLHost;
-  };
   const graphqlHostname = devtoolsGraphQLHost();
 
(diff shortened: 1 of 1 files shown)


packages/dev-tools/server/DevToolsServer.js
```diff
@@ -52,7 +52,31 @@ export async function startAsync(projectDir) {
...
+    if (process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS) {
+      devtoolHost = process.env.EXPO_DEVTOOLS_LISTEN_ADDRESS.trim();
+    } else {
+      devtoolHost = `localhost`;
+    }
+    return devtoolHost;
+  };
+  const listenHostname = devtoolsHost();
+
+  const devtoolsGraphQLHost = () => {
```
brentvatne: i think we can move the `devtoolsGraphQLHost` and `devtoolsHost` functions definition outside of `startAsync`, this would make the code easier to read.

kerbe: Moved function definitions outside of `startAsync`, further down in the file.

PR #1257, opened 2019-11-22
Show NOTE when running fetch:android:hashes

awinograd:
This bit me when trying to include a google maps api key that was restricted to our app hash. While this notice already shows in the `--help` for this command, it would be useful to show it when running the command directly to avoid confusion.


brentvatne:
@wkozyra95 - curious what you think about this

awinograd:
@wkozyra95 updated

PR #1269, opened 2019-11-26
[metro][xdl] Move custom log reporter to XDL

EvanBacon:
Not sure why half of the log reporter was included in the expo package and the other half is in XDL. Moving this will help simplify the `expo/metro-config` package I'm working on.

## Test Plan

- `expod start --ios`
- test that `console.logs` show in the terminal
- test that `console.logs` show in the dev tools screen under the device
- ensure Metro custom logging looks like before (loading bar)

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -291,9 +291,9 @@ async function _resolveManifestAssets(
 ) {
   try {
     // Asset fields that the user has set
-    const assetSchemas = (
-      await ExpSchema.getAssetSchemasAsync(manifest.sdkVersion)
-    ).filter((assetSchema: ExpSchema.AssetSchema) => get(manifest, assetSchema.fieldPath));
+    const assetSchemas = (await ExpSchema.getAssetSchemasAsync(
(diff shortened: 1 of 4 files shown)


packages/xdl/src/Project.ts
```diff
@@ -1687,15 +1687,7 @@ export async function startReactNativeServerAsync(
...
-  let customLogReporterPath: string | undefined;
-
-  const possibleLogReporterPath = projectHasModule('expo/tools/LogReporter', projectRoot, exp);
-  if (possibleLogReporterPath) {
-    customLogReporterPath = possibleLogReporterPath;
-  } else {
-    // TODO: Bacon: Prompt to install expo?
-    logger.global.warn(`Expo is not installed: Using default reporter to format logs.`);
-  }
+  const customLogReporterPath: string = require.resolve(path.join(__dirname, 'reporter'))
```
ide:
```suggestion
  const customLogReporterPath: string = require.resolve(__dirname, 'reporter');
```

EvanBacon: this didn't work

PR #1273, opened 2019-11-27
feat(expo-cli): add force flag to ios and android build

### LINKED ISSUE #1224: Non interactive mode asks for input
## Description

Input is asked when running build command with `--non-interactive` flag.

## Expected Behavior

Built an ios archive, then re-run command in a bitbucket pipeline. Expo-cli should not ask for input.

## Observed Behavior

```
There were no new changes from the last build, you can download that build from here: https://exp-shell-app-assets.s3.us-west-1.amazonaws.com/ios/%40lcde/onestaff-d38289d9-9e0e-4f8d-8d9a-88123d5278f1-archive.ipa
 Input is required, but Expo CLI is in non-interactive mode.
Required input:
> Do you want to build app anyway?
```

Pipeline fails

## Environment

```
Expo CLI 3.4.1 environment info:
    System:
      OS: Linux 4.19 Debian GNU/Linux 9 (stretch) 9 (stretch)
      Shell: 4.4.12 - /bin/bash
    Binaries:
      Node: 10.15.3 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
    npmPackages:
      expo: ^33.0.0 => 33.0.7 
      react: 16.8.3 => 16.8.3 
      react-native: https://github.com/expo/react-native/archive/sdk-33.0.0.tar.gz => 0.59.8 
      react-navigation: ^3.12.1 => 3.12.1 
```

## Reproducible Demo

Build once, then re-run the command: `expo build:ios --release-channel staging --no-publish --no-wait --non-interactive`

byCedric:
Fixes #1224, it adds a simple `--force` flag to `expo build:android` and `expo build:ios` commands like Ville mentioned:
> Looks like we need to add a --force flag for non-interactive mode. It would correspond to choosing “yes build anyway” in the prompt.

It might be better if we expand the `prompt` script, and add a force flag there. That way we can handle forces within the prompt and we don't have to add if statements over the questions? I think that might work just fine, you can even include a default value for the questions right? Maybe we can auto-pick that as answer?

I also added another PR, with an alternative solution for this exact problem (the reuse question in the base builder). You can check it out here, PR #1272. We should pick either this one, or that one to fix the current issue.

### DIFF
--- a/packages/expo-cli/src/commands/build/BaseBuilder.js
+++ b/packages/expo-cli/src/commands/build/BaseBuilder.js
@@ -40,6 +40,7 @@ export default class BaseBuilder {
     clearCredentials: false,
     releaseChannel: 'default',
     publish: false,
+    force: false,
   };
   manifest: Object = {};
   user: Object;
@@ -164,18 +165,17 @@ Please see the docs (${chalk.underline(
         )}`
       );
 
-      let questions = [
-        {
+      if (!this.options.force) {
+        const answer = await prompt({
           type: 'confirm',
           name: 'confirm',
           message: 'Do you want to build app anyway?',
-        },
-      ];
+        });
 
-      const answers = await prompt(questions);
-      if (!answers.confirm) {
-        log('Stopping the build process');
-        process.exit(0);
+        if (!answer.confirm) {
+          log('Stopping the build process');
+          process.exit(0);
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


brentvatne:
i think it's better to just warn people and not block. we'll probably want to clarify the message as well, because the current heuristics aren't accurate enough to make the warning super useful

PR #1275, opened 2019-11-27
[config] Created getConfig methods for future config evaluation requirement

EvanBacon:
## Why

Split out of https://github.com/expo/expo-cli/pull/1213 to simplify the transition and PR.

- This helps to maintain functionality without changing how every module works.
- Deprecate `readConfigJsonAsync` and `readConfigJson` so we can cleanly move to a system where the Expo config isn't required and can also be evaluated from a file like `app.config.js`

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -543,6 +543,9 @@ export async function getConfigAsync(
   };
 }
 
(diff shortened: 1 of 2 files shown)


EvanBacon:
Stale.

PR #1296, opened 2019-12-04
[expo-cli] Run prettier on TypeScript in precommit hook

wschurman:
why
===
As mentioned by @brentvatne in #1239 we should run prettier on precommit for both TypeScript and JS files.

how
===
To accomplish this, the eslint config must specify that the typescript parser is used for typescript files. The `eslint-config-universe` config (updated version) specifies this.

test plan
===
Add  a lint error to a `.ts` file, amend this commit, see that the lint error is fixed automatically.

discussion
===
Note that there are a lot of prettier discrepancies in this repo in TypeScript files, so it may be a good idea to just run a prettier fix commit alongside this commit so that a slew of future commits don't include automatic prettier changes as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     }
   },
   "lint-staged": {
-    "*.js": [
+    "*.{js,ts}": [
       "eslint --fix",
       "prettier --write",
       "git add"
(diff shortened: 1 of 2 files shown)


brentvatne:
thanks!

> Note that there are a lot of prettier discrepancies in this repo in TypeScript files, so it may be a good idea to just run a prettier fix commit alongside this commit so that a slew of future commits don't include automatic prettier changes as well.

will do!

PR #1298, opened 2019-12-04
refactor(cli) create a dist cert selector

quinlanj:
# Overview
This is part of the effort to refactor the credentials codebase - we want all our credentials code to live in `packages/expo-cli/src/credentials` and be written in Typescript. This PR:
- creates a distribution certificate selector that we will use to replace the selector in adhoc builds, and will eventually be used as the selector for `expo build:ios`. 
- polls Apple to check the validity of certificates on Expo server when possible (ie) upload, selecting existing credentials, etc
- works when user is not logged in (adhoc builds need to support anonymous users). The `Doctor.validateLowLatencyAsync` we call in `context.ts` throws an error when not in a project directory instead of returning fatal. We change `Doctor.ts` to return `FATAL` in this case.
- the selector gives the user the option to 'go back' to the previous UI
- give the user the option to see more information when presented with 'high risk' actions. For example, users are hesitant to revoke credentials and would like to see the consequences of their actions in that case.

### DIFF
--- a/packages/expo-cli/src/credentials/views/IosDistCert.ts
+++ b/packages/expo-cli/src/credentials/views/IosDistCert.ts
@@ -24,30 +24,53 @@ Please revoke the old ones or reuse existing from your other apps.
 Please remember that Apple Distribution Certificates are not application specific!
 `;
 
-export class CreateIosDist implements IView, ISelect<DistCert> {
+export class CreateIosDistSelector implements ISelect<DistCert> {
   async create(ctx: Context): Promise<IosDistCredentials> {
     const newDistCert = await this.provideOrGenerate(ctx);
     return await ctx.ios.createDistCert(newDistCert);
   }
 
-  async open(ctx: Context): Promise<IView | null> {
+  // TODO(quin): support the option to go 'back'
+  async select(
+    ctx: Context,
+    options: Partial<DistCertOptions> = {}
+  ): Promise<ISelect<DistCert> | DistCert | null> {
+    return ctx.user
+      ? await this._selectAuthenticated(ctx, options)
+      : await this._selectUnauthenticated(ctx, options);
+  }
+
+  async _selectAuthenticated(ctx: Context, options: Partial<DistCertOptions> = {}) {
     const distCert = await this.create(ctx);
 
     log(chalk.green('Successfully created Distribution Certificate\n'));
     displayIosUserCredentials(distCert);
     log();
-    return null;
+
+    const { experienceName, bundleIdentifier } = options;
+    if (experienceName && bundleIdentifier) {
+      await ctx.ios.useDistCert(experienceName, bundleIdentifier, distCert.id);
+      log(
+        chalk.green(
+          `Successfully assigned Distribution Certificate to ${experienceName} (${bundleIdentifier})`
+        )
+      );
+    }
+    return distCert;
   }
 
-  // TODO(quin): support the option to go 'back'
-  async select(ctx: Context, options: DistCertOptions): Promise<ISelect<DistCert> | DistCert> {
-    return ctx.user ? await this.create(ctx) : await this.provideOrGenerate(ctx);
+  async _selectUnauthenticated(ctx: Context, options: Partial<DistCertOptions> = {}) {
+    const distCert = await this.provideOrGenerate(ctx);
+
+    log(chalk.green('Successfully created Distribution Certificate\n'));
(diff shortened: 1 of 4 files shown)


packages/expo-cli/src/credentials/views/IosDistCert.ts
```diff
@@ -36,10 +39,23 @@ export class CreateIosDist implements IView {
     return null;
   }
 
+  // TODO(quin): support the option to go 'back'
+  async select(ctx: Context, options: DistCertOptions): Promise<ISelect<DistCert> | DistCert> {
+    return ctx.user ? await this.create(ctx) : await this.provideOrGenerate(ctx);
```
wkozyra95: select in UseExistingDistributionCert is updating db and her it's not? I'm not sure what would be the desired behavior here, but it should be the same.
Are you want to update all credentials at the end or everything separately?

packages/expo-cli/src/credentials/views/IosDistCert.ts
```diff
@@ -202,29 +228,87 @@ export class UseExistingDistributionCert implements IView {
     }
     return null;
   }
+
+  async select(ctx: Context, options: DistCertOptions): Promise<ISelect<DistCert> | DistCert> {
```
wkozyra95: select and open In most cases should have the same behavior, it would be great to merge those two mechanisms
In all cases open could be implemented using select (if select exists)

I'm assuming that idea of applying changes in db(mentioned on the notion) was droped

packages/expo-cli/src/credentials/views/SelectDistributionCert.ts
```diff
@@ -0,0 +1,96 @@
...
+  async handleAction(
+    ctx: Context,
+    action: string,
+    options: DistCertOptions
+  ): Promise<ISelect<DistCert> | DistCert> {
+    switch (action) {
+      case 'CHOOSE_EXISTING':
+        return await new iosDistView.UseExistingDistributionCert().select(ctx, options);
+      case 'GENERATE':
+        return await new iosDistView.CreateIosDist().select(ctx, options);
```
wkozyra95: Just to make sure. Are you aware that this will only create distribution certificate? It will be added to user credentials but won't be assigned to any experience

quinlanj: Good catch :O I've changed this so that if there is a `bundleIdentifier` and `experienceName`, we assign it to an experience. 

I noticed that `credentials:manager` relies on this as well. With the new changes, if a user runs the manager and chooses to `create distribution certificate` in project context, it will assign it to an experience. Else it just creates the certificate. Lmk if that's ok!

packages/expo-cli/src/credentials/views/IosDistCert.ts
```diff
@@ -299,19 +397,35 @@ async function generateDistCert(ctx: Context): Promise<DistCert> {
...
-          choices: certs.map((cert, index) => ({
-            value: index,
-            name: formatDistCertFromApple(cert, ctx.ios.credentials),
-          })),
+          choices,
           pageSize: Infinity,
         },
       ]);
 
+      if (revoke.includes(MORE_INFO)) {
```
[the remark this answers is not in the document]
quinlanj: agreed... let me see if i can find smth better :/

packages/expo-cli/src/credentials/views/IosDistCert.ts
quinlanj: It should be ok to update the credentials separately. 

The situation we wanted to avoid is generating a cert, running into an unrecoverable error and requiring the user to go to Apple Development Portal to manually revoke credentials. As long as we give people the option to recover from errors (ie) revoke certs when the max has been generated, and we let ppl use previously generated credentials (even from a past failed attempt to build), this should be an acceptable user experience.

packages/expo-cli/src/credentials/views/SelectDistributionCert.ts
wkozyra95: I would prefer to keep old behavior in `credentials:manager`, I agree that in most cases users would want to create and assign to experience, but I think actions there should be as basic as possible. Also, all the credentials are displayed in one place and what is connected to what should be clearly visible

quinlanj:
closing this in favour of https://github.com/expo/expo-cli/pull/1381

PR #1305, opened 2019-12-06
[expo-cli][xdl] add instant timeout to requests when `--offline` flag is passed

cruzach:
When you are working on a project, and have poor network connection (i.e. in-flight wifi), `expo start --offline` will still try to make these requests, and it takes a lot of extra time. If the user explicitly says "I might as well be offline," we should just timeout of these requests.

Right now- `expo start --offline` with poor network takes ~30 seconds. With these changes, it takes < 3 seconds

Tested changes with good wifi, really poor wifi, and no wifi


packages/expo-cli/src/commands/start.js
```diff
@@ -52,6 +52,10 @@ function parseStartOptions(projectDir: string, options: Object): Object {
     startOpts.maxWorkers = options.maxWorkers;
   }
 
+  if (options.offline) {
+    startOpts.offline = true;
```
brentvatne: what if instead of passing this arg around we instead make a module that is responsible for storing the connection status? we could call it `ConnectionStatus` i guess and then here we could do `ConnectionStatus.setIsOffline(true)` and whenever we want to know the status from another module we can use `ConnectionStatus.isOffline()` or `ConnectionStatus.isOnline()`

cruzach: Yeah, definitely cleaner than passing the flag about 5 calls deep 

Sorry I'm just getting around to addressing this feedback now

packages/xdl/src/Project.ts
```diff
@@ -1891,9 +1892,14 @@ export async function startExpoServerAsync(projectRoot: string): Promise<void> {
       extended: true,
     })
   );
-  if ((await Doctor.validateWithNetworkAsync(projectRoot)) === Doctor.FATAL) {
+  if (
+    (ConnectionStatus.isOffline()
+      ? await Doctor.validateLowLatencyAsync(projectRoot)
```
brentvatne: this method has a bit of a strange name. i think it should be renamed to `validateHighLatencyAsync` or `validateWithoutNetworkAsync`. i know it wasn't added in this pr but it could be good to clean up. will require updating other callsites ofc

packages/xdl/src/Versions.ts
```diff
@@ -50,7 +51,12 @@ type Versions = {
 export async function versionsAsync(): Promise<Versions> {
   const api = new ApiV2Client();
   const versionCache = new Cacher(
-    () => api.getAsync('versions/latest'),
+    () =>
+      api.getAsync(
+        'versions/latest',
+        undefined,
+        ConnectionStatus.isOffline() ? { timeout: 1 } : undefined
```
brentvatne: maybe we can update the api client itself to default to `{ timeout: 1 }` when the connection is offline, so we don't have to be careful to add this to each api call.

some psuedocode for the api module:

```
options = options || {};
if (!options.hasOwnProperty('timeout') && ConnectionStatus.isOffline()) {
  options.timeout = 1;
}
```

packages/xdl/src/ConnectionStatus.ts
```diff
@@ -0,0 +1,9 @@
+let offline: boolean = false;
+
+export function setIsOffline(bool: boolean): void {
+  offline = bool;
+}
+
+export function isOffline(): boolean {
+  return offline;
+}
```
brentvatne: not something important for now but in the future we might want to add a listener to this in case we want to let people switch between online/offline (or listen to network connection and do this automatically) while the cli is running already

cruzach: yeah, i think listening to the network connection and doing it automatically is better. Need to find out the best way to do that where it makes almost no diff in expo-cli performance though

packages/xdl/src/Api.ts
```diff
@@ -87,19 +88,22 @@ async function _callMethodAsync(
...
-      let { data } = convertedFormData;
-      options.headers = {
-        ...options.headers,
-        ...requestOptions.formData.getHeaders(),
-      };
-      options = { ...options, data };
-    } else {
-      options = { ...options, ...requestOptions };
-    }
+  if (!requestOptions.timeout && ConnectionStatus.isOffline()) {
```
brentvatne:
```suggestion
  if (!requestOptions.hasOwnProperty('timeout') && ConnectionStatus.isOffline()) {
```

this is important because if `timeout` is `0` then `!0` also returns `true`, so people wouldn't be able to set `timeout` to 0.

brentvatne: to be clear on the semantics - if the connection is offline then we are saying we will still respect the timeout property if it is provided

packages/xdl/src/Api.ts
```diff
@@ -87,19 +88,22 @@ async function _callMethodAsync(
...
+  if (requestOptions.formData) {
+    let convertedFormData = await _convertFormDataToBuffer(requestOptions.formData);
+    let { data } = convertedFormData;
+    options.headers = {
+      ...options.headers,
+      ...requestOptions.formData.getHeaders(),
+    };
+    options = { ...options, data };
+  } else {
+    options = { ...options, ...requestOptions };
```
brentvatne: need to be careful here not to clobber our timeout config from above. i think it should be fine but i think ideally we would move the timeout stuff after this

packages/xdl/src/Api.ts
```diff
@@ -87,19 +88,22 @@ async function _callMethodAsync(
...
+    options.timeout = 1;
   }
+
+  if (requestOptions.formData) {
+    let convertedFormData = await _convertFormDataToBuffer(requestOptions.formData);
+    let { data } = convertedFormData;
+    options.headers = {
+      ...options.headers,
+      ...requestOptions.formData.getHeaders(),
+    };
```
brentvatne: if formData and timeout are provided the timeout property will never be used

packages/xdl/src/Api.ts
cruzach: 0 means no timeout, though. Should we allow that if they're offline? That would result in an endless hang

brentvatne: i think if someone passes in a timeout we should respect the option, and it is up to the caller to decide if they want to change the timeout depending on the connection status

packages/xdl/src/Api.ts
cruzach: moved 👍

Issue #1301, opened 2019-12-05
Remove sharp dependency from expo-cli installation


brentvatne:
## Background

- sharp installations cause problems/confusion in a variety of situations on various platforms and configurations (https://github.com/expo/expo-cli/issues/1297#issuecomment-561993283)
- xdl depends on image-utils currently, but doesn't actually use it
- webpack-pwa-manifest-plugin also depends on image-utils
- image-utils has an optional dependency on sharp-cli, which has a dependency on sharp

## Proposed action

I think we should remove the image-utils dependency xdl (easy) from the webpack-pwa-manifest-plugin (more complicated) or move that plugin elsewhere so it can be installed on demand rather than needing to be installed along with each expo-cli installation.

GollyJer:
Windows 10, 64 bit. 
I get a sharp error on every install.

EvanBacon:
@GollyJer it would helpful if you could report or share these errors. Alternatives to sharp are +8x slower.

GollyJer:
@EvanBacon Just FYI. [This is the only combination of tools that works without error.](https://forums.expo.io/t/unable-to-start-expo/30460/2?u=gollyjer)
However, expo 3.10.2 actually installs even though `sharp` doesn't build and install; which is good.

I'm currently running 
- Windows 10, 64 bit
- Node 13.10

I think most of the expo-cli install errors, for Windows users, boil down to two issues.
1. The developer of a package doesn't have a pre-built binary released for Windows x64 and Node (newest).
2. When the a pre-built binary isn't found the build script isn't written correctly for Win64.

You can see this happening below... a prebuilt isn't found then a function call throws an error during the build process.

Thanks,
Jeremy


```bash
[3/4] Linking dependencies...
[4/4] Building fresh packages...
[-/7] ⢀ waiting...
[2/7] ⢀ sharp
[-/7] ⢀ waiting...
[-/7] ⢀ waiting...
warning Error running install script for optional dependency: "C:\\Users\\Jeremy\\AppData\\Local\\Yarn\\Data\\global\\node_modules\\sharp-cli\\node_modules\\sharp: Command failed.
Exit code: 1
Command: (node install/libvips && node install/dll-copy && prebuild-install) || (node-gyp rebuild && node install/dll-copy)
Arguments:
Directory: C:\\Users\\Jeremy\\AppData\\Local\\Yarn\\Data\\global\\node_modules\\sharp-cli\\node_modules\\sharp
Output:
[...]

PR #1318, opened 2019-12-09
Remove unused dependencies from xdl

EvanBacon:
search for all dependencies in the `package.json`, if they don't exist in the package, remove them.

Related #1301


brentvatne:
ya gone

PR #1335, opened 2019-12-10
fix(expo-cli): refer current sdk from empty-checked version string

### LINKED ISSUE #1331: expo client:install:ios/android should work outside of a project directory
In this case it should just install the latest version of the client. cc @byCedric

byCedric:
This fixes #1331

### DIFF
--- a/packages/expo-cli/src/commands/client/index.js
+++ b/packages/expo-cli/src/commands/client/index.js
@@ -273,14 +273,18 @@ export default program => {
 
       const latestSdk = await Versions.newestSdkVersionAsync();
       const sdkVersions = await Versions.sdkVersionsAsync();
-      const currentSdkVersion = currentSdk.sdkVersion ? sdkVersions[currentSdk.sdkVersion] : null;
+      const currentSdkVersionString =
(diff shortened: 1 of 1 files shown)


brentvatne:
i'm not sure this would address the problem, because the command will exit before this code can execute due to this: https://github.com/expo/expo-cli/blob/ac26c9c4f06fe3ba76f30b81f818ac9661de777b/packages/expo-cli/src/commands/client/index.js#L256-L269

i think the easiest fix here is to just call `Simulator.upgradeExpoAsync()` when no config is found

byCedric:
I see your point, but the no-project exception is thrown from the `findProjectRootAsync` (line 257), and skips line 258 + 259. So the part you highlighted will log and continue below.

I'm actually trying to refactor this as we speak, to make it way more readable. 😄

byCedric:
I broke up the functionality and put some logic into a utility file. If you want a quick fix, merge this. [PR 1358 fixes this issue too](https://github.com/expo/expo-cli/pull/1358), but in a nicer way 😄

brentvatne:
:) i'll go with 1358, thanks @byCedric!

PR #1338, opened 2019-12-10
[xdl] Replace pattern with Facebook scheme

sjchmiela:
Prerequisite for https://github.com/expo/expo/pull/6475.

This change won't affect other SDKs since there will be no such pattern in the `AndroidManifest` there.


brentvatne:
thanks 🙏

PR #1342, opened 2019-12-10
[config] Create dynamic config "app.config.js"

EvanBacon:
## Why 

Problems with the existing solution: `app.json`
- can't swap between modes like development and production
- no way to comment and document your config
- can't share parts of your config across multiple projects
- hard to discover what features are available 
- hard to know what everything in the config does
- config types aren't in the monorepo and can't be easily changed

## How

If a project has an `app.config.js`, that will take priority over an `app.json`. You can use the `app.config.js` to modify your `app.json` using environment variables or custom JS plugins. Here is how you might modify your project's `app.json` using the new `app.config.js`:
```js
// config is generated from the package.json and the app.json.
module.exports = function({ projectRoot, config, mode }) {
  // Use the mode to control how the app works in dev vs prod
  config.name = mode === 'development' ? 'dev app' : 'prod app';
  // Return the modified config
  return config;
}
```

Also add support for the following config types:
- `app.config.js`
- `app.config.json`
- ~~`app.config.json5`~~ - disabled via @ide request
- Added but disabled: 
  - ~~`app.config.ts`~~
  - ~~`app.config.yml`~~
  - ~~`app.config.yaml`~~
  - ~~`app.config.toml`~~

- I'm somewhat against the idea of having multiple different languages for the config. At the very least we could just add them later after `app.config.js` has migrated.
- Only allow sync configs to maintain usability with Next.js.

## Test Plan

- Add tests for reading from and validating each language
- Add a test for ensuring that `app.config` has higher priority than `app.json`
- Add a test to ensure the `app.config.js` method is passed the `app.json` and a `mode` option
- Ensure that other language `app` files don't get read (like `app.yml`, `app.json5`)
- Enforce that the config is always serialized to JSON so it can be used for the manifest.


brentvatne:
one tradeoff worth noting here is that we won't be able to automate updating config for people after this, at least not with the `.js` extension.

so when we run `expo upgrade` we can't change the sdk version, and automating updating things like version/buildNumber etc will be more difficult in the future and for packages like @byCedric's [semantic-release-expo](https://github.com/bycedric/semantic-release-expo). 

re: sdk version, honestly we should move the source of truth for this to package.json anyways, and just use whatever sdk version corresponds to the expo package. that might be something we want to do before landing this.

are there any other tradeoffs you can think of?

EvanBacon:
Currently we do read the sdkVersion from the expo/package.json. sdkVersion is only used for locking down a version or using UNVERSIONED. Having a js config opens up room for plugins so we could potentially migrate scripts to dynamic classes that you add in the config.

brentvatne:
oh i did not realize we had made that change, nice!

byCedric:
Don't worry about the semantic release, it wasn't using Expo components anyways. I'll rework that library.

Also, the `getExpoSDKVersion` is 🔥

brentvatne:
this is pretty big so i don't think i'll deploy it until the new year. will review it then as well :)

EvanBacon:
Maybe we should also add expo icon to VSCode for `app.config.js` https://github.com/vscode-icons/vscode-icons/wiki/ListOfFiles

EvanBacon:
Added a method to ensure the config is always serialized before being read.

packages/config/src/getConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+  // result = await Promise.resolve(result);
+
+  if (result instanceof Promise) {
+    throw new ConfigError(`Config file ${configFile} cannot return a Promise.`, 'INVALID_CONFIG');
+  }
+
+  // Add a generated field.
+  if (!result._expo) result._expo = {};
+  // Set the config format
+  result._expo.configType = format;
```
brentvatne: can we add some documentation on why this field is useful?

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+ */
+export function getConfig(
+  projectRoot: string,
+  options: {
+    mode: 'development' | 'production';
+    configPath?: string;
+    skipSDKVersionRequirement?: boolean;
+  }
+): ProjectConfig {
+  if (!['development', 'production'].includes(options.mode)) {
```
brentvatne: we should document if this is impacted by toggling development mode in expo-cli

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+  }
+
+  // TODO(Bacon): This doesn't support changing the location of the package.json
+  const packageJsonPath = getRootPackageJsonPath(projectRoot, {});
+  const pkg = JsonFile.read(packageJsonPath);
+
+  const configPath = options.configPath || customConfigPaths[projectRoot];
+
+  // If the app.json exists, we'll read it and pass it to the app.config.js for further modification
+  const { configPath: appJsonConfigPath } = findConfigFile(projectRoot);
```
brentvatne: i'd pull all of the logic for getting the static app.json config and ensuring default values to a separate function so it's easier to follow the logic in `getConfig`

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+
+  const { exp: configFromPkg } = ensureConfigHasDefaultValues(projectRoot, rawConfig, pkg, true);
+
+  const context = {
+    mode: options.mode,
+    projectRoot,
+    configPath,
+    config: configFromPkg,
+  };
+  const config = findAndEvalConfig(context);
```
brentvatne:
```suggestion
  const config = findAndEvalConfig(context) ?? context.config;
```

brentvatne: with this change you can get rid of `finalConfig`

brentvatne: also i'd change this to `let` so we don't need to create `configs` below

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+
+  const finalConfig = config || context.config;
+
+  const configs = ensureConfigHasDefaultValues(
```
brentvatne: can just do `config = ensureConfigHasDefaultValues(....` if we change from const to let above on config

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+  if (projectHasModule('react-native', projectRoot, exp)) {
+    platforms.push('ios', 'android');
+  }
+  if (projectHasModule('react-native-web', projectRoot, exp)) {
+    platforms.push('web');
+  }
+  return platforms;
+}
+
+export function serialize(val: any): any {
```
brentvatne: this should be part of `findAndEval` because if any property is a function we are executing it. it also makes `getConfig` a bit cleaner because it can just receive the fully evaluated config object when it falls `findAndEvalConfig` rather than having to operate over properties on it. this could be problematic if `findAndEvalConfig` ever gets used elsewhere and people forget to use `serialize`

packages/config/src/Config.ts
```diff
@@ -1,12 +1,137 @@
...
+ * - `app.config.json`
+ * - `app.json`
+ *
+ * @param projectRoot the root folder containing all of your application code
+ * @param options enforce criteria for a project config
+ */
+export function getConfig(
+  projectRoot: string,
+  options: {
+    mode: 'development' | 'production';
```
brentvatne: do we need to limit modes to just these two?

packages/config/src/getConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import { ConfigContext, ExpoConfig } from './Config.types';
+import { ConfigError } from './Errors';
+import { fileExists } from './Modules';
+
+// support all common config types
+export const allowedConfigFileNames: string[] = (() => {
+  const prefix = 'app';
+  return [
+    // order is important
```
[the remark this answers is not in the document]
EvanBacon: because you would want `app.config.js` to be used instead of `app.config.json` if both existed.

packages/config/src/Config.ts
EvanBacon: would be better to limit them, then add support for more in the future.

packages/config/src/Config.ts
EvanBacon: This PR doesn't actually implement any of that. Future PRs that migrate to this config will need to account for this.

packages/config/src/Config.ts
```diff
@@ -1,12 +1,118 @@
...
+  projectRoot: string,
+  options: GetConfigOptions
+): { context: ConfigContext; pkg: JSONObject } {
+  // TODO(Bacon): This doesn't support changing the location of the package.json
+  const packageJsonPath = getRootPackageJsonPath(projectRoot, {});
+  const pkg = JsonFile.read(packageJsonPath);
+
+  const configPath = options.configPath || customConfigPaths[projectRoot];
+
+  // If the app.json exists, we'll read it and pass it to the app.config.js for further modification
```
fson: Maybe we want to get rid of this capability, and make the JSON/JS configurations mutually exclusive, for a few reasons:
- Having the configuration split across two files is more confusing for new users or anyone coming across a project that does this. Now you must know two ways to do the same thing. Now you need to check two files to confirm what configuration value is used.
- When responding to issues the common question "please show us your app.config.js" becomes "please show us your app.config.js and app.json (and package.json)"
- JS can include JSON inside it. It can also import configuration from an arbitrary JSON file, if a static format is desired for some values.

packages/config/src/__tests__/fixtures/language-support/js/app.json
```diff
@@ -0,0 +1,4 @@
+{
+  "foo": "invalid",
+  "slug": "someslug"
```
fson:
```suggestion
  "expo": {
    "slug": "someslug"
  }
```
(This is the format we support right now.)

packages/config/src/Config.ts
```diff
@@ -1,12 +1,118 @@
...
+ * If a function is exported from the `app.config.js` then a partial config will be passed as an argument.
+ * The partial config is composed from any existing app.json, and certain fields from the `package.json` like name and description.
+ *
+ * You should use the supplied `mode` option in an `app.config.js` instead of `process.env.NODE_ENV`
+ *
+ * **Example**
+ * ```js
+ * module.exports = function({ config, mode }) {
+ *   // mutate the config before returning it.
+ *   config.slug = 'new slug'
```
fson: It seems that the `config` object currently passed in doesn't have `config.slug`, but `config.expo.slug`?

packages/config/src/getConfig.ts
```diff
@@ -0,0 +1,99 @@
...
+    const output: { [key: string]: any } = {};
+    for (const property in val) {
+      if (val.hasOwnProperty(property)) {
+        output[property] = serializeAndEvaluate(val[property]);
+      }
+    }
+    return output;
+  }
+
+  throw new Error(`Unhandled item type: ${typeof val}`);
```
fson: This error should probably show to the path of the file and location within the file too.

packages/config/src/getConfig.ts
```diff
@@ -0,0 +1,99 @@
...
+export function findAndEvalConfig(request: ConfigContext): ExpoConfig | null {
+  // TODO(Bacon): Support custom config path with `findConfigFile`
+  // TODO(Bacon): Should we support `expo` or `app` field with an object in the `package.json` too?
+
+  function testFileName(configFilePath: string): null | Partial<ExpoConfig> {
+    if (!fileExists(configFilePath)) return null;
+
+    try {
+      return evalConfig(configFilePath, request);
+    } catch (error) {
```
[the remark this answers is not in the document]
EvanBacon: Isn't this package for bundlers, I don't see how it should work with a node require.

packages/config/src/Config.ts
EvanBacon: > Having the configuration split across two files

We support this functionality for values that are generated by Expo CLI like `bundleIdentifier`. I'm totally open to better ideas though. I imagine a new project's config would look something like:

```js
module.exports = function({ config }) {
  return {
  ...config,
  icon: 'assets/icon.png',
  /* etc... */
  }
}
```

The spread of another config is somewhat unpleasant.

> responding to issues

We can add a mechanism for printing out the serialized config after it's been returned from `app.config.js`.

> if a static format is desired for some values.

I imagine eventually `app.json` will be the static location for generated values.

brentvatne: > When responding to issues the common question "please show us your app.config.js" becomes "please show us your app.config.js and app.json (and package.json)"

maybe we could add this to `expo diagnostics`? and we could automatically filter out any potentially sensitive fields

packages/config/src/getConfig.ts
```diff
@@ -0,0 +1,124 @@
...
+
+// support all common config types
+export const allowedConfigFileNames: string[] = (() => {
+  const prefix = 'app';
+  return [
+    // order is important
+    // TODO: Bacon: Slowly rollout support for other config languages: ts, yml, toml
+    `${prefix}.config.js`,
+    `${prefix}.config.json`,
+    `${prefix}.config.json5`,
```
ide: With .js support I don't think we should encourage JSON5 more. I also think the cost of supporting YAML and TOML isn't worth it. Supporting TS might be worthwhile since it feels more consistent with the rest of a project.

packages/config/src/Config.ts
EvanBacon: all we'd have to do to support this is switch the method used to read the config file.

PR #1344, opened 2019-12-11
[config] Add stronger types to ExpoConfig object

EvanBacon:
It'll be easier to enforce type safety across the entire monorepo if the `ExpoConfig` has all of the types in it.


packages/xdl/src/detach/IosNSBundle.ts
```diff
@@ -36,7 +36,7 @@ const KERNEL_URL = 'https://expo.io/@exponent/home';
 
 function _configureInfoPlistForLocalDevelopment(config: any, exp: ExpoConfig): ExpoConfig {
   // add detached scheme
-  if (exp.isDetached && exp.detach.scheme) {
+  if (exp.isDetached && exp.detach && exp.detach.scheme) {
```
brentvatne: we can use optional chaining now in this repo! so `exp.detach?.scheme` will be a better way to do this. same applies to some other things in this pr

PR #1345, opened 2019-12-11
[xdl][web] Refactor Webpack module

EvanBacon:
For projects that need to startup web from a node API, they shouldn't have to depend on the massive XDL package. This is the first step towards moving web support into its own package `@expo/webpack-adapter`.  At the moment that can't be done because the logger is in XDL so this PR simply refactors the package and cleans up the outward API.


packages/xdl/src/DevSession.ts
```diff
@@ -5,6 +5,7 @@ import Config from './Config';
 import logger from './Logger';
 import * as UrlUtils from './UrlUtils';
 import UserManager from './User';
+import { Webpack } from './xdl';
```
brentvatne: why not `import * as Webpack from './Webpack';`?

packages/xdl/src/Webpack/WebpackConfig.ts
```diff
@@ -0,0 +1,92 @@
+import { AnyConfiguration } from '@expo/webpack-config/webpack/types';
+import chalk from 'chalk';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+import { CLIWebOptions, WebEnvironment } from './Webpack.types';
+import { isDebugModeEnabled } from './WebpackEnv';
+
+export function setMode(mode: 'development' | 'production' | 'test' | 'none'): void {
```
brentvatne: we support more "modes" here than we do for app.config.js https://github.com/expo/expo-cli/pull/1342/files#diff-bf738e4e35c755a9c5f9377a673b89e6R85

packages/xdl/src/Webpack/WebpackInstance.ts
```diff
@@ -0,0 +1,101 @@
+import { getNameFromConfig, readConfigJsonAsync } from '@expo/config';
+import internalIp from 'internal-ip';
+import openBrowser from 'react-dev-utils/openBrowser';
+import { Urls } from 'react-dev-utils/WebpackDevServerUtils';
+
+import { printInstructions } from './createWebpackCompiler';
+import { DevServer, Protocol, WebpackSettings } from './Webpack.types';
+
+function getIP(): string {
+  return internalIp.v4.sync() || '127.0.0.1';
```
brentvatne: should we use the `WEB_HOST` first if it's provided? https://github.com/expo/expo-cli/pull/1345/files#diff-85a83054c319410f82d717da5681df9fR3

EvanBacon:
stale

PR #1352, opened 2019-12-12
  Convert publish endpoint to apiV2

jkhales:
Following @ide's advice, added a EXPO_LEGACY_API variable that we can tell people to set to true while a fix is deployed.

(the server side APIv2 is being reviewed here: https://github.com/expo/universe/pull/3938)

### DIFF
--- a/packages/xdl/src/ApiV2.ts
+++ b/packages/xdl/src/ApiV2.ts
@@ -186,8 +186,7 @@ export default class ApiV2Client {
       reqOptions.data = options.body;
     }
 
-    reqOptions = merge({}, reqOptions, extraRequestOptions);
-    reqOptions = merge({}, reqOptions, uploadOptions);
+    reqOptions = merge({}, reqOptions, extraRequestOptions, uploadOptions);
 
     let response;
     let result;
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Project.ts
```diff
@@ -951,9 +951,15 @@ async function _uploadArtifactsAsync({
   formData.append('iosBundle', iosBundle, 'iosBundle');
   formData.append('androidBundle', androidBundle, 'androidBundle');
   formData.append('options', JSON.stringify(options));
-  let response = await Api.callMethodAsync('publish', null, 'put', null, {
-    formData,
-  });
+
+  let response: any;
+  if (process.env.EXPO_LEGACY_API) {
```
brentvatne: can we invert this so we have a flag on `EXPO_NEXT_API` or something and then we can land this and keep using the current/legacy api for now?

packages/xdl/src/ApiV2.ts
```diff
@@ -163,6 +187,8 @@ export default class ApiV2Client {
     }
 
     reqOptions = merge({}, reqOptions, extraRequestOptions);
```
brentvatne:
```suggestion
    reqOptions = merge({}, reqOptions, extraRequestOptions, uploadOptions);
```

then remove the next line. see https://lodash.com/docs/#merge for more info and https://snack.expo.io/@notbrent/joyous-popcorn for an example

PR #1358, opened 2019-12-12
refactor(expo-cli): break functionality in client install for reusabilty

byCedric:
This is a refactor for the `expo client:install:{android|ios}` command. Should make it way more readable and separates some functionality for reuse in both install commands. Right now, I only refactored the simulator (ios) version. Mainly to check if I'm doing the right things here.

#### Highlights
- No more hard-to-read stacked `if ... else if` statements
- Gone with some duplicate code, like getting a list of available clients from sdk versions
- Bye to excessive fallback-values for possibly undefined properties

#### Other changes
- Added a "recommended" flag, next to "latest", to point developers to the most probable-to-work version. This was one of the nested if-else-if logic, by moving it here we make it easier for us and the developer I think. But I'm not sure if the "recommended" text is the best one, its based on their current sdk version. (if this is something "not desired", i'd be happy to revert it)
- Removed some extraneous logging related to "you are using the latest sdk" and "pick one of these clients, where this one is latest"

#### Scenarios

<details>
<summary>1. run from outside a project (with installing latest early etc)</summary>
scenario-1-no-sdk
</details>

<details>
<summary>2. run from sdk 32 project</summary>
scenario-2-sdk-32
</details>

<details>
<summary>3. run from sdk 33 project</summary>
scenario-3-sdk-33
</details>

<details>
<summary>4. run from sdk 34 project</summary>
scenario-4-sdk-34
</details>

<details>
<summary>5. run from sdk 35 project</summary>
scenario-5-sdk-35
</details>

<details>
<summary>6. run from sdk 36 project</summary>
scenario-6-sdk-36
</details>

### DIFF
--- a/packages/expo-cli/src/commands/client/index.js
+++ b/packages/expo-cli/src/commands/client/index.js
@@ -334,132 +334,83 @@ export default program => {
     .command('client:install:android')
     .description('Install the Expo Client for Android on a connected device or emulator')
     .asyncAction(async () => {
-      let currentSdk;
-      const getAndroidVersion = sdkVersion =>
-        sdkVersion.androidClientVersion || sdkVersion.androidVersion || 'version unknown';
+      const currentSdkConfig = await ClientUpgradeUtils.getExpoSdkConfig(process.cwd());
+      const currentSdkVersion = currentSdkConfig ? currentSdkConfig.sdkVersion : undefined;
 
-      try {
-        const { projectRoot } = await findProjectRootAsync(process.cwd());
-        const { exp } = await ConfigUtils.readConfigJsonAsync(projectRoot);
-        currentSdk = exp;
-      } catch (error) {
-        // When a developer runs the installs command outside the project,
-        // they should be able to install any version
-        if (error.code !== 'NO_PROJECT') {
-          throw error;
-        }
+      if (!currentSdkVersion) {
         log(
-          'Could not find your Expo project. If you run this command from a project, we can help pick the right Expo Client version!'
+          'Could not find your Expo project. If you run this from a project, we can help pick the right Expo Client version!'
         );
       }
 
-      let targetSdkVersion;
-      let targetSdkVersionString;
-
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/client/index.js
```diff
@@ -249,132 +250,84 @@ export default program => {
...
-          `Done! We installed the client for SDK ${targetSdkVersionString}, version ${getIosVersion(
-            targetSdkVersion
-          )}`
-        );
+      const { targetClientUrl } = await prompt({
+        type: 'list',
+        name: 'targetClientUrl',
+        message: 'Choose an SDK version to install the client for:',
+        pageSize: 20,
+        choices: availableClients.map(client => {
```
byCedric: I think it would be good if we extract this too, although it's just asking questions it contains some logic on what to present to the developer.

packages/expo-cli/src/commands/client/index.js
```diff
@@ -249,132 +250,84 @@ export default program => {
     .command('client:install:ios')
     .description('Install the Expo Client for iOS on the simulator')
     .asyncAction(async () => {
-      let currentSdk;
-      const getIosVersion = sdkVersion =>
-        sdkVersion.iosClientVersion || sdkVersion.iosVersion || 'version unknown';
+      const currentSdkConfig = await ClientUpgradeUtils.getExpoSdkConfig(process.cwd());
+      const currentSdkVersion = currentSdkConfig ? currentSdkConfig.sdkVersion : undefined;
```
byCedric: I'm not sure if we should replace this with `ConfigUtils.getExpoSDKVersion`. Is this something we should add now? (I only know about this because it was included in PR 1337 - dont ask...)

packages/expo-cli/src/commands/client/index.js
byCedric: Did it anyway, the logic is definitely reusable.

PR #1362, opened 2019-12-14
add apiV2 option for send

jkhales:
To test use universe branch: @jkhales/make-apiv2-endpoint-send 

To use set: `EXPO_NEXT_API=true`

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.js
+++ b/packages/expo-cli/src/commands/start/TerminalUI.js
@@ -201,6 +201,7 @@ export const startAsync = async (projectDir, options) => {
               let sent = false;
               try {
                 await Exp.sendAsync(sendTo, lanAddress);
+                sent = true;
                 log(`Sent link successfully.`);
               } catch (err) {
                 log(`Could not send link. ${err}`);
--- a/packages/expo-cli/src/commands/start/TerminalUI.js
+++ b/packages/expo-cli/src/commands/start/TerminalUI.js
@@ -200,9 +200,9 @@ export const startAsync = async (projectDir, options) => {
 
               let sent = false;
               try {
-                await Exp.sendAsync(sendTo, lanAddress);
+                const result = await Exp.sendAsync(sendTo, lanAddress);
+                sent = result.sent;
                 log(`Sent link successfully.`);
-                sent = true;
               } catch (err) {
                 log(`Could not send link. ${err}`);
               }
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/commands/start/TerminalUI.js
```diff
@@ -200,9 +200,9 @@ export const startAsync = async (projectDir, options) => {
 
               let sent = false;
               try {
-                await Exp.sendAsync(sendTo, lanAddress);
+                const result = await Exp.sendAsync(sendTo, lanAddress);
+                sent = result.sent;
```
brentvatne: looks like we're not regressing anything here but sent/result seem unused :O

packages/xdl/src/Exp.ts
```diff
@@ -270,7 +271,19 @@ export async function getPublishInfoAsync(root: string): Promise<PublishInfo> {
 }
 
 export async function sendAsync(recipient: string, url_: string, allowUnauthed: boolean = true) {
-  let result = await Api.callMethodAsync('send', [recipient, url_, allowUnauthed]);
+  let result;
+  if (process.env.EXPO_NEXT_API) {
+    const user = await UserManager.ensureLoggedInAsync();
+    const api = ApiV2.clientForUser(user);
+    result = await api.postAsync('send-project', {
+      //keys are what the values are called in the server.
```
brentvatne: this comment probably isn't necessary

packages/expo-cli/src/commands/start/TerminalUI.js
jkhales: This is the only place result.sent is being used. And I forced it. 
I am going to just delete line 204 and remove result.sent from the server response as well.

`Sent link successfully` is gated by sendAsync throwing an error if it's not. (sent is hardcoded 'true' on the server side).

packages/expo-cli/src/commands/start/TerminalUI.js
```diff
@@ -202,7 +202,6 @@ export const startAsync = async (projectDir, options) => {
               try {
                 await Exp.sendAsync(sendTo, lanAddress);
                 log(`Sent link successfully.`);
-                sent = true;
```
brentvatne: i think we should actually keep this since we use `sent` below in https://github.com/expo/expo-cli/pull/1362/files#diff-dfafc98a4c32ef6db0217cf6b6fcc7c7R210

jkhales: OK cool.

Issue #985, opened 2019-09-03
Provide more information when downloading/installing client for iOS and Android


brbledsoe:
I'm trying to open a React Native project on an iOS Simulator through Xcode (Expo-Cli 3.0.10 - iOS 11.4 iPhone 8 - Xcode 10.3) and it gets hung up at:
```
Downloading the latest version of Expo client app

Installing Expo client on iOS simulator
```

I'm expecting the Simulator to open with the Expo app installed with my project running but no luck. 

Here's the diagnostics: 

 Expo CLI 3.0.10 environment info:
    System:
      OS: macOS 10.14.6
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.16.2 - /usr/local/bin/node
      npm: 6.11.2 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Xcode: 10.3/10G8 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^34.0.1 => 34.0.4
      react: 16.8.3 => 16.8.3
      react-native: https://github.com/expo/react-native/archive/sdk-34.0.0.tar.gz => 0.59.8
    npmGlobalPackages:
      expo-cli: 3.0.10

I have read similar issues to this one on your GitHub page but none have helped. I'm wondering if there are problems with the versions I am using of either Xcode or the iOS Simulator.

TheInternet:
I am also having this issue.

aarshaw:
Has anyone found a workaround for this?

dashcraft:
Same issue, can we get some help with this?

lukett89:
+1

lukett89:
I had the same issue and I solved it by upgrading the version of expo-cli to 3.3.0.
Hope it will work for you too!

brentvatne:
We should show updates on the download progress and if some time passes without progress we should show some information on how to work around this -- you can download the clients from the website at https://expo.io/tools

brentvatne:
cc @byCedric this could be a good one to work on that would provide lots of value

EvanBacon:
fixed in #1364

PR #1364, opened 2019-12-15
fix(xdl): add warning when install expo is taking longer than expected

byCedric:
Adds the workaround [described by Brent here](https://github.com/expo/expo-cli/issues/985#issuecomment-562909709).

- I have no clue if the current "30 seconds"-timeout is correct. Are there any metrics available on average install time? Or does anyone knows a better threshold?
- Alternative option was to use a non-awaiting `delayAsync`. But there is no cancel feature in js promises. Using timeouts works around that. If an exception is thrown, the timeout is also not triggered because the code is exiting.

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -22,6 +22,8 @@ let _lastUrl: string | null = null;
 const BEGINNING_OF_ADB_ERROR_MESSAGE = 'error: ';
 const CANT_START_ACTIVITY_ERROR = 'Activity not started, unable to resolve Intent';
 
+const INSTALL_WARNING_TIMEOUT = 30 * 1000;
+
 export function isPlatformSupported(): boolean {
   return (
     process.platform === 'darwin' || process.platform === 'win32' || process.platform === 'linux'
@@ -127,6 +129,11 @@ export async function downloadApkAsync(url?: string) {
 }
 
 export async function installExpoAsync(url?: string) {
+  const warningTimer = setTimeout(() => {
+    Logger.global.info(
+      'This takes longer than expected, you can also download the clients from the website at https://expo.io/tools'
+    );
+  }, INSTALL_WARNING_TIMEOUT);
   Logger.global.info(`Downloading latest version of Expo`);
   Logger.notifications.info({ code: NotificationCode.START_LOADING });
   let path = await downloadApkAsync(url);
@@ -135,6 +142,7 @@ export async function installExpoAsync(url?: string) {
   Logger.notifications.info({ code: NotificationCode.START_LOADING });
   let result = await getAdbOutputAsync(['install', path]);
   Logger.notifications.info({ code: NotificationCode.STOP_LOADING });
+  clearTimeout(warningTimer);
   return result;
 }
 
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -139,18 +139,32 @@ export async function installExpoAsync(url?: string) {
     total: 100,
     width: 40,
   });
-  const warningTimer = setTimeout(() => {
-    Logger.global.info(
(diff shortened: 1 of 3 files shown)


brentvatne:
this is a good start. we should probably listen to download progress events and update the user on the % complete, and then just warn people if the download appears to have frozen.

packages/xdl/src/Simulator.ts
```diff
@@ -25,6 +25,8 @@ const SUGGESTED_XCODE_VERSION = `8.2.0`;
 const XCODE_NOT_INSTALLED_ERROR =
   'Simulator not installed. Please visit https://developer.apple.com/xcode/download/ to download Xcode and the iOS simulator. If you already have the latest version of Xcode installed, you may have to run the command `sudo xcode-select -s /Applications/Xcode.app`.';
 
+const INSTALL_WARNING_TIMEOUT = 30 * 1000;
```
brentvatne:
```suggestion
const INSTALL_WARNING_TIMEOUT = 60 * 1000;
```

packages/xdl/src/Simulator.ts
```diff
@@ -355,6 +357,11 @@ export async function _downloadSimulatorAppAsync(url?: string) {
 
 // url: Optional URL of Exponent.app tarball to download
 export async function _installExpoOnSimulatorAsync(url?: string) {
+  const warningTimer = setTimeout(() => {
+    Logger.global.info(
+      'This takes longer than expected, you can also download the clients from the website at https://expo.io/tools'
```
brentvatne:
```suggestion
      'This download is taking longer than expected. You can also try downloading the clients from the website at https://expo.io/tools'
```

byCedric:
Thanks Brent, I'll dive into it asap!

byCedric:
@brentvatne How about this, I copied the ProgressBar code from the [`AndroidCredentials.ts`](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/credentials/AndroidCredentials.ts#L124-L129). The current bunyan logger doesn't have a progress indicator or bar I think. If this is ok, I can implement it for Android too 😄

packages/xdl/src/Android.ts
```diff
@@ -121,20 +127,31 @@ export async function downloadApkAsync(url?: string) {
...
+  }, INSTALL_WARNING_TIMEOUT);
   Logger.notifications.info({ code: NotificationCode.START_LOADING });
-  let path = await downloadApkAsync(url);
+  let path = await downloadApkAsync(url, progress => bar.tick(1, progress));
   Logger.notifications.info({ code: NotificationCode.STOP_LOADING });
   Logger.global.info(`Installing Expo on device`);
   Logger.notifications.info({ code: NotificationCode.START_LOADING });
   let result = await getAdbOutputAsync(['install', path]);
   Logger.notifications.info({ code: NotificationCode.STOP_LOADING });
+  clearTimeout(warningTimer);
```
brentvatne: i think we should clear / reset the timeout on every progress event

packages/xdl/src/Android.ts
```diff
@@ -139,18 +139,32 @@ export async function installExpoAsync(url?: string) {
...
+        Logger.global.info(
+          'This download is taking longer than expected. You can also try downloading the clients from the website at https://expo.io/tools'
+        ),
+      INSTALL_WARNING_TIMEOUT
     );
-  }, INSTALL_WARNING_TIMEOUT);
+  };
+
   Logger.notifications.info({ code: NotificationCode.START_LOADING });
+  warningTimer = setWarningTimer();
```
byCedric: It's a bit weird that I have to set it here, couldn't do it in `setWarningTimer`. Mostly because TypeScript told me the value was uninitialised. 🙆‍♂

Issue #1363, opened 2019-12-15
After upgrading to Expo SDK 36, expo-cli crashes with "TypeError: chunk.msg.match is not a function"


kevgrig:
## Description

After upgrading to Expo SDK 36, I'm always getting the following error immediately after launching in my emulator. I never received it before. Issue #519 is similar but `expo start --clear` didn't help.

## Observed Behavior

```
/@expo/xdl@57.2.1/src/logs/PackagerLogsStream.ts:493
    if (chunk.msg.match(/\(node:.\d*\)/)) {
                  ^
TypeError: chunk.msg.match is not a function
    at PackagerLogsStream.chunk [as _cleanUpNodeErrors] (/@expo/xdl@57.2.1/src/logs/PackagerLogsStream.ts:493:19)
    at Object.write (/@expo/xdl@57.2.1/src/logs/PackagerLogsStream.ts:211:24)
    at Logger._emit (/usr/lib/node_modules/expo-cli/node_modules/@expo/bunyan/lib/bunyan.js:928:22)
    at Logger.info (/usr/lib/node_modules/expo-cli/node_modules/@expo/bunyan/lib/bunyan.js:1053:24)
    at Object.logInfo (/@expo/xdl@57.2.1/src/project/ProjectUtils.ts:92:27)
    at _logPackagerOutput (/@expo/xdl@57.2.1/src/Project.ts:1573:18)
    at Stream.packagerProcess.stdout.pipe.on.data (/@expo/xdl@57.2.1/src/Project.ts:1797:7)
    at Stream.emit (events.js:198:13)
    at drain (/usr/lib/node_modules/expo-cli/node_modules/through/index.js:36:16)
    at Stream.stream.queue.stream.push (/usr/lib/node_modules/expo-cli/node_modules/through/index.js:45:5)
```

## Environment

```
$ expo diagnostics

  Expo CLI 3.11.1 environment info:
    System:
      OS: Linux 5.3 Fedora 30 (Workstation Edition) 30 (Workstation Edition)
      Shell: 5.0.7 - /bin/bash
    Binaries:
      Node: 10.16.3 - /usr/bin/node
      Yarn: 1.19.2 - /usr/bin/yarn
      npm: 6.9.0 - /usr/bin/npm
    npmPackages:
      expo: ^36.0.0 => 36.0.0 
      react: 16.9.0 => 16.9.0 
[...]

PR #1368, opened 2019-12-16
Potential fix for chunk.msg.match is not a function (#1363)

kevgrig:


brentvatne:
tweaked it slightly, thanks!

kevgrig:
@brentvatne Thanks! Your tweak is good. Just last night, I was thinking that this issue probably happens when a chunk of a message happens to be, for example, an integer, and gets coerced into a non-String, so I'm more confident this will work. Do you know of any estimate when this might get pushed into a new version of the Expo client (I'm presuming a local update to expo-cli won't be sufficient to pick it up?)?

brentvatne:
this change only requires me to republish @expo/xdl and expo-cli, i plan on doing this later today after reviewing several more prs

kevgrig:
@brentvatne Awesome, thanks! Cheers.

kevgrig:
@brentvatne I upgraded to expo-cli@3.11.2 but still have the same problem; however, I noticed in my error message (below) that it still shows as xdl@57.2.1 even though the [latest publish](https://github.com/expo/expo-cli/commit/a19d624798eb2e5b8b27274aa6503477ec9e53af) seems to bring in xdl@57.4.1. I'm upgrading expo-cli with `sudo npm install --unsafe-perm -g expo-cli`; should I be doing something else to ensure xdl is upgraded?

```
/@expo/xdl@57.2.1/src/logs/PackagerLogsStream.ts:493
    if (chunk.msg.match(/\(node:.\d*\)/)) {
```

kevgrig:
Nevermind, it looks like it works after `expo start -c`! Seemingly the hypothesis is confirmed, thanks for the quick turnaround!

PR #1381, opened 2019-12-18
refactor(cli) create a dist cert selector: 2nd design

quinlanj:
second pass at https://github.com/expo/expo-cli/pull/1298

# Overview
This is part of the effort to refactor the credentials codebase - we want all our credentials code to live in `packages/expo-cli/src/credentials` and be written in Typescript. This PR:
- creates a distribution certificate selector that we will use to replace the selector in adhoc builds, and will eventually be used as the selector for `expo build:ios`. 
- polls Apple to check the validity of certificates on Expo server when possible (ie) upload, selecting existing credentials, etc
- works when user is not logged in (adhoc builds need to support anonymous users). The `Doctor.validateLowLatencyAsync` we call in `context.ts` throws an error when not in a project directory instead of returning fatal. We change `Doctor.ts` to return `FATAL` in this case.
- the selector gives the user the option to 'go back' to the previous UI
- give the user the option to see more information when presented with 'high risk' actions. For example, users are hesitant to revoke credentials and would like to see the consequences of their actions in that case.

# Testing

- [x]  cred manager add/remove dist cert works still 
- [x]  adhoc dist cert anon works
- [x]  adhoc dist cert non anon works


packages/expo-cli/src/credentials/route.ts
```diff
@@ -1,24 +1,88 @@
...
+        const previousView = this._currentView;
+        this._currentView =
+          (await this._currentView.open(this._ctx)) || (await this._quit(this._mainView));
+        this.addToHistory(previousView);
+      } catch (error) {
+        if (error instanceof QuitError) {
+          return null;
+        } else if (error instanceof GoBackError) {
+          this._currentView = this.popFromHistory() || (await this._quit(this._mainView));
+          continue;
```
wkozyra95: maybe add another `else` block instead of using `continue`

packages/expo-cli/src/credentials/views/IosDistCert.ts
```diff
@@ -39,7 +53,15 @@ export class CreateIosDist implements IView {
   async provideOrGenerate(ctx: Context): Promise<DistCert> {
     const userProvided = await promptForDistCert(ctx);
     if (userProvided) {
-      return userProvided;
+      if (!ctx.hasAppleCtx()) {
+        log(
+          "WARNING! Unable to validate Distribution Certificate due to insufficient Apple Credentials. Please double check that you're uploading valid files for your app otherwise you may encounter strange errors!"
```
wkozyra95:
```suggestion
          chalk.yellow("WARNING! Unable to validate Distribution Certificate due to insufficient Apple Credentials. Please double check that you're uploading valid files for your app otherwise you may encounter strange errors!")
```

packages/expo-cli/src/credentials/views/IosDistCert.ts
```diff
@@ -166,61 +193,257 @@ export class UpdateIosDist implements IView {
...
```
wkozyra95:
```suggestion
          chalk.yellow("WARNING! Unable to validate Distribution Certificate due to insufficient Apple Credentials. Please double check that you're uploading valid files for your app otherwise you may encounter strange errors!")
```

PR #1382, opened 2019-12-18
[expo-cli] direct users to dashboard builds page

FiberJW:
## Why

We shouldn't continue to reference the legacy /builds page now that the dashboard builds page has been out for a while and has proven to be stable.

## How

I updated the build page url generator function to return a dashboard link if a username is passed in. The output looks like what is shown below:
 
Screen Shot 2019-12-18 at 2 28 27 PM

---

Thanks for checking this out!

smile_1

### DIFF
--- a/packages/expo-cli/src/commands/build/BaseBuilder.js
+++ b/packages/expo-cli/src/commands/build/BaseBuilder.js
@@ -159,7 +159,7 @@ export default class BaseBuilder {
       return;
     }
 
-    this.logBuildStatuses(buildStatus);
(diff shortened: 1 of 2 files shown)


brentvatne:
does this work with the `owner` field? cc @tcdavis

FiberJW:
Our existing /builds page and the dashboard page use the same query that allows a user to see a build from any team that they are a part of, so yes 👍

tcdavis:
Assuming dashboard/username/builds are just showing projects under the @username namespace and not all projects username has access to, looks like no.

PR #1409, opened 2020-01-07
Set bundle identifier on an ejected bare project

### LINKED ISSUE #1327: ejecting to bare causes bundle ID to change
If you have the bundle ID defined in your `app.json`, we should use that value in the bare project after ejecting.

## Repro

1. bootstrap a new managed expo project
2. add bundle ID to the app.json
```json
{
  "expo": {
    "ios": {
      "bundleIdentifier": "com.important.thing"
    }
  }
}
```
3. eject to bare
4. iOS project now has a random bundle ID `org.reactjs.native.example.aasa`

EvanBacon:
fix #1327


brentvatne:
thanks 🙏

PR #1424, opened 2020-01-09
refactor(cli) replace push key selector in adhoc builds

quinlanj:
The ios push key version of https://github.com/expo/expo-cli/pull/1299


packages/expo-cli/src/commands/client/index.js
```diff
@@ -127,7 +128,18 @@ export default program => {
...
+
+      let pushKey;
+      if (user) {
+        await runCredentialsManager(
+          context,
+          new SetupIosPush({ experienceName, bundleIdentifier })
+        );
+        pushKey = await context.ios.getPushKey(experienceName, bundleIdentifier);
+      } else {
+        pushKey = await new CreateIosPush().provideOrGenerate(context);
```
wkozyra95: push key is server-side credential so how can the anonymous user make any use of it?

quinlanj: good catch, thanks!

PR #1432, opened 2020-01-13
[webpack] Remove ModuleScopePlugin

EvanBacon:
Seems like this causes a lot of mismatch between Metro config and Webpack config. This also throws an incorrect error pointing to the `src/` folder which is something we don't enforce. 
We can investigate adding a solution like this back in the future.


brentvatne:
would be useful to refer to some issues that this causes in the pr description so it's clear if someone wants to refer back to it later

Issue #1436, opened 2020-01-14
expo build:ios, SSL verify failed


ayakayakak:
## Description

I tried to build ios standalone app, but it failed with SSL_connect error.
Same error is reported here:
https://forums.expo.io/t/expo-build-ios-ssl-verify-failed/32102

## Expected Behavior

I ran the command below.
```
expo build:ios -c

Checking if there is a build in progress...

Removed existing credentials from expo servers
Please enter your Apple Developer Program account credentials. These
credentials are needed to manage certificates, keys and provisioning profiles
in your Apple Developer account.
The password is only used to authenticate with Apple and never stored.
? Apple ID: xxxxxxxxxxx
? Password (for xxxxxxxxxxx): [hidden]
```

## Observed Behavior

```
Trying to authenticate with Apple Developer Portal...
Authentication with Apple Developer Portal failed!
Reason: Unknown reason, raw: "SSL_connect returned=1 errno=0 state=error: certificate verify failed"
Error: Reason: Unknown reason, raw: "SSL_connect returned=1 errno=0 state=error: certificate verify failed"
    at runAction (/expo-cli@3.11.2/src/appleApi/fastlane.ts:28:17)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

## Environment

expo diagnostics result:
```
Expo CLI 3.11.2 environment info:
    System:
      OS: macOS 10.15.2
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 8.13.0 - ~/.nodebrew/current/bin/node
      Yarn: 1.19.2 - /usr/local/bin/yarn
      npm: 6.13.1 - ~/.npm-global/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 11.2/11B52 - /usr/bin/xcodebuild
    npmPackages:
      @storybook/react-native: ^5.2.6 => 5.2.6 
      @types/expo: ^32.0.11 => 32.0.13 
      @types/react: ^16.8.14 => 16.9.11 
      @types/react-native: ^0.57.51 => 0.57.65 
      @types/react-navigation: ^3.4.0 => 3.4.0 
      expo: ^35.0.0 => 35.0.1 
      react: 16.8.3 => 16.8.3 
      react-native: https://github.com/expo/react-native/archive/sdk-35.0.0.tar.gz => 0.59.8 
[...]

PR #1446, opened 2020-01-15
fix(traveling-fastlane) cURL cURL's pem

quinlanj:
# Problem
Traveling ruby is a self-contained ruby with no dependencies. This means they also bundle their own ca certs. Unfortunately many of these certs are expired, causing https://github.com/expo/expo-cli/issues/1436

# Assumes
- we trust Mozilla and cURL

# How
- Mozilla has public Certificate Authorities (CA) available in its source tree: https://hg.mozilla.org/releases/mozilla-release/raw-file/default/security/nss/lib/ckfw/builtins/certdata.txt
- cURL parses this certdata.txt and extracts CA Root Certificates into PEM format, and makes it publicly available on their website
- Every time we release traveling fastlane, we replace `traveling-ruby`'s expired ca bundle with cURL's most up to date bundle.

# Test

- [x] ca bundle is not expired
```
$ openssl x509 -enddate -noout -in ca-bundle.crt 
notAfter=Jan 28 12:00:00 2028 GMT
```

### DIFF
--- a/packages/traveling-fastlane/Rakefile
+++ b/packages/traveling-fastlane/Rakefile
@@ -118,7 +118,7 @@ def create_package target
   sh "tar -xzf packaging/traveling-ruby-#{TRAVELING_RUBY_VERSION}-#{target}.tar.gz -C #{package_dir}/lib/ruby"
 
   # Hack to replace traveling ruby's expired ca-bundle.crt with cURL's most recent one
-  sh "wget #{CA_CERT_URL} -O #{package_dir}/lib/ruby/lib/ca-bundle.crt"
+  sh "curl -o #{package_dir}/lib/ruby/lib/ca-bundle.crt #{CA_CERT_URL}"
 
   for script in SCRIPTS do
     if script != 'funcs' && script != 'preload'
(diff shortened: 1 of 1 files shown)


packages/traveling-fastlane/Rakefile
```diff
@@ -115,6 +116,10 @@ def create_package target
   end
   sh "mkdir #{package_dir}/lib/ruby"
   sh "tar -xzf packaging/traveling-ruby-#{TRAVELING_RUBY_VERSION}-#{target}.tar.gz -C #{package_dir}/lib/ruby"
+
+  # Hack to replace traveling ruby's expired ca-bundle.crt with cURL's most recent one
+  sh "wget #{CA_CERT_URL} -O #{package_dir}/lib/ruby/lib/ca-bundle.crt"
```
ide: Since we rely on curl elsewhere in this file, I'd prefer to use curl here unless there's some bootstrapping issue (what copy of curl/wget runs here?)

quinlanj: the person who is releasing traveling-fastlane will have their local copy of `curl`/`wget` run here. I used wget before just because it downloads recursively (see [here](https://unix.stackexchange.com/questions/47434/what-is-the-difference-between-curl-and-wget))

quinlanj: the person who is releasing traveling-fastlane will have their local copy of `curl`/`wget` run here. I used wget before just because it downloads recursively (see [here](https://unix.stackexchange.com/questions/47434/what-is-the-difference-between-curl-and-wget))

PR #1457, opened 2020-01-17
fix(webpack-config): register general service worker in development

jaulz:
Currently, no service worker is registered at all in development which makes it difficult to test service worker related functionality. This PR will always register the `expo-service-worker` but will still avoid Workbox in development.


jaulz:
In development we have the noop service worker instead: 
https://github.com/expo/expo-cli/blob/fd2d5bd4d8d9964921c1854af834c3a3e982d4dc/packages/webpack-config/src/addons/withDevServer.ts#L109

Or are there any other implications?

PR #1461, opened 2020-01-20
fix(webpack-config): remove compressed versions from manifest

jaulz:
Currently, even the compressed versions of all files are included in the manifest which causes the files to be downloaded twice by the service worker. This fix will remove them from the manifest.

### DIFF
--- a/packages/webpack-config/src/webpack.config.ts
+++ b/packages/webpack-config/src/webpack.config.ts
@@ -260,6 +260,10 @@ export default async function(
       new ManifestPlugin({
         fileName: 'asset-manifest.json',
         publicPath,
+        filter = ({ path }) => {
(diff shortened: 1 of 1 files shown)


EvanBacon:
Could you please supply some documentation regarding why this is a requirement.

jaulz:
As far as I know the gzipped version is relevant for the server and not for the client, i.e. the server will usually automatically offer the gzipped version if it finds the ".gz" suffix. If we keep both the gzipped file name and the uncompressed file name in the manifest, both will be downloaded but the client actually only needs the latter.

jaulz:
I just noticed that also the service worker files are included and we could remove them as well from the manifest to avoid unnecessary requests.

PR #1479, opened 2020-01-25
feat(expo-cli): add menu for build type

### LINKED ISSUE #1377: Present options when running expo build:[ios/android]
The goal here is to make the various options more discoverable - so if you run `expo build:android` without any flags you will get to choose between an APK or an AAB and there will be some information about each. Similarly, we want to show people that it's possible to do simulator builds on iOS and we can expose that nicely in this kind of menu.

byCedric:
Fixes #1377

This is a pretty simple setup for the menu Brent described in #1377. The `askBuildType` accepts both the type (from flag) and available build types with a description (plain object).

### What did you change?
1. Moved the `build:ios` type check to `askBuildType` method in `build/utils.js`
2. Added menu when build type is not defined
3. Implemented `askBuildType` in both `build:ios` and `build:android`
4. Removed the default values (in interactive mode, for non-interactive it uses the first default built type)
5. Added nonInteractive check for old behavior to avoid breaking changes in people's CI/CD

> By adding point 4, this change won't be a breaking change for non-internactive environments. Removing this is possible, but that would probably conflict with CI/CD systems where no type is defined.

### Default option value and validation
It might be better to remove the option regex. Right now, it uses this regex and prevents bad input. I noticed that this does not work as expected with default values. I added an example below, but basically this happens:

- `expo build:android --type app-bundle` -> `{ type: app-bundle }`
- `expo build:android --type polymer` -> `{ type: apk }`
- `expo build:ios` -> `{ type: archive }`

### Examples

<details>
<summary><code>expo build:android</code></summary>
Android example
</details>

<details>
<summary><code>expo build:ios</code></summary>
iOS example
</details>

<details>
<summary>Weird result I mentioned</summary>
iOS example
</details>

### DIFF
--- a/packages/expo-cli/src/commands/build/index.ts
+++ b/packages/expo-cli/src/commands/build/index.ts
@@ -29,12 +29,7 @@ export default function(program: Command) {
       '--apple-id <login>',
       'Apple ID username (please also set the Apple ID password as EXPO_APPLE_PASSWORD environment variable).'
     )
-    .option(
-      '-t --type <build>',
-      'Type of build: [archive|simulator].',
-      /^(archive|simulator)$/i,
-      'archive'
-    )
+    .option('-t --type <build>', 'Type of build: [archive|simulator].', 'archive')
     .option('--release-channel <channel-name>', 'Pull from specified release channel.', 'default')
     .option('--no-publish', 'Disable automatic publishing before building.')
     .option('--no-wait', 'Exit immediately after scheduling build.')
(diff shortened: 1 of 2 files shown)


brentvatne:
yeah the regex is weird, we should just support only the exact inputs. otherwise this looks great to me! sorry i forgot about this!

byCedric:
No worries! 😄 I'll rebase and update it this weekend!

byCedric:
@brentvatne finished it! 😄

PR #1482, opened 2020-01-28
refactor(cli,fastlane) add provisioning profile selector

quinlanj:
# Overview
The provisioning profile version of https://github.com/expo/expo-cli/pull/1381 plus some additional things (see below)

# Traveling Fastlane
- added `list()` and `use-existing()` profile methods
- added `name` parameter to `create()` so there is no naming clash error on Apple Portal
- changed `extraArgs` to be a splat operator so we can put in more easily add >1 extra params

# Provisioning Profile Selector
Unlike dist certs and apns, Apple doesn't cap the max number of profiles you can make. However, we try to re-use profiles where possible to avoid creating a bunch of unnecessary ones and cluttering up the user's UI.

We select the provisioning profile the following way:
- If there is a configured profile, use it
- Check user's apple account for existing profiles, preferring the ones with the same distCert
- Give user option to choose from existing profiles or to create/upload a new one
- Once we've chosen a profile, we attempt to update with our distCert and validate where possible

# TODO

- [x] land and publish this to fastlane https://github.com/expo/expo-cli/pull/1606 , upgrade expo-cli deps


packages/expo-cli/src/appleApi/provisioningProfile.ts
```diff
@@ -12,15 +22,51 @@ export class ProvisioningProfileManager {
...
+  ): Promise<ProvisioningProfile> {
+    const args = [
+      'use-existing',
+      this.ctx.appleId,
+      this.ctx.appleIdPassword,
+      this.ctx.team.id,
+      String(this.ctx.team.inHouse),
+      bundleIdentifier,
+      provisioningProfile.provisioningProfileId,
+      distCert.distCertSerialNumber || '__last__',
```
wkozyra95: is fallback to `__last__` correct behavior here?
I'm not 100% sure but older dist cert and provided manually might not have a serial number, we should probably generate it
https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/PKCS12Utils.ts#L4

I'm not exactly sure where to handle that, it could be done every place like this where we relly on serial number or on get api calls.

packages/expo-cli/src/appleApi/provisioningProfile.ts
```diff
@@ -12,15 +22,51 @@ export class ProvisioningProfileManager {
...
     const args = [
       'create',
       this.ctx.appleId,
       this.ctx.appleIdPassword,
       this.ctx.team.id,
-      this.ctx.team.inHouse,
+      String(this.ctx.team.inHouse),
       bundleIdentifier,
-      metadata.distCertSerialNumber || '__last__',
+      distCert.distCertSerialNumber || '__last__',
```
wkozyra95: the same as before

PR #1492, opened 2020-01-30
Always show web in dev tools

EvanBacon:
Now that web support is [approaching V1](https://github.com/expo/expo/issues/6782), we should make it always visible in the dev tools.

- add web to the Terminal UI all the time
- add ability to start Webpack from dev-tools page

### DIFF
--- a/packages/dev-tools/common/state.js
+++ b/packages/dev-tools/common/state.js
@@ -166,6 +166,62 @@ function updateLayout(client, id, input) {
   });
 }
 
+export const openBrowser = async props => {
+  const id = new Date().getTime();
+
+  // NOTE(jim): Breaks out of publishing modal.
+  props.dispatch({
+    type: 'UPDATE',
+    state: {
+      isPublishing: false,
+    },
+  });
+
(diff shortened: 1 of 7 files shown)


EvanBacon:
Not sure how to test the dev-tools properly. Seems like it should always work but instead I get the following error:

Screen Shot 2020-01-29 at 6 51 26 PM

To test I'm:
- running `expod start` in a project
- running `yarn start <path to project>` in `packages/dev-tools`

brentvatne:
I think this should say "Run in web browser" rather than "Start Expo web"

cc @fson re: testing

EvanBacon:
@brentvatne the title switches after the Webpack dev server starts.

brentvatne:
@EvanBacon - do we need to differentiate between the state where webpack is started or not? I think if we just say open in web browser and then handle starting up webpack in order to do it that should be fine.

brentvatne:
we should also show this in the cli
Screen Shot 2020-01-30 at 13 58 28

PR #1501, opened 2020-01-31
feat: add tentative support for Yarn v2 using node-modules plugin

fson:
Yarn 2 "Berry" uses plug'n'play instead of a `node_modules` folder by default. This is not yet supported by React Native tooling, including Metro, and Expo CLI.

To make Berry work in a backwards compatible way with a `node_modules` folder, it includes the node-modules plugin. This PR enables the node-modules plugin in projects created with `expo init`.

**Note:** *Yarn 2.0.0-rc.27 has a bug (https://github.com/yarnpkg/berry/issues/820) where installing packages will fail on the first try. This has already been fixed in Berry `master`. Testing with the `master` version is possible by running `yarn set version from sources` in the project folder after creating a project.*

------

For existing projects, we will need to document how to enable Yarn 2 support, i.e. one must add this line to a `.yarnrc.yml` file in the project folder:
```yml
nodeLinker: node-modules
```

In the future, we will want to support plug'n'play as well. As a first step towards plug'n'play support, I'll try to enable it in the `expo-cli` repository, so we can get more experience using it and learn more about potential issues.


packages/xdl/src/Exp.ts
```diff
@@ -184,16 +186,16 @@ async function initGitRepoAsync(root: string) {
...
+  const options = { cwd: projectRoot };
   if (packageManager === 'yarn') {
-    await spawnAsync('yarnpkg', ['install'], {
-      cwd: projectRoot,
-      stdio: 'inherit',
-    });
+    const yarn = new YarnPackageManager(options);
+    const version = await yarn.versionAsync();
+    if (semver.satisfies(version, '>=2.0.0-rc.24')) {
+      await fs.writeFile(path.join(projectRoot, '.yarnrc.yml'), 'nodeLinker: node-modules\n');
```
brentvatne: we should probably:
- let the user know that we're doing this
- bail out if the file already exists, check if it contains `nodeLinker: node-modules`, warn them if not

fson:
Added this warning:
Screen Shot 2020-02-10 at 18 00 36

Also modified it so that if an existing `.yarnrc.yml` is found, it gets modified to use the `nodeLinker: node-modules` setting.

> bail out if the file already exists, check if it contains nodeLinker: node-modules, warn them if not

In the end, I chose not to do this, because running Yarn 2 without this setting will create a project that won't work correctly.

Issue #1494, opened 2020-01-30
Eject to bare workflow umbrella issue


cruzach:
## Description

Ejecting to bare is going to get a lot more popular, this is an umbrella issue for all the improvements we can make to provide a smooth ejection process

- [x] Better error checking on `expo eject` (check for spaces, even if in non-interactive) 
- [x] escape characters for XML (values that go into `info.plist` or `AndroidManifest.xml`)
- [x] alert if you're importing anything from the `expo` package (no modules left to import from the expo package now 👍 )
- [ ] warn if you're relying on `Constants.manifest` anywhere (maybe this *will* be supported with the bare OTA module?)
- [x] if you try to run a command that expo doesn't support for bare, it should let you know

Apply app-specific config, if one of these keys are present in `app.json`:
- [x] version
- [x] orientation
- [x] userInterfaceStyle - also iOS and android specific values. default to `light`
- [x] Splash- also iOS and android specific values
- [x] backgroundColor - also iOS and android specific values
- [x] primaryColor
- [x] icon  also iOS and android specific values
- [x] androidStatusBar
- [x] androidNavigationBar 
- [x] scheme
- [x] locales
- [x] owner

iOS: 
- [x] buildNumber
- [x] “Config” object - branch, usesNonExemptEncryption, googleMapsApiKey, googleMobileAdsAppId, googleMobileAdsAutoInit,  googleSignIn
- [x] googleServicesFile
- [x] supportsTablet
- [x] istabletonly
- [x] infoPlist
- [x] associatedDomains
- [x] usesAppleSignIn
- [x] usesIcloudStorage
- [x] accessesContactNotes

Android:
- [x] versionCode
- [x] can't use java reserved words for package name
- [x] adaptiveIcon
- [x] Permissions
- [x] googleServicesFile
- [x] “Config” object - branch, usesNonExemptEncryption, googleMapsApiKey, googleMobileAdsAppId, googleMobileAdsAutoInit,  googleSignIn
- [x] Splash
- [x] intentfilters

Facebook:
- [x] facebookappid
- [x] Facebookautoinitenabled
- [x] facebookAutoLogAppEventsEnabled
- [x] facebookAdvertiserIDCollectionEnabled
- [x] facebookDisplayName
- [x] facebookScheme

brentvatne:
>  alert if you're importing anything from the expo package

it'd be nice to explore a better way to handle this, so the app still loads but we just give a meaningful warning about it

cruzach:
[...]

PR #1502, opened 2020-01-31
fix: apply version, buildNumber, and versionCode on eject

cruzach:
ref https://github.com/expo/expo-cli/issues/1494


brentvatne:
I think we want to take a different approach here. We can continue to use this `Transformer` mechanism for modifying the `app.json` file, but I think that we should do something similar to the [setBundleIdentifier](https://github.com/expo/expo-cli/blob/354a5f50d7abbec91537f440281249e306b12e1b/packages/xdl/src/detach/IosSetBundleIdentifier.ts) function for applying the app config to the iOS and Android projects. See where we use that function [here](https://github.com/expo/expo-cli/blob/354a5f50d7abbec91537f440281249e306b12e1b/packages/expo-cli/src/commands/eject/Eject.ts#L330-L334).

PR #1520, opened 2020-02-05
[xdl] Migrate simulator and emulator to use new getConfig method

EvanBacon:


packages/xdl/src/Android.ts
```diff
@@ -407,7 +410,10 @@ const splashScreenDPIConstraints = [
  * @since SDK33
  */
 export async function checkSplashScreenImages(projectDir: string): Promise<void> {
-  const { exp } = await readConfigJsonAsync(projectDir);
+  const { dev } = await ProjectSettings.readAsync(projectDir);
+  const { exp } = getConfig(projectDir, {
+    mode: dev ? 'development' : 'production',
```
fson: Having to do this every time you need the config seems a bit cumbersome... Perhaps there's a way we could default to the current setting, or add some helper method to get the config for currently enabled mode? (Or at least a function somewhere that returns the current mode `development`/`production` instead of the `dev` boolean?)

PR #1521, opened 2020-02-05
[xdl] Migrate getSlug to use getConfig

EvanBacon:
Add support for `app.config.js`

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -3,6 +3,7 @@ import {
   PackageJSONConfig,
   Platform,
   configFilename,
+  getConfig,
   projectHasModule,
   readConfigJson,
   readConfigJsonAsync,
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Project.ts
```diff
@@ -357,14 +358,17 @@ function _requireFromProject(modulePath: string, projectRoot: string, exp: ExpoC
 }
 
 // TODO: Move to @expo/config
-export async function getSlugAsync(projectRoot: string, options = {}): Promise<string> {
-  const { exp } = await readConfigJsonAsync(projectRoot);
+export async function getSlugAsync(
+  projectRoot: string,
+  options: { mode: 'production' | 'development'; [key: string]: any }
+): Promise<string> {
+  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true, mode: options.mode });
```
fson: Maybe we should just hard-code `mode: production` here? I think distinguishing between development slug / production slug makes `getSlugAsync` too complicated for what it is.

EvanBacon: the ability to toggle may be valuable for people who want to test it on a different slug locally ¯\_(ツ)_/¯

PR #1524, opened 2020-02-05
[xdl][Project] Support app.config.js

EvanBacon:
- _getExpAsync(): use slugify to transform config name from ThirdParty module
- Update types
- Remove unused variables
- Add support for app.config.js when starting a project on any platform
- Add support for app.config.js when getting build status, and starting a build.

## Notes

- I guess you can make config changes then refresh the client (⌘R) to see them.

# Test Plan

This effects:
- `expo start`
- `expo start:web`
- `expo publish`
- `expo url`

Each should be run with an `app.config.js` and `app.json`.

### Test Project

`app.json`
```json5
{
  "expo": {
    "version": "1.0.0"
    // ...
  }
}
```

`app.config.js`
```js
module.exports = function({ config, mode }) {
    return {
        ...config,
        name: "My Project - " + mode
    }
}
```

`App.js`
```js
import React from 'react';
import { Text } from 'react-native';
import Constants from 'expo-constants';

export default () => <Text>Data: {Constants.manifest.name}</Text>
```

### DIFF
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -85,6 +85,7 @@
     "semver": "5.5.0",
     "serialize-error": "^5.0.0",
     "slugid": "1.1.0",
+    "slugify": "^1.3.6",
     "source-map-support": "0.4.18",
     "split": "1.0.1",
     "tar": "4.4.6",
--- a/packages/xdl/src/Config.ts
+++ b/packages/xdl/src/Config.ts
@@ -13,7 +13,7 @@ import * as ProjectSettings from './ProjectSettings';
  */
 export async function getProjectConfigAsync(
   projectRoot: string,
-  options: Partial<GetConfigOptions>
+  options: Partial<GetConfigOptions> = {}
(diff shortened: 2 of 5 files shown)


EvanBacon:
I'm thinking it might make sense to set an environment variable which we use for getting the environment in expo-cli. CC @fson

fson:
What would this look like for a user of Expo CLI? Where would this environment variable usually be set?

rungwe:
Hi @EvanBacon, in the app.config.js, there are two properties which we can extract,  
**_config_** and **_mode_**
I do understand what config is, I am not sure what **mode** is? is this the release channel.

 I am trying to get the release channel info so that I can apply some logic in the function to determine which config properties to use 

`module.exports = function({ config, mode }) {
    return {
        ...config,
        name: "My Project - " + mode
    }
}`

arlyon:
@rungwe `mode: packagerOpts.dev ? 'development' : 'production'`

PR #1532, opened 2020-02-08
fix typo options.mode -> mode

jkhales:
`expod ph` was broken:
Screen Shot 2020-02-07 at 4 04 43 PM

Seems like `option.mode` was used before...  `mode` is defined by destructing now.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -365,7 +365,7 @@ export async function getSlugAsync(
     ...options
   }: { mode?: 'production' | 'development'; [key: string]: any } = {}
 ): Promise<string> {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Project.ts
```diff
@@ -365,7 +365,7 @@ export async function getSlugAsync(
     ...options
   }: { mode?: 'production' | 'development'; [key: string]: any } = {}
 ): Promise<string> {
-  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true, mode: options.mode });
```
EvanBacon: mode is destructured from options on line 364, it doesn't exist in options

jkhales: Yea, isn't that was this PR is fixing?

PR #1533, opened 2020-02-08
autofill expo-updates properties on `expo publish`

esamelson:
partial step of expo/expo#6517

In order to preserve the current functionality of ExpoKit updates in the new expo-updates unimodule, we want to set certain properties in the native iOS and Android projects when running `expo publish`. This PR adds that functionality for the properties defined in expo/expo#6625 and expo/expo#6787.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -938,9 +938,9 @@ export async function publishAsync(
           'AndroidManifest.xml'
         );
         let androidManifestXmlFile = fs.readFileSync(androidManifestXmlPath, 'utf8');
-        let expoUpdateUrlRegex = /<meta-data android:name="expo.modules.updates.EXPO_UPDATE_URL" android:value="[^"]*" \/>/;
-        let expoSdkVersionRegex = /<meta-data android:name="expo.modules.updates.EXPO_SDK_VERSION" android:value="[^"]*" \/>/;
-        let expoReleaseChannelRegex = /<meta-data android:name="expo.modules.updates.EXPO_RELEASE_CHANNEL" android:value="[^"]*" \/>/;
+        let expoUpdateUrlRegex = /<meta-data[^>]+"expo.modules.updates.EXPO_UPDATE_URL"[^>]+\/>/;
+        let expoSdkVersionRegex = /<meta-data[^>]+"expo.modules.updates.EXPO_SDK_VERSION"[^>]+\/>/;
+        let expoReleaseChannelRegex = /<meta-data[^>]+"expo.modules.updates.EXPO_RELEASE_CHANNEL"[^>]+\/>/;
 
         let expoUpdateUrlTag = `<meta-data android:name="expo.modules.updates.EXPO_UPDATE_URL" android:value="${fullManifestUrl}" />`;
         let expoSdkVersionTag = `<meta-data android:name="expo.modules.updates.EXPO_SDK_VERSION" android:value="${exp.sdkVersion}" />`;
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -932,12 +932,12 @@ export async function publishAsync(
           'AndroidManifest.xml'
         );
         let androidManifestXmlFile = fs.readFileSync(androidManifestXmlPath, 'utf8');
-        let expoAppUrlRegex = /<meta-data android:name="expo.modules.updates.EXPO_APP_URL" android:value="[^"]*" \/>/;
-        if (androidManifestXmlFile.search(expoAppUrlRegex) < 0) {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Project.ts
```diff
@@ -892,26 +902,87 @@ export async function publishAsync(
...
+          'android',
+          'app',
+          'src',
+          'main',
+          'AndroidManifest.xml'
+        );
+        let androidManifestXmlFile = fs.readFileSync(androidManifestXmlPath, 'utf8');
+        let expoUpdateUrlRegex = /<meta-data android:name="expo.modules.updates.EXPO_UPDATE_URL" android:value="[^"]*" \/>/;
+        let expoSdkVersionRegex = /<meta-data android:name="expo.modules.updates.EXPO_SDK_VERSION" android:value="[^"]*" \/>/;
+        let expoReleaseChannelRegex = /<meta-data android:name="expo.modules.updates.EXPO_RELEASE_CHANNEL" android:value="[^"]*" \/>/;
```
fson: Modifying the Android manifest using regular expressions like these seems very prone to error. Because the manifest is also modified by other tools, even a line break in the middle of one of these tags or a different order of XML attributes could break this, even though both are valid in XML.
 
Could we adopt a different approach that makes use of an XML parser instead of regexes to make these modifications in the file? You could look in https://github.com/expo/expo-cli/blob/4472a15a54468be7dae3e477d2dce7c7be21f60e/packages/config/src/android/Manifest.ts for an existing example of this in this codebase.

packages/xdl/src/Project.ts
```diff
@@ -892,26 +902,87 @@ export async function publishAsync(
...
+          constantsPath
+        );
+        await ExponentTools.regexFileAsync(
+          /RELEASE_CHANNEL = "[^"]*"/,
+          `RELEASE_CHANNEL = "${options.releaseChannel}"`,
+          constantsPath
+        );
+      } else {
+        // This is an app with expo-updates installed, so set the applicable meta-data properties in
+        // AndroidManifest.xml
```
fson: As far as I know, managed apps don't have `AndroidManifest.xml` (an `android` folder) and I haven't been aware that `expo publish` might be used with bare apps.

The `if` branch above says it's for ExpoKit apps. Then what kind of apps can land in this `else` branch? Additionally, what does the workflow enabled by this code look like for the user?

packages/xdl/src/Project.ts
esamelson: Thanks for pointing this out. I tried 2 different libraries for parsing and writing xml files -- xml2js (which is used in @expo/config) and [xml-js](https://github.com/nashwaan/xml-js), which is more fully featured and offers a non-compact mode. This work is visible in the [`@eric/expo-updates-wip`](https://github.com/expo/expo-cli/tree/%40eric/expo-updates-wip) branch.

Both approaches had issues -- in particular, parsing and rewriting the entire AndroidManifest.xml file created a much larger diff than I wanted. xml2js removed comments and changed the order of elements. xml-js was better, but despite offering some customization options, made many whitespace changes to the file. I suppose this is only ever an issue the first time someone runs this command, but I suspect it will affect almost everyone, since the default Android Studio XML style is not possible to achieve with this library. I could not find another library for building XML from JS with more whitespace customization options.

My opinion is that since this is supposed to be a convenient side-effect of running a command with an entirely different purpose (`expo publish`), it's better for us to have the brittle regexes that sometimes fail than to unintentionally create large and confusing diffs in developers' project files, which feels icky to me. Developers can always change these settings themselves (and there is clear documentation on how to do so). I did make a change to the regexes to make them more forgiving in terms of whitespace and attribute order. Let me know what you think!

packages/xdl/src/Project.ts
esamelson: This `else` branch is for bare workflow apps with `expo updates`, in which case `expo publish` will need to be used in order to host an update on Expo's servers. The workflow is similar to how OTA updates with ExpoKit apps work right now, but I expect to improve it over the coming months after pushing out this initial version with SDK 37.

esamelson: I've changed this to just check for `expo-updates` instead of using this `else` branch -- this should be more clear. (thanks for the suggestion @sjchmiela !)

packages/xdl/src/Project.ts
fson: That seems like an acceptable tradeoff for now. If we at some point end up developing a tool to update AndroidManifest.xml without regexes, this code is quite clear and should be easy to adapt to use that tool. (I suspect that `Manifest.ts`, given that it’s also using xml2js, might also be suffering from some of the problems you mentioned, so we might visit this in the future.)

PR #1553, opened 2020-02-18
[expo-cli]  Add firebase web-config support

IjzerenHein:
This PR add support for putting the firebase web configuration in `app.json`. This is uses for auto-configuration of firebase on Web and to make Firebase Analytics possible on the standard Expo Client, through a pure JS Firebase Analytics tracker which requires the web-config `measurementId`.

### DIFF
--- a/packages/config/src/Config.types.ts
+++ b/packages/config/src/Config.types.ts
@@ -520,6 +520,14 @@ export type WebPlatformConfig = {
    * Configuration for PWA splash screens.
    */
   splash?: WebSplashScreen;
+  config?: {
+    /**
+     * [Firebase Configuration Object](https://support.google.com/firebase/answer/7015592) Firebase web configuration.
+     */
+    firebase?: {
+      [key: string]: any;
+    };
+  };
 };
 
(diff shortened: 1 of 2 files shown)


packages/config/src/Config.types.ts
```diff
@@ -520,6 +520,14 @@ export type WebPlatformConfig = {
    * Configuration for PWA splash screens.
    */
   splash?: WebSplashScreen;
+  config?: {
+    /**
+     * [Firebase Configuration Object](https://support.google.com/firebase/answer/7015592) Firebase web configuration.
+     */
+    firebase?: {
```
EvanBacon: Perhaps `googleServicesFile` so it matches iOS and Android?

IjzerenHein: I initially considered that as well, but was having trouble parsing the file through the webpack plugin. My code wasn't being called in places where I expected it.  And in `createEnvironmentConstants` I couldn't access (or don't know how to safely) the project-root.

Do you have any recommendations for me?

Issue #1557, opened 2020-02-18
Split XDL detach folder into a new module


EvanBacon:
- XDL `detach/` is very large
- detach is hard to test and needs unit tests at the very least
- detach is not fully converted to TypeScript 
- When we deprecate ExpoKit we will have a lot of dead code that no one in particular is a code owner to. #1513

EvanBacon:
closing in favor of deprecating the whole thing

PR #1559, opened 2020-02-19
[xdl] Migrate `xdl/src/detach/*` to TypeScript

EvanBacon:
## Why

- Simplify all of XDL by finishing the TS migration.
- Make testing easier and changing eject command safer.
- https://github.com/expo/expo-cli/issues/1557

## How

- Convert the contents of `xdl/src/detach` to TypeScript.
- [x] delete unused gulp build, and watch tasks.
  - This will reduce the build time from `0m14.007s` to `0m9.331s` (rebuild times from `0m9.811s` to `0m4.367s`)
  - Reduces the package size from `4.6 MB` to `4.5 MB` (due to the lack of `build/__sourcemaps__/*`)
- [x] delete manual source-maps in favor of TypeScript source maps.

## Test Plan

- TS build is successful.
- Manually ran `EXPO_DEBUG=true expo eject --eject-method expokit --force`.
- Ran `expod bundle-assets` in `ios/` and `android/`.
- [ ] Not sure how to test `turtle`.
- [ ] Net sure how to test `expo/tools`.
- [ ] Unit/E2E tests - there don't seem to be any.


wkozyra95:
If you don't want to fully setup everything for local turtle to work, the simplest way would be to
- remove `declare module` from src/types/declaration.d.ts and make sure `yarn build`
- release xdl as alpha on npm, update it in turtle package.json and deploy turtle to staging (CI job under PR). After that just run a build from expo-cli on staging.

Setting up it locally is not much harder, it would require setting up .secrets.local file (you could use staging AWS SQS queues, so there is no need to setup it);

PR #1575, opened 2020-02-19
fix: prompt before overwriting credential files

cruzach:
Using the `expo fetch:android:*` commands, users can overwrite previously downloaded keystores, so we should prompt them to rename the new file, or confirm the overwrite

Defaults to overwrite in `non-interactive` mode

### DIFF
--- a/packages/expo-cli/src/commands/fetch/android.ts
+++ b/packages/expo-cli/src/commands/fetch/android.ts
@@ -8,7 +8,7 @@ import { Context } from '../../credentials';
 
 import log from '../../log';
 
-async function maybeRenameExistingFile(projectDir, filename) {
+async function maybeRenameExistingFile(projectDir: string, filename: string) {
   let desiredFilePath = path.resolve(projectDir, filename);
 
   if (await fs.pathExists(desiredFilePath)) {
--- a/packages/expo-cli/src/commands/fetch/android.ts
+++ b/packages/expo-cli/src/commands/fetch/android.ts
@@ -1,18 +1,42 @@
 import path from 'path';
 import fs from 'fs-extra';
 import get from 'lodash/get';
+import program from 'commander';
 
 import { AndroidCredentials } from '@expo/xdl';
 import { DownloadKeystore } from '../../credentials/views/AndroidCredentials';
 import { Context } from '../../credentials';
+import prompt from '../../prompt';
 
 import log from '../../log';
+
+async function promptIfFileExists(projectDir, filename) {
+  while (
+    (await fs.pathExists(await path.resolve(projectDir, filename))) &&
+    !program.nonInteractive
+  ) {
+    let question = {
+      type: 'input',
+      name: 'newName',
+      message: `You already have a file in your project directory named "${filename}"\n\n  Press enter to overwrite this file, or provide a different filename:`,
+      default: 'overwrite',
+    };
+    const answer = await prompt(question);
+    if (answer.newName === 'overwrite') {
+      break;
+    } else {
+      filename = answer.newName;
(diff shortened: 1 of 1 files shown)


cruzach:
I thought the same thing, and ended up going with this option just because it seemed better to let the user know that they now have 2 different keystores in their project's directory (wouldn't want someone surprised by that) and they can decide if they want to overwrite it, or if they don't they can name it whatever helps them remember it's new (e.g. `myapp-theDate.jks` or whatever)

Auto-appending a number and writing to a new file by default is definitely easier, but then there's the tiny chance we break someone's current workflow where they are somehow relying on this automatic overwrite in CI, maybe? 

In the end, I think both options are fine, so if you feel that appending `-1.jks` is better, I'm happy to change it!

esamelson:
Gotcha. That's a good point that my suggestion would actually be a breaking change. What about if we rename the old file, instead, to something like `<app>-old-1.jks`? I think that might be the best of both worlds, especially if we log a message to the console when that's happening. (They need to read the console anyway, to get the key password/alias.)

My main concern/motivation here is to prevent the accidental overwrite possibility at all reasonable costs. 😅  If you would rather prompt the user, I'm fine with that, but would prefer that the default option be *not* to overwrite.

skevy:
# [Codecov](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=h1) Report
> Merging [#1575](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=desc) into [master](https://codecov.io/gh/expo/expo-cli/commit/ee26c1f8b45ddafa09b2da7ee3eadef1db83018d&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##           master    #1575   +/-   ##
=======================================
  Coverage   60.14%   60.14%           
=======================================
  Files          19       19           
  Lines         833      833           
  Branches      219      217    -2     
=======================================
  Hits          501      501           
- Misses        327      331    +4     
+ Partials        5        1    -4     
```


| [Impacted Files](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [packages/config/src/Web.ts](https://codecov.io/gh/expo/expo-cli/pull/1575/diff?src=pr&el=tree#diff-cGFja2FnZXMvY29uZmlnL3NyYy9XZWIudHM=) | `17.05% <0.00%> (ø)` | :arrow_up: |
| [packages/android-manifest/src/Manifest.ts](https://codecov.io/gh/expo/expo-cli/pull/1575/diff?src=pr&el=tree#diff-cGFja2FnZXMvYW5kcm9pZC1tYW5pZmVzdC9zcmMvTWFuaWZlc3QudHM=) | `69.79% <0.00%> (ø)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=footer). Last update [ee26c1f...e2ec040](https://codecov.io/gh/expo/expo-cli/pull/1575?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

PR #1584, opened 2020-02-20
feat: make breaking app.json changes automatically

cruzach:
ref: https://github.com/expo/expo/pull/7049#issuecomment-588334127

This will make any breaking changes to app.json in the future much easier/safer


packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -100,6 +100,20 @@ async function getUpdatedDependenciesAsync(
   return result;
 }
 
+async function upgradeAppJson(projectRoot: string, targetSdkVersionString: string) {
+  let { exp } = await ConfigUtils.readConfigJsonAsync(projectRoot);
+  exp.sdkVersion = targetSdkVersionString;
+  switch (targetSdkVersionString) {
+    case '37.0.0':
+      if (exp.androidNavigationBar?.visible === false) {
```
brentvatne: we should log something to users to let them know what change we are making if we remove the `androidNavigationBar.visible` key or change it

cruzach:
followed suggestions [here](https://github.com/expo/expo-cli/pull/1587#discussion_r384087081) for the functionality- if key exists in custom config but not app.json, then warn them to change. Otherwise, update the value in app.json

Issue #1302, opened 2019-12-06
Update Next in dev-tools


brentvatne:
It currently uses Next 5, which is on Babel 6 and React 16.2. Sad times.

EvanBacon:
Just putting it out there but react-native-web is faster than emotion https://necolas.github.io/react-native-web/benchmarks/

brentvatne:
That also seems like a reasonable option but I'm hesitant to spend more time than necessary on upgrading this.

PR #1594, opened 2020-02-25
fix unit tests for xdl

EvanBacon:
add xdl to coverage


EvanBacon:
Dev server tests aren't working because it's using a much older version of babel. Related #1302

PR #1598, opened 2020-02-26
[plist] Add a Plist parser

quinlanj:
# why
- was using the `plist` parser in the credentials manager for `expo build:ios`. However, `plist` parser uses a DOMParser that spews a bunch of non fatal errors to stderr and there's no good way to turn it off (other people have had this same [problem](https://www.npmjs.com/package/xmldom-silent) 
- This is a bad user experience and surfaced in user testing

# how
- take `plist` parser and it's tests and use it in our repo
- convert to typescript, add our own error handler to DOMParser so we dont verbosely print to stderr
- ported over the `plist` tests, and ensure they all pass

# TODO
- [ ] publish package to npm


EvanBacon:
BTW test coverage for xdl is not yet enabled. Related: https://github.com/expo/expo-cli/pull/1594

packages/xdl/src/tools/Plist/build.ts
```diff
@@ -0,0 +1,150 @@
+/* (The MIT License)
```
EvanBacon: Might be better to create a new package since XDL is already pretty big

brentvatne: yeah let's move this out into a separate package, maybe call it `@expo/plist`

PR #1602, opened 2020-02-27
Upgrades pacote

arcanis:
Pacote has unlisted dependencies that I'll fix as part of https://github.com/npm/pacote/pull/32. This diff upgrades to `^11.1.0` so that Expo can benefit from the fixes as soon as they land.

### DIFF
--- a/packages/xdl/src/Exp.ts
+++ b/packages/xdl/src/Exp.ts
@@ -129,8 +129,8 @@ export async function extractTemplateAppAsync(
 ) {
   await pacote.tarball.stream(
     templateSpec,
-    () => {
-      return extractTemplateAppAsyncImpl(templateSpec, targetPath, config);
+    tarStream => {
+      return extractTemplateAppAsyncImpl(templateSpec, targetPath, config, tarStream);
     },
     {
       cache: path.join(UserSettings.dotExpoHomeDirectory(), 'template-cache'),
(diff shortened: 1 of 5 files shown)


packages/xdl/src/Exp.ts
```diff
@@ -126,9 +127,25 @@ export async function extractTemplateAppAsync(
   targetPath: string,
   config: AppJSONConfig | BareAppConfig
 ) {
-  let tarStream = await pacote.tarball.stream(templateSpec, {
-    cache: path.join(UserSettings.dotExpoHomeDirectory(), 'template-cache'),
-  });
+  await pacote.tarball.stream(
+    templateSpec,
+    () => {
+      return extractTemplateAppAsyncImpl(templateSpec, targetPath, config);
```
fson: The 4th argument (`tarStream`) seems to be missing here?

```
@expo/xdl: src/Exp.ts(133,14): error TS2554: Expected 4 arguments, but got 3.
```

arcanis: Woops indeed! Are there tests I can easily run to double-check that everything looks fine?

fson: The tests for `expo-cli` (they can be run with `yarn test` in `packages/expo-cli`) test the `expo init` command, and they all passed in CircleCI, so it looks fine now. 👍

PR #1615, opened 2020-02-29
[xdl] Automatically start an emulator when no android device is connected

EvanBacon:
# Why

New users mostly don't know how to open an Emulator without opening Android Studio. Opening Android Studio can immediately kill your vibe so we should minimize reasons to do that.
Also `expo start --android` should work the same as `expo start --ios`.

# Test Plan

- `expo start --android` should open an emulator is no device is connected, then install the expo client and open the project.

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -325,10 +325,9 @@ async function attemptToStartEmulatorOrAssertAsync() {
   if (!(await _isDeviceAttachedAsync())) {
     // If no devices or emulators are attached we should attempt to open one.
     if (!(await maybeStartAnyEmulatorAsync())) {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Android.ts
```diff
@@ -235,9 +321,23 @@ async function _openUrlAsync(url: string) {
   return output;
 }
 
+async function attemptToStartEmulatorOrAssertAsync() {
+  if (!(await _isDeviceAttachedAsync())) {
+    // If no devices or emulators are attached we should attempt to open one.
+    if (!(await maybeStartAnyEmulatorAsync())) {
+      // TODO(Bacon): Maybe add instructions for easily creating an emulator.
```
brentvatne: we can link to these instructions: https://docs.expo.io/versions/latest/workflow/android-studio-emulator/

PR #1614, opened 2020-02-29
Remove gulp from expo-cli

EvanBacon:
Use babel to compile files and TSC to watch them.

### Before

```
real	0m10.446s
user	0m14.956s
sys	0m1.219s
```

### After

```
real	0m4.897s
user	0m6.160s
sys	0m0.539s
```

## Reduce bundle size

This estimate removes the source maps. They're larger source maps now but in time we shouldn't need them.

### Before

```
=== Tarball Details === 
name:          expo-cli                                
version:       3.19.1                                  
filename:      expo-cli-3.19.1.tgz                     
package size:  125.3 kB                                
unpacked size: 611.2 kB                                
shasum:        8bf604ddc31b8553646ea828c26f41e2d03df720
integrity:     sha512-JiZNDnNHdLtVs[...]L7rZIBI7xLIbw==
total files:   117  
```

### After

```
=== Tarball Details === 
name:          expo-cli                                
version:       3.19.1                                  
filename:      expo-cli-3.19.1.tgz                     
package size:  124.4 kB                                
unpacked size: 603.1 kB                                
shasum:        517a2d27525cd73df0821fb37dcdcded9ce5fac3
integrity:     sha512-e2yM5N/zPsGoF[...]6wfdBPM9kce/A==
total files:   117 
```

# Test Plan

### Before

```
GitHub $ time expo whoami
Not logged in

real	0m0.323s
user	0m0.288s
sys	0m0.060s
```

### After

```
GitHub $ time expod whoami
Not logged in

real	0m0.321s
user	0m0.287s
sys	0m0.059s
```

PR #1625, opened 2020-03-02
Migrate all imports of package.json to use node resolution

EvanBacon:
# Why

TypeScript will attempt to bundle the `package.json` when it's imported relatively which breaks the bundle. This approach prevents that.

- Moved from #1614

### DIFF
--- a/packages/expo-optimize/src/index.ts
+++ b/packages/expo-optimize/src/index.ts
@@ -2,7 +2,7 @@
 import chalk from 'chalk';
 import { Command } from 'commander';
 import { resolve } from 'path';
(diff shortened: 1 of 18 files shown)


packages/expo-optimize/src/index.ts
```diff
@@ -2,16 +2,16 @@
 import chalk from 'chalk';
 import { Command } from 'commander';
 import { resolve } from 'path';
+// @ts-ignore: expo-cli is not listed in its own dependencies
```
brentvatne:
```suggestion
// @ts-ignore: expo-optimize is not listed in its own dependencies
```

PR #1626, opened 2020-03-03
Extract standalone app config to @expo/config and apply on eject

brentvatne:
# iOS Config

### General

- [x] version
- [x] bundleIdentifier
- [x] buildNumber
- [x] infoPlist custom entries
- [x] scheme
- [x] userInterfaceStyle - (expo.userInterfaceStyle and expo.ios.userInterfaceStyle)
- [x] supportsTablet / isTabletOnly
- [x] orientation
- [x] usesNonExemptEncryption
- ~[ ] owner *I don't think this applies to ejecting? seems just like an app.json concern*~ @tcdavis can you verify?

### Assets

~- [ ] locales~ defer this until later
~- [ ] icon (expo.icon and expo.ios.icon)~ defer this until later
- ~[ ] Splash - also iOS specific values~ defer this until later, will depend on expo-splash-screen
- ~[ ] backgroundColor - also iOS specific value~ defer this

## Branch

- [x] branch.apiKey

### Google

- [x] googleMapsApiKey
- [x] googleMobileAdsAppId
- [x] googleMobileAdsAutoInit
- [x] googleSignIn
- [x] googleServicesFile

### Entitlements

- [x] associatedDomains
- [x] usesAppleSignIn
- [x] usesIcloudStorage
- [x] accessesContactNotes

### Facebook

- [x] facebookAppId
- [x] facebookAutoInitEnabled
- [x] facebookAutoLogAppEventsEnabled
- [x] facebookAdvertiserIDCollectionEnabled
- [x] facebookDisplayName
- [x] facebookScheme

# Android config

### General

- [x] version
- [x] versionCode
- [x] scheme
- [x] Permissions
- [x] userInterfaceStyle - also iOS and android specific values. default to `light`
- [x] orientation
- [x] backgroundColor - also iOS and android specific values
- [x] primaryColor
- [x] androidStatusBar
- [x] androidNavigationBar 
- [x] googleServicesFile
- [x] “Config” object - branch, googleMapsApiKey, googleMobileAdsAppId, googleMobileAdsAutoInit
- [x] intentfilters

### Facebook

- [x] facebookappid
- [x] Facebookautoinitenabled
- [x] facebookAutoLogAppEventsEnabled
- [x] facebookAdvertiserIDCollectionEnabled
- [x] facebookDisplayName
- [x] facebookScheme

### Log warnings

- [x] adaptiveIcon
- [x] icon
- [x] Splash 

# End-to-end test

- [x] Apply on eject
~- [ ] Apply on standalone apps~ this will technically be possible but it will be behind a flag because we need that the properties can all be applied idempotently


packages/config/src/android/Facebook.ts
```diff
@@ -42,6 +49,31 @@ export function getFacebookAdvertiserIDCollection(config: ExpoConfig) {
...
+
+  let stringsJSON = await readStringsXMLAsync(stringsPath);
+  let stringItemToAdd: XMLItem[] = [{ _: appId, $: { name: 'facebook_app_id' } }];
+  stringsJSON = setStringItem(stringItemToAdd, stringsJSON);
+
+  try {
+    await writeStringsXMLAsync(stringsPath, stringsJSON);
+  } catch (e) {
+    throw new Error(
+      `Error setting Android navigation bar color. Cannot write strings.xml to ${stringsPath}.`
```
brentvatne:
```suggestion
      `Error setting facebookAppId. Cannot write strings.xml to ${stringsPath}.`
```

packages/config/src/android/GoogleMapsApiKey.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // add meta-data item
+  let existingMetaDataItem;
+  const metaDataItem = {
+    $: {
+      'android:name': 'com.google.android.geo.API_KEY',
+      'android:value': apiKey,
+    },
+  };
+  if (mainApplication.hasOwnProperty('meta-data')) {
```
EvanBacon:
```suggestion
  if ('meta-data' in mainApplication) {
```

packages/config/src/android/GoogleMobileAds.ts
```diff
@@ -0,0 +1,67 @@
+import { ExpoConfig } from '../Config.types';
+import { Document } from './Manifest';
+
+export function getGoogleMobileAdsAppId(config: ExpoConfig) {
```
EvanBacon:
```suggestion
export function getGoogleMobileAdsAppId(config: ExpoConfig): string | null {
```

packages/config/src/android/Facebook.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export function getFacebookAppId(config: ExpoConfig) {
+  return config.facebookAppId ?? null;
+}
+
+export function getFacebookDisplayName(config: ExpoConfig) {
+  return config.facebookDisplayName ?? null;
+}
+export function getFacebookAutoInitEnabled(config: ExpoConfig) {
+  return config.hasOwnProperty('facebookAutoInitEnabled') ? config.facebookAutoInitEnabled : null;
```
EvanBacon:
```suggestion
  return config.facebookAutoInitEnabled ?? null;
```

If I'm not mistaken, ?? will only convert undefined or null to null.

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
```diff
@@ -0,0 +1,90 @@
...
+// * Overall this function needs to be revamped, just a placeholder for now!
+function getIOSPaths(projectRoot: string) {
+  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+
+  let projectName = exp.name;
+  if (!projectName) {
+    throw new Error('Need a name ;O');
+  }
+
+  const iosProjectDirectory = path.join(projectRoot, 'ios', sanitizedName(projectName));
```
EvanBacon: Maybe we can just look up the xcodeproj with `glob`, something like this (from #1752):
```ts
const reactNativeIos = sync(path.join(projectRoot, 'ios', '*', 'Info.plist'));
path.join(reactNativeIos[0], '..');
```

Probably want to check if the `AppDelegate.m` exists in the same folder, otherwise glob sync may resolve the Test folder in iOS.

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
```diff
@@ -0,0 +1,90 @@
...
+async function modifyInfoPlistAsync(projectRoot: string, callback: (plist: any) => any) {
+  const { iosProjectDirectory } = getIOSPaths(projectRoot);
+  await IosPlist.modifyAsync(iosProjectDirectory, 'Info', callback);
+  await IosPlist.cleanBackupAsync(iosProjectDirectory, 'Info', false);
+}
+
+export default async function configureIOSProjectAsync(projectRoot: string) {
+  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+
+  IOSConfig.BundleIdenitifer.setBundleIdentifierForPbxproj(projectRoot, exp.ios!.bundleIdentifier!);
```
EvanBacon: Probably should validate and throw an error here about the `bundleIdentifier`

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    return infoPlist;
+  });
+
+  // Configure entitlements/capabilities
+  await modifyEntitlementsPlistAsync(projectRoot, entitlementsPlist => {
+    // TODO: We don't have a mechanism for getting the apple team id here yet
+    entitlementsPlist = IOSConfig.Entitlements.setICloudEntitlement(
+      exp,
+      'TODO-GET-APPLE-TEAM-ID',
```
EvanBacon: Probably want to use something like [appleApi/authenticate](https://github.com/expo/expo-cli/blob/c4b46396c3cb947fdcc5f17a3b587b71815f7ae3/packages/expo-cli/src/appleApi/authenticate.ts#L122)

packages/expo-cli/src/commands/utils/maybeBailOnGitStatusAsync.ts
```diff
@@ -0,0 +1,33 @@
+import program from 'commander';
+
+import { validateGitStatusAsync } from './ProjectUtils';
+import log from '../../log';
+import prompt from '../../prompt';
+
+export default async function maybeBailOnGitStatusAsync(): Promise<boolean> {
+  const isGitStatusClean = await validateGitStatusAsync();
```
EvanBacon: Interesting that we don't pass a `projectRoot` to this.

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -474,7 +480,7 @@ export async function upgradeAsync(
     log.addNewLineIfNone();
     log(
       chalk.bold(
-        `The following packages were ${chalk.underline(
+        `🚨 The following packages were ${chalk.underline(
```
EvanBacon: +1 Good use of emojis

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
brentvatne: yeah, going to defer this one until after release to follow up with

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -394,22 +378,34 @@ export async function upgradeAsync(
...
     yarn: options.yarn,
     log,
+    silent: true,
   });
 
   log.addNewLineIfNone();
-  log(chalk.underline.bold('Installing the expo package...'));
+  let installingPackageStep = logNewSection('Installing the expo package...');
   log.addNewLineIfNone();
-  await packageManager.addAsync(`expo@^${targetSdkVersionString}`);
```
EvanBacon: Maybe we should create a superset of PackageManager that implements `expo install`

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
brentvatne: validate that we have a bundle identifier or what?

PR #1653, opened 2020-03-09
fix(xdl): clean install with android clients

### LINKED ISSUE #1637: expo client:install:android uninstalling but not installing
```
╰─$ expo client:install:android
Could not find your Expo project. If you run this from a project, we can help pick the right Expo client version!
? Do you want to install the latest client? Yes
Uninstalling Expo from Android device.

Done!
╭─~
```

It doesn't actually download the client app or install it. The only way to install it is to run the Android app from dev tools or expo-cli interactive prompt.

cc @byCedric

byCedric:
Fixes #1637

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -274,7 +274,7 @@ export async function uninstallExpoAsync() {
     return await getAdbOutputAsync(['uninstall', 'host.exp.exponent']);
   } catch (e) {
     Logger.global.error(
-      'Could not uninstall Expo from your device, please uninstall Expo manually and try again.'
+      'Could not uninstall Expo client from your device, please uninstall Expo client manually and try again.'
     );
     throw e;
   }
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -125,7 +125,7 @@ export async function getAdbOutputAsync(args: string[]): Promise<string> {
     let result = await spawnAsync(adb, args);
     return result.stdout;
   } catch (e) {
-    let errorMessage = _.trim(e.stderr);
+    let errorMessage = _.trim(e.stderr || e.stdout);
     if (errorMessage.startsWith(BEGINNING_OF_ADB_ERROR_MESSAGE)) {
       errorMessage = errorMessage.substring(BEGINNING_OF_ADB_ERROR_MESSAGE.length);
     }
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Android.ts
```diff
@@ -263,7 +263,21 @@ export async function installExpoAsync(url?: string) {
...
+    return '';
+  }
+
+  try {
+    return await getAdbOutputAsync(['uninstall', 'host.exp.exponent']);
+  } catch (e) {
+    Logger.global.error(
+      'Could not uninstall Expo from your device, please uninstall Expo manually and try again.'
+    );
+    throw e;
```
byCedric: I put this into the `uninstallExpoAsync` method since this might be used somewhere else. It shows some a generic way that might fix the issue, but that might require some more info. By throwing the error, the execution stops and the error is dumped to the logs.

packages/xdl/src/Android.ts
```diff
@@ -263,7 +263,21 @@ export async function installExpoAsync(url?: string) {
...
+  const isInstalled = await _isExpoInstalledAsync();
+  if (!isInstalled) {
+    return '';
+  }
+
+  try {
+    return await getAdbOutputAsync(['uninstall', 'host.exp.exponent']);
+  } catch (e) {
+    Logger.global.error(
+      'Could not uninstall Expo from your device, please uninstall Expo manually and try again.'
```
brentvatne:
```suggestion
      'Could not uninstall Expo client from your device, please uninstall Expo client manually and try again.'
```

packages/xdl/src/Android.ts
```diff
@@ -263,7 +263,21 @@ export async function installExpoAsync(url?: string) {
 
 export async function uninstallExpoAsync() {
   Logger.global.info('Uninstalling Expo from Android device.');
-  return await getAdbOutputAsync(['uninstall', 'host.exp.exponent']);
+
+  // we need to check if its installed, else we might bump into "Failure [DELETE_FAILED_INTERNAL_ERROR]"
+  const isInstalled = await _isExpoInstalledAsync();
+  if (!isInstalled) {
+    return '';
```
brentvatne:
```suggestion
    return;
```

packages/xdl/src/Android.ts
```diff
@@ -263,7 +263,21 @@ export async function installExpoAsync(url?: string) {
 
 export async function uninstallExpoAsync() {
```
brentvatne:
```suggestion
export async function uninstallExpoAsync(): string | undefined {
```

PR #1655, opened 2020-03-10
remove link for netinfo

chrisdrackett:
no longer exported as of RNW 0.12.0.

https://github.com/react-native-community/react-native-netinfo/tree/v5.5.0

### DIFF
--- a/packages/webpack-config/src/env/alias.ts
+++ b/packages/webpack-config/src/env/alias.ts
@@ -7,7 +7,6 @@
 export const aliases = {
   // Alias direct react-native imports to react-native-web
   'react-native$': 'react-native-web',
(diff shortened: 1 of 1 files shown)


brentvatne:
we don't use this version of react-native-web yet but assigned @EvanBacon for when we do

chrisdrackett:
@brentvatne might want to update the docs then, at the moment they just say to install `react-native-web` without a version which will cause errors on build:

https://docs.expo.io/versions/v36.0.0/guides/running-in-the-browser/#adding-web-support-to-expo-projects

EvanBacon:
Thanks for tracking this @chrisdrackett

brentvatne:
handled by https://github.com/expo/expo-cli/pull/1721

PR #1656, opened 2020-03-10
[cli] better worded prompt

quinlanj:
make cli prompt more clear - original question can be kind of confusing because it implies that an app with the bundle id already exists on the store.

### DIFF
--- a/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
+++ b/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
@@ -92,7 +92,7 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
       {
         type: 'confirm',
         name: 'confirm',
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -92,7 +92,7 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
       {
         type: 'confirm',
         name: 'confirm',
-        message: `Do you have access to the Apple Account that owns the app with the bundle identifier ${bundleIdentifier}?`,
+        message: `Do you have access to the Apple Account that will be used for submitting this app to the App Store?`,
```
brentvatne:
```suggestion
        message: `Do you have access to the Apple account that will be used for submitting this app to the App Store?`,
```

brentvatne:
🙏

PR #1661, opened 2020-03-10
convert publishInfo endpoint

jkhales:
Missed a publishInfo -> publish/history endpoint.

Since this endpoint has already been converted in other places, it's relatively low risk and can go out in the same publish as another conversion.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -403,7 +403,6 @@ export async function getLatestReleaseAsync(
     result = await api.postAsync('publish/history', {
       owner: options.owner,
       slug: await getSlugAsync(projectRoot),
-      version: '2',
       releaseChannel: options.releaseChannel,
       count: '1',
       platform: options.platform,
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -403,9 +403,9 @@ export async function getLatestReleaseAsync(
     result = await api.postAsync('publish/history', {
       owner: options.owner,
       slug: await getSlugAsync(projectRoot),
-      version: '2',
+      version: 2,
       releaseChannel: options.releaseChannel,
-      count: '1',
+      count: 1,
       platform: options.platform,
     });
   }
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -403,7 +403,6 @@ export async function getLatestReleaseAsync(
     result = await api.postAsync('publish/history', {
       owner: options.owner,
       slug: await getSlugAsync(projectRoot),
-      version: 2,
       releaseChannel: options.releaseChannel,
       count: 1,
       platform: options.platform,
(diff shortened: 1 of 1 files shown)


packages/xdl/src/Project.ts
```diff
@@ -381,20 +381,35 @@ export async function getLatestReleaseAsync(
...
+    });
+  } else {
+    const user = await UserManager.ensureLoggedInAsync();
+    const api = ApiV2.clientForUser(user);
+    result = await api.postAsync('publish/history', {
+      owner: options.owner,
+      slug: await getSlugAsync(projectRoot),
+      version: '2',
+      releaseChannel: options.releaseChannel,
+      count: '1',
```
fson: Is there a reason why we use the string `'1'` instead of the number `1` here?

packages/xdl/src/Project.ts
```diff
@@ -381,20 +381,35 @@ export async function getLatestReleaseAsync(
...
+    result = await Api.callMethodAsync('publishInfo', [], 'post', null, {
+      formData,
+    });
+  } else {
+    const user = await UserManager.ensureLoggedInAsync();
+    const api = ApiV2.clientForUser(user);
+    result = await api.postAsync('publish/history', {
+      owner: options.owner,
+      slug: await getSlugAsync(projectRoot),
+      version: '2',
```
fson: I believe the `version` field should be removed from APIV2 code of this endpoint. It was used to denote two backwards incompatible versions of the APIV1 endpoint. For APIV2, we've never released any client code that uses the older variant, so this field is redundant.

We should also remove it from the backend here (since there are no clients calling the APIV2 history endpoint with version `1`:
https://github.com/expo/universe/blob/85f7202d0f1b295ed87d6a32b96186266032e3e9/server/www/src/apiv2/PublishRouter.ts#L499-L504

packages/xdl/src/Project.ts
jkhales: I just copied line 392. Do you prefer 1?

packages/xdl/src/Project.ts
jkhales: Nice. Will do.

jkhales: Actually, regarding the backend code, there maybe some old clients that still call it. So as long as we have the deprecated apiv1 endpoints up, we should keep the code.

jkhales: And if the backend code stays the same, then we need to keep the version code here as well, otherwise it defaults to version===1 and deletes fields.

packages/xdl/src/Project.ts
fson: Line 392
https://github.com/expo/expo-cli/blob/d97f20b9d080434a5f0247b61bc64982b74f2cdf/packages/xdl/src/Project.ts#L392
used the string because the V1 publish endpoint took thzput as `multipart/form-data`, which, unlike JSON, only supports strings and not numbers.

I think `1` is better because it's more self-describing for numeric data.

packages/xdl/src/Project.ts
fson: The `version === 2` parameter was added to Expo CLI more than 2 years ago in https://github.com/expo/expo-cli/commit/3c4b73b98891b90c3a25e67bd16502fc93a387d1.

It's safe to say that any Expo CLI version older than that has been deprecated and is no longer supported. We should remove support for `version === 1`.

jkhales: OK, great.

PR #1669, opened 2020-03-11
[cli] better provisioning profile validation

quinlanj:
# Why
Have better validation for provisioning profiles, especially if:
- the profile was uploaded by a user 
- there is no access to apple

# Tests
- [x] Upload provisioning profile with Dist Cert A, then run `expo build:ios` with Dist Cert B. Expect to detect that uploaded profile is not compatible and recover.

### DIFF
--- a/packages/expo-cli/src/credentials/views/IosProvisioningProfile.ts
+++ b/packages/expo-cli/src/credentials/views/IosProvisioningProfile.ts
@@ -3,6 +3,8 @@ import fs from 'fs-extra';
 import find from 'lodash/find';
 import ora from 'ora';
 import every from 'lodash/every';
+import plist, { PlistObject } from '@expo/plist';
+import { IosCodeSigning, PKCS12Utils } from '@expo/xdl';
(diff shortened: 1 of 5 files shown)


packages/expo-cli/src/credentials/views/IosProvisioningProfile.ts
```diff
@@ -342,8 +346,28 @@ export async function validateProfileWithoutApple(
     return false;
   }
 
-  if (ProvisioningProfileManager.isExpired(base64EncodedProfile)) {
-    spinner.fail('Provisioning profile on file is expired');
+  const buffer = Buffer.from(base64EncodedProfile, 'base64');
+  const profile = buffer.toString('utf-8');
+  const profilePlist = plist.parse(profile) as PlistObject;
+
+  const distCertFingerprint = await PKCS12Utils.getP12CertFingerprint(
```
wkozyra95: I think those two functions can throw an exception, probably it's best to extend `try` blockto to the entire function

Issue #1434, opened 2020-01-14
Use react-native-cli Metro HTTP server instance directly


brentvatne:
This will require changes to react-native-cli. We should be able to inject our middleware to react-native-cli's express server. This has some benefits:

- don't need to start an other express server on another port
- don't need to start up a tunnel on both servers, just one
- don't need to serialize config and pass it in awkwardly in command line flags
- many more

idea via @EvanBacon

EvanBacon:
Fixed in SDK 40 or now with `EXPO_USE_DEV_SERVER=true`.

PR #1678, opened 2020-03-12
fix(xdl): ping metro bundler when fetching current status

### LINKED ISSUE #1677: Validate that server is actually running in Project.currentStatus and handle cases when it is not
## Description

The [Project.currentStatus(projectDir)](https://github.com/expo/expo-cli/blob/a26baf39c19cefd016eb0acc0445c4a3cba16915/packages/xdl/src/Project.ts#L176-L185) function determines the project status by looking at the packager info in the project `.expo` directory.

Sometimes expo-cli can close in an unexpected way (not exactly sure which steps will repro this, I suspect maybe a hard reboot without closing expo-cli would do it) and the information in `.expo` will be stale. We should validate that the status in this file is actually correct, eg: by sending a request to packager/expo server.  

## Expected Behavior

expo-cli should gracefully handle the information in `.expo` being stale/incorrect.

## Observed Behavior

Cryptic errors like this can happen when running `expo publish`:

```
Publishing to channel 'default'...
Building iOS bundle
connect ECONNREFUSED 127.0.0.1:19001
Error: connect ECONNREFUSED 127.0.0.1:19001
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1129:14)
```

byCedric:
Fixes #1677

This calls the Metro bundler's `/status` endpoint to determine if the project is running. It's performing a request, so that might slow the method down (with a current max time out of 2 seconds). But it provides a more accurate response about the current status. The code is [copied from `startReactNativeServerAsync`](https://github.com/expo/expo-cli/blob/a26baf39c19cefd016eb0acc0445c4a3cba16915/packages/xdl/src/Project.ts#L1925)

I did not check for the expo server, it [closes itself when you start it](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/Project.ts#L1985). So that means it restarts every time you run `expo start`, right? And I think both `expo publish` and `expo build` doesn't require the Expo server to run.

Thanks for thinking along @quinlanj 😄

### DIFF
--- a/packages/expo-cli/e2e/__tests__/start-test.ts
+++ b/packages/expo-cli/e2e/__tests__/start-test.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 
 import spawnAsync from '@expo/spawn-async';
+import { ProjectSettings } from '@expo/xdl';
 import temporary from 'tempy';
 
 import { EXPO_CLI, runAsync } from '../TestUtils';
@@ -26,4 +27,35 @@ test('start --offline', async () => {
     }
   });
   cli.stderr.pipe(process.stderr);
+});
+
+test('start --offline with existing packager info', async () => {
+  await ProjectSettings.setPackagerInfoAsync(tempDir, {
+    devToolsPort: 19002,
+    expoServerPort: 19001,
+    packagerPort: 19000,
+    packagerPid: 1337,
+  });
+
+  const promise = spawnAsync(EXPO_CLI, ['start', '--offline'], { cwd: projectDir });
+  const cli = promise.child;
+
+  cli.stderr.pipe(process.stderr);
+  cli.stdout.on('data', data => {
+    process.stdout.write(data);
+    if (/Your native app is running/.test(data.toString())) {
+      cli.kill('SIGINT');
+    }
+  });
+
+  try {
+    await promise;
+  } finally {
+    await ProjectSettings.setPackagerInfoAsync(tempDir, {
+      devToolsPort: null,
+      expoServerPort: null,
+      packagerPort: null,
+      packagerPid: null,
+    });
+  }
 });
--- a/packages/expo-cli/e2e/__tests__/start-test.ts
+++ b/packages/expo-cli/e2e/__tests__/start-test.ts
@@ -17,16 +17,19 @@ beforeAll(async () => {
 });
 
 test('start --offline', async () => {
-  const promise = spawnAsync(EXPO_CLI, ['start', '--offline'], {
-    cwd: projectDir,
-  });
+  const promise = spawnAsync(EXPO_CLI, ['start', '--offline'], { cwd: projectDir });
   const cli = promise.child;
-  cli.stdout.pipe(process.stdout).on('data', data => {
-    if (/Your native app is running/.test(data.toString())) {
-      cli.kill('SIGINT');
-    }
-  });
+
   cli.stderr.pipe(process.stderr);
+  cli.stdout
+    .on('data', data => {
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Project.ts
```diff
@@ -1560,7 +1578,7 @@ export async function buildAsync(
 ): Promise<BuildStatusResult | BuildCreatedResult> {
   /**
     This function corresponds to an apiv1 call and is deprecated.
-    Use 
+    Use
```
byCedric: Oh my bad, I think this is my vscode removing trailing whitespaces. I can revert this if you want!

brentvatne:
this seems reasonable! one thing i want to be sure of is that whenever we check the status, if it is bad then we reset it to a clean state.

byCedric:
Sounds good, I wasn't sure if the `currentStatus` was supposed to do this. I'll make it happen asap!

packages/xdl/src/Project.ts
brentvatne: nbd

brentvatne:
@byCedric - yeah maybe not a good fit for `currentStatus` exactly, maybe make another function that calls `currentStatus` and also takes care of reseting and use that wherever it makes sense

fson:
This pinging of Metro won't be necessary once https://github.com/expo/expo-cli/issues/1434 lands, as it will put Metro into the same process/server as the manifest handler.

The old way to run Metro as a subprocess will still be around for a while to support older SDK versions, so I guess it's still worth it to land this.

brentvatne:
I think it would be very valuable to try to set up an e2e test for this so that when #1434 lands we have an easy way to verify that we don't regress on this. @byCedric - want to give it a try?

packages/expo-cli/e2e/__tests__/start-test.ts
```diff
@@ -27,3 +28,34 @@ test('start --offline', async () => {
...
+  cli.stderr.pipe(process.stderr);
+  cli.stdout.on('data', data => {
+    process.stdout.write(data);
+    if (/Your native app is running/.test(data.toString())) {
+      cli.kill('SIGINT');
+    }
+  });
+
+  try {
+    await promise;
```
byCedric: I had to wait until the command exits with `SIGINT`, else the test would just validate even though it never closes. Not sure if that's just my computer, or if this is also missing from the basic `start --offline` test.

packages/expo-cli/e2e/__tests__/start-test.ts
```diff
@@ -27,3 +28,34 @@ test('start --offline', async () => {
...
+    packagerPort: 19000,
+    packagerPid: 1337,
+  });
+
+  const promise = spawnAsync(EXPO_CLI, ['start', '--offline'], { cwd: projectDir });
+  const cli = promise.child;
+
+  cli.stderr.pipe(process.stderr);
+  cli.stdout.on('data', data => {
+    process.stdout.write(data);
```
byCedric: On my computer, the `on('data')` event is not triggered if chained with `.pipe(process.stdout)`. This achieves the same result without doing additional stuff.

byCedric: Ville helped me out, I'll change the other test as well using `.pipe` after `.on` 😄

packages/expo-cli/e2e/__tests__/start-test.ts
byCedric: Talked this over with Ville, we should `await` else Jest will skip the test and not wait for the results.

Issue #1647, opened 2020-03-07
Expo-Cli 3.13.2 won't let me build ios app and throws 404 error


edumorlom:
## Description
`expo build:ios` throws an error **Request failed with status code 404** when I use the following PNG in my build directory.

![swipeup-gif](https://user-images.githubusercontent.com/11444894/76150004-c15fad80-6073-11ea-922a-5f40192e9d52.png)

## Expected Behavior
Should build without a problem, it shouldn't ask me to login to my Apple Developer account every time I build.

What action did you perform, and what did you expect to happen?
`expo build:ios` should build without a problem but instead throws a 404 error.

## Observed Behavior
Weirdly, when I run the build command. It makes me sign into my Apple Developer account. Always.... It never used to do this, it would always remember my account. After downgrading to 3.13.1 the issue no longer happens and I am able to build the exact same directory without any issues. It seems to be with 3.13.2 Also, if I removed that specific PNG, it builds just fine.

What actually happened when you performed the above actions?
```
Publishing to channel 'default'...

? Do you have access to the Apple Account that owns the app with the bundle identifier edu.fiu.moms-infants-healthy?

..............

Building iOS bundle
Building Android bundle
Analyzing assets
Uploading assets
Uploading /swipeup-gif.png
Request failed with status code 404

```

edumorlom:
Uploading any other image works, it seems to be related to that image in specific... That's the weird part. Also, why does it make me log in to the Apple Developer account anytime I build with this new version? Downgrading to 3.13.1 works just fine.

fnfilho:
I am also having issues with 3.13.2 and it is throwing a 404 error. Was able to downgrade to 3.13.1 and have the build work fine again.

```
Nery:app fernando$ expo ba
Checking if there is a build in progress...


? Would you like to upload a keystore or have us generate one for you?
If you don't know what this means, let us handle it! :)
 false
Unable to find an existing Expo CLI instance for this directory, starting a new one...
Starting Metro Bundler on port 19001.
Successfully ran `adb reverse`. Localhost URLs should work on the connected Android device.
Tunnel ready.
Publishing to channel 'default'...
Building iOS bundle
Finished building JavaScript bundle in 2553ms.
Building Android bundle
Analyzing assets
Building JavaScript bundle [====================================================================================================] 100%Finished building JavaScript bundle in 1793ms.
Finished building JavaScript bundle in 1080ms.
Finished building JavaScript bundle in 1127ms.
Uploading assets
[...]

PR #1680, opened 2020-03-12
Increase max upload size of apiv2 to same as in apiv1

jkhales:
# why
https://github.com/expo/expo-cli/issues/1647#issuecomment-598273532

# how
The default upload limit is 10, it was increased to 100 in api v1. This increases it in api v2 as well.

# test

Set the limit to 1 byte to repro the error.


NikolaNikushev:
Awesome!

PR #1688, opened 2020-03-13
[config] Added watcher for app.config.js

EvanBacon:
# Why 

It's confusing that `app.json` configs can be reloaded while the CLI is running, but `app.config.js` configs cannot. This logs out a warning for the user, notifying them to restart the CLI if they want to see changes.


EvanBacon:
Closing in favor of #1689

PR #1696, opened 2020-03-17
[config] Added TypeScript support to app.config

EvanBacon:
# Why 

TypeScript is dope, having first-class support for `app.config.ts` will be great.

# How

Use `@expo/babel-preset-cli` (also adds nullish coalescing and optional chaining support).

# Test Plan

Unskip the TypeScript format test.


brentvatne:
❤️ it

PR #1698, opened 2020-03-17
Extract Android standalone app config to @expo/config and apply on eject

brentvatne:
- [x] version
- [x] versionCode
- [x] scheme
- [x] Permissions
- [x] userInterfaceStyle - also iOS and android specific values. default to `light`
- [x] orientation
- [x] backgroundColor - also iOS and android specific values
- [x] primaryColor
- [x] androidStatusBar
- [x] androidNavigationBar 
- [x] googleServicesFile
- [x] “Config” object - branch, googleMapsApiKey, googleMobileAdsAppId, googleMobileAdsAutoInit
- [x] intentfilters

Facebook:
- [x] facebookappid
- [x] Facebookautoinitenabled
- [x] facebookAutoLogAppEventsEnabled
- [x] facebookAdvertiserIDCollectionEnabled
- [x] facebookDisplayName
- [x] facebookScheme

Log Warnings for:
- [x] adaptiveIcon
- [x] icon
- [x] Splash

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "start": "yarn run bootstrap && lerna run watch --parallel --ignore @expo/dev-tools --ignore expo-optimize --ignore @expo/next-adapter --ignore @expo/electron-adapter",
     "tsc": "echo 'You are trying to run \"tsc\" in the workspace root. Run it from an individual package instead.' && exit 1",
     "lint": "eslint . --ext js,ts",
-    "test-coverage": "jest --coverage && codecov"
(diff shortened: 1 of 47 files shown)


cruzach:
I'm not so familiar with jest's inner workings, but it looks like having a pool of processes running the tests causes problems with creating/deleting/finding the `tmp` directory these tests write to. I've added the flag to run the tests serially, one after the other, and this resolved the issue. Not sure if we **don't** want to do that for a specific reason, or if there's a better way to test writing/reading files with jest concurrently

PR #1707, opened 2020-03-20
[cli] fix publish:rollback

quinlanj:
# why
fixes https://github.com/expo/expo-cli/issues/596

# how
- change implementation to do a `publish:set` of the second most recent thing in history for a given sdkVersion, platform and channel instead of actually deleting anything in the history
- add confirmation with the publication that clients will receive as a result of the rollback
- move shared methods into `PublishUtils`

# tests
- [ ] repro on a client
- [x] publish rollback unit tests pass

PR #1724, opened 2020-03-23
[cli] tests for release channels

quinlanj:
tests 4 https://github.com/expo/expo-cli/pull/1707

### DIFF
--- a/packages/expo-cli/src/commands/__tests__/publish-info-test.ts
+++ b/packages/expo-cli/src/commands/__tests__/publish-info-test.ts
@@ -1,0 +1,128 @@
+import { vol } from 'memfs';
+import { ApiV2 } from '@expo/xdl';
+
+import { getPublicationDetailAsync, getPublishHistoryAsync } from '../utils/PublishUtils';
+
+jest.mock('fs');
+jest.mock('resolve-from');
+jest.mock('ora', () =>
+  jest.fn(() => {
+    return {
+      start: jest.fn(() => {
+        return { stop: jest.fn(), succeed: jest.fn() };
+      }),
+    };
+  })
+);
+jest.mock('@expo/xdl', () => {
+  const user = {
+    kind: 'user',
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/__tests__/publish-info-test.ts
```diff
@@ -0,0 +1,119 @@
...
+    kind: 'user',
+    username: 'test-username',
+    nickname: 'test-nickname',
+    userId: 'test-id',
+    picture: 'test-pic',
+    currentConnection: 'Username-Password-Authentication',
+    sessionSecret: 'test-session-secret',
+  };
+  const pkg = jest.requireActual('@expo/xdl');
+  pkg.UserManager.getCurrentUserAsync = jest.fn(() => user);
```
brentvatne: this will mutate the actual @expo/xdl module can could lead to some confusing errors - when tests are run in parallel there will be failures in random places depending on the load order of the tests relative to this test.

a safer way to do this is:

```js
return {
  ...pkg,
  UserManager: {
    ...pkg.UserManager,
   getCurrentUserAsync: jest.fn(() => user),
  },
  // etc..
}
```

packages/expo-cli/src/commands/__tests__/publish-modify-test.ts
```diff
@@ -0,0 +1,276 @@
...
+    kind: 'user',
+    username: 'test-username',
+    nickname: 'test-nickname',
+    userId: 'test-id',
+    picture: 'test-pic',
+    currentConnection: 'Username-Password-Authentication',
+    sessionSecret: 'test-session-secret',
+  };
+  const pkg = jest.requireActual('@expo/xdl');
+  pkg.UserManager.getCurrentUserAsync = jest.fn(() => user);
```
brentvatne: same comment as in other file re: mutating @expo/xdl

Issue #1256, opened 2019-11-22
expo cli crash while rebuilding web package


pihomeserver:
## Description

Using the expo-cli to work a project, tested as a web package
The build works the first time, the live refresh crashes the client and it should be restarted manually

## Expected Behavior

Previous version had live refresh full operational

## Observed Behavior

Here is the stack trace of the first run after the build (tunnel is up and web page opens successfully) : 

```
To run the app with live reloading, choose one of:
  • Sign in as @pihomeserver in Expo Client on Android or iOS. Your projects will automatically appear in the "Projects" tab.
  • Scan the QR code above with the Expo app (Android) or the Camera app (iOS).
  • Press a for Android emulator, or i for iOS simulator.
  • Press e to send a link to your phone with email.
┌──────────────────────────────────────────────────────────────────────────┐
│                                                                          │
│   Web support in Expo is experimental and subject to breaking changes.   │
│   Do not use this in production yet.                                     │
│                                                                          │
└──────────────────────────────────────────────────────────────────────────┘

 React  You can now view AuthTest in the browser.

  Local:            http://localhost:19006/
  On Your Network:  http://192.168.1.10:19006/

Note that the development build is not optimized.

 › To create a production build, run expo build:web
 › Press Ctrl+C to exit.

 Expo  Press ? to show a list of all available commands.
```

Then if i change a source file : 
```
Compiling...

  build [========            ] 40%TypeError: compilation.getPathWithInfo is not a function
    at /Users/chris/.nvm/versions/node/v13.0.1/lib/node_modules/expo-cli/node_modules/@expo/webpack-config/node_modules/webpack/lib/HotModuleReplacementPlugin.js:287:26
    at SyncHook.eval (eval at create (/Users/chris/.nvm/versions/node/v13.0.1/lib/node_modules/expo-cli/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:14:1)
    at SyncHook.lazyCompileHook (/Users/chris/.nvm/versions/node/v13.0.1/lib/node_modules/expo-cli/node_modules/tapable/lib/Hook.js:154:20)
    at /Users/chris/.nvm/versions/node/v13.0.1/lib/node_modules/expo-cli/node_modules/webpack/lib/Compilation.js:1361:37
[...]

PR #1726, opened 2020-03-24
chore: update webpack

jaulz:
The current version of `webpack` and `webpack-dev-server` are already a year old and we should update them.


EvanBacon:
Thanks for this, I'll need to take a closer look at this when I have some time. We've been waiting on this due to the breadth of testing involved with making sure it's a stable change.

rhino88:
@jaulz this seemed to reintroduce the following issue for me: https://github.com/expo/expo-cli/issues/1256

PR #1734, opened 2020-03-25
[expo-cli][xdl][config] add expo publish --target managed|bare

esamelson:
# Why

https://www.notion.so/expo/App-entry-points-94433d55f7524837be7b37bccb2154a1

TL;DR In order to make `expo publish` able to create bundles for both the managed Expo client environment and bare native client environments, we need to be able to choose at runtime whether or not to resolve `.expo.*` files in the project source.

This PR adds a `--target` flag to `expo publish`; for `--target managed`, the packager resolves `.expo.*` files, and for `--target bare` the packager ignores them.

This PR also changes the default behavior for bare projects. Previously, `expo publish` **always** resolved `.expo.*` files, but now it only does so by default for managed workflow projects.

This is a **breaking change** for people who rely on `expo publish` resolving `.expo.*` files in bare workflow projects -- these developers will need to add the explicit `--target managed` flag to the command after this commit.

# How

- Added a new `getBareExtensions` to @expo/config to match `getManagedExtensions` and thread `target` param through to `startReactNativeServerAsync`
- Save the `target` in the packager info on disk; if there is an existing server instance running we only use it if the `target` matches, and close and restart otherwise
- Added `Project.isBareWorkflowProject` util for distinguishing between managed & bare projects

# Test Plan

Tested on a new bare workflow project from https://github.com/expo/expo/pull/7466
- [x] `--target` field overrides default
- [x] publishing with `--target managed` produces a build that errors with `The Expo SDK requires Expo to run...`
- [x] publishing with `--target bare` or unspecified target produces a build that runs
- [x] if `expo start` is running in a different process, publishing with `--target managed` uses the same server instance
- [x] if `expo start` is running in a different process, publishing with `--target bare` closes that instance and starts a new one

### DIFF
--- a/packages/expo-cli/src/commands/publish.ts
+++ b/packages/expo-cli/src/commands/publish.ts
@@ -43,7 +43,7 @@ export async function action(projectDir: string, options: Options = {}) {
   }
 
   const target =
-    options.target ?? ((await Project.isBareWorkflowProject(projectDir)) ? 'bare' : 'managed');
+    options.target ?? ((await Project.isBareWorkflowProjectAsync(projectDir)) ? 'bare' : 'managed');
 
   const status = await Project.currentStatus(projectDir);
   let shouldStartOurOwn = false;
--- a/packages/expo-cli/src/commands/publish.ts
+++ b/packages/expo-cli/src/commands/publish.ts
@@ -1,6 +1,6 @@
 import { getConfig } from '@expo/config';
 import simpleSpinner from '@expo/simple-spinner';
-import { Exp, Project } from '@expo/xdl';
+import { Exp, Project, ProjectSettings } from '@expo/xdl';
 import chalk from 'chalk';
 import { Command } from 'commander';
 import fs from 'fs';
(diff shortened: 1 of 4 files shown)


packages/xdl/src/Project.ts
```diff
@@ -184,6 +186,30 @@ export async function currentStatus(projectDir: string): Promise<ProjectStatus>
   }
 }
 
+export async function isBareWorkflowProject(projectDir: string): Promise<boolean> {
```
brentvatne:
```suggestion
export async function isBareWorkflowProjectAsync(projectDir: string): Promise<boolean> {
```

packages/expo-cli/src/commands/publish.ts
```diff
@@ -40,16 +41,37 @@ export async function action(projectDir: string, options: Options = {}) {
       )}.`
     );
   }
+
+  const target =
+    options.target ?? ((await Project.isBareWorkflowProject(projectDir)) ? 'bare' : 'managed');
```
brentvatne:
```suggestion
    options.target ?? ((await Project.isBareWorkflowProjectAsync(projectDir)) ? 'bare' : 'managed');
```

PR #1740, opened 2020-03-27
[xdl] improve compatibility of export with expo-updates

esamelson:
# Why

Tested out expo-updates with `expo export` (I'd been testing mostly with `expo publish` before) and noticed a few things that would make the experience better.

# How

- added `releaseId` to exported manifests (we add this on the server for published manifests)
- added writing to configuration files & embedding manifest/bundle into native projects, the same way `expo publish` does (originally used for ExpoKit & repurposed for expo-updates). To do this I extracted all of this functionality into `_configureEmbeddedFilesAsync`.

# Test plan

- [x] Exported manifest loads & runs in expo-updates app (and updates work)
- [x] Embedded manifest + bundle written correctly by export
- [x] Embedded manifest + bundle still written correctly by publish
- [x] Configuration files (Expo.plist, AndroidManifest.xml) written correctly by export
- [x] Configuration files still written correctly by publish

PR #1745, opened 2020-03-29
[xdl] add default paths for saving embedded assets to disk

esamelson:
(depends on #1740)

This PR adds default paths for saving embedded assets to disk in projects with `expo-updates` installed. This removes the need for developers to configure the `ios.publishManifestPath`, `ios.publishBundlePath`, `android.publishManifestPath`, and `android.publishBundlePath` app.json fields upon installation of expo-updates; they can still configure these fields if they want a different location than the default.

The logic here assumes that all ExpoKit apps **will** still have these fields specified (ok since we write them to app.json automatically upon eject), but if `expo-updates` is installed, it will write the files whether or not these fields are specified.

Also, if the iOS files are being created anew, the command will log a reminder to add the files to the Xcode project. This only needs to be done once.

# Test Plan

On `expo publish` or `expo export`...

- [x] project with expo-updates installed writes to default location if nothing is specified in app.json
- [x] project with expo-updates installed writes to custom location if specified in app.json
- [x] project without expo-updates installed writes to location if specified in app.json
- [x] project without expo-updates installed does not write files if not specified in app.json
- [x] if writing the iOS files and they do not already exist, warning message about adding them to Xcode project is logged

# Todos (not blocking)

- [x] expo/fyi PR adding the referenced doc (https://github.com/expo/fyi/pull/3)
- [x] remove README instructions to add app.json fields
- [x] remove from template bare projects (https://github.com/expo/expo/pull/7533)

### DIFF
--- a/packages/xdl/src/EmbeddedAssets.ts
+++ b/packages/xdl/src/EmbeddedAssets.ts
@@ -35,6 +35,55 @@ export async function configureAsync(config: EmbeddedAssetsConfiguration) {
   await _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config);
 }
 
+export function getEmbeddedManifestPath(
+  platform: 'ios' | 'android',
+  projectRoot: string,
+  exp: PublicConfig
+): string {
+  if (platform === 'ios') {
+    return exp.ios && exp.ios.publishManifestPath
+      ? exp.ios.publishManifestPath
+      : _getDefaultEmbeddedManifestPath(platform, projectRoot, exp);
+  } else if (platform === 'android') {
+    return exp.android && exp.android.publishManifestPath
+      ? exp.android.publishManifestPath
+      : _getDefaultEmbeddedManifestPath(platform, projectRoot, exp);
+  }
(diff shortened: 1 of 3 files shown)


packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -46,57 +48,100 @@ async function _maybeWriteArtifactsToDiskAsync(config: EmbeddedAssetsConfigurati
...
+  // set defaults
+  if (pkg.dependencies['expo-updates']) {
+    const context = StandaloneContext.createUserContext(projectRoot, exp);
+    const { supportingDirectory } = IosWorkspace.getPaths(context);
+    const androidAssetsDir = path.join(projectRoot, 'android', 'app', 'src', 'main', 'assets');
+
+    await fs.ensureDir(supportingDirectory);
+    await fs.ensureDir(androidAssetsDir);
+
+    androidBundlePath = path.join(androidAssetsDir, 'shell-app.bundle');
```
brentvatne: would be nice to change these to names that make more sense for non-shell-app projects

PR #1761, opened 2020-03-31
Warn before ejecting that some config needs to be set on dynamic config

brentvatne:
This PR adds some checks before proceeding with ejecting to ensure that it's compatible with app.config.[ts/js]

```
✖ We are unable to continue with ejecting until you make the following changes to app.config.ts:
- name: A project name is required.
- ios.bundleIdentifier: An iOS bundle identifier is required. Details.
- android.package: An Android package is required. Details.
```


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -164,56 +169,97 @@ async function createNativeProjectsFromTemplateAsync(projectRoot: string): Promi
...
+        'An Android package is required.',
+        'https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#2-configure-appjson',
+      ]);
+    }
 
-  let bundleIdentifier = await getOrPromptForBundleIdentifier(projectRoot);
-  appJson.expo.ios = appJson.expo.ios ?? {};
-  appJson.expo.ios.bundleIdentifier = bundleIdentifier;
+    if (entryPoint) {
+      warnings.push(['entryPoint', 'Remove the entryPoint field.', undefined]);
```
EvanBacon, line 205: This might be the only warning that isn't covered now

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -290,11 +288,9 @@ async function createNativeProjectsFromTemplateAsync(projectRoot: string): Promi
...
 
-  // TODO: remove entryPoint and log about it for sdk 37 changes
-  if (exp.entryPoint && exp.entryPoint !== EXPO_APP_ENTRY) {
-    log(`- expo.entryPoint is already configured, we recommend using "${EXPO_APP_ENTRY}`);
-  } else {
-    exp.entryPoint = EXPO_APP_ENTRY;
+  if (exp.entryPoint) {
+    delete exp.entryPoint;
+    log(`- expo.entryPoint is not needed and has been removed.`);
   }
```
EvanBacon, line 294: I've modified this so the entryPoint is simply deleted upon ejection. In a static config, this should be persisted, otherwise it won't matter because every eject will delete the field.

PR #1773, opened 2020-04-01
[eject] Fix `GoogleService-Info.plist` file after ejecting to bare

IjzerenHein:
# Why

When ejecting to bare, the iOS `GoogleService-Info.plist` file is not copied to the correct location, nor is the `target membership` flag enabled. This will cause the Firebase SDKs to not be initialized.

# How

- Rename `GoogleServices-Info.plist` to `GoogleService-Info.plist`. 
- Enable the `target membership` flag for `GoogleService-Info.plist`.

# Test plan

## Before

image

image

image

## After fix

image

image


brentvatne:
thank you for fixing this!

Issue #1701, opened 2020-03-18
Easier metro config by making Expo defaults available


brentvatne:
Idea from https://github.com/expo/expo-cli/issues/1019#issuecomment-600082709

fson:
I'm going to do this as a part of https://github.com/expo/expo-cli/issues/1434.

PR #1777, opened 2020-04-01
feat: add @expo/metro-config

fson:
This PR adds the default Metro configuration for Expo apps as a separate `@expo/metro-config` package. The configuration is a minimal Metro config required to run React Native apps.

# Why

This is the first PR towards rolling out `@expo/dev-server`, which will replace the previous setup where we run the React Native CLI as a subprocess. The new dev server will use `@expo/metro-config` as a configuration, when it sets up a Metro server directly.

# Testing

TBD, still working on this, planning to check that an app can be bundled with Metro CLI when using this config. 

Fixes https://github.com/expo/expo-cli/issues/1701.

### DIFF
--- a/packages/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/metro-config/src/ExpoMetroConfig.ts
@@ -58,7 +58,13 @@ export interface LoadOptions {
   sourceExts?: string[];
 }
 
-export async function loadAsync(projectRoot: string, options: LoadOptions = {}): Promise<ConfigT> {
-  const defaultConfig = getDefaultConfig(projectRoot);
-  return loadConfig({ cwd: projectRoot, projectRoot, ...options }, defaultConfig);
+export async function loadAsync(
+  projectRoot: string,
+  { reporter, ...options }: LoadOptions = {}
+): Promise<ConfigT> {
+  let defaultConfig = getDefaultConfig(projectRoot);
+  if (reporter) {
+    defaultConfig = { ...defaultConfig, reporter };
+  }
+  return await loadConfig({ cwd: projectRoot, projectRoot, ...options }, defaultConfig);
 }
--- a/packages/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/metro-config/src/ExpoMetroConfig.ts
@@ -58,7 +58,7 @@ export interface LoadOptions {
   sourceExts?: string[];
 }
 
-export async function load(projectRoot: string, options: LoadOptions = {}): Promise<ConfigT> {
+export async function loadAsync(projectRoot: string, options: LoadOptions = {}): Promise<ConfigT> {
   const defaultConfig = getDefaultConfig(projectRoot);
   return loadConfig({ cwd: projectRoot, projectRoot, ...options }, defaultConfig);
 }
--- a/packages/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/metro-config/src/ExpoMetroConfig.ts
@@ -1,0 +1,64 @@
+import path from 'path';
+
+import { getConfig, resolveModule } from '@expo/config';
+import { Reporter } from 'metro';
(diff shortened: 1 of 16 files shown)


packages/metro-config/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ExpoMetroConfig';
```
EvanBacon, line 1: Maybe just make the main file point to `ExpoMetroConfig.ts`?

packages/metro-config/transformer.js
```diff
@@ -0,0 +1,14 @@
+const transformer = require('metro-react-native-babel-transformer');
+
+exports.getCacheKey = transformer.getCacheKey;
+
+exports.transform = function({ filename, options, src, plugins }) {
+  return transformer.transform({
+    filename,
+    options: Object.assign({}, options, {
+      extendsBabelConfigPath: require.resolve('babel-preset-expo'),
```
EvanBacon, line 9: nice

packages/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export interface LoadOptions {
+  config?: string;
+  maxWorkers?: number;
+  port?: number;
+  reporter?: Reporter;
+  resetCache?: boolean;
+  // TODO(ville): make this a part of the default config?
+  sourceExts?: string[];
```
brentvatne, line 58: yeah we should do this. our current approach of passing in sourceExts in flag will clobber any value provided in metro.config.js, and people need to pass it in to packageOpts in app.json, which we should remove in favor of telling people to just use metro.config.js.

one tricky thing here is that the sourceExts need to be different depending on whether we are starting the packager for managed or bare workflow -- we don't use .expo.js in bare

packages/metro-config/transformer.js
fson: The transformer is copy pasted from your Metro PR, so thank you 😄

packages/metro-config/src/ExpoMetroConfig.ts
fson: That is true. It needs to be possible to override it in `metro.config.js` and the interface for doing that could be something like this:
```js
// metro.config.js
const { getDefaultConfig } = require('@expo/metro-config');

const defaultConfig = getDefaultConfig(__dirname);

module.exports = {
  resolver: {
    sourceExts: [...defaultConfig.resolver.sourceExts, "svg"]
  }
};
```

This is indeed made a bit trickier by the fact that the default is different between bare and managed. We have access to the app configuration in `getDefaultConfig`, so we should be able to tell if it's a managed or bare app, *but* now it's also possible to override that setting for bare apps using the `--target` flag, at runtime: https://github.com/expo/expo-cli/pull/1734.

We might have to add an environment variable like `EXPO_WORKFLOW` or `EXPO_TARGET` to be able to access this information when `getDefaultConfig` is called from `metro.config.js`.

Or as an alternative, it kind of seems that ideally we wouldn't bake the `expo` "platform extension" into `sourceExts` at all (`expo.js`, `expo.jsx`, `expo.ts`, `expo.tsx`...), , but instead it would be separate, like the `resolver.platforms` config in Metro is separate from `sourceExts`. Then you would, regardless of workflow, simply have something like
```
sourceExts: ['ts', 'tsx', 'js', 'jsx', 'json', 'wasm']
```
and it would be possible to prefix each of them with `expo` for the managed target. To enable this, we might have to define a `CustomResolver` (the `resolve.resolveRequest` setting) or we could see if defining platforms like this would work:
```
platforms: ['expo.android', 'expo.ios', 'expo.web', 'android', 'ios', 'web']
```

fson: I'll open a new pull request for this change, once it's ready.

Issue #1760, opened 2020-03-31
[Android] Error when using custom keystore: "Cannot read property 'errors' of undefined"


ShimShamSam:
## Description

We are receiving an error when trying to build using our own keystore (not managed or created by Expo).

The issue seems to have been caused by #1700

We do not receive this error if we let Expo manage the keystore for us. In our case, this simply isn't an option.

For now, we have set the `EXPO_LECACY_API` environment variable to use the old version.

## Expected Behavior

The Android build should complete without error

## Observed Behavior

```
Cannot read property 'errors' of undefined
TypeError: Cannot read property 'errors' of undefined
    at Object.updateCredentialsForPlatform (/@expo/xdl@57.7.1/src/credentials/credentials.ts:135:21)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Environment

```
Expo CLI 3.16.1 environment info:
    System:
      OS: macOS Mojave 10.14.6
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
      npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
    IDEs:
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^36.0.0 => 36.0.2 
      react: 16.9.0 => 16.9.0 
      react-native: https://github.com/expo/react-native/archive/sdk-36.0.1.tar.gz => 0.61.4 
    npmGlobalPackages:
      expo-cli: 3.16.1
```

byCedric:
Hi @ShimShamSam! Thanks for the report and for checking out where the issue is located. I was able to reproduce this using a new project and the latest expo cli (`3.17.1`). As you said, using `EXPO_LECACY_API=true expo build:android` would work as a temporary workaround until this is resolved. Again, thanks for the issue! 😄

nangot:
Hi same problem here with expo cli 3.17.11 !

Issue #1774, opened 2020-04-01
error to upload my own keystore


jsoares568974:
## Description

error to upload my own keystore

## Expected Behavior

I do constant uploads of my apk, I went to do the same process as always, I managed to generate the ipa for ios, the android presented errors

Command: `expo build:android --config "variants/rocketpizza.json" --clear-credentials`

Selected: `2) I want to upload my own keystore!`

Path to keystore: /Users/macbookpro/Documents/projetos/plom-beta/plom-lanches.bak.jks
? Keystore Alias: delivery-key-alias
? Keystore Password: [hidden]
? Key Password: [hidden]

## Observed Behavior

Cannot read property 'errors' of undefined
TypeError: Cannot read property 'errors' of undefined
    at Object.updateCredentialsForPlatform (/@expo/xdl@57.7.1/src/credentials/Credentials.ts:135:21)
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
    at AndroidBuilder.collectAndValidateCredentials (/expo-cli@3.16.1/src/commands/build/AndroidBuilder.ts:204:9)
    at AndroidBuilder.run (/expo-cli@3.16.1/src/commands/build/AndroidBuilder.ts:30:5)
    at AndroidBuilder.command (/expo-cli@3.16.1/src/commands/build/BaseBuilder.ts:29:7)
    at Command.<anonymous> (/expo-cli@3.16.1/src/exp.ts:85:7)`

## Environment

Expo CLI 3.16.1 environment info:
    System:
      OS: macOS 10.15.4
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 12.6.0 - /usr/local/bin/node
      Yarn: 1.17.3 - /usr/local/bin/yarn
      npm: 6.14.4 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5014246
      Xcode: 10.2/10E125 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^36.0.0 => 36.0.2 
      react: 16.12.0 => 16.12.0 
      react-native: https://github.com/expo/react-native/archive/sdk-36.0.1.tar.gz => 0.61.4 
      react-navigation: ^4.1.1 => 4.3.6 
    npmGlobalPackages:
      expo-cli: 3.16.1

jkhales:
Hi! Would you mind trying the same process but with 
[...]

PR #1798, opened 2020-04-03
[credentials][apiv2] fix credentials bug

jkhales:
# why
Close: https://github.com/expo/expo-cli/issues/1760 and https://github.com/expo/expo-cli/issues/1774

# how 
The cli was trying to read `result.data.errors` instead of `result.errors`.

Fixed this and made `EXPO_NEXT_API` the default again.

# test

- checked it built with user supplied keystore. 
- threw a dummy ApiError server side and confirmed it handled it.

# extra

The solution and the testing of this was suggested @byCedric 🙏

### DIFF
--- a/packages/xdl/src/credentials/Credentials.ts
+++ b/packages/xdl/src/credentials/Credentials.ts
@@ -113,7 +113,7 @@ export async function updateCredentialsForPlatform(
   userCredentialsIds: Array<number>,
   metadata: CredentialMetadata
 ): Promise<void> {
-  if (process.env.EXPO_LECACY_API) {
+  if (process.env.EXPO_LEGACY_API) {
     const { err, credentials } = await Api.callMethodAsync('updateCredentials', [], 'post', {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/credentials/Credentials.ts
```diff
@@ -135,6 +124,17 @@ export async function updateCredentialsForPlatform(
...
+    const { experienceName } = metadata;
+    const user = await UserManager.ensureLoggedInAsync();
+    const api = ApiV2.clientForUser(user);
+    const result = await api.putAsync(`credentials/android/keystore/${experienceName}`, {
+      credentials: newCredentials,
+    });
+
+    if (result.errors) {
+      throw new Error(`Error updating credentials: ${JSON.stringify(result.errors)}}`);
+    }
```
jkhales, line 137: This is were the error was: `result.data.errors` instead of `result.errors`

Issue #1742, opened 2020-03-28
Logging should support %s format


brentvatne:
Some logs show up like this:

`Warning: %s: Error boundaries should implement getDerivedStateFromError(). In that method, return a state update to display an error message or fallback UI., RootErrorBoundary`

I'm not too familiar with how this works in JS but I suspect it's like printf

fson:
We implemented this with @byCedric in #1799.

PR #1799, opened 2020-04-03
feat: handle device logs with format strings (e.g. %s)

fson:
Fixes https://github.com/expo/expo-cli/issues/1742.


brentvatne:
nice one! we should probably add tests around logger, worth doing in follow up pr

Issue #1786, opened 2020-04-01
Configuration of sourceExts is awkward


brentvatne:
We override sourceExts provided through metro.config.js, so people need to pass them in through `packagerOpts.sourceExts` in app.json. *But* then they need to make sure to pass in all of our defaults too! See: https://github.com/kristerkari/react-native-svg-transformer#step-3-configure-the-react-native-packager

fson:
Duplicate of https://github.com/expo/expo-cli/issues/2199

Issue #1019, opened 2019-08-05
"sourceExts" in the Metro config are ignored


zyxel648:
### Environment
  Expo CLI 3.0.8 environment info:
    System:
      OS: Windows 10
    Binaries:
      npm: 6.10.0 - C:\Program Files\nodejs\npm.CMD
Android, Expo Client v.2.12.0

```
npm list -g --depth=0

+-- @vue/cli@3.9.3
+-- @vue/cli-init@3.9.0
+-- create-react-native-app@2.0.2
+-- exp@57.2.1
+-- expo-cli@3.0.8
+-- npm@6.10.0
+-- react-native-cli@2.0.1
+-- vue-native-cli@0.1.0
+-- webpack@4.29.6
`-- webpack-cli@3.3.0

```
package.json

```
{
  "main": "node_modules/expo/AppEntry.js",
  "scripts": {
    "start": "expo start",
    "android": "expo start --android",
    "ios": "expo start --ios",
    "web": "expo start --web",
    "eject": "expo eject"
  },
  "dependencies": {
    "expo": "^34.0.1",
    "react": "16.8.3",
    "react-dom": "^16.8.6",
    "react-native": "https://github.com/expo/react-native/archive/sdk-34.0.0.tar.gz",
    "react-native-web": "^0.11.4",
    "vue-native-core": "^0.1.0",
    "vue-native-helper": "^0.1.0",
    "vue-native-scripts": "0.0.16"
  },
  "devDependencies": {
    "@babel/core": "^7.0.0",
    "babel-preset-expo": "^6.0.0"
  },
  "private": true
}
```

`metro.config.js`

```
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */

const { getDefaultConfig } = require("metro-config");

module.exports = (async () => {
  const {
    resolver: { sourceExts }
  } = await getDefaultConfig();
  return {
    transformer: {
      babelTransformerPath: require.resolve("./vueTransformerPlugin.js"),
      getTransformOptions: async () => ({
        transform: {
          experimentalImportSupport: false,
          inlineRequires: false,
        },
      })
    },
    resolver: {
      sourceExts: [...sourceExts, "vue"]
    }
  };
})();
```

### Steps to Reproduce
Download [Repository](https://github.com/zyxel648/vue-native-sample)

1. run `npm install`
2. run `npm start`
3. open in expo client on Android

### Expected Behavior

Expected to see label "My Vue Native App"

[...]

PR #1821, opened 2020-04-06
feat: add sourceExts to default Metro config

fson:
Earlier discussion: https://github.com/expo/expo-cli/pull/1777#discussion_r402875727

Configuring the `sourceExts` setting of Metro bundler has previously been difficult with Expo CLI (see https://github.com/expo/expo-cli/issues/1019, https://github.com/expo/expo-cli/issues/1786), because we've been passing the `sourceExts` option to RN CLI, which overrides the setting specified in `metro.config.js`.

This PR adds `sourceExts` to the default Metro configuration added in https://github.com/expo/expo-cli/pull/1777. Once Expo CLI has been updated to load the configuration using the new package (I'm working on this), it will be possible to make use of these defaults when writing a custom `metro.config.js` file.

#### Usage with `metro.config.js`

For example, to add an extension to `sourceExts` in a managed/bare Expo app, one will be able to do this:
```js
// metro.config.js
const { getDefaultConfig } = require('@expo/metro-config');

const defaultConfig = getDefaultConfig(__dirname);

module.exports = {
  resolver: {
    sourceExts: [...defaultConfig.resolver.sourceExts, "svg"]
  }
};
```
~However, a notable corner case are *bare apps that can also run in Expo Client* (when published with `expo publish --target managed`). For those apps, instead of~
```js
const defaultConfig = getDefaultConfig(__dirname);
```
~you would have to use~
```js
const defaultConfig = getDefaultConfig(__dirname, { target });
```
~where `target` is the value of the `--target` flag passed to `expo publish`. One way to do this could be to write~
```js
const defaultConfig = getDefaultConfig(__dirname, { target: process.env.TARGET });
```
~and publish the app with `TARGET=managed expo publish --target managed`.~
UPDATE: bare apps publishing to a `managed` target are also handled, see https://github.com/expo/expo-cli/pull/1821#issuecomment-610434473

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,14 @@ workflows:
           e: node-10
           requires:
             - build10
+      - test-metro-config:
+          e: node-10
+          requires:
+            - build10
+      - test-metro-config:
+          e: node-12
+          requires:
+            - build12
 jobs:
   build:
     parameters:
@@ -114,7 +122,7 @@ jobs:
           name: Bootstrap (Link) Internal Dependencies
           command: yarn bootstrap
       - persist_to_workspace:
-          root: '~'
+          root: "~"
           paths:
             - expo-cli
       - save_cache:
@@ -130,7 +138,7 @@ jobs:
     working_directory: ~/expo-cli/packages/dev-tools
     steps:
       - attach_workspace:
-          at: '~'
+          at: "~"
       - run: yarn test
   test-config:
     parameters:
@@ -140,7 +148,7 @@ jobs:
     working_directory: ~/expo-cli/packages/config
     steps:
       - attach_workspace:
-          at: '~'
+          at: "~"
       - run: yarn test
   test-json-file:
     parameters:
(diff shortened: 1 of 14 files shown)


brentvatne:
awesome! the target part is definitely a bit awkward but i think it's an edge case for what is already an edge case. that said, the obscurity of it makes it hard for us to document. i think it would be useful for us to think about how we can flag the need for handling target in metro.config.js automatically. for example, when you run `expo publish --managed` in a bare app with a metro.config.js and metro.config.js does not match `/target/i` we should point them to a expo.fyi page about this

fson:
@brentvatne I've changed how target is handled now. In the end I decided against adding the potential trap for the bare app `--target managed` + `metro.config.js` use-case. It would have been very hard to document, not least because it's so rare to actually need that info. Trying to detect that scenario could also have been prone to error.

Instead of having to pass `target` to `getDefaultConfig` from `metro.config.js`, if you run `expo publish` with the `--target` flag, we will now automatically set the `EXPO_TARGET` environment variable and `getDefaultConfig` will determine the target by looking at a few things in this order:
1. `target` option passed to `getDefaultConfig` (Expo CLI will use this)
2. `EXPO_TARGET` env variable (automatically set when `--target` is used)
3. the default target for the project

I believe this approach, while it makes `getDefaultConfig` not so "pure" as a function, will provide a better experience as it normally should Just Work™, without adding a new thing you must know about as a developer.

PR #1832, opened 2020-04-08
Feature/upgrade codemod

byCedric:
The basics of adding `expo-codemod` to `expo upgrade` are here, but I think its good to polish some of these parts.

**Stuff to polish**
1. We need to know the exact transform name to apply (e.g. `sdk37-imports`). Right now I added a `expo-codemod --list 37` option (also requires a codemod publish when merging PR), to get a comma separated list of all imports available for the major version of the SDK. Would definitely prefer either adding option to run `expo-codemod` by sdk name, or fetch a list of codemods to apply from the versions API.
2. Probably need to add a range of all major SDK versions between start and end and perform each codemod per sdk. Before I start on this, I first want to fine tune point 1.
3. Me and @fson talked about invoking codemod programmatically, but that requires adding a specific codemod version to the CLI. When we add more codemods, we would need to update the CLI too. I avoided this by using `spawnAsync('npx', ['expo-codemod', ...]);`, maybe there is a better way?

**Examples**

<details>
<summary>Successfully upgrading from 36 to 37</summary>

</details>

<details>
<summary>Successfully upgrading from 35 to 36 (no codemod)</summary>

</details>

<details>
<summary>Failure when checking if a codemod is available</summary>

</details>

<details>
<summary>Failure during a codemod</summary>

</details>

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -2,6 +2,7 @@ import * as ConfigUtils from '@expo/config';
 import JsonFile from '@expo/json-file';
 import * as PackageManager from '@expo/package-manager';
 import { Android, Project, Simulator, Versions } from '@expo/xdl';
+import spawnAsync from '@expo/spawn-async';
 import chalk from 'chalk';
 import program, { Command } from 'commander';
 import _ from 'lodash';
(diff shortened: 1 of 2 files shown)


brentvatne:
i think this is a neat feature but we haven't needed codemods for a while, and ideally we won't in the future :) let's keep it alive and we can come back to it if we need to run codemods on a future sdk

EvanBacon:
stale, but maybe useful for the next upgrade command

PR #1845, opened 2020-04-09
refactor: add @expo/dev-server

fson:
`@expo/dev-server` will use the new `dev-server-api` and `@expo/metro-config` packages to set up a Metro dev server compatible with React Native projects. In addition to the React Native development endpoints, it will include additional endpoints e.g. for serving Expo manifests and receiving logs from Expo client.

### DIFF
--- a/packages/dev-server/LICENSE
+++ b/packages/dev-server/LICENSE
@@ -1,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2015-present 650 Industries, Inc. (aka Expo)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
--- a/packages/dev-server/LICENSE
+++ b/packages/dev-server/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015-present 650 Industries, Inc. (aka Expo)
+Copyright (c) 2020-present 650 Industries, Inc. (aka Expo)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
@@ -19,4 +19,3 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
-
--- a/packages/dev-server/README.md
+++ b/packages/dev-server/README.md
@@ -1,0 +1,1 @@
+# dev-server
--- a/packages/dev-server/README.md
+++ b/packages/dev-server/README.md
@@ -1,1 +1,5 @@
 # dev-server
+
+A React Native development server.
+
+This package is used by Expo CLI to run a development server for React Native apps. It is built on [Metro](https://facebook.github.io/metro/) (a JavaScript bundler) and [`@react-native-community/cli-server-api`](https://github.com/react-native-community/cli/tree/master/packages/cli-server-api) (the HTTP and WebSocket API for React Native apps to interface with the development server).
--- a/packages/dev-server/src/LogReporter.ts
+++ b/packages/dev-server/src/LogReporter.ts
@@ -1,0 +1,21 @@
+import Log from '@expo/bunyan';
+import { serializeError } from 'serialize-error';
+
+export default class LogReporter {
+  logger: Log;
+  reportEvent: (event: any) => void;
+
+  constructor(logger: Log, reportEvent: (event: any) => void = () => {}) {
+    this.logger = logger;
+    this.reportEvent = reportEvent;
+  }
+
+  update(event: any) {
+    if ('error' in event && event.error instanceof Error) {
+      event.error = serializeError(event.error);
+    }
+    // TODO(ville): replace xdl.PackagerLogsStream with a reporter to avoid serializing to JSON.
+    this.logger.info({ tag: 'metro' }, JSON.stringify(event));
+    this.reportEvent(event);
+  }
+}
--- a/packages/dev-server/src/LogReporter.ts
+++ b/packages/dev-server/src/LogReporter.ts
@@ -2,10 +2,7 @@ import Log from '@expo/bunyan';
 import { serializeError } from 'serialize-error';
 
 export default class LogReporter {
-  logger: Log;
-  reportEvent: (event: any) => void;
-
-  constructor(logger: Log, reportEvent: (event: any) => void = () => {}) {
+  constructor(public logger: Log, public reportEvent: (event: any) => void = () => {}) {
     this.logger = logger;
     this.reportEvent = reportEvent;
   }
--- a/packages/dev-server/src/LogReporter.ts
+++ b/packages/dev-server/src/LogReporter.ts
@@ -11,7 +11,7 @@ export default class LogReporter {
   }
 
   update(event: any) {
-    if ('error' in event && event.error instanceof Error) {
+    if (event.error instanceof Error) {
       event.error = serializeError(event.error);
     }
     // TODO(ville): replace xdl.PackagerLogsStream with a reporter to avoid serializing to JSON.
--- a/packages/dev-server/src/MetroDevServer.ts
+++ b/packages/dev-server/src/MetroDevServer.ts
@@ -4,10 +4,9 @@ import Log from '@expo/bunyan';
 import * as ExpoMetroConfig from '@expo/metro-config';
 import clientLogsMiddleware from './middleware/clientLogsMiddleware';
 
-export async function runMetroDevServerAsync(
-  projectRoot: string,
-  options: ExpoMetroConfig.LoadOptions & { logger: Log }
-) {
+export type MetroDevServerOptions = ExpoMetroConfig.LoadOptions & { logger: Log };
+
+export async function runMetroDevServerAsync(projectRoot: string, options: MetroDevServerOptions) {
   let reportEvent: ((event: any) => void) | undefined;
   // TODO(ville): implement a reporter
   const reporter = {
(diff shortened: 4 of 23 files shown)


packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function _isAppRegistryStartupMessage(body: any[]) {
+  return (
+    body.length === 1 &&
+    (/^Running application "main" with appParams:/.test(body[0]) ||
+      /^Running "main" with \{/.test(body[0]))
+  );
+}
+
+type ConsoleLogLevel = 'info' | 'warn' | 'error' | 'debug';
```
EvanBacon, line 36: Move to the top

packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function _handleDeviceLogs(logger: Log, deviceId: string, deviceName: string, logs: any) {
```
EvanBacon, line 38: Remove underscore

packages/dev-server/src/middleware/__tests__/clientLogsMiddleware-test.ts
```diff
@@ -0,0 +1,115 @@
...
+import clientLogsMiddleware from '../clientLogsMiddleware';
+
+const headers = {
+  'content-type': 'application/json',
+  'device-id': '11111111-CAFE-0000-0000-111111111111',
+  'session-id': '22222222-C0DE-0000-0000-222222222222',
+  'device-name': 'iPhone',
+};
+
+it('logs messages from the device', async () => {
```
EvanBacon, line 17: Nice

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const { middleware, attachToServer } = createDevServerMiddleware({
+    port: metroConfig.server.port,
+    watchFolders: metroConfig.watchFolders,
+  });
+  middleware.use(clientLogsMiddleware(options.logger));
+
+  const customEnhanceMiddleware = metroConfig.server.enhanceMiddleware;
+  // @ts-ignore can't mutate readonly config
+  metroConfig.server.enhanceMiddleware = (metroMiddleware: any, server: Metro.Server) => {
```
EvanBacon, line 33: I think perhaps this could be added to the `@expo/metro-config`:

```js
server: {
  enhanceMiddleware: (metroMiddleware: any, server: Metro.Server) => {
    if (customEnhanceMiddleware) {
      metroMiddleware = customEnhanceMiddleware(metroMiddleware, server);
    }
    return middleware.use(metroMiddleware);
    },
}
```

packages/dev-server/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2015-present 650 Industries, Inc. (aka Expo)
```
EvanBacon, line 3:
```suggestion
Copyright (c) 2020-present 650 Industries, Inc. (aka Expo)
```

packages/dev-server/src/MetroDevServer.ts
fson: I'll look into it. I think this might open some interesting possibilities, e.g. bare Metro CLI could work out of the box.

packages/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -69,7 +71,7 @@ export function getDefaultConfig(
       },
     },
     transformer: {
-      babelTransformerPath: require.resolve('../transformer'),
+      babelTransformerPath: require.resolve('metro-react-native-babel-transformer'),
```
fson, line 74: Couldn't get the custom transformer to work for now, for some reason when using it something goes wrong and I start getting "missing class properties transform" errors... I think we'll want to do it as a follow up, though (it allows removing `babel.config.js` while defaulting to `babel-preset-expo`).

EvanBacon: definitely would be worth adding.

packages/dev-server/src/LogReporter.ts
```diff
@@ -0,0 +1,21 @@
...
+  logger: Log;
+  reportEvent: (event: any) => void;
+
+  constructor(logger: Log, reportEvent: (event: any) => void = () => {}) {
+    this.logger = logger;
+    this.reportEvent = reportEvent;
+  }
+
+  update(event: any) {
+    if ('error' in event && event.error instanceof Error) {
```
EvanBacon, line 14: ??
```suggestion
    if (event.error instanceof Error) {
```

packages/dev-server/src/LogReporter.ts
```diff
@@ -0,0 +1,21 @@
+import Log from '@expo/bunyan';
+import { serializeError } from 'serialize-error';
+
+export default class LogReporter {
+  logger: Log;
+  reportEvent: (event: any) => void;
+
+  constructor(logger: Log, reportEvent: (event: any) => void = () => {}) {
```
EvanBacon, line 8:
```suggestion
  constructor(public logger: Log, public reportEvent: (event: any) => void = () => {}) {
```

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as ExpoMetroConfig from '@expo/metro-config';
+import bodyParser from 'body-parser';
+
+import clientLogsMiddleware from './middleware/clientLogsMiddleware';
+import LogReporter from './LogReporter';
+
+export type MetroDevServerOptions = ExpoMetroConfig.LoadOptions & { logger: Log };
+
+export async function runMetroDevServerAsync(projectRoot: string, options: MetroDevServerOptions) {
+  const { exp } = getConfig(projectRoot);
```
EvanBacon, line 13:
```suggestion
  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
```
It'd be nice to change how we do `nodeModulesPath` so we can avoid having to read the config every time we want to resolve a string that rarely exists.

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -0,0 +1,45 @@
...
+  const reporter = new LogReporter(options.logger);
+
+  const metroConfig = await ExpoMetroConfig.loadAsync(projectRoot, { reporter, ...options });
+
+  const { middleware, attachToServer } = createDevServerMiddleware({
+    port: metroConfig.server.port,
+    watchFolders: [...metroConfig.watchFolders],
+  });
+  middleware.use(bodyParser.json());
+  middleware.use('/logs', clientLogsMiddleware(options.logger));
```
EvanBacon, line 25: Maybe add comments here explaining what these middleware features do

packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+    res.end('Success');
+  };
+}
+
+function isIgnorableBugReportingExtraData(body: any[]) {
+  return body.length === 2 && body[0] === 'BugReporting extraData:';
+}
+
+function isAppRegistryStartupMessage(body: any[]) {
```
EvanBacon, line 40: Add return types

packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,93 @@
+import http from 'http';
+import Log from '@expo/bunyan';
+
+type ConsoleLogLevel = 'info' | 'warn' | 'error' | 'debug';
+
+export default function clientLogsMiddleware(logger: Log) {
+  return function(
+    req: http.IncomingMessage & { body?: any },
+    res: http.ServerResponse,
+    next: (err?: any) => void
```
EvanBacon, line 10:
```suggestion
    next: (err?: Error) => void
```
??

packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,93 @@
...
+function isAppRegistryStartupMessage(body: any[]) {
+  return (
+    body.length === 1 &&
+    (/^Running application "main" with appParams:/.test(body[0]) ||
+      /^Running "main" with \{/.test(body[0]))
+  );
+}
+
+function handleDeviceLogs(logger: Log, deviceId: string, deviceName: string, logs: any) {
+  for (let i = 0; i < logs.length; i++) {
```
EvanBacon, line 49:
```suggestion
  for (const log of logs) {
```

packages/dev-server/src/middleware/clientLogsMiddleware.ts
```diff
@@ -0,0 +1,93 @@
...
+    const log = logs[i];
+    let body = typeof log.body === 'string' ? [log.body] : log.body;
```
EvanBacon, line 51:
```suggestion
    let body = typeof log.body === 'string' ? [log.body] : log.body;
```
Is it always `string | string[]`? Might want to change this to `Array.isArray()`

packages/dev-server/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "@expo/babel-preset-cli/tsconfig.base",
+  "include": ["src/**/*.ts"],
```
EvanBacon, line 3: Be sure to exclude tests:
```json
{
  "extends": "@expo/babel-preset-cli/tsconfig.base",
  "compilerOptions": {
    "outDir": "build",
    "rootDir": "src"
  },
  "include": ["src/**/*.ts"],
  "exclude": ["**/__mocks__/*", "**/__tests__/*"]
}
```

packages/xdl/src/Project.ts
```diff
@@ -2105,6 +2127,32 @@ export async function stopExpoServerAsync(projectRoot: string): Promise<void> {
...
+    logger: ProjectUtils.getLogger(projectRoot),
+  };
+  if (startOptions.reset) {
+    options.resetCache = true;
+  }
+  if (startOptions.maxWorkers != null) {
+    options.maxWorkers = startOptions.maxWorkers;
+  }
+  if (startOptions.target) {
+    options.target = startOptions.target;
```
EvanBacon, line 2150: Thoughts on moving `EXPO_TARGET` here to keep the dev server package more agnostic?

packages/xdl/src/Project.ts
```diff
@@ -2346,6 +2394,9 @@ export async function startAsync(
     await Webpack.restartAsync(projectRoot, options);
     DevSession.startSession(projectRoot, exp, 'web');
     return exp;
+  } else if (getenv.boolish('EXPO_USE_DEV_SERVER', false)) {
+    await startDevServerAsync(projectRoot, options);
+    DevSession.startSession(projectRoot, exp, 'native');
```
EvanBacon, line 2399: IIRC we talked about potentially putting this in another file so we could identify and deprecate the legacy code more easily.

packages/dev-server/README.md
```diff
@@ -0,0 +1 @@
+# dev-server
```
EvanBacon, line 1: Maybe add some short info about the use case.

packages/dev-server/src/MetroDevServer.ts
fson: Can you explain why `skipSDKVersionRequirement` is needed? Given that most call sites of `getConfig` pass in `skipSDKVersionRequirement: true`, maybe it should be the default? Then in those cases where it's actually needed we could call `getExpoSDKVersion(projectRoot, exp)`, which throws for missing `sdkVersion`? I think this would be a cleaner API, since normally you just want the config without caring too much about how the `sdkVersion` is treated.

packages/dev-server/src/middleware/clientLogsMiddleware.ts
fson: I believe in practice it's always `LogData[]`, which means it's never anything else than an array and this check might be unnecessary (https://github.com/expo/expo/blob/e8ae25a7c9646adbb5641d0a792542a6d351e88e/packages/expo/src/logs/RemoteLogging.ts#L10-L25). Changed this to `isArray` anyway, as that's a more robust way to check this.

packages/dev-server/tsconfig.json
fson: I want `__tests__` to be type checked however, as far as I know Jest compiles them with Babel which doesn't do type checking? Maybe we can only ignore them in `.npmignore` or `files` when publishing to npm?

packages/xdl/src/Project.ts
fson: I'd rather do this as a part of splitting up XDL, I'll start working on it soon.

packages/xdl/src/Project.ts
fson: Good idea 👍

PR #1877, opened 2020-04-14
fix(expo-cli): terminate bundle progress bar when finished

### LINKED ISSUE #1770: Fix progress bars when publishing
When running `expo publish` without an existing CLI instance open (and sometimes with it open) it's really easy to reproduce this:

Screen Shot 2020-03-31 at 8 54 35 PM

byCedric:
Fixes #1770

We can also keep the status bar, and just add a new line. But because [the progress bar is initialized explicitly with `{ clear: true }`](https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/exp.ts#L245) I thought this was the most appropriate fix.

> Edit: If we want to keep the status bar listed as completed, removing `{ clear: true }` will fix that. [Terminate checks if it should be removed or "new-lined"](https://github.com/visionmedia/node-progress/blob/master/lib/node-progress.js#L227-L236)

### DIFF
--- a/packages/expo-cli/src/exp.ts
+++ b/packages/expo-cli/src/exp.ts
@@ -261,6 +261,7 @@ Command.prototype.asyncActionProjectDir = function(
 
         if (bar) {
           log.setBundleProgressBar(null);
+          bar.terminate();
           bar = null;
 
           if (err) {


brentvatne:
hopefully this fixes it! :)

Issue #824, opened 2019-07-08
expo init disregards --android-package and --ios-bundle-identifier arguments


kokeksibir:
## Description

I am trying to set android package name and ios bundle identifier while initializing the project. Here is the commands I execute

```
$ expo --version
2.21.2
$ expo init testApp --android-package com.test.app --ios-bundle-identifier com.test.app
...
```

## Expected Behavior

In `app.json` file I expect having `android.package` and `ios.bundleIdentifier` values to be set at least. When I choose a bare minimum project, I expect the given package name is also injected to relevant project files like `android/app/build.gradle`, `AndroidManifest.xml` and  `ios/<projectName>.xcodeproj`.

## Observed Behavior

Expected behavior defined above is not present at all. CLI disregards these arguments.

## Environment

```
  Expo CLI 2.21.2 environment info:
    System:
      OS: macOS 10.14.5
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.3 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.8 => 0.59.8 
    npmGlobalPackages:
      expo-cli: 2.21.2
```

rg4real:
Confirming.

brentvatne:
indeed there appears to have been a regression here, thanks for the report

te3:
expo v3.13.1, '--android-package' not working.

EvanBacon:
https://github.com/expo/expo-cli/pull/1878#issuecomment-679325012

PR #1878, opened 2020-04-14
feat(expo-cli): add package and bundle identifier to init

### LINKED ISSUE #432: Flags `--android-package` and `--ios-bundle-identifier` of `init` command not updating app.json
## Description

The `--android-package` and `--ios-bundle-identifier` flags of the `init` command do not set the corresponding properties in **app.json**.

## Expected Behavior

When passing the `--android-package com.expo.test` and `--ios-bundle-identifier com.expo.test` flags to the `init` command, the `expo.android.package` and `expo.ios.bundleIdentifier` properties, respectively, should both be set to `"com.expo.test"` all in **app.json**.

## Observed Behavior

When passing the `--android-package com.expo.test` and `--ios-bundle-identifier com.expo.test` flags to the `init` command, the `expo.android.package` and `expo.ios.bundleIdentifier` properties, respectively, do not appear at all in **app.json**.

## Environment

Output of `expo diagnostics`: 

>   Expo CLI 2.11.9 environment info:
>     System:
>       OS: Windows 10
>     Binaries:
>       Yarn: 1.13.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
>       npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
>     IDEs:
>       Android Studio: Version  3.3.0.0 AI-182.5107.16.33.5314842

byCedric:
Fixes #432

### DIFF
--- a/packages/expo-cli/e2e/__tests__/init-test.ts
+++ b/packages/expo-cli/e2e/__tests__/init-test.ts
@@ -29,7 +29,7 @@ xtest('init', async () => {
   expect(appJson).toHaveProperty(['expo', 'slug'], 'hello-world');
 });
 
-test('init with android package and bundle identifier', async () => {
+xtest('init with android package and bundle identifier', async () => {
   jest.setTimeout(60000);
   const cwd = temporary.directory();
   const { stdout } = await runAsync(
--- a/packages/expo-cli/e2e/__tests__/init-test.ts
+++ b/packages/expo-cli/e2e/__tests__/init-test.ts
@@ -28,3 +28,27 @@ xtest('init', async () => {
   expect(appJson).toHaveProperty(['expo', 'name'], 'Hello');
   expect(appJson).toHaveProperty(['expo', 'slug'], 'hello-world');
 });
+
+test('init with android package and bundle identifier', async () => {
+  jest.setTimeout(60000);
+  const cwd = temporary.directory();
+  const { stdout } = await runAsync(
+    [
+      'init',
+      'hello-world',
+      '--template',
+      'blank',
+      '--name',
+      'Hello',
+      '--android-package',
+      'host.expo.android',
+      '--ios-bundle-identifier',
+      'host.expo.ios',
+    ],
+    { cwd, env: { ...process.env, YARN_CACHE_FOLDER: path.join(cwd, 'yarn-cache') } }
+  );
+  expect(stdout).toContain(`Your project is ready`);
+  const appJson = await JsonFile.readAsync(path.join(cwd, 'hello-world/app.json'));
+  expect(appJson).toHaveProperty(['expo', 'android', 'package'], 'host.expo.android');
+  expect(appJson).toHaveProperty(['expo', 'ios', 'bundleIdentifier'], 'host.expo.ios');
+});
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/commands/init.ts
```diff
@@ -472,5 +523,7 @@ export default function(program: Command) {
     .option('--yarn', 'Use Yarn to install dependencies. (default when Yarn is installed)')
     .option('--name [name]', 'The name of your app visible on the home screen.')
     .option('--yes', 'Use default options. Same as "expo init . --template blank')
+    .option('--android-package [name]', 'The package name for your Android standalone app')
+    .option('--ios-bundle-identifier [name]', 'The bundle identifier for your iOS standalone app')
```
EvanBacon, line 527: Maybe something like this 🤔 
```suggestion
    .option('--bundle-id [name]', 'The bundle identifier for your iOS standalone app')
```

packages/expo-cli/e2e/__tests__/init-test.ts
```diff
@@ -23,8 +23,32 @@ test('init', async () => {
...
     { cwd, env: { ...process.env, YARN_CACHE_FOLDER: path.join(cwd, 'yarn-cache') } }
   );
-  expect(stdout).toMatch(`Your project is ready at ${cwd}`);
+  expect(stdout).toContain(`Your project is ready`);
   const appJson = await JsonFile.readAsync(path.join(cwd, 'hello-world/app.json'));
   expect(appJson).toHaveProperty(['expo', 'name'], 'Hello');
   expect(appJson).toHaveProperty(['expo', 'slug'], 'hello-world');
 });
+
+test('init with android package and bundle identifier', async () => {
```
EvanBacon, line 32: awesome 😍

packages/expo-cli/src/commands/init.ts
byCedric: Yeah, it's a bit long for ios 😄 We could do `--package`/`--package-name` and `--bundle-id`?

byCedric:
Oh wait, I'll patch the start e2e test. I know whats up with that 😄

brentvatne:
we should ensure that when a bare project is initialized that it uses the bundle id and package flags as well

byCedric:
Yeah, I noticed the `--name` flag is also not picked up in bare. I'll add some additional e2e tests for bare.

byCedric:
I'm doing a follow up PR for the e2e tests, but I noticed more stuff is going on in bare.

- The iOS `project.pbxproj` still contains `productName = "Hello World";`
- Android namespaces are still just the name, but lowercased.

Is this what you meant that needs to be tested @brentvatne?

brentvatne:
@byCedric - perhaps we can reuse the @expo/config code that handles setting bundleIdentifier and package? we use this only in eject currently and it should handle the cases you mentioned

packages/expo-cli/src/commands/init.ts
EvanBacon: Didn't realize it was already a feature, we should probably keep it how it was https://github.com/expo/expo-cli/issues/824

EvanBacon:
fix https://github.com/expo/expo-cli/issues/824

fson:
@byCedric can we ship this?

byCedric:
@fson I need to reimplement this using `@expo/config`, I'm afraid it's not yet mergeable. I'll get going on this on Monday or maybe this weekend, sorry!

brentvatne:
@EvanBacon plans to use the bundle id and package prompts that he wrote for eject here too

PR #1879, opened 2020-04-14
feat(expo-cli): add shortcut to open editor

### LINKED ISSUE #1762: Hotkey to open code at project root in editor
It'd be neat if I could go into a project and type `expo start` then open app in simulator in one keypress and open the code in visual studio code in another keypress. I usually end up opening multiple terminal tabs, so I run `expo start` and then switch to new tab and write `code .`

byCedric:
Fixes #1762

This uses Sindre's `env-editor` and `spawnAsync` over some other options I've tried. The env editor basically tries to find the default editor, if that's not found it will fallback to vscode. If none of these 2 options work, it will log with an error and describes how to fix it.

I tested this on _MacOS_, _Windows_ and _Windows WSL_ 😄

**Fun facts(tm)**

- `open-editor` didn't work that well for folders on both MacOS and Windows. On MacOS it thought it was a file and appended line and colon (`/my/path:1:1`) to this. On Windows there were some errors with `--goto` flag.
- `open` didn't work on MacOS, apparently MacOS can't open vscode with `open /my/path -a vscode` (but does work with `open /my/path -a "Visual Studio Code"`). Also `open` didn't reject the first attempt, it resolved without errors.

### DIFF
--- a/packages/expo-cli/src/commands/utils/EditorUtils.ts
+++ b/packages/expo-cli/src/commands/utils/EditorUtils.ts
@@ -21,7 +21,7 @@ export async function startEditorAsync(path: string, options: { editor?: string 
   }
 
   log.error(
-    'Could not open editor, you can set it by defining the $EDITOR env var with the binary of your editor. (e.g. "code" or "atom")'
+    'Could not open editor, you can set it by defining the $EDITOR environment variable with the binary of your editor. (e.g. "code" or "atom")'
   );
   return false;
 }
--- a/packages/expo-cli/src/commands/utils/EditorUtils.ts
+++ b/packages/expo-cli/src/commands/utils/EditorUtils.ts
@@ -1,5 +1,5 @@
+import spawnAsync from '@expo/spawn-async';
 import editors from 'env-editor';
-import open from 'open';
 import log from '../../log';
 
 export function guessEditor() {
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -4178,11 +4178,6 @@
   dependencies:
     "@types/node" "*"
 
-"@types/caseless@*":
```
byCedric, line 4181: Not sure what's going on here, looks like the yarn lockfile could be slightly outdated?

packages/expo-cli/src/commands/utils/EditorUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (editor) {
+    try {
+      await spawnAsync(editor.binary, [path]);
+      return true;
+    } catch {}
+  }
+
+  log.error(
+    'Could not open editor, you can set it by defining the $EDITOR env var with the binary of your editor. (e.g. "code" or "atom")'
```
brentvatne, line 24:
```suggestion
    'Could not open editor, you can set it by defining the $EDITOR environment variable with the binary of your editor. (e.g. "code" or "atom")'
```

brentvatne:
we have previously used the EXPO_EDITOR env var for this, i can't remember exactly where though, i think it might have been in XDE. the old script for this is here: https://github.com/expo/expo-cli/blob/cc9e1d8b9f8f5834fc21b67c2414a13a516f2ea8/packages/xdl/src/FileSystem.ts#L47-L55

i wonder why we chose to use `osascript` 🤔

PR #1893, opened 2020-04-15
Convert web console color to node console color

jrobber:
In `_handleDeviceLogs` we want to parse incoming logs to convert hex codes and special web characters to node friendly command line characters.

This includes support for using commas to separate console logs into multiple parts and adding color to parts individually by tracking historical color additions as we loop through the input.

I used ChalkJS because it is already used in other parts of the application. No new dependencies needed.

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -1670,7 +1670,7 @@ function convertWebColorToChalk(logObj: MessageWithColorSupport): string {
     .split('%c')
     .filter((x: string) => x)
     .map((part: string, idx: number) => {
-      return colors[idx] ? chalk.hex(colors[idx])(part) : part;
+      return colors[idx] ? chalk.hex(colors[idx]).inverse(part) : part;
     })
     .join('');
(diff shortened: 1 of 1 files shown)


brentvatne:
hello! i love this idea. could you extract the larger chunks of logic here out into functions and write some tests?

jrobber:
@brentvatne I'd be happy to do a re-factor and write some tests, but I'm not sure where you'd like them.  I found the __tests__ and __integrationTests__ folders in xdl.  Both of those are pretty barebones, and neither seems actively maintained.  

Do you mind pointing me at where you'd like the test and if there is an example format/style somewhere else that I can mimic?  I don't need a lot, just a pointer of where to look.  Thanks!

brentvatne:
hey @jrobber - yeah there has not been a good history of writing tests in xdl, this package is on its way to deprecation as we slowly take parts out. a better example to refer to would be the tests in @expo/config - https://github.com/expo/expo-cli/tree/master/packages/config/src/__tests__

PR #1907, opened 2020-04-17
fix(image-utils): add minimal required sharp cli version back

### LINKED ISSUE #1900: expo-optimize no longer working. Can't find sharp-cli
## Description

Any attempt of trying to run expo-optimize results in the error [below](#Observed-Behavior)

```sh
This will overwrite the original assets.

› Optimizing assets...
› Cannot optimize images without sharp-cli.
› Run this command again after successfully installing sharp with `npm install -g sharp-cli`
```

After digging a bit more, it seems the actual culprit is `@exp/image-util` as this is where the sharp detection logic lies.

Potential issue might be due this [commit](https://github.com/expo/expo-cli/commit/8b9b581c732b5a5e8c7f6907de5f7f141f9cf585) and this [findSharpBinAsync](https://github.com/expo/expo-cli/blob/3d849b51b8176065285fe65a3c461522f28037aa/packages/image-utils/src/sharp.ts#L95)

## Expected Behavior

Ran `yarn expo-optimize` and expected it to optimize my expo assets

## Observed Behavior

Expo did not optimize, but rather error-ed out

```sh
This will overwrite the original assets.

› Optimizing assets...
› Cannot optimize images without sharp-cli.
› Run this command again after successfully installing sharp with `npm install -g sharp-cli`
```

## Environment

```sh
  Expo CLI 3.18.2 environment info:
    System:
      OS: macOS 10.15.3
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
      Yarn: 1.22.4 - ~/.nvm/versions/node/v12.16.2/bin/yarn
      npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
    IDEs:
      Android Studio: 3.4 AI-183.5429.30.34.5452501
      Xcode: 11.4.1/11E503a - /usr/bin/xcodebuild
    npmPackages:
      expo: 37.0.7 => 37.0.7 
      react: 16.9.0 => 16.9.0 
      react-native: https://github.com/expo/react-native/archive/sdk-37.0.1.tar.gz => 0.61.4 
    npmGlobalPackages:
      expo-cli: 3.18.2
```

Also specify:

- expo-optimize: v0.1.19
- sharp-cli: v1.13.1

## Reproducible Demo

- Install `sharp-cli`
- init managed workflow expo project
- Install `expo-optimize` as dev dependency
- Run `yarn expo-optimize`
- DO NOT PROFIT.. :(

byCedric:
Fixes #1900

This hardcodes the minimal required sharp cli version back to image-utils. As reported in the issue, the problematic [commit is this](https://github.com/expo/expo-cli/commit/8b9b581c732b5a5e8c7f6907de5f7f141f9cf585). It removes the sharp-cli peer dependency from the package, which is actually used by the sharp detection code. As alternatives, we could also revert the change and add the peer dependency back. Or even remove the version required fully.

### DIFF
--- a/packages/image-utils/src/sharp.ts
+++ b/packages/image-utils/src/sharp.ts
@@ -4,6 +4,7 @@ import resolveFrom from 'resolve-from';
 import { Options, SharpCommandOptions, SharpGlobalOptions } from './sharp.types';
 
 const SHARP_HELP_PATTERN = /\n\nSpecify --help for available options/g;
+const SHARP_REQUIRED_VERSION = '^1.10.0';
 
 export async function resizeBufferAsync(buffer: Buffer, sizes: number[]): Promise<Buffer[]> {
   const sharp = await findSharpInstanceAsync();
@@ -96,15 +97,12 @@ async function findSharpBinAsync(): Promise<string> {
   if (_sharpBin) {
     return _sharpBin;
   }
-  const requiredCliVersion = require('@expo/image-utils/package.json').peerDependencies[
-    'sharp-cli'
-  ];
   try {
     const sharpCliPackage = require('sharp-cli/package.json');
     const libVipsVersion = require('sharp').versions.vips;
     if (
       sharpCliPackage &&
-      semver.satisfies(sharpCliPackage.version, requiredCliVersion) &&
+      semver.satisfies(sharpCliPackage.version, SHARP_REQUIRED_VERSION) &&
       typeof sharpCliPackage.bin.sharp === 'string' &&
       typeof libVipsVersion === 'string'
     ) {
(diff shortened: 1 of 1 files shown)


packages/image-utils/src/sharp.ts
```diff
@@ -12,7 +13,7 @@ export async function resizeBufferAsync(buffer: Buffer, sizes: number[]): Promis
   // Create buffer for each size
   const resizedBuffers = await Promise.all(
     sizes.map(dimension => {
-      const density = (dimension / Math.max(metadata.width, metadata.height)) * metadata.density;
+      const density = dimension / Math.max(metadata.width, metadata.height) * metadata.density;
```
byCedric, line 16: "I don't take responsibility at all" for this Prettier change, but I can revert it if required 😄

packages/image-utils/src/sharp.ts
```diff
@@ -4,6 +4,7 @@ import resolveFrom from 'resolve-from';
 import { Options, SharpCommandOptions, SharpGlobalOptions } from './sharp.types';
 
 const SHARP_HELP_PATTERN = /\n\nSpecify --help for available options/g;
+const SHARP_REQUIRED_VERSION = '^1.10.0';
```
byCedric, line 7: Taken from the previous `peerDependency`

packages/image-utils/src/sharp.ts
EvanBacon: we need to fix our husky config so it stops doing this.

EvanBacon: partially fixed https://github.com/expo/expo-cli/issues/1920

byCedric: Thanks, just rebased and removed these 2 changes

PR #1908, opened 2020-04-17
[expo-cli][android] Add helpful error when incorrectly uploading keystore

jkhales:


packages/expo-cli/src/commands/build/AndroidBuilder.ts
```diff
@@ -213,12 +213,30 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
...
-      !!process.env.EXPO_ANDROID_KEY_PASSWORD
-    );
+      !!process.env.EXPO_ANDROID_KEY_PASSWORD;
+
+    if (allEnvSet) {
+      return true;
+    }
+
+    // Check if user was trying to upload keystore incorretly and supply an helpful error message if so.
+    if (!!this.options.keystorePath || !!this.options.keystoreAlias) {
```
fson, line 227:
```suggestion
    if (this.options.keystorePath || this.options.keystoreAlias) {
```
Manually coercing these to boolean is not necessary, since it will happen automatically in boolean context. I think removing the `!!` makes the `if` statement slightly easier to scan quickly.

packages/expo-cli/src/commands/build/AndroidBuilder.ts
```diff
@@ -213,12 +213,30 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
...
+      throw Error(
+        'When uploading your own keystore you must provide:\n' +
+          '\t--keystore-path /path/to/your/keystore.jks \n' +
+          '\t--keystore-alias PUT_KEYSTORE_ALIAS_HERE \n' +
+          'And set the enviroment variables:\n' +
+          '\tEXPO_ANDROID_KEYSTORE_PASSWORD\n' +
+          '\tEXPO_ANDROID_KEY_PASSWORD\n' +
+          'For details, see:\n' +
+          '\thttps://docs.expo.io/versions/latest/distribution/turtle-cli/#building-for-android'
```
fson, line 236: This URL seems to be incorrect: this page is not about Expo CLI, it's about Turtle CLI and talks about building apps on your own server, not using our build service. The page about using the build service is Building Standalone Apps (https://docs.expo.io/versions/latest/distribution/building-standalone-apps/), but unfortunately it doesn't have many details about uploading a keystore. Please consider expanding [Building Standalone Apps](https://docs.expo.aio/versions/latest/distribution/building-standalone-apps/) and using that URL instead.

Also the [Expo CLI documentation](https://docs.expo.io/versions/latest/workflow/expo-cli/) is lacking: the `--keystore-alias` option is missing from the list of options there. I think we should add it and also mention in `--keystore-path` description that it must be used together with `--keystore-alias` and the environment variables must be set.

jkhales: Good catch

jkhales:
Updated the link and documentation.

PR #1911, opened 2020-04-17
[credentials][android][apiv2] Fix upload keystore

jkhales:
# why

Android credentials were not being updated on the apiv2 endpoint when running `expo build:android -c`.

# how

The cli was sending incorrectly formatted credentials to the endpoint.

# test
Followed @wkozyra95 detalied repro:
```
export EXPO_NEXT_API=true
expo init testProject
cd testProject
expo credentials:manager  // and generate generate upload kesytore
expo fetch:android:hashes // returns some hashes
expo fetch:android:keystore // backup keystore (it's better to move it somwhere because it can be overwritten)
expo credentials:manager  // and generate generate upload kesytore new one
expo fetch:android:hashes // verified that hashes changed
expo build:android -c // and provided keystore generated first time
expo fetch:android:hashes // hashes should change back to orignal ones but stayed the same, you don't need to wait for the end of the build to run this command, if build is in queue keystore should be updated already```

### DIFF
--- a/packages/expo-cli/src/commands/build/AndroidBuilder.ts
+++ b/packages/expo-cli/src/commands/build/AndroidBuilder.ts
@@ -1,7 +1,7 @@
 import fs from 'fs-extra';
 import path from 'path';
 import untildify from 'untildify';
-import { Android, Credentials } from '@expo/xdl';
+import { Android, AndroidCredentials, Credentials } from '@expo/xdl';
 import chalk from 'chalk';
 import get from 'lodash/get';
 
@@ -195,9 +195,9 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
         // read the keystore
         const keystoreData = await fs.readFile(keystorePath);
 
-        const credentials = {
+        const credentials: AndroidCredentials.Keystore = {
           keystore: keystoreData.toString('base64'),
-          keystoreAlias,
+          keyAlias: keystoreAlias,
           keystorePassword,
           keyPassword,
         };
--- a/packages/xdl/src/credentials/Credentials.ts
+++ b/packages/xdl/src/credentials/Credentials.ts
@@ -1,5 +1,5 @@
 import { Platform, getConfig } from '@expo/config';
-import { ApiV2 } from '../xdl';
+import { AndroidCredentials, ApiV2 } from '../xdl';
 
 import Api from '../Api';
 import UserManager from '../User';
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: packages/expo-cli/src/commands/build/AndroidBuilder.ts [hunk 1]
```diff
@@ -195,9 +195,9 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
         // read the keystore
         const keystoreData = await fs.readFile(keystorePath);
 
-        const credentials = {
+        const credentials: AndroidCredentials.Keystore = {
           keystore: keystoreData.toString('base64'),
-          keystoreAlias,
+          keyAlias: keystoreAlias,
```
wkozyra95, line 200: is this fix for apiv1 or v2 ?

Current version of expo-cli have default api set to v1 for know, but in PR description you mentioned that this is fix for apiv2? 

If this is fix for apiv1 then both `keyAlias` and `keystoreAlias` should still be supported in apiv1, because otherwise old clients will stop working. If this is intentional we should deprecate them.

If this is fix for apiv2 why default api is still on apiv1?

packages/xdl/src/credentials/Credentials.ts
```diff
@@ -1,5 +1,5 @@
 import { Platform, getConfig } from '@expo/config';
-import { ApiV2 } from '../xdl';
+import { AndroidCredentials, ApiV2 } from '../xdl';
```
wkozyra95, line 2: there is a import cycle here, you shoudl import this from respecitve files `../ApiV2` and `./AndroidCredentials`

packages/expo-cli/src/commands/build/AndroidBuilder.ts
jkhales: This is a fix for apiv2.

I am keeping apiv2 default until the teams feature is supported on the backend.

packages/xdl/src/credentials/Credentials.ts
jkhales: great point

packages/expo-cli/src/commands/build/AndroidBuilder.ts
jkhales: I updated the testing to clarify that it should be done with `EXPO_NEXT_API=true`.

jkhales: > 
> If this is fix for apiv1 then both `keyAlias` and `keystoreAlias` should still be supported in apiv1, because otherwise old clients will stop working. If this is intentional we should deprecate them.
> 

The object with renamed keys is only sent to an apiv2 endpoint, so we don't need to support `keystoreAlias`: `api.putAsync('credentials/android/keystore/${experienceName}'`

Issue #1919, opened 2020-04-18
expo init (What is the name of your project?) should have a default value 


EvanBacon:
Maybe default to a silly-willy random name like we do with snack. We should read file structure to ensure we don't recommend a copy.

PR #1923, opened 2020-04-18
added default name for projects

EvanBacon:
fix https://github.com/expo/expo-cli/issues/1919

### DIFF
--- a/packages/expo-cli/src/commands/init.ts
+++ b/packages/expo-cli/src/commands/init.ts
@@ -415,7 +415,7 @@ async function promptForBareConfig(
   } else {
     ({ projectName } = await prompt({
       name: 'projectName',
-      message: 'What is your app called?',
+      message: 'What would you like to name your app?',
       default: 'my-app',
       filter: (name: string) => name.trim(),
       validate: (name: string) => validateProjectName(name),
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/init.ts
```diff
@@ -415,7 +415,8 @@ async function promptForBareConfig(
   } else {
     ({ projectName } = await prompt({
       name: 'projectName',
-      message: 'What is the name of your project?',
+      message: 'What is your app called?',
```
brentvatne, line 418: not crazy about this wording change

`What would you like to name your app?` maybe seems better?

EvanBacon, line 418:
```suggestion
      message: 'What would you like to name your app?',
```

packages/expo-cli/src/commands/init.ts
```diff
@@ -441,7 +442,8 @@ async function promptForManagedConfig(
   } else {
     ({ slug } = await prompt({
       name: 'slug',
-      message: 'What is the name of your project?',
+      message: 'What is your app called?',
```
EvanBacon, line 445:
```suggestion
      message: 'What would you like to name your app?',
```

PR #1928, opened 2020-04-19
[cli] better build errors

quinlanj:
based on weekend slack chat https://expo-developers.slack.com/archives/C04Q3JTSV/p1587257406295000?thread_ts=1586737385.022700&cid=C04Q3JTSV

# why
- throw credentials errors in catch block so we dont continue if there is a fatal error
- Add context specific failures (ie) user knows they failed to setup a particular credential at a certain step

## context specific failures: before
Screen Shot 2020-04-23 at 2 26 40 PM

## context specific failures: after
Screen Shot 2020-04-23 at 2 26 04 PM

### DIFF
--- a/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
+++ b/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
@@ -35,6 +35,7 @@ import {
   getProvisioningProfileFromParams,
   useProvisioningProfileFromParams,
 } from '../../../credentials/views/IosProvisioningProfile';
+import { IosAppCredentials, IosDistCredentials } from '../../../credentials/credentials';
 
 class IOSBuilder extends BaseBuilder {
   appleCtx?: apple.AppleCtx;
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -144,21 +145,102 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
...
+      if (pushKeyFromParams) {
+        await usePushKeyFromParams(ctx, appCredentials, pushKeyFromParams);
+      } else {
+        await runCredentialsManager(
+          ctx,
+          new SetupIosPush({ experienceName, bundleIdentifier, nonInteractive })
+        );
+      }
+    } catch (e) {
+      log.error('Failed to set up Push Key');
```
brentvatne, line 203: is there any more context that we can provide in this error?

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -144,21 +145,102 @@ See https://docs.expo.io/versions/latest/distribution/building-standalone-apps/#
...
+          new SetupIosProvisioningProfile({
+            experienceName,
+            bundleIdentifier,
+            distCert: distributionCert,
+            nonInteractive,
+          })
+        );
+      }
+    } catch (e) {
+      log.error('Failed to set up Provisioning Profile');
```
brentvatne, line 238: similar to above, is there any more context we can provide here? maybe a expo.fyi link that explains why this can happen, for example

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
quinlanj: The more context-specific error with the `expo.fyi` link is caught and rethrown (which will then print to stderr). I catch it at this point and print out `Failed to set up Push Key` so we know which part of the credential process we failed at. The error logging would look something like this:

```
Failed to set up Push Key
Additional information needed to setup credentials in non-interactive mode.
Learn more about how to resolve this: expo.fyi/credentials-non-interactive (​https://expo.fyi/credentials-non-interactive​).
```

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
quinlanj: see above comment :)

PR #1941, opened 2020-04-21
[cli] add support back for owner field when cred checking

quinlanj:
# why
We used to support builds where if User A gave User B permission to build their app, User B would be able to build User A's app. However, User B would need to specify `UserA` was the owner of the app in the owner field of the `app.json`.

When we moved to the new credential checker, we overlooked this use case. We add it back now by doing the following:

- Add `username` field to credential's IosApi
- Add `owner` in the api's request body or query parameter as appropriate

# todo
- [x] add tests
- [x] land https://github.com/expo/universe/pull/4807 
- [x] deploy https://github.com/expo/universe/pull/4807  to prod
- [x]  sanity test for User A with Developer role on User B's account
- [x] sanity test for regular users

### DIFF
--- a/packages/expo-cli/src/commands/fetch/android.ts
+++ b/packages/expo-cli/src/commands/fetch/android.ts
@@ -3,6 +3,7 @@ import fs from 'fs-extra';
 import get from 'lodash/get';
 
 import { AndroidCredentials } from '@expo/xdl';
+import invariant from 'invariant';
 import { DownloadKeystore } from '../../credentials/views/AndroidCredentials';
 import { Context } from '../../credentials';
 
@@ -31,7 +32,8 @@ export async function fetchAndroidKeystoreAsync(projectDir: string): Promise<voi
   await maybeRenameExistingFile(projectDir, keystoreFilename);
   const backupKeystoreOutputPath = path.resolve(projectDir, keystoreFilename);
 
-  const view = new DownloadKeystore(ctx.manifest.slug);
+  invariant(ctx.manifest.slug, 'app.json slug field must be set');
+  const view = new DownloadKeystore(ctx.manifest.slug as string);
   await view.fetch(ctx);
   await view.save(ctx, backupKeystoreOutputPath, true);
 }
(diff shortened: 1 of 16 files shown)


packages/expo-cli/src/commands/fetch/android.ts
```diff
@@ -31,7 +32,8 @@ export async function fetchAndroidKeystoreAsync(projectDir: string): Promise<voi
   await maybeRenameExistingFile(projectDir, keystoreFilename);
   const backupKeystoreOutputPath = path.resolve(projectDir, keystoreFilename);
 
-  const view = new DownloadKeystore(ctx.manifest.slug);
+  invariant(ctx.manifest.slug, 'app.json slug field must be set');
```
brentvatne, line 35: in current expo-cli versions we set the slug automatically if it's not provided in app.json, so it would certainly be weird to get into this state! good to have that invariant

brentvatne: were we encountering cases where slug was not set?

quinlanj: i changed the type of `manifest` from `any` to `ExpoConfig`, and `slug` could be undefined in this case - so i added invariant checks to get rid of the errors

packages/expo-cli/src/credentials/context.ts
```diff
@@ -64,6 +70,9 @@ export class Context {
       const { exp } = getConfig(projectDir);
       this._manifest = exp;
       this._hasProjectContext = true;
+      this._iosApiClient = new IosApi(this.user).withProjectContext(this);
```
brentvatne, line 73: could use some more context on what's going on here - why do create an api client if the status is fatal? and why not with the project context?

quinlanj: i updated this with better comments -- basically we use `Doctor.validateWithoutNetworkAsync` to check whether there is a manifest in the project directory. If there is (we get a non fatal response), we initialize the necesary project context things, else we do a normal initialization

PR #1942, opened 2020-04-21
[webpack-config] fix: update workbox

jaulz:
The workbox plugin is already a year old. We should update it to a recent version as it supports more functionalities.


jaulz:
@EvanBacon I think once we have this we could also use `workbox-window` in `expo-updates` in order to check if new updates are available. Another thing is, that we should try to make it feasible to use Typescript in service worker files. What do you think?

EvanBacon:
I'm ok with the idea of TS support for SW files as long as we can accurately add them to the E2E tests.

PR #1943, opened 2020-04-21
Fix pasting path from finder

glennreyes:
When one copies the path from finder, it adds a trailing space. This PR trims that space away that leads to better user experience.


fson:
Hey @glennreyes! 👋  Thanks for the PR, looks good to me 😄

PR #1949, opened 2020-04-22
[DO NOT MERGE][cli] jest unit config

quinlanj:
made jest unit config analogous to the one we have for xdl. i always had this shim when developing but would probably be useful if committed

### DIFF
--- a/packages/expo-cli/jest/unit-test-config.js
+++ b/packages/expo-cli/jest/unit-test-config.js
@@ -4,13 +4,11 @@ module.exports = {
   rootDir: path.resolve(__dirname, '..'),
   roots: ['__mocks__', 'src'],
   displayName: require('../package.json').name,
-  testRegex: '/__(tests|testfixtures)__/.*(test|spec)\\.(j|t)sx?$',
   moduleNameMapper: {
     '^jest/(.*)': path.join(__dirname, '../../../jest/$1'),
   },
   transform: {
-    '^.+\\.[jt]sx?$': ['babel-jest', { configFile: require.resolve('../babel.config.js') }],
+    '^.+\\.[jt]sx?$': ['babel-jest', { configFile: require.resolve(__dirname, '../babel.config.js') }],
   },
   testEnvironment: 'node',
-  resetModules: false,
 };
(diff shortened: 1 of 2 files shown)


packages/expo-cli/jest/unit-test-config.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = {
+  rootDir: path.resolve(__dirname, '..'),
+  roots: ['__mocks__', 'src'],
+  displayName: require('../package.json').name,
+  testRegex: '/__(tests|testfixtures)__/.*(test|spec)\\.(j|t)sx?$',
+  moduleNameMapper: {
+    '^jest/(.*)': path.join(__dirname, '../../../jest/$1'),
+  },
+  transform: {
+    '^.+\\.[jt]sx?$': ['babel-jest', { configFile: require.resolve('../babel.config.js') }],
```
fson, line 12:
```suggestion
    '^.+\\.[jt]sx?$': ['babel-jest', { configFile: require.resolve(__dirname, '../babel.config.js') }],
```
(Otherwise `require.resolve` will resolve starting from `process.cwd()`.)

packages/expo-cli/jest/unit-test-config.js
```diff
@@ -0,0 +1,16 @@
...
+  },
+  testEnvironment: 'node',
+  resetModules: false,
```
fson, line 15:
```suggestion
```
([`false` is the default](https://jestjs.io/docs/en/configuration#resetmodules-boolean).)

packages/expo-cli/jest/unit-test-config.js
```diff
@@ -0,0 +1,16 @@
+const path = require('path');
+
+module.exports = {
+  rootDir: path.resolve(__dirname, '..'),
+  roots: ['__mocks__', 'src'],
+  displayName: require('../package.json').name,
+  testRegex: '/__(tests|testfixtures)__/.*(test|spec)\\.(j|t)sx?$',
```
fson, line 7: The default setting is `(/__tests__/.*|(\\.|/)(test|spec))\\.[jt]sx?$`, which should work. `__testfixtures__` is a special path used by `testUtils` of `jscodeshift` (we use this in `expo-codemod`), but that directory should not include regular tests.
```suggestion
```

PR #1950, opened 2020-04-22
Add codecov action

EvanBacon:
This introduces flags so in theory each report would be linked to a package

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
             node_modules
             */*/node_modules
           key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
-      - run: yarn lerna bootstrap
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Cache node_modules
+        id: cache_node_modules
+        uses: actions/cache@master
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
+
+      - run: yarn
```
fson, line 31:
```suggestion
      - run: yarn
```
It should be possible to omit this line because `lerna bootstrap` will run `yarn install`.

EvanBacon: But `lerna` won't be available until after yarn is run 🙃 
qTpo0MB

PR #1964, opened 2020-04-23
[expo-codemod] fix jest config

EvanBacon:
- root was in wrong place, this caused coverage to be generated in `src/coverage`.
- add codemod test to GH actions


brentvatne:
i guess you mean "add codemod test to GH actions" rather than "add metro test to GH actions" 😅

PR #1975, opened 2020-04-24
Always print stack trace

EvanBacon:
kinda silly that we make you rerun the command to see what failed the first time. Maybe there's some reason I'm not aware of?


fson:
Let's do this. Whoever added `EXPO_DEBUG` probably had good intentions, but it just makes reporting issues (and fixing them) a lot harder than it needs to be.

PR #1978, opened 2020-04-24
[xdl] Get xdl-manifests from apiv2 endpoint

jkhales:
# How

There are two locations xdl-manifests are fetched. For the first one in `packages/xdl/gulp/build-tasks.js` replaced the url. Confirmed this worked by repopulating the schema cache (git diff is empty):
```
cd ~/expo-cli/packages/xdl
rm caches/*
yarn prepack
git diff
```

For the second location, Line 69 in `packages/xdl/src/project/ExpSchema.ts`:
1) Refactored the Cache out of the Api v1 class and into its own: `XdlSchemaClient`. 
2) Checked it worked in a similar way to the first point, but this time checking that`~/.expo/caches/` was repopulated by running `expo start` in an expo project. (Note: you have to delete `~/expo-cli/packages/xdl/caches/*` before running this test to force the repopulation to happen from the endpoint instead of the local backup.)

### DIFF
--- a/packages/xdl/src/Utils.ts
+++ b/packages/xdl/src/Utils.ts
@@ -1,4 +1,9 @@
 import { ncp } from 'ncp';
+import path from 'path';
+import { JSONObject } from '@expo/json-file';
+import { Cacher } from './tools/FsCache';
+
+import { ApiV2 } from './xdl';
 
 export function ncpAsync(source: string, dest: string, options: any = {}) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


packages/xdl/src/Utils.ts
```diff
@@ -44,3 +49,22 @@ export class Semaphore {
     }
   }
 }
+
+export class XdlSchemaClient {
```
fson, line 53: A better name for this would be `ConfigurationSchemaClient`. (The old method name `xdlSchemaAsync`  was incorrect, or at least outdated by now.)

But since this "singleton" only has static properties and this is not Java, you might as well remove the class and have a function like `getConfigurationSchemaAsync`. (`_schemaCaches` can be a top-level variable in the module.) Please also move it to `ExpSchema.ts`, `Utils` only has some helper functions that are not really bound to our APIs or anything like that.

PR #1989, opened 2020-04-24
[expo-cli] Improve readability of asyncActionProjectDir

EvanBacon:
- `asyncActionProjectDir` accepts three parameters, we've been adding comment blocks to tell what those parameters do. This is bad practice.
- Mostly we disable the validation checks that are done, so I've flipped the condition where you enable the check instead of disabling it.
- Auth check wasn't being used so I removed it.

### DIFF
--- a/packages/expo-cli/src/commands/build/index.ts
+++ b/packages/expo-cli/src/commands/build/index.ts
@@ -49,24 +49,27 @@ export default function(program: Command) {
     .description(
       'Build a standalone IPA for your project, signed and ready for submission to the Apple App Store.'
     )
-    .asyncActionProjectDir(async (projectDir: string, options: IosOptions) => {
-      if (options.publicUrl && !UrlUtils.isHttps(options.publicUrl)) {
-        throw new CommandError('INVALID_PUBLIC_URL', '--public-url must be a valid HTTPS URL.');
-      }
-      let channelRe = new RegExp(/^[a-z\d][a-z\d._-]*$/);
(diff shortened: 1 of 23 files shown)


packages/expo-cli/src/commands/build/index.ts
```diff
@@ -117,9 +123,7 @@ export default function(program: Command) {
           ...options,
           dev: typeof options.dev === 'undefined' ? false : options.dev,
         });
-      },
-      /* skipProjectValidation: */ false,
```
EvanBacon, line 121: Im purposefully skipping this check now as it doesn't need to happen.

packages/expo-cli/src/commands/publish.ts
```diff
@@ -37,13 +37,19 @@ export async function action(projectDir: string, options: Options = {}) {
...
     );
     log.warn(
-      `These two packages are incompatible and ${chalk.bold('publishing updates with expo-updates will not work if expokit is installed.')}`
+      `These two packages are incompatible and ${chalk.bold(
+        'publishing updates with expo-updates will not work if expokit is installed.'
+      )}`
     );
     log.warn(
-      `If you intent to use ${chalk.bold('expo-updates')}, please remove ${chalk.bold('expokit')} from your dependencies.`
+      `If you intent to use ${chalk.bold('expo-updates')}, please remove ${chalk.bold(
```
brentvatne, line 50:
```suggestion
      `If you intend to use ${chalk.bold('expo-updates')}, please remove ${chalk.bold(
```

PR #1992, opened 2020-04-25
Print warning when running on newer version of Node.js

LinusU:
Right now, both Node.js 13 and 14 is "Current release"

Screenshot 2020-04-25 at 13 25 13

### DIFF
--- a/packages/expo-cli/bin/expo.js
+++ b/packages/expo-cli/bin/expo.js
@@ -23,7 +23,7 @@ const versionIsHigherThanSupported = semver.gtr(version, maxSupportedVersion.ran
 if (versionIsHigherThanSupported) {
   console.error(
     chalk.yellow(
-      `WARNING: Node.js version ${version} has not yet been tested.\n\n` +
+      `WARNING: expo-cli has not yet been tested against Node.js version ${version}. If you encounter any issues, please report them to https://github.com/expo/expo-cli/issues\n\n` +
         'expo-cli supports following Node.js versions:\n' +
         versionInfo
     )
--- a/packages/expo-cli/bin/expo.js
+++ b/packages/expo-cli/bin/expo.js
@@ -13,23 +13,29 @@ const isSupported = supportedVersions.some(function(supported) {
   return semver.satisfies(version, supported.range);
 });
 
-if (isSupported) {
+const versionInfo = supportedVersions
+  .map(supported => '* ' + supported.range + ' (' + supported.name + ')')
+  .join('\n');
+
+const maxSupportedVersion = supportedVersions[supportedVersions.length - 1];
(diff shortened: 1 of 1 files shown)


brentvatne:
we haven't tried using it with node 14 yet, perhaps we should add a new class of warning that this is a new release and there may be issues, but let people proceed with using expo-cli with it. can you update the pr to do that?

LinusU:
@brentvatne does this look good?

btw, `chalk` doesn't even import on Node.js <8, so effectively the error is only printed for 8.x, 11.x.  Maybe that isn't a problem though? I'm not really sure what the motivation behind the warning is ☺️

packages/expo-cli/bin/expo.js
```diff
@@ -13,23 +13,29 @@ const isSupported = supportedVersions.some(function(supported) {
...
 
-  const maxSupportedVersion = supportedVersions[supportedVersions.length - 1];
-  const versionIsHigherThanSupported = semver.gtr(version, maxSupportedVersion.range);
+const maxSupportedVersion = supportedVersions[supportedVersions.length - 1];
+const versionIsHigherThanSupported = semver.gtr(version, maxSupportedVersion.range);
 
+if (versionIsHigherThanSupported) {
+  console.error(
+    chalk.yellow(
+      `WARNING: Node.js version ${version} has not yet been tested.\n\n` +
```
brentvatne, line 26:
```suggestion
      `WARNING: expo-cli has not yet been tested against Node.js version ${version}. If you encounter any issues, please report them to https://github.com/expo/expo-cli/issues\n\n` +
```

brentvatne:
@LinusU - i didn't know that :O

the purpose of this is that we don't want to have people creating issues about problems they encounter with unsupported versions of expo-cli, and we found that if we don't block it from starting people will just ignore the warnings and try anyways. we can be pretty sure that on node < 10 people are going to have problems, so we just save them the trouble of encountering an obscure error and stop them from starting it at all.

thanks for the pr!

LinusU:
> the purpose of this is that we don't want to have people creating issues about problems they encounter with unsupported versions of expo-cli, and we found that if we don't block it from starting people will just ignore the warnings and try anyways.

Thanks for the explanation, that makes sense!

> i didn't know that :O

`semver` also only supports Node.js 10 and up, so we probably need to do this without dependencies if you want it to work on any old version. We are doing something similar in [StandardJS here](https://github.com/standard/standard/blob/37d8bb202bf7f250f7d403c4c35d0bd097d6b39f/bin/cmd.js). I'd be happy to submit a PR if you want to...

brentvatne:
that would be great, thank you @LinusU!

LinusU:
Thanks for merging! Follow up PR is here: #2030

PR #2007, opened 2020-04-27
feat: suggest closest match to an unknown command

### LINKED ISSUE #1926: Integrate something like didyoumean to suggest command when typo
```
╰─$ expo publis
"publis" is not an expo command. See "expo --help" for the full list of commands.
```

this could instead be

```
╰─$ expo publis
"publis" is not a valid command -- did you mean "expo publish"?
See "expo --help" for the full list of commands.
```

jamesgeorge007:
closes #1926

### DIFF
--- a/packages/expo-cli/src/exp.ts
+++ b/packages/expo-cli/src/exp.ts
@@ -4,6 +4,7 @@ import url from 'url';
 
 import ProgressBar from 'progress';
 import last from 'lodash/last';
+import leven from 'leven';
 import compact from 'lodash/compact';
 import findLastIndex from 'lodash/findLastIndex';
 import boxen from 'boxen';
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/exp.ts
```diff
@@ -364,9 +365,15 @@ function runAsync(programName: string) {
     });
 
     program.on('command:*', subCommand => {
-      log.warn(
-        `"${subCommand}" is not an ${programName} command. See "${programName} --help" for the full list of commands.`
+      let msg = `"${subCommand}" is not an expo command. See "expo --help" for the full list of commands.`;
+      const availableCommands = program.commands.map(cmd => cmd._name);
+      const suggestion = availableCommands.find(
+        cmd => leven(cmd, subCommand[0]) < cmd.length * 0.4
```
brentvatne, line 371: what's the logic behind `cmd.length * 0.4`? might be useful to add in a comment

jamesgeorge007: `leven` returns the distance between 2 given strings based on the `Levenshtein distance algorithm`. Here we're setting the threshold value to `40%` of the command length.

brentvatne: right, any reason why you chose 40% as the threshold or just a guess at what might work well?

jamesgeorge007: https://github.com/dcporter/didyoumean.js/#threshold

jamesgeorge007: CI is red for an unknown reason 🤔

Issue #2013, opened 2020-04-27
Add --target flag to start


brentvatne:
This will allow @nikki93 to test dev client with bare workflow

fson:
We shouldn't complicate the developer facing interface with features we only use internally. https://github.com/expo/expo-cli/issues/1980 is another example of this.

fson:
Closing this issue, please use `EXPO_TARGET` environment variable if you need to experiment with this, for now.

PR #2014, opened 2020-04-27
Add --target flag to expo start

brentvatne:
This is useful for @nikki93 to test dev client. Closes https://github.com/expo/expo-cli/issues/2013

### DIFF
--- a/packages/expo-cli/src/commands/start.ts
+++ b/packages/expo-cli/src/commands/start.ts
@@ -1,8 +1,8 @@
 import {
   ExpoConfig,
   PackageJSONConfig,
+  ProjectTarget,
   getConfig,
-  getProjectConfigDescription,
   resolveModule,
 } from '@expo/config';
 // @ts-ignore: not typed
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/start.ts
```diff
@@ -139,6 +140,11 @@ function parseStartOptions(options: NormalizedOptions): Project.StartOptions {
     startOpts.maxWorkers = options.maxWorkers;
   }
 
+  // NOTE(brentvatne): Unlike in publish we default to managed when running `expo start` as there
+  // is currently no good reason why anyone would want to do this outside of someone who is working
+  // on the dev client
```
fson, line 145: This makes me think we shouldn't add this to the public command-line interface, at least not yet. Every option we add is a new thing for us to document and for the user to learn and figure out if they need to pass it in. Let's not unnecessarily clutter the interface. We should *minimize decision burden*.

The `EXPO_TARGET` environment variable that `@expo/metro-config` supports might be a reasonable way to support this internal use case for developing the development client. https://github.com/expo/expo-cli/blob/f6d426a803d0b7c13d489457ee8a98b145dd85b6/packages/metro-config/src/ExpoMetroConfig.ts#L32-L45

fson:
Closing this PR, since the `EXPO_TARGET` environment variable should be sufficient for internal/experimental use.

PR #2024, opened 2020-04-28
chore(deps): migrate to cli-table3

jamesgeorge007:
`cli-table3` is based on (and api compatible with) the original `cli-table`, and `cli-table2`, which are both unmaintained.


brentvatne:
thanks!

PR #2028, opened 2020-04-28
[xdl] clean up Expo.plist artifacts left behind by IosPlist util

esamelson:
The `IosPlist` util leaves behind some artifacts (`*.json`, `*.plist.bak`) which may be confusing to users who aren't expecting them.

In particular, users running `expo publish` in a bare workflow project with `expo-updates` may find these extra files (which come from modifying the data in the Expo.plist config file) confusing. Their utility is limited -- since most people are using version control there's no need to have separate backup files in the same folder. So in this case we should just delete those files after the `IosPlist.modifyAsync` operation completes successfully.

## Test plan

`expo init`, bare-minimum template, `expo publish`

- [x] latest expo-cli `expo publish` produces `Expo.json` and `Expo.plist.bak` files in the Supporting dir
- [x] after this change, `expo publish` no longer produces those files
- [x] deleting files that don't exist with `fs.removeSync` is a no-op

### DIFF
--- a/packages/xdl/src/EmbeddedAssets.ts
+++ b/packages/xdl/src/EmbeddedAssets.ts
@@ -272,6 +272,7 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
       }
       return configPlist;
     });
+    await IosPlist.cleanBackupAsync(supportingDirectory, 'Expo', false);
   }
 
   // Android expo-updates


packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -272,6 +272,10 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
       }
       return configPlist;
     });
+
+    // delete artifacts sometimes left behind by IosPlist util
+    fs.removeSync(path.join(supportingDirectory, 'Expo.json'));
+    fs.removeSync(path.join(supportingDirectory, 'Expo.plist.bak'));
```
brentvatne, line 278: should be able to use https://github.com/expo/expo-cli/blob/6972d777022ce078fd8041517441894b23126052/packages/xdl/src/detach/IosPlist.ts#L95-L107 instead

PR #2030, opened 2020-04-28
Make Node.js version gate work on older versions

LinusU:
Follow up to #1992, this makes the guard that prints a nice error when using an older version of Node.js work on older versions of Node.js.

It even supports 0.10 now 😄 

Screenshot 2020-04-28 at 21 11 58

(ping @brentvatne)


brentvatne:
thank you!

PR #2034, opened 2020-04-29
[expo-cli][xdl] changes for expo-updates no-publish workflow

esamelson:
Depends on https://github.com/expo/expo/pull/8066 to land and be released

This PR removes warnings in expo-cli for expo-updates, regarding publishing before making release builds.

It also stops automatically embedding manifests + bundles in Xcode/Android projects on `expo publish` and `expo export`, except for those projects using expo-updates `0.1.x` or when the files already exist (to support developers who have upgraded to new expo-updates but have not switched over to the no-publish workflow).

Also added a new warning to mitigate the following confusing workflow:
- init new project
- make test release build
- publish OTA update (which automatically sets URL)
- previous build did not have URL yet --> cannot download update --> ???

# Test plan

- [x] `expo publish` with `expo-updates@0.1.2` in package.json embeds assets automatically
- [x] `expo publish` with `expo-updates@0.2.0` in package.json does not embed assets automatically
- [x] `expo publish` with `expo-updates@0.2.0` in package.json, but WITH app.manifest/app.bundle files already created, does embed assets automatically

### DIFF
--- a/packages/xdl/src/EmbeddedAssets.ts
+++ b/packages/xdl/src/EmbeddedAssets.ts
@@ -287,6 +287,8 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
     return;
   }
 
+  let isLikelyFirstPublish = false;
+
   const { projectRoot, exp, releaseChannel, iosManifestUrl, androidManifestUrl } = config;
 
   const { iosSupportingDirectory: supportingDirectory } = getIOSPaths(projectRoot);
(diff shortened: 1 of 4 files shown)


packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -327,6 +362,15 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
       androidManifestXmlPath
     );
   }
+
+  if (isLikelyFirstPublish) {
+    logger.global.warn(
+      '🚀 It looks like this your first publish for this project! ' +
+        "We've automatically set some configuration values in Expo.plist and AndroidManifest.xml. " +
```
brentvatne, line 369: we might want to point to the full paths to Expo.plist and AndroidManifest.xml here so people can easily see if the path is incorrect (has sometimes happened when people rename their projects on ios) and open it with just cmd+clicking it

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -265,6 +297,9 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
   if (fs.existsSync(path.join(supportingDirectory, 'Expo.plist'))) {
```
brentvatne, line 297: should we warn if the user has `expo-updates` installed but we can't find `Expo.plist`?

brentvatne:
going to do a deploy right now, can address those comments in a follow-up pr if you like but i think it'd be good to get this in

PR #2037, opened 2020-04-29
Migrate all packages to use globby

EvanBacon:
We were using both glob and globby, this uses just one.


packages/config/package.json
```diff
@@ -49,10 +49,10 @@
     "@expo/plist": "0.0.4",
     "find-yarn-workspace-root": "^1.2.1",
     "fs-extra": "^7.0.1",
-    "globby": "^11.0.0",
+    "lodash": "4.17.15",
+    "globby": "11.0.0",
     "invariant": "^2.2.4",
     "jest-message-util": "^25.1.0",
-    "plist": "^3.0.1",
```
fson, line 55: Also the types should be removed https://github.com/expo/expo-cli/blob/46c419af3fe6993ac53c45f05008e8439f7c6126/packages/config/package.json#L65

packages/config/package.json
```diff
@@ -49,10 +49,10 @@
     "@expo/plist": "0.0.4",
     "find-yarn-workspace-root": "^1.2.1",
     "fs-extra": "^7.0.1",
-    "globby": "^11.0.0",
+    "lodash": "4.17.15",
```
fson, line 52: Removed lodash from `@expo/config` here: https://github.com/expo/expo-cli/commit/fd6223ef0aad7e14c2720c82528bd77a2d9919bf (it was only used in one function)

PR #2048, opened 2020-04-29
fix: resolve and import metro-config from the project

fson:
Fixes "Error: Cannot find module 'metro-config'" when running
```
EXPO_USE_DEV_SERVER=true expo start
```

### DIFF
--- a/packages/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/metro-config/src/ExpoMetroConfig.ts
@@ -1,6 +1,12 @@
 import path from 'path';
 
-import { ProjectTarget, getConfig, getDefaultTarget, resolveModule } from '@expo/config';
+import {
+  ProjectTarget,
+  getConfig,
+  getDefaultTarget,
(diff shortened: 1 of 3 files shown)


fson:
I initially missed this because `@expo/metro-config` had `metro-config` as a `devDependency`.

packages/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -96,5 +96,7 @@ export async function loadAsync(
   if (reporter) {
     defaultConfig = { ...defaultConfig, reporter };
   }
+  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const { loadConfig } = require(resolveModule('metro-config', projectRoot, exp));
```
EvanBacon, line 100: Would be useful to get the path with `projectHasModule('metro-config', projectRoot, exp)` and if the returned path is nullish then throw an error informing the user how to install the package.

PR #1876, opened 2020-04-14
fix(xdl): try to stop adb when shutting down

byCedric:
Fixes #283, #175

This PR requires a fix for the graceful shutdown on Windows, see #1749.

**What did I do**
- Add failure-permitted `Android.maybeStopAdbDaemonAsync` in the XDL
- Add `Android.maybeStopAdbDaemonAsync` in `Project._stopInternalAsync`

**Why did I do**
If you install the CLI locally you have the shipped ADB binaries in your project's `node_modules`. On Windows keeping the daemon running results in a "blocking" node modules folder, you can't remove the file while the process is running.

**How does this do for others**
The impact should be minimal. Even if the developer is using a custom installed ADB, the daemon is restarted automagically by ADB when executing commands 😄

PR #2064, opened 2020-05-04
@sviande/adb kill

sviande:
Stop ADB daemon only when expo launched it.

Fix a behavior introduce by #1876 it kills ADB daemon even if another process uses it.

Test 1
```
#ADB is not running
expo start
#ctrl-c expo
#adb is not running
```

Test 2
```
adb start-server
#use ADB (adb push, scrcpy....)
expo start
#ctrl-c expo
#ADB is always running
```

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -20,6 +20,8 @@ import * as Versions from './Versions';
 import { getUrlAsync as getWebpackUrlAsync } from './Webpack';
 
 let _lastUrl: string | null = null;
+let _isAdbOwner: boolean | null = null;
+
 const BEGINNING_OF_ADB_ERROR_MESSAGE = 'error: ';
 const CANT_START_ACTIVITY_ERROR = 'Activity not started, unable to resolve Intent';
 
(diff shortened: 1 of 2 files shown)


brentvatne:
@byCedric - what do you think?

byCedric:
Looks good! It's a similar approach to [our `publish` command](https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/commands/publish.ts#L202-L204). I didn't think it would matter a lot, because once you run a new ADB command it will start up the daemon again.

packages/xdl/src/Android.ts
```diff
@@ -20,6 +20,8 @@ import * as Versions from './Versions';
 import { getUrlAsync as getWebpackUrlAsync } from './Webpack';
 
 let _lastUrl: string | null = null;
+let _isAdbOwner: boolean | null = null;
```
byCedric, line 23: Could we store this in the local `.expo` project settings? Or is that overkill @brentvatne

brentvatne: i think the risk there is that the situation on the users' machine changes and then we have some cached value that's no longer accurate

PR #2073, opened 2020-05-05
chore: fix the failing build

fson:


fson:
It's the second time this has happended within a week... looks like we're no longer get failures in CI, if the TypeScript build is failing. I will add this to CI.

PR #2078, opened 2020-05-06
fix(config): add postPublishHook type 

cruzach:
`postPublishHook` aren't strings, just moving this type out of `xdl` and into `config`

### DIFF
--- a/packages/config/src/Config.types.ts
+++ b/packages/config/src/Config.types.ts
@@ -35,6 +35,24 @@ export type ProjectConfig = {
 };
 export type AppJSONConfig = { expo: ExpoConfig; [key: string]: any };
 export type BareAppConfig = { name: string; [key: string]: any };
+export type HookArguments = {
+  config: any;
+  url: any;
+  exp: ExpoConfig;
+  iosBundle: string;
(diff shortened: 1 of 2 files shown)


packages/config/src/Config.types.ts
```diff
@@ -35,6 +35,24 @@ export type ProjectConfig = {
...
+  androidBundle: string;
+  androidSourceMap: string | null;
+  androidManifest: any;
+  projectRoot: string;
+  log: (msg: any) => void;
+};
+export type PostPublishHook = {
+  file: string;
+  config: any;
+  _fn: (input: HookArguments) => any;
```
fson, line 54:
```suggestion
```
I believe `_fn` should not be included here, since it's not a part of the `hooks` field in `ExpoConfig`, it's only added in `Project.ts` when pre-processing the hooks. You could add another type in `Project.ts` for this:
```ts
type LoadedPostPublishHook = PostPublicHook & {
  _fn: (input: HookArguments) => any;
}
```

cruzach: oh good catch, made those changes in the most recent commit 👍

PR #2087, opened 2020-05-07
[config] automate android icon

cruzach:
## Why?
This would automate setting the android icon and adaptive icons (Android 8+) on eject, and would enable developers to change their app icons much more easily when using the bare workflow. Plus, we could use the same code when building shell apps. 

## Testing
Run `yarn test`. CodeCov report should go up even more if we replace the current [android icons code](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/AndroidIcons.ts) with calls to `AndroidConfig.setIconAsync` 😉 

## Additional info
This should replace the current [android icons code](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/AndroidIcons.ts), and would fix https://github.com/expo/expo/issues/7374 (if we remove the old AndroidIcons code and replace with a call to `Config.setIconAsync`)


cruzach:
hm for some reason [this call](https://github.com/expo/expo-cli/blob/b6ca2457961c2f08eaa992bbc485d420fdd39687/packages/config/src/android/Icon.ts#L123-L126) is causing those `expo-cli` package tests to fail... still trying to investigate why

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    foregroundImage: config.android?.adaptiveIcon?.foregroundImage ?? null,
+    backgroundColor: config.android?.adaptiveIcon?.backgroundColor ?? '#FFFFFF',
+    backgroundImage: config.android?.adaptiveIcon?.backgroundImage ?? null,
+  };
 }
 
 export async function setIconAsync(config: ExpoConfig, projectRoot: string) {
-  let icon = getIcon(config);
+  const { foregroundImage, backgroundColor, backgroundImage } = getAdaptiveIcon(config);
+  let icon = foregroundImage || getIcon(config);
```
dsokal, line 38:
```suggestion
  const icon = foregroundImage ?? getIcon(config);
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
   if (!icon) {
-    return;
+    return null;
+  }
+
+  const iconPath = resolve(projectRoot, icon);
+  let length: number;
+
+  // Legacy Icon for Android 7 and earlier
+  for (let dpi in dpiValues) {
```
dsokal, line 48:
```suggestion
  for (const dpi in dpiValues) {
```
Screenshot 2020-07-02 at 10 12 34
I know that generally it doesn't matter but I've noticed you're mixing `let` with `const` without explicit consistency.

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+      if (backgroundImage && foregroundImage) {
+        // if a background image is supplied, layer the foreground on top of that image.
+        let resizedBackgroundImage = await resizeImageAsync(backgroundImage, length);
+        iconImage = resizedBackgroundImage.composite(iconImage, 0, 0);
+      } else if (backgroundColor && foregroundImage) {
+        // if a background color is supplied, layer the foreground on top of that color.
+        let resizedBackgroundImage = new Jimp(length, backgroundColor);
+        iconImage = resizedBackgroundImage.composite(iconImage, 0, 0);
+      }
+      iconImage.write(resolve(dpiFolderPath, IC_LAUNCHER_PNG));
```
dsokal, line 64:
```suggestion
      await iconImage.write(resolve(dpiFolderPath, IC_LAUNCHER_PNG));
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+      iconImage.circle().write(resolve(dpiFolderPath, IC_LAUNCHER_ROUND_PNG));
```
dsokal, line 65:
```suggestion
      await iconImage.circle().write(resolve(dpiFolderPath, IC_LAUNCHER_ROUND_PNG));
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  setBackgroundColor(projectRoot, backgroundColor);
+
+  for (let dpi in dpiValues) {
+    const { folderName, scale } = dpiValues[dpi];
+    length = 48 * scale;
+    let dpiFolderPath = resolve(projectRoot, ANDROID_RES_PATH, folderName);
+
+    try {
+      let finalAdpativeIconForeground = await resizeImageAsync(foregroundImage, length);
+      finalAdpativeIconForeground.write(resolve(dpiFolderPath, IC_LAUNCHER_FOREGROUND_PNG));
```
dsokal, line 86:
```suggestion
      await finalAdpativeIconForeground.write(resolve(dpiFolderPath, IC_LAUNCHER_FOREGROUND_PNG));
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+
+      if (backgroundImage) {
+        let finalAdpativeIconBackground = await resizeImageAsync(backgroundImage, length);
+        finalAdpativeIconBackground.write(resolve(dpiFolderPath, IC_LAUNCHER_BACKGROUND_PNG));
```
dsokal, line 90:
```suggestion
        await finalAdpativeIconBackground.write(resolve(dpiFolderPath, IC_LAUNCHER_BACKGROUND_PNG));
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    backgroundImage
   );
+  await createAdaptiveIconXmlFiles(projectRoot, icLauncherXmlString);
+  return true;
+}
+
+async function resizeImageAsync(imagePath: string, length: number) {
+  let imageBuffer = (await Jimp.read(imagePath)).resize(length, length).quality(100);
+
+  return imageBuffer;
```
dsokal, line 109:
```suggestion
  return (await Jimp.read(imagePath)).resize(length, length).quality(100);
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  }
+  let colorsJson = await Colors.readColorsXMLAsync(colorsXmlPath);
+  let colorItemToAdd = [
+    {
+      _: backgroundColor,
+      $: { name: ICON_BACKGROUND },
+    },
+  ];
+  colorsJson = Colors.setColorItem(colorItemToAdd, colorsJson);
+  Colors.writeColorsXMLAsync(colorsXmlPath, colorsJson);
```
dsokal, line 128:
```suggestion
  await Colors.writeColorsXMLAsync(colorsXmlPath, colorsJson);
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  backgroundColor: string | null,
+  backgroundImage: string | null
+) => `<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    ${
+      backgroundImage
+        ? `<background android:drawable="@mipmap/ic_launcher_background"/>`
+        : backgroundColor
+        ? `<background android:drawable="@color/iconBackground"/>`
+        : null
```
dsokal, line 141: Please add some parenthesis, I think most people are not sure about the operator precedence (`... ?  ... : ...`).

packages/config/package.json
```diff
@@ -40,6 +40,7 @@
     "glob": "7.1.6",
     "globby": "^11.0.0",
     "invariant": "^2.2.4",
+    "jimp": "^0.10.3",
```
bbarthec, line 43: AFAIK we would like to use our own package https://github.com/expo/expo-cli/tree/master/packages/image-utils for any `sharp`- or `jimp`-based image processing.
We should depend on this package and remove any direct `sharp` or `jimp` dependencies. Even further: if any new better-performing package would appear we would like to support in via `image-utils` package.

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
+import { resolve } from 'path';
+import Jimp from 'jimp';
+import fs from 'fs-extra';
 import { ExpoConfig } from '../Config.types';
-import { addWarningAndroid } from '../WarningAggregator';
+import * as Colors from './Colors';
+
+type dpiMap = { [dpistring: string]: { folderName: string; scale: number } };
```
bbarthec, line 7: I would go with sth like this here: https://www.staging-typescript.org/play?#code/MYewdgzgLgBAIgBQJIGUoCcCWYDmEYC8MA2gOQC2AJgA6akA0MpAFjXY6QB6u0NOfc2fLgJ50AujACG+UJCgBuAFBQAntQCm8ZGiy5CMNZpAAzbagzY8xMAFdyAIw3pxyo1rYBZKdQMAlDVB0SgAeRAs9HEYAbxgTEAAbSmcAOSlyDQAuGGhIhRzgKQSsmDtHZxgAXwA+BSA
```suggestion
const DPIStrings = ['mdpi', 'hdpi', 'xhdpi', 'xxhdpi', 'xxxhdpi'] as const;
type DPIString = typeof DPIStrings[number];
type dpiMap = Record<DPIString, { folderName: string; scale: number }>;
```
💅

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
-    return;
+    return null;
+  }
+
+  const iconPath = resolve(projectRoot, icon);
+  let length: number;
+
+  // Legacy Icon for Android 7 and earlier
+  for (let dpi in dpiValues) {
+    const { folderName, scale } = dpiValues[dpi];
```
bbarthec, line 49:
```suggestion
  for (const [dpi, { folderName, scale }] in Object.entries(dpiValues)) {
```

bbarthec: That's my preference and I don't know any performance issues with `Object.entries`, so it's just a suggestion 💅  😉

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    let dpiFolderPath = resolve(projectRoot, ANDROID_RES_PATH, folderName);
```
bbarthec, line 50:
```suggestion
    let dpiFolderPath = path.resolve(projectRoot, ANDROID_RES_PATH, folderName);
```
IMO, `resolve` method from `path` module is not self-descriptive enough 😞 
When I was reading the code and I saw `resolve` function I started to wonder what is it? On the other hand if I saw `path.resolve` (very common method in NodeJS) I would recognise it immediately and continue reading.
My suggestion for build-in NodeJS modules is to import whole module and then invoke method on a module object instead of importing these methods directly via `named imports`. It's all about readability, proper naming and context of invocation 😉

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    length = 48 * scale;
```
bbarthec, line 51:
```suggestion
    const length = SOME_MEANINGFUL_CONSTANT_HERE_PLEASE * scale;
```
I wish such variables are constants defined at top level of the module and ideally with some comment describing their semantics 😉

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+
+export function getAdaptiveIcon(config: ExpoConfig) {
+  return {
+    foregroundImage: config.android?.adaptiveIcon?.foregroundImage ?? null,
+    backgroundColor: config.android?.adaptiveIcon?.backgroundColor ?? '#FFFFFF',
+    backgroundImage: config.android?.adaptiveIcon?.backgroundImage ?? null,
+  };
 }
 
 export async function setIconAsync(config: ExpoConfig, projectRoot: string) {
```
bbarthec, line 36: As this method is rather a complex one, what would you say for writing a small method comment what's happening here and why? 🤔 
I wish that would became sort of practise when we write logic related code 🙏 😉

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    }
+  }
+
+  if (!foregroundImage) {
+    // If no foreground image, we shouldn't configure the Android Adaptive Icon
+    return null;
+  }
+
+  // set background color in colors.xml
+  setBackgroundColor(projectRoot, backgroundColor);
```
bbarthec, line 77:
```suggestion
  await setBackgroundColor(projectRoot, backgroundColor);
```
🤔

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  const { foregroundImage, backgroundColor, backgroundImage } = getAdaptiveIcon(config);
+  let icon = foregroundImage || getIcon(config);
+
   if (!icon) {
-    return;
+    return null;
+  }
+
+  const iconPath = resolve(projectRoot, icon);
+  let length: number;
```
bbarthec, line 45: I would remove this variable from here and create for-loop scoped variables instead.
```suggestion
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  if (!foregroundImage) {
+    // If no foreground image, we shouldn't configure the Android Adaptive Icon
+    return null;
+  }
+
+  // set background color in colors.xml
+  setBackgroundColor(projectRoot, backgroundColor);
+
+  for (let dpi in dpiValues) {
+    const { folderName, scale } = dpiValues[dpi];
```
bbarthec, line 80:
```suggestion
  for (const [dpi, { folderName, scale }] in Object.entries(dpiValues)) {
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    length = 48 * scale;
```
bbarthec, line 81:
```suggestion
    const length = SOME_MEANINGFUL_CONSTANT * scale;
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+    } catch (e) {
+      throw new Error('Encountered an issue resizing adaptive app icon: ' + e);
+    }
   }
 
-  addWarningAndroid(
-    'icon',
-    'This is the image that your app uses on your home screen, you will need to configure it manually.'
+  // create ic_launcher.xml and ic_launcher_round.xml
+  let icLauncherXmlString = createAdaptiveIconXmlString(
```
bbarthec, line 98:
```suggestion
  const icLauncherXmlString = createAdaptiveIconXmlString(
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+}
+
+async function resizeImageAsync(imagePath: string, length: number) {
+  let imageBuffer = (await Jimp.read(imagePath)).resize(length, length).quality(100);
+
+  return imageBuffer;
+}
+
+export async function setBackgroundColor(projectDir: string, backgroundColor: string) {
+  let colorsXmlPath = await Colors.getProjectColorsXMLPathAsync(projectDir);
```
bbarthec, line 113:
```suggestion
  const colorsXmlPath = await Colors.getProjectColorsXMLPathAsync(projectDir);
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+  if (!colorsXmlPath) {
+    console.warn(
+      'Unable to find a colors.xml file in your android project. Background color is not being set.'
+    );
+    return;
+  }
+  let colorsJson = await Colors.readColorsXMLAsync(colorsXmlPath);
+  let colorItemToAdd = [
```
bbarthec, line 121:
```suggestion
  const colorItemToAdd = [
```

packages/config/src/android/Icon.ts
bbarthec: @dsokal I'm not sure about this parenthesis idea.
Our linters are removing parenthesis in ternary operators chains and instead they're adding another level of indentation if needed. 🤷 
Moreover I consider below code as readable:
```ts
someConditional
  ? valueIfTruthy
  : anotherConditionalIfFalsy
  ? valueIfTruthy
  : valueIfFalsy
```

packages/config/src/android/Icon.ts
```diff
@@ -1,24 +1,152 @@
...
+        ? `<background android:drawable="@mipmap/ic_launcher_background"/>`
+        : backgroundColor
+        ? `<background android:drawable="@color/iconBackground"/>`
+        : null
+    }
+    <foreground android:drawable="@mipmap/ic_launcher_foreground"/>
+</adaptive-icon>`;
+
+async function createAdaptiveIconXmlFiles(projectRoot: string, icLauncherXmlString: string) {
+  let anyDpiV26Directory = resolve(projectRoot, ANDROID_RES_PATH, 'mipmap-anydpi-v26');
```
bbarthec, line 147:
```suggestion
  const anyDpiV26Directory = pathresolve(projectRoot, ANDROID_RES_PATH, 'mipmap-anydpi-v26');
```

packages/config/src/android/Icon.ts
dsokal: I strongly disagree that this is readable. We should change the eslint rules.

packages/image-utils/src/Image.ts
```diff
@@ -35,6 +35,9 @@ async function resizeAsync(imageOptions: ImageOptions): Promise<Buffer> {
     if (imageOptions.removeTransparency) {
       jimp.colorType(2);
     }
+    if (imageOptions.circle) {
+      jimp.circle();
```
cruzach, line 39: would like to do something similar here to how we're doing this same operation with `sharp`, that way this could be a setting to round the corners instead of just cut out a circle, but was running into some mimeType issues trying to create a `Sharp` image. but can look into that soon

EvanBacon: perhaps a borderRadius property would scale better

packages/image-utils/src/Image.ts
```diff
@@ -158,3 +177,31 @@ export async function generateFaviconAsync(
...
+
+/**
+ * Layers the provided foreground image over the provided background image.
+ *
+ * @param foregroundImageBuffer
+ * @param foregroundImageBuffer
+ * @param x pixel offset from the left edge, defaults to 0.
+ * @param y pixel offset from the top edge, defaults to 0.
+ */
+export async function layerImageAsync(
```
EvanBacon, line 189:
```suggestion
export async function compositeImagesAsync(
```

packages/image-utils/src/Image.ts
cruzach: changed it to `borderRadius: number`. Right now if using Jimp any number will just make the image a circle though

packages/config/src/android/Colors.ts
```diff
@@ -56,3 +56,14 @@ export function setColorItem(itemToAdd: XMLItem[], colorFileContentsJSON: Docume
   }
   return colorFileContentsJSON;
 }
+
+export function removeColorItem(keyToRemove: string, colorFileContentsJSON: Document) {
```
EvanBacon, line 60: used?

cruzach: nope not used, we default to setting the background color to white so we don't need this, I'll take it out

PR #2088, opened 2020-05-07
[xdl]  Start Ngrok only when running expo start --tunnel

### LINKED ISSUE #335: Start Ngrok only when running expo start --tunnel
Right now Ngrok will be started when using LAN, Localhost or Tunnel. The only way to avoid Ngrok starting is to use the --offline flag. It would be best to only start Ngrok when using tunnel.

FiberJW:
Fixes #335

### DIFF
--- a/packages/dev-tools/server/graphql/GraphQLSchema.ts
+++ b/packages/dev-tools/server/graphql/GraphQLSchema.ts
@@ -576,7 +576,22 @@ const resolvers = {
     },
     async setProjectSettings(parent, { settings }, context) {
       const currentProject = context.getCurrentProject();
+      let previousSettings = await ProjectSettings.readAsync(currentProject.projectDir);
       let updatedSettings = await ProjectSettings.setAsync(currentProject.projectDir, settings);
+
+      // If 'tunnel' wasn't previously configured and it will be as a result of this request, start tunnels.
+      if (previousSettings.hostType !== 'tunnel' && updatedSettings.hostType === 'tunnel') {
+        try {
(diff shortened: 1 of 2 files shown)


fson:
I think we might have to also start it, if you run `expo start --tunnel` and later enable the tunnel from Expo Dev Tools (https://github.com/expo/expo-cli/blob/d33851b9b109f52ba9903225b634d11258e68050/packages/dev-tools/server/graphql/GraphQLSchema.ts#L577-L584).

packages/dev-tools/server/graphql/GraphQLSchema.ts
```diff
@@ -576,7 +576,22 @@ const resolvers = {
...
       const currentProject = context.getCurrentProject();
+      let previousSettings = await ProjectSettings.readAsync(currentProject.projectDir);
       let updatedSettings = await ProjectSettings.setAsync(currentProject.projectDir, settings);
+
+      // If 'tunnel' wasn't previously configured and it will be as a result of this request, start tunnels.
+      if (previousSettings.hostType !== 'tunnel' && updatedSettings.hostType === 'tunnel') {
+        try {
+          await Project.startTunnelsAsync(currentProject.projectDir);
+        } catch (e) {
+          ProjectUtils.logDebug(
```
fson, line 587:
```suggestion
          ProjectUtils.logWarning(
```
`logDebug` isn't shown to the user and not being able to start the tunnel will result in missing functionality so I we should probably surface it.

PR #2089, opened 2020-05-08
[xdl] expose certData PKCS function, add tests

quinlanj:
# why
im writing something in www that needs the `getCertData` function exposed

# tests

- [x] new ones pass


packages/xdl/src/detach/PKCS12Utils.ts
```diff
@@ -20,9 +20,20 @@ export function findP12CertSerialNumber(
   p12Buffer: Buffer | string,
   passwordRaw: string | null
 ): string | null {
+  const { serialNumber } = getCertData(p12Buffer, passwordRaw);
+  return serialNumber;
+}
+
+function _processSerialNumber(maybeSerialNumber: any) {
```
brentvatne, line 27:
```suggestion
function _processSerialNumber(maybeSerialNumber: string | null | undefined) {
```

maybe this is a more appropriate type?

PR #2117, opened 2020-05-13
[expo-cli] unify and remove duplicated jest configuration

byCedric:
This is an improved version of the jest simplification, just like #1971. It's the next iteration of the previous PR, but with feedback from the PR processed.

Some thoughts:

1. Because we compile everything with typescript now, I think we can remove the `babel.config.js` from most of the packages.
2. It would be great if we can auto-ignore e2e or integration tests in `jest/unit-test-config.js`, and add a `jest/e2e-test-config.js` that only looks in the package's `e2e` folder for tests. That would clean up the custom `roots: [...]` in some projects.


fson:
Would it make sense to add a `yarn test` script in the workspace root `package.json`, to make it easy to run all tests with `yarn test` (or even `yarn test --watch`)? It seems like this would already work.

byCedric:
No worries! 😄 Yes! That definitely works, I tried it with `yarn jest`. It should make it easier and maybe even faster (if you work on multiple packages at the same time). I'll add it!

byCedric:
Done, I'll merge it after the CI checks. That way, @dsokal can reuse this PR 😄

Issue #2109, opened 2020-05-13
build:* warns (fails in --non-interactive) when --config option used


christynwegner:
## Description

We have an app that  allows for different operators to skin, enable features, and to customize their experience via different `app.json` configuration files. These files have been defined in `config/<appName>.json` which we pass into the expo-cli commands via the `--config` option. As such, we have removed the `app.json` from the project root directory.

As of the expo-cli version `3.20.3`, the `expo build:* --config ./config/<appName>.json` commands warn that the app is not using the expo managed workflow; however, the app has never been ejected and thus should be an expo managed workflow app.

Quickly digging into the code base I found that the managed workflow check requires that the `app.json` exist in the project root directory, ignoring the path declared in the `--config` option: https://github.com/expo/expo-cli/blob/expo-cli%403.20.3/packages/expo-cli/src/commands/utils/ProjectUtils.ts#L25

The check that was added in version `3.20.3` is `maybeBailOnWorkflowWarning` https://github.com/expo/expo-cli/blob/expo-cli%403.20.3/packages/expo-cli/src/commands/build/index.ts#L15

The current workaround for this to apply the `--skip-workflow-check` option to both `build:ios` and `build:android` commands

## Expected Behaviour

`expo build:* --config config/path.json` passes the expo managed workflow check using the `app.json` file declared via the `--config` option

## Observed Behaviour

CI/CD build failed due to failing the expo managed workflow check (with `--non-interactive` enabled)
```
[16:42:06] ⚠️  expo build:android currently only supports managed workflow apps.
[16:42:06] If you proceed with this command, we can run the build for you but it will not include any custom native modules or changes that you have made to your local native projects.
[16:42:06] Unless you are sure that you know what you are doing, we recommend aborting the build and doing a native release build through Android Studio.
[16:42:06] Input is required, but Expo CLI is in non-interactive mode.
Required input:
> Would you like to proceed?
```

## Environment

```
$ expo diagnostics

  Expo CLI 3.20.3 environment info:
    System:
      OS: macOS 10.15.3
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 13.10.1 - /usr/local/bin/node
      Yarn: 1.22.4 - /usr/local/bin/yarn
      npm: 6.13.7 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #2119, opened 2020-05-13
fix: Rewrite ProjectUtils.findProjectRootAsync to use @expo/config

brentvatne:
Fixes https://github.com/expo/expo-cli/issues/2109

`findProjectRootAsync` was not respecting the config and `build:*` was not calling `setCustomConfigPath` with the value of the `--config` flag. This resolves both of those issues.

Note: there are other places where we call `findProjectRootAsync`: `install` and `upgrade`. I didn't think it made sense to support the config flag in these places, but I am open to being convinced.

### TODO

- [x] Add test coverage for `findProjectRootAsync`

### DIFF
--- a/packages/expo-cli/src/commands/build/index.ts
+++ b/packages/expo-cli/src/commands/build/index.ts
@@ -1,4 +1,5 @@
 import { UrlUtils, Webpack } from '@expo/xdl';
+import { setCustomConfigPath } from '@expo/config';
 import { Command } from 'commander';
 import chalk from 'chalk';
 import BaseBuilder from './BaseBuilder';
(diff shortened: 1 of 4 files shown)


packages/expo-cli/src/commands/build/index.ts
```diff
@@ -125,8 +149,18 @@ export default function (program: Command) {
     )
     .asyncActionProjectDir(
       async (projectDir: string, options: AndroidOptions) => {
+        if (options.config) {
+          setCustomConfigPath(projectDir, options.config);
```
fson, line 153: This should not be necessary because the universal `--config` option is handled and `setCustomConfigPath` called in `asyncActionProjectDir`.

https://github.com/expo/expo-cli/blob/d1e02e31a4091ec43dd77053758f64c4b83ea00f/packages/expo-cli/src/exp.ts#L131-L137

PR #2132, opened 2020-05-15
[expo-cli] copy metro config in eject

esamelson:
Metro config with the `hashAssetFiles` plugin is now needed for expo-updates in the bare workflow, so we need to add this to projects when ejecting.

We now check all three places metro config may exist (metro.config.js, metro.config.json, and package.json['metro']), and we copy metro.config.js from the template projects if there's no existing config. If there is existing config, we log a warning instead, linking to the instructions/example in the expo-updates README.

- [x] eject with no existing metro config copies metro.config.js
- [x] eject with existing metro config logs a warning

### DIFF
--- a/packages/expo-cli/src/commands/eject/Eject.ts
+++ b/packages/expo-cli/src/commands/eject/Eject.ts
@@ -316,27 +316,33 @@ async function createNativeProjectsFromTemplateAsync(projectRoot: string): Promi
    */
 
   let updatingMetroConfigStep = logNewSection('Adding Metro bundler configuration');
-  if (
-    fse.existsSync(path.join(projectRoot, 'metro.config.js')) ||
-    fse.existsSync(path.join(projectRoot, 'metro.config.json')) ||
-    pkg.metro
-  ) {
+  try {
+    if (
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -309,6 +310,35 @@ async function createNativeProjectsFromTemplateAsync(projectRoot: string): Promi
...
+      `- You will need to add the ${chalk.bold(
+        'hashAssetFiles'
+      )} plugin to your existing Metro configuration. ${terminalLink(
+        'Example.',
+        'https://github.com/expo/expo/blob/master/packages/expo-updates/README.md#metroconfigjs'
+      )}`
+    );
+    log.newLine();
+  } else {
+    fse.copySync(path.join(tempDir, 'metro.config.js'), path.join(projectRoot, 'metro.config.js'));
```
brentvatne, line 338: we might want to re-organize this a bit so that if fse.copySync throws an error for some reason we show the "Metro bundler configuration not applied" msg

Issue #2094, opened 2020-05-09
Turn off watching file changes (Metro) during builds and publishes


brentvatne:
Should be easier to do after we control metro config more closely

Related: https://expo.canny.io/admin/board/feature-requests/p/turn-off-watching-file-changes-metro-during-builds-and-publishes

Issue #277, opened 2018-12-28
Handle ENOSPC error (fs.inotify.max_user_watches reached)


fson:
Expo CLI should catch the `ENOSPC` error (https://github.com/expo/expo/issues/2944) from exceeding the `inotify` limits on Linux and suggest a fix to the user (increase the `fs.inotify.max_user_watches` setting).

The corresponding error in Watchman: https://github.com/facebook/watchman/blob/02d1d89b093bdbacdaa5e37471cef00080954ac5/error_category.cpp#L30-L41

fson:
Unfortunately the best we can do is suggest increasing the amount of watches. Expo CLI can't automatically do this, because changing the setting requires superuser privileges.

You only need to configure the setting once per server, it can be done by running this command:
```
echo fs.inotify.max_user_instances=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
echo fs.inotify.max_queued_events=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```

smithmr:
We are using GitLab CI/CD When we run `expo publish --non-interactive --release-channel develop` We get the same error 
`[06:27:32] Building iOS bundle
[06:27:42] fs.js:1384
[06:27:42]     throw error;
[06:27:42]     ^
[06:27:42] 
[06:27:42] Error: watch /builds/divdot/mobile-app/node_modules/react-native/ReactAndroid/src/main/libraries/fbcore/src/test/java/com/facebook/powermock ENOSPC
[06:27:42]     at FSWatcher.start (fs.js:1382:19)
[06:27:42]     at Object.fs.watch (fs.js:1408:11)
[06:27:42]     at NodeWatcher.watchdir (/builds/divdot/mobile-app/node_modules/sane/src/node_watcher.js:175:20)
[06:27:42]     at Walker.<anonymous> (/builds/divdot/mobile-app/node_modules/sane/src/common.js:116:12)
[06:27:42]     at emitTwo (events.js:126:13)
[06:27:42]     at Walker.emit (events.js:214:7)
[06:27:42]     at /builds/divdot/mobile-app/node_modules/walker/lib/walker.js:69:16
[06:27:42]     at go$readdir$cb (/builds/divdot/mobile-app/node_modules/graceful-fs/graceful-fs.js:162:14)
[...]

PR #2149, opened 2020-05-20
feat: bundle JS without file watching or HTTP

fson:
## Why

Starting file watching sometimes causes errors like running out of `inotify` watches (https://github.com/expo/expo-cli/issues/277), requiring additional configuration settings on CI servers. File watching is unnecessary for a one-off build of the JS bundles, where we don't need to run a development server. And while we're at it, why not also stop running a HTTP server for Metro, because we can just call the JS APIs of Metro. This should require less privileges and consume less CI server resources.

As a bonus, doing this allows us to refactor and get rid of a quite a lot of now unnecessary code doing things like fetching bundles, assets and source maps over HTTP, or checking if a server is already running and trying to re-use a running HTTP server.

Closes https://github.com/expo/expo-cli/issues/2094. There's also a Canny request: https://expo.canny.io/feature-requests/p/turn-off-watching-file-changes-metro-during-builds-and-publishes

## How

Instead of starting Metro with a HTTP server using `Metro.runServer`, we now use the `Metro.runMetro` API to get a `MetroServer` instance and call its `build` and `getAssets` methods to get the JS bundles, source maps and asset maps.

## Testing

We should probably run some kind of diff to check that the code/assets we output from `expo publish` and `expo export` remains unchanged.


packages/dev-server/src/MetroDevServer.ts
```diff
@@ -1,15 +1,35 @@
-import { getConfig, projectHasModule } from '@expo/config';
+import http from 'http';
+import { Platform, getConfig, projectHasModule } from '@expo/config';
 import { createDevServerMiddleware } from '@react-native-community/cli-server-api';
 import Log from '@expo/bunyan';
 import * as ExpoMetroConfig from '@expo/metro-config';
 import bodyParser from 'body-parser';
+import type Metro from 'metro';
```
EvanBacon, line 7: I wasn't aware you could do this in TypeScript.

packages/expo-cli/src/commands/export.ts
```diff
@@ -80,49 +80,15 @@ export async function action(projectDir: string, options: Options) {
     console.warn(`Dev Mode: publicUrl ${options.publicUrl} does not conform to HTTP format.`);
   }
 
-  const target = options.target ?? getDefaultTarget(projectDir);
```
EvanBacon, line 83: I don't see any tests for this command, we should probably add at least one e2e test before merging this so we can verify that the change isn't unexpected.

packages/expo-cli/src/commands/publish.ts
```diff
@@ -147,61 +110,54 @@ export async function action(projectDir: string, options: Options = {}) {
...
-        `The manifest URL is: ${terminalLink(url, url)}. ${terminalLink(
-          'Learn more.',
-          'https://expo.fyi/manifest-url'
-        )}`
-      );
-    }
+  log('Publish complete');
+  log.newLine();
 
-    if (target === 'managed') {
```
EvanBacon, line 182: Again I think this should have at least one test before creating such a hefty change. Because it's somewhat of a long running task it needs to be continuously tested.

packages/xdl/src/Project.ts
```diff
@@ -1160,11 +1029,15 @@ async function _configureExpForAssets(projectRoot: string, exp: ExpoConfig, asse
   return exp;
 }
 
-async function _fetchAndUploadAssetsAsync(projectRoot: string, exp: PublicConfig) {
+async function publishAssetsAsync(
+  projectRoot: string,
+  exp: PublicConfig,
+  bundles: { android: BundleOutput; ios: BundleOutput }
```
EvanBacon, line 1035: Split this type out and reuse it across methods

packages/dev-server/src/MetroDevServer.ts
fson: It was added in TS 3.8. I guess it’s not needed very often, but here it’s invaluable 👍

fson:
I've added back the old implementation that requests bundles from a separate Metro process, so now there are two modes:
- Legacy mode: starts Metro in a separate process (with file watching and a HTTP server) and requests bundles from it with HTTP.
- `EXPO_USE_DEV_SERVER=true`: starts Metro in-process and builds bundles using Metro's JS API, no HTTP or file watching used.

Because the changes are only applied if `EXPO_USE_DEV_SERVER` is set, this should now be safe to land and release anytime, so it can be tested in real projects and then later changed so that it's enabled by default (for SDK37 and up?).

@EvanBacon mind giving this one more look before merging?

brentvatne:
oh no we have so many conflicts now 😱

FLGMwt:
👋 Just tried this out in the new expo-cli release in a github action, the `expo publish` step went from 4m24s to 2m56s. Awesome work!

mymattcarroll:
This is awesome! Thank you so much Expo Team. Couple of things...

Will this be documented here: https://docs.expo.io/workflow/expo-cli/

It's not really clear how to prevent file watches when running `expo publish`. Do you set `EXPO_USE_DEV_SERVER=true` to enable file watches, or to disable file watches? I can understand if you want to keep file watches by default to avoid a breaking changes to the CLI (although, I am not sure if there are any use cases to have files being watched while doing a publish), but the environment variable is not very clear and is undocumented.

Could this be added as a flag to the publish and export commands? e.g. `expo publish --no-watch`?

Also, I think it would be worth checking for the `CI` environment variable to prevent file watches. This is quite common for CLI's that have commands that are expected to be used in CI.

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

brentvatne:
@mymattcarroll - the plan is to always bundle this way in the future. once we're more confident that this works well for people who opt in to using it with `EXPO_USE_DEV_SERVER=true` then we will flip the switch

mymattcarroll:
Sounds great! Again, thank you for the quick turn around on this feature.

PR #2164, opened 2020-05-24
log message suggesting windows users to use other terminal instead of

Fahad-Mahmood:
cmd.exe added


brentvatne:
hi there! does `os.platform() === 'win32'` not return `true` in powershell as well?

Fahad-Mahmood:
> hi there! does `os.platform() === 'win32'` not return `true` in powershell as well?

I don't think this piece of code is executed after adding it to the file you mentioned.

brentvatne:
i don't understand :(

brentvatne:
i'm going to close this because it doesn't detect if the developer is using cmd.exe, it only detects if they are using windows

Issue #1979, opened 2020-04-24
NCC Expo CLI


EvanBacon:
Because TS doesn't minify libraries, we need to minify the end result of our packages which in this case is expo-cli.

fson:
I evaluated this in #2168, but because of the large size of the codebase and its dependencies, I wasn't able to work around the increased time to interaction caused by bundling all this code into a large single file that needs to be parsed before the command does anything (see https://github.com/expo/expo-cli/pull/2168#issuecomment-648221147).

`ncc` currently does *not* appear to have an option for using dynamic imports as code splitting points (similar to https://webpack.js.org/guides/code-splitting/#dynamic-imports), which would have been useful for solving this problem.

`ncc` does support multiple entry points with a shared "common bundle", which could be useful, but moving every command to be a separate entry point would be a major project and possibly require custom code for argument parsing and `--help` to work.

I'm closing this issue for now, but I think it would be worth it to re-evaluate, if ncc at some points gains more capabilities for code-splitting or if we decide to extract smaller parts of Expo CLI into separate tools that we could compile with `ncc`.

PR #2168, opened 2020-05-25
enchancement: build expo-cli with ncc

fson:
https://github.com/expo/expo-cli/issues/1979

### DIFF
--- a/packages/expo-cli/.gitignore
+++ b/packages/expo-cli/.gitignore
@@ -29,4 +29,6 @@ node_modules
 /build
 /expo-cli/build
 
+/dist
+
 dist-bin/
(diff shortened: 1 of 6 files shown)


fson:
This can't be used as is because of the increase in startup/execution time increase it causes over our previous setup with lazy loading of modules (using `@babel/plugin-transform-modules-commonjs` with the `lazy` option enabled.)

For example, time to just get the help page (`--help`) is 3-5x slower using the ncc build:
```
expo --help  0.22s user 0.05s system 109% cpu 0.249 total
./bin/expo.js --help  1.08s user 0.13s system 105% cpu 1.147 total
```

I've tried using dynamic imports to make this faster (so we'd only import the command that's executing), but it seems that the problem might be parsing the ~20MB JS bundle, not evaluating the JS, so it doesn't really help).

PR #2170, opened 2020-05-25
chore: clean up lodash usage

fson:
- Removed every `import _ from 'lodash'` usage and replaced it with individual imports (which is the prevailing convention in this codebase). This way we don't end up bundling/parsing the whole library, when we only use a subset. This makes it also easier to later switch to individual `lodash.*` packages, in cases where doing that makes sense.
- Replaced most uses of `lodash/get` with plain TypeScript, because the return value of `get()` is `any`, which hides critical type errors. (Found a few potential errors this way.)
- Replaced `lodash/concat`, `lodash/every` etc, with plain TypeScript.

### DIFF
--- a/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
+++ b/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
@@ -1,10 +1,8 @@
 import os from 'os';
 import chalk from 'chalk';
 import pickBy from 'lodash/pickBy';
-import get from 'lodash/get';
 import { XDLError } from '@expo/xdl';
 
-import { Dictionary } from 'lodash';
 import terminalLink from 'terminal-link';
 import semver from 'semver';
 import BaseBuilder from '../BaseBuilder';
@@ -39,6 +37,14 @@ import {
 } from '../../../credentials/views/IosProvisioningProfile';
 import { IosAppCredentials, IosDistCredentials } from '../../../credentials/credentials';
 
+function missingBundleIdentifierError() {
+  return new XDLError(
+    'INVALID_OPTIONS',
+    `Your project must have a bundleIdentifier set in app.json.
+See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appjson`
+  );
+}
+
 class IOSBuilder extends BaseBuilder {
   appleCtx?: apple.AppleCtx;
 
@@ -59,17 +65,13 @@ class IOSBuilder extends BaseBuilder {
   }
 
   async validateProject() {
-    const bundleIdentifier = get(this.manifest, 'ios.bundleIdentifier');
+    const bundleIdentifier = this.manifest.ios?.bundleIdentifier;
     const sdkVersion = this.manifest.sdkVersion;
 
     await this.validateIcon();
 
     if (!bundleIdentifier) {
-      throw new XDLError(
-        'INVALID_OPTIONS',
-        `Your project must have a bundleIdentifier set in app.json.
-See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appjson`
-      );
+      throw missingBundleIdentifierError();
     }
     await utils.checkIfSdkIsSupported(sdkVersion!, PLATFORMS.IOS);
   }
@@ -116,7 +118,8 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
     // TODO: Fix forcing the username to be valid
     const username = this.manifest.owner ?? this.user?.username!;
     const experienceName = `@${username}/${this.manifest.slug}`;
-    const bundleIdentifier = get(this.manifest, 'ios.bundleIdentifier');
+    const bundleIdentifier = this.manifest.ios?.bundleIdentifier;
+    if (!bundleIdentifier) throw missingBundleIdentifierError();
     const context = new Context();
     await context.init(this.projectDir);
     await this.bestEffortAppleCtx(context, bundleIdentifier);
@@ -301,7 +304,7 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
     ctx: Context,
     experienceName: string,
     bundleIdentifier: string,
-    credsToClear: Dictionary<boolean>
+    credsToClear: { [name: string]: boolean }
   ): Promise<void> {
     const shouldRevokeOnApple = this.options.revokeCredentials;
     const nonInteractive = this.options.parent && this.options.parent.nonInteractive;
@@ -336,7 +339,7 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
     }
   }
 
-  determineCredentialsToClear(): Dictionary<boolean> {
+  determineCredentialsToClear(): { [name: string]: boolean } {
     const {
       clearCredentials,
       clearDistCert,
@@ -369,7 +372,11 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
   // validates whether the icon doesn't have transparency
   async validateIcon() {
     try {
-      const icon = get(this.manifest, 'ios.icon', this.manifest.icon);
+      const icon = this.manifest.ios?.icon ?? this.manifest.icon;
+      if (!icon) {
+        // icon is optional
+        return;
+      }
       await ensurePNGIsNotTransparent(icon);
     } catch (err) {
       if (err instanceof XDLError) {
--- a/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
+++ b/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
@@ -339,7 +339,7 @@ class IOSBuilder extends BaseBuilder {
     }
   }
 
-  determineCredentialsToClear(): { [name: string]: boolean } {
+  determineCredentialsToClear(): Record<string, boolean> {
     const {
       clearCredentials,
       clearDistCert,
--- a/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
+++ b/packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
@@ -304,7 +304,7 @@ class IOSBuilder extends BaseBuilder {
     ctx: Context,
     experienceName: string,
     bundleIdentifier: string,
-    credsToClear: { [name: string]: boolean }
+    credsToClear: Record<string, boolean>
   ): Promise<void> {
     const shouldRevokeOnApple = this.options.revokeCredentials;
     const nonInteractive = this.options.parent && this.options.parent.nonInteractive;
--- a/packages/expo-cli/src/commands/client/index.ts
+++ b/packages/expo-cli/src/commands/client/index.ts
@@ -2,7 +2,7 @@ import { Android, Simulator, UserManager, Versions } from '@expo/xdl';
 import chalk from 'chalk';
 import CliTable from 'cli-table3';
 import fs from 'fs-extra';
-import _ from 'lodash';
+import has from 'lodash/has';
 import ora from 'ora';
 import path from 'path';
 import { Command } from 'commander';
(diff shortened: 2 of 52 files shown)


packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -336,7 +339,7 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
     }
   }
 
-  determineCredentialsToClear(): Dictionary<boolean> {
+  determineCredentialsToClear(): { [name: string]: boolean } {
```
EvanBacon, line 342:
```suggestion
  determineCredentialsToClear(): Record<string, boolean> {
```
nit

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -301,7 +304,7 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
     ctx: Context,
     experienceName: string,
     bundleIdentifier: string,
-    credsToClear: Dictionary<boolean>
+    credsToClear: { [name: string]: boolean }
```
EvanBacon, line 307:
```suggestion
    credsToClear: Record<string, boolean>
```

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -61,19 +61,19 @@ export default function (program: Command) {
         }
         if (!exp.ios) exp.ios = {};
 
-        if (!_.has(exp, 'facebookAppId') || !_.has(exp, 'facebookScheme')) {
+        if (!has(exp, 'facebookAppId') || !has(exp, 'facebookScheme')) {
```
EvanBacon, line 64:
```suggestion
        if (!exp.facebookAppId || !exp.facebookScheme) {
```
Maybe something like this?

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -61,19 +61,19 @@ export default function (program: Command) {
...
           const disabledReason = exp
             ? `facebookAppId or facebookScheme are missing from app configuration. `
             : 'No custom configuration file could be found. You will need to provide a json file with valid facebookAppId and facebookScheme fields.';
           disabledServices.facebookLogin = { name: 'Facebook Login', reason: disabledReason };
         }
-        if (!_.has(exp, 'ios.config.googleMapsApiKey')) {
+        if (!has(exp, 'ios.config.googleMapsApiKey')) {
```
EvanBacon, line 70:
```suggestion
        if (!exp.ios?.config?.googleMapsApiKey) {
```
Maybe?

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -61,19 +61,19 @@ export default function (program: Command) {
...
           const disabledReason = exp
             ? `ios.config.googleMapsApiKey does not exist in the app configuration.`
             : 'No custom configuration file could be found. You will need to provide a json file with a valid ios.config.googleMapsApiKey field.';
           disabledServices.googleMaps = { name: 'Google Maps', reason: disabledReason };
         }
-        if (_.has(exp, 'ios.googleServicesFile')) {
+        if (has(exp, 'ios.googleServicesFile')) {
```
EvanBacon, line 76:
```suggestion
        if (exp.ios.googleServicesFile) {
```

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -2,7 +2,7 @@ import { Android, Simulator, UserManager, Versions } from '@expo/xdl';
 import chalk from 'chalk';
 import CliTable from 'cli-table3';
 import fs from 'fs-extra';
-import _ from 'lodash';
+import has from 'lodash/has';
```
EvanBacon, line 5: Prolly don't need this

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -494,11 +496,11 @@ export async function upgradeAsync(
   let updates = await getUpdatedDependenciesAsync(projectRoot, workflow, targetSdkVersion);
 
   // Split updated packages by dependencies and devDependencies
-  let devDependencies = _.pickBy(updates, (_version, name) => _.has(pkg.devDependencies, name));
+  let devDependencies = pickBy(updates, (_version, name) => has(pkg.devDependencies, name));
   let devDependenciesAsStringArray = Object.keys(devDependencies).map(
     name => `${name}@${updates[name]}`
   );
-  let dependencies = _.pickBy(updates, (_version, name) => _.has(pkg.dependencies, name));
+  let dependencies = pickBy(updates, (_version, name) => has(pkg.dependencies, name));
```
EvanBacon, line 503:
```suggestion
  let dependencies = pickBy(updates, (_version, name) => pkg.dependencies[name]);
```
or 
```suggestion
  let dependencies = pickBy(updates, (_version, name) => name in pkg.dependencies);
```

packages/expo-cli/src/credentials/views/IosAppCredentials.ts
```diff
@@ -1,12 +1,4 @@
-import chalk from 'chalk';
-import get from 'lodash/get';
-
 import { Context, IView } from '../context';
-import { IosAppCredentials, IosCredentials, IosPushCredentials } from '../credentials';
-import { askForUserProvided } from '../actions/promptForCredentials';
-import { displayIosUserCredentials } from '../actions/list';
-import prompt from '../../prompt';
-import log from '../../log';
```
EvanBacon, line 9: dang

packages/expo-cli/src/credentials/views/Select.ts
```diff
@@ -135,7 +134,7 @@ export class SelectAndroidExperience implements IView {
     this.askAboutProjectMode = false;
 
     if (this.androidCredentials.length === 0) {
-      this.androidCredentials = get(await ctx.api.getAsync('credentials/android'), 'credentials');
+      this.androidCredentials = (await ctx.api.getAsync('credentials/android')).credentials;
```
EvanBacon, line 137:
```suggestion
      this.androidCredentials = (await ctx.api.getAsync('credentials/android'))?.credentials;
```

Maybe? Just guessing from the `get` command.

packages/xdl/src/Android.ts
```diff
@@ -609,7 +613,7 @@ See https://docs.expo.io/guides/splash-screens/#splash-screen-api-limitations-on
     return;
   }
 
-  if (_.some(androidSplashImages, ({ sizeMatches }) => !sizeMatches)) {
+  if (some(androidSplashImages, ({ sizeMatches }) => !sizeMatches)) {
```
EvanBacon, line 616:
```suggestion
  if (androidSplashImages.some({ sizeMatches }) => !sizeMatches)) {
```
maybe use built-in.

packages/xdl/src/detach/AndroidIntentFilters.ts
```diff
@@ -26,7 +26,7 @@ export default function renderIntentFilters(intentFilters: any) {
 }
 
 function renderIntentFilterDatumEntries(datum: any) {
-  return _.toPairs(datum)
+  return toPairs(datum)
```
EvanBacon, line 29:
```suggestion
  return Object.entries(datum)
```
Hard to tell from the `any` types, but prolly can use `Object.entries`.

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -132,9 +132,9 @@ function getRemoteOrLocalUrl(manifest, key, isDetached) {
   // in detached apps, `manifest` is actually just app.json, so there are no remote url fields
   // we should return a local url starting with file:// instead
   if (isDetached) {
-    return _.get(manifest, key);
+    return get(manifest, key);
```
EvanBacon, line 135:
```suggestion
    return manifest?.[key];
```

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -132,9 +132,9 @@ function getRemoteOrLocalUrl(manifest, key, isDetached) {
...
   }
-  return _.get(manifest, `${key}Url`);
+  return get(manifest, `${key}Url`);
```
EvanBacon, line 137:
```suggestion
  return manifest?.[`${key}Url`];
```

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -719,7 +715,7 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
   }
 
   // Add app-specific intent filters
-  const intentFilters = _.get(manifest, 'android.intentFilters');
+  const intentFilters = get(manifest, 'android.intentFilters');
```
EvanBacon, line 718:
```suggestion
  const intentFilters = manifest?.android?.intentFilters;
```

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1355,7 +1351,7 @@ async function buildShellAppAsync(context, sdkVersion, buildType, buildMode) {
     });
     await fs.copy(
       outputPath,
-      _.get(context, 'build.android.outputFile') || `/tmp/shell-debug.${ext}`
+      get(context, 'build.android.outputFile') || `/tmp/shell-debug.${ext}`
```
EvanBacon, line 1354:
```suggestion
      context?.build?.android?.outputFile || `/tmp/shell-debug.${ext}`
```

packages/xdl/src/detach/ExponentTools.ts
```diff
@@ -125,8 +126,8 @@ async function spawnAsync(
 
 function createSpawner(buildPhase: string, logger?: Logger) {
   return (command: string, ...args: any[]) => {
-    const lastArg = _.last(args);
-    const optionsFromArg = _.isObject(lastArg) ? args.pop() : {};
+    const lastArg = last(args);
```
EvanBacon, line 129:
```suggestion
    const { length, [length - 1]: lastArg } = args;
```
or 

```js
const lastArg = args[args.length - 1];
```

packages/xdl/src/detach/IosIPABuilder.js
```diff
@@ -124,8 +125,8 @@ export default function createIPABuilder(buildParams) {
       }
     );
     const plistRaw = output.join('');
-    const plistData = _.attempt(plist.parse, plistRaw);
-    if (_.isError(plistData)) {
+    const plistData = attempt(plist.parse, plistRaw);
```
EvanBacon, line 128:
```suggestion
    const plistData = typeof plist.parse === 'function' ? plist.parse(plistRaw) : undefined;
```
Something like this?

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
fson: sure

packages/expo-cli/src/credentials/views/Select.ts
fson: Good call. The return value was `any` so it was masking the error 👿

fson: Although I'm still not sure if it's necessary... might just push the error down the line ...

packages/xdl/src/Android.ts
fson: Thanks, missed this earlier.

packages/xdl/src/detach/AndroidShellApp.js
fson: Yeah, I though of this too, but then realized this file is JS 😮 🙀

packages/xdl/src/detach/AndroidShellApp.js
fson: This is JS

packages/xdl/src/detach/AndroidShellApp.js
EvanBacon: ☠️

packages/xdl/src/detach/IosIPABuilder.js
fson: Went with 
```ts
    try {
      return plist.parse(plistRaw);
    } catch (error) {
      throw new Error(`Error when parsing plist: ${error.message}`);
    }
```
`attempt` followed by `isError` must be the weirdest way to catch errors I've seen this week 😄

PR #2171, opened 2020-05-25
Remove fs extra from package-manager

EvanBacon:
Due to the varying versions of fs-extra, the net bundle size in an NCC'd package is larger https://github.com/expo/create-react-native-app/pull/820 notice that multiple copies are included (each ~25 Kb). This PR reuses `expo/json-file` which is always at the latest in expo packages.


packages/json-file/src/JsonFile.ts
```diff
@@ -10,7 +10,7 @@ import { codeFrameColumns } from '@babel/code-frame';
 
 import JsonFileError from './JsonFileError';
 
-const readFileAsync = promisify(fs.readFile);
+const readFileAsync = promisify(readFile);
```
fson, line 13: I wish we could use `fs.promises`, but in node 10.x.x accessing the property produces a `ExperimentalWarning: The fs.promises API is experimental`.

fson:
In any case we should ensure we only use one version of `fs-extra` within the codebase.

PR #2181, opened 2020-05-26
improvement: wrap osascript.execAsync in try catch

### LINKED ISSUE #1597: Allow for expo start -i in CI environments
Thanks for the project, I really appreciate the work that has been done in the last few years on expo!

## Description

I want to be able to run automated integration tests in my CI environment (GH actions for me). I ran into an issue where the iOS app cannot open because the expo-cli uses applescript commands that will prompt for permissions. Apple does not allow these types of permissions to be set in the terminal.

## Expected Behavior

I would expect the iOS app to be able to open on a macOS build runner.

## Observed Behavior

When running `osascript -e 'tell app "System Events" to count processes whose name is "Simulator"'`, Apple prompts the user to giver permissions like this:

Screen Shot 2020-02-26 at 13 14 36

In a CI environment, this will time out after a few minutes with the error:

```
execution error: System Events got an error: AppleEvent timed out. (-1712)
```
I looked initially to set the permissions through the terminal but Apple's permissions util `tccutil` only allows clearing perms. In this case `System Events` needs to be given permission to control `Simulator`. 

## Reproducible Demo

This is specific to GH actions, but using this `yml` shows the timeout after two minutes.

```
name: Applescript Test

on: push

jobs:
  applescript-test:
    name: Applescript test
    runs-on: macos-latest
    steps:
      - uses: actions/checkout@v1
      - run: osascript -e 'tell app "System Events" to count processes whose name is "Simulator"'
```

I am not sure of the best way to support this but it is clear we won't be able to run the iOS expo app in CI without updating these scripts: https://github.com/expo/expo-cli/blob/b810269440bfc4a1bd28dd70e53f6741086ac1f6/packages/osascript/src/index.ts

watchinharrison:
fix #1597

on exception, return false from _isSimulatorRunningAsync
to allow script to continue running on timeout in CI env


EvanBacon:
Closing since this was implemented with multiple simulator support

PR #2184, opened 2020-05-26
[xdl] Deprecate Web module in favor of Webpack module

EvanBacon:
- Don't need two modules that basically do the same stuff. 
- Merge into one and simplify the exports of the remaining one.


packages/expo-cli/src/commands/start.ts
```diff
@@ -2,13 +2,13 @@ import {
   ExpoConfig,
   PackageJSONConfig,
   getConfig,
-  getProjectConfigDescription,
+  // getProjectConfigDescription,
```
fson, line 5:
```suggestion
```
maybe?

Issue #2172, opened 2020-05-25
Remove util.promisify


EvanBacon:
It doesn't seem like we still need this package. https://www.npmjs.com/package/util.promisify

- From `@expo/json-file`

PR #2185, opened 2020-05-26
chore: update dependencies

fson:
- `glob` dependency was missing in `@expo/config` (fixes https://github.com/expo/expo-cli/issues/2185)
- `plist` was unused in `@expo/config`
- a few other unused deps like the `util.promisify` package that was needed prior to Node 8.0.0 (fixes https://github.com/expo/expo-cli/issues/2172)
- added some missing dependencies – these sometimes go unnoticed because another package in the workspace pulls them in
- remove `stream` and `url` (browser copies of Node modules, these were possibly previously used in XDE)


fson:
@brentvatne what was the issue?

fson:
I reverted the code changes related to `glob`/`globby`, so that this can be merged. Hopefully @brentvatne can clarify why https://github.com/expo/expo-cli/pull/2037 had to be reverted in 4bdb81f, so I can sort that out in a follow up PR!

PR #2201, opened 2020-05-29
[expo-cli][eject] Reduce dead code, improve testing

EvanBacon:
- Currently doctor will check if node modules are installed, and throw a full error if they aren't. This will skip that and log a formatted error message if the SDK Version cannot be inferred from the Expo config.
- disable "making sure project is setup correctly..." - doctor check with various random checks
- remove extra package.json writing
- remove jetifier stuff which never gets invoked
- test that gitignores are merged


packages/expo-cli/src/commands/eject.ts
```diff
@@ -19,7 +20,15 @@ async function action(
   projectDir: string,
   options: LegacyEject.EjectAsyncOptions | Eject.EjectAsyncOptions
 ) {
-  let { exp } = getConfig(projectDir);
+  let exp: ExpoConfig;
+  try {
+    exp = getConfig(projectDir).exp;
```
EvanBacon, line 25: This is a duplicate error message :[

Issue #1983, opened 2020-04-24
Migrate from inquirer to prompts


EvanBacon:
Size difference
- [prompts](https://packagephobia.now.sh/result?p=prompts)
- [inquirer](https://packagephobia.now.sh/result?p=inquirer)

PR #2209, opened 2020-06-01
[cli] Replace inquirer with prompts in the AndroidBuilder

byCedric:
~~If you use the `rawlist`, name or short names aren't displayed. Instead, it uses the value of the question. In this case, it would be either `true` or `false`, which is not a clear definition of what's chosen. This switches to the `list` to show the user what his selection is, without modifying the actual returned value. I double-checked if there weren't any other `rawlist`-questions, but there aren't~~ 😄

~~See https://github.com/SBoudrias/Inquirer.js/issues/889~~

See comments 😄

### DIFF
--- a/packages/expo-cli/src/commands/build/AndroidBuilder.ts
+++ b/packages/expo-cli/src/commands/build/AndroidBuilder.ts
@@ -1,5 +1,4 @@
 import { Android } from '@expo/xdl';
-import pick from 'lodash/pick';
 
 import { Context } from '../../credentials';
 import { runCredentialsManager } from '../../credentials/route';
@@ -11,8 +10,6 @@ import {
 import { SetupAndroidKeystore } from '../../credentials/views/SetupAndroidKeystore';
 import BuildError from './BuildError';
 import BaseBuilder from './BaseBuilder';
-import prompt, { Question } from '../../prompt';
-import log from '../../log';
 import * as utils from './utils';
 import { PLATFORMS, Platform } from './constants';
 
--- a/packages/expo-cli/src/prompts.ts
+++ b/packages/expo-cli/src/prompts.ts
@@ -1,5 +1,5 @@
 import program from 'commander';
-import prompts, { Choice, PromptType, PromptObject as Question } from 'prompts';
+import prompts, { Choice, Options, PromptType, PromptObject as Question } from 'prompts';
 
 import CommandError from './CommandError';
 
@@ -7,7 +7,7 @@ export { PromptType, Question };
 
 export default function prompt(
   questions: Question | Question[],
-  { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
+  { nonInteractiveHelp, ...options }: { nonInteractiveHelp?: string } & Options = {}
 ) {
   questions = Array.isArray(questions) ? questions : [questions];
   if (program.nonInteractive && questions.length !== 0) {
@@ -29,6 +29,7 @@ export default function prompt(
     onCancel() {
       throw new CommandError('ABORTED');
     },
+    ...options,
   });
 }
 
--- a/packages/expo-cli/src/prompts.ts
+++ b/packages/expo-cli/src/prompts.ts
@@ -1,0 +1,34 @@
+import program from 'commander';
+import prompts, { Choice, PromptObject, PromptType } from 'prompts';
+
+import CommandError from './CommandError';
+
+export { PromptType, PromptObject };
+
+type CliPrompts = PromptObject | PromptObject[];
+
+export default function prompt(
+  cliPrompts: CliPrompts,
+  { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
+) {
+  const nPrompts = Array.isArray(cliPrompts) ? cliPrompts.length : 1;
+  if (program.nonInteractive && nPrompts !== 0) {
+    let message = `Input is required, but Expo CLI is in non-interactive mode.\n`;
+    if (nonInteractiveHelp) {
+      message += nonInteractiveHelp;
+    } else {
+      const question: any = Array.isArray(cliPrompts) ? cliPrompts[0] : cliPrompts;
+      message += `Required input:\n${(question.message || '').trim().replace(/^/gm, '> ')}`;
+    }
+    throw new CommandError('NON_INTERACTIVE', message);
+  }
+  return prompts(cliPrompts, {
+    onCancel() {
+      throw new CommandError('ABORTED');
+    },
+  });
+}
+
+// todo: replace this workaround, its still selectable by the cursor
+// see: https://github.com/terkelg/prompts/issues/254
+prompt.separator = (title: string): Choice => ({ title, disable: true, value: undefined });
--- a/packages/expo-cli/src/prompts.ts
+++ b/packages/expo-cli/src/prompts.ts
@@ -1,28 +1,31 @@
 import program from 'commander';
-import prompts, { Choice, PromptObject, PromptType } from 'prompts';
+import prompts, { Choice, PromptType, PromptObject as Question } from 'prompts';
 
 import CommandError from './CommandError';
 
-export { PromptType, PromptObject as Question };
-
-type CliPrompts = PromptObject | PromptObject[];
+export { PromptType, Question };
 
 export default function prompt(
   questions: Question | Question[],
   { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
 ) {
   questions = Array.isArray(questions) ? questions : [questions];
-  if (program.nonInteractive && nPrompts !== 0) {
+  if (program.nonInteractive && questions.length !== 0) {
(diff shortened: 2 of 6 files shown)


fson:
We're slowly migrating from `inquirer` to `prompts` (https://github.com/expo/expo-cli/issues/1983). Would you mind updating this particular prompt to use [`prompts](https://github.com/terkelg/prompts) (if it's not too much trouble). Since we're changing this code anyway and doing so we'll to have test the changes, it might be good to switch over at the same time?

Feel free to create a copy of https://github.com/expo/expo-cli/blob/5d29fe67353a54b0257b5a5b013428c24275f252/packages/expo-cli/src/prompt.ts that uses `prompts` so we can switch modules over to it one by one.

byCedric:
Of course! I'll get on this after standup!

byCedric:
Ok, I wrote up the first draft and made some screenshots. I'll comment on some specific code to highlight it and provide more context.

<details>
<summary>Simple list state (question 1 - how to manage keystore)</summary>
upload or generate?
</details>

<details>
<summary>Selected list state (question 1 - how to manage keystore)</summary>
let expo handle it
</details>

<details>
<summary>Validation error input state (question 2 - path to keystore)</summary>
invalid path to keystore
</details>

<details>
<summary>Question aborted state (question 1 - how to manage keystore)</summary>
canceled
</details>

packages/expo-cli/src/commands/build/AndroidBuilder.ts
```diff
@@ -118,58 +118,50 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
...
           choices: [
-            { name: 'Let Expo handle the process!', value: false },
-            { name: 'I want to upload my own keystore!', value: true },
+            { title: 'Let Expo handle the process!', value: false },
+            { title: 'I want to upload my own keystore!', value: true },
           ],
         },
         {
-          type: 'input',
+          type: (prev, values) => values.uploadKeystore && 'text',
```
byCedric, line 132: This conditional type is the replacement of the `when` method. ([see this example](https://github.com/terkelg/prompts/blob/master/example.js#L33))

packages/expo-cli/src/commands/build/AndroidBuilder.ts
```diff
@@ -118,58 +118,50 @@ See https://docs.expo.io/distribution/building-standalone-apps/#2-configure-appj
...
-          validate: async (keystorePath: string): Promise<boolean> => {
+          validate: async (keystorePath: string): Promise<boolean | string> => {
             try {
               const keystorePathStats = await fs.stat(keystorePath);
               return keystorePathStats.isFile();
             } catch (e) {
-              // file does not exist
-              console.log('\nFile does not exist.');
-              return false;
+              return 'File does not exist.';
```
byCedric, line 140: - Returning `true` passes validation
- Returning `false` shows the default error message
- Returning any `string` shows the string as the error message

packages/expo-cli/src/prompts.ts
```diff
@@ -0,0 +1,34 @@
...
+  return prompts(cliPrompts, {
+    onCancel() {
+      throw new CommandError('ABORTED');
+    },
+  });
+}
+
+// todo: replace this workaround, its still selectable by the cursor
+// see: https://github.com/terkelg/prompts/issues/254
+prompt.separator = (title: string): Choice => ({ title, disable: true, value: undefined });
```
byCedric, line 34: As stated above, there is no separator just yet. They are working on a [plugin API to make this happen](https://github.com/terkelg/prompts/issues/254). For now, I'd figure this kinda brings the same functionality but the cursor will select it. (If you select it, a warning appears with "this answer is disabled")

packages/expo-cli/src/prompts.ts
```diff
@@ -0,0 +1,34 @@
...
+      message += nonInteractiveHelp;
+    } else {
+      const question: any = Array.isArray(cliPrompts) ? cliPrompts[0] : cliPrompts;
+      message += `Required input:\n${(question.message || '').trim().replace(/^/gm, '> ')}`;
+    }
+    throw new CommandError('NON_INTERACTIVE', message);
+  }
+  return prompts(cliPrompts, {
+    onCancel() {
+      throw new CommandError('ABORTED');
```
byCedric, line 27: I had to implement this by default. `prompts` allows the user to "cancel" input and continue the execution. This is not how all code was built, adding this would make refactoring _a lot easier_. After the initial refactoring, we can do another iteration to remove this ofc!

packages/expo-cli/src/commands/build/AndroidBuilder.ts
byCedric: Maybe I should use `answers`, to clarify this a bit more?

byCedric:
If you want I can help refactor the rest. This should be a good start for the AndroidBuilder 😄

packages/expo-cli/src/commands/build/AndroidBuilder.ts
fson:
```suggestion
          type: (prev, answers) => {
            if (answers.uploadKeystore) {
              return 'text';
            } else {
              return false;
            }
          }
```
Maybe?

packages/expo-cli/src/prompts.ts
fson: I think this makes sense. If some code requires cancellation, it can either catch this error (can add a `code` property it then to make it distinguishable from other errors) or we can allow passing `onCancel` to this function.

Either way, I think this is fine for now!

packages/expo-cli/src/prompts.ts
fson: Nice trick 😄

packages/expo-cli/src/prompts.ts
```diff
@@ -0,0 +1,34 @@
+import program from 'commander';
+import prompts, { Choice, PromptObject, PromptType } from 'prompts';
+
+import CommandError from './CommandError';
+
+export { PromptType, PromptObject };
```
fson, line 6:
```suggestion
export { PromptType, PromptObject as Question };
```
Maybe to make it easier to refactor? Plus the name `PromptObject` doesn't come from the original library (they're called ["questions"](https://github.com/terkelg/prompts/blob/master/lib/index.js#L10) in `prompts` source code), but from `@types/prompts` and the `Object` suffix is redundant.

packages/expo-cli/src/prompts.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type CliPrompts = PromptObject | PromptObject[];
+
+export default function prompt(
+  cliPrompts: CliPrompts,
+  { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
+) {
+  const nPrompts = Array.isArray(cliPrompts) ? cliPrompts.length : 1;
```
fson, line 14:
```suggestion
  questions: Question | Question[],
  { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
) {
  questions = Array.isArray(questions) ? questions : [questions];
```

packages/expo-cli/src/prompts.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (program.nonInteractive && nPrompts !== 0) {
+    let message = `Input is required, but Expo CLI is in non-interactive mode.\n`;
+    if (nonInteractiveHelp) {
+      message += nonInteractiveHelp;
+    } else {
+      const question: any = Array.isArray(cliPrompts) ? cliPrompts[0] : cliPrompts;
```
fson, line 20:
```suggestion
      const question = questions[0];
```

packages/expo-cli/src/prompts.ts
byCedric: Ah, gotcha, sure I'll export them with the existing names 😄

packages/expo-cli/src/commands/build/AndroidBuilder.ts
byCedric: Yes, that's also a good one. But I'm afraid this might cause a lot of additional code. Going from 1 line, to mark something as optional, to 7 might be a bit much if we have to do this for everything. We can also go for an inline approach using the same explicit return values?

```
type: (prev, answers) => (values.uploadKeystore ? 'text' : false)
```

packages/expo-cli/src/prompts.ts
byCedric: Yeah, I think adding the prompt's `Option` interface would be great. With that, you can overwrite the default `onCancel` behavior by simply using `prompts([], { onCancel() { } });` If this is too verbose, we could also add a `abortOnCancel: false` option?

packages/expo-cli/src/prompts.ts
fson: sounds good to me

packages/expo-cli/src/prompts.ts
fson: `onCancel` seems fine and is more flexible

PR #2214, opened 2020-06-02
Add a deprecation notice for expo generate-module

EvanBacon:
We discussed this internally on slack. This will help make the `--help` command more readable, and generally make expo-cli lighter.

Screen Shot 2020-06-02 at 2 15 24 PM

### DIFF
--- a/packages/expo-cli/src/commands/generate-module.ts
+++ b/packages/expo-cli/src/commands/generate-module.ts
@@ -1,8 +1,10 @@
-import { Command } from 'commander';
+import { Command, CommandOptions } from 'commander';
+import chalk from 'chalk';
+import boxen from 'boxen';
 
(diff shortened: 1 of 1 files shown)


fson:
We may need to do something about this first: https://github.com/expo/expo/blob/14d9d35beeff6eaaa26f56cb3ba3973decb707db/tools/expotools/src/commands/CreateUnimodule.ts#L18-L21

fson:
I think it might make sense to extract `generate-module` into a standalone tool, similar to `@expo/configure-splash-screen`, that you can run with `npx @expo/generate-module`. @bbarthec @brentvatne any thoughts on this?

bbarthec:
@fson, I like this idea! 👍
But, beforehand we need to set up some convention for this kind of scenario I believe 🤔

PR #2217, opened 2020-06-02
[expo-cli] Updated syntax for send command

EvanBacon:
- Updated syntax to modern

### DIFF
--- a/packages/expo-cli/src/askUser.ts
+++ b/packages/expo-cli/src/askUser.ts
@@ -2,21 +2,20 @@ import { UserSettings } from '@expo/xdl';
 import prompt from './prompt';
 
 async function askForSendToAsync(): Promise<string> {
-  var sendToFromSettings = await UserSettings.getAsync('sendTo', null);
+  const cachedValue = await UserSettings.getAsync('sendTo', null);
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/askUser.ts
```diff
@@ -26,7 +25,7 @@ export default {
 };
 
 if (require.main === module) {
-  askForSendToAsync().then(function (sendTo) {
+  askForSendToAsync().then(function(sendTo) {
     console.log('Your email address is', sendTo);
   });
 }
```
fson, line 31: Maybe remove this part? It seems to enable something like running `node ./build/askUser.js` to manually test this module, but I don't believe anyone is actually doing this 😅

packages/expo-cli/src/askUser.ts
```diff
@@ -2,21 +2,20 @@ import { UserSettings } from '@expo/xdl';
 import prompt from './prompt';
 
 async function askForSendToAsync(): Promise<string> {
-  var sendToFromSettings = await UserSettings.getAsync('sendTo', null);
+  const cachedValue = await UserSettings.getAsync('sendTo', null);
```
fson, line 5: 👍

PR #2218, opened 2020-06-03
refactor: unify glob usage

fson:
Follow up to https://github.com/expo/expo-cli/pull/2185.

Unify `glob` usage across the codebase.

- Use `glob` not `globby`, it's smaller and mostly sufficient for our needs.
  * `globby`: 20 dependencies, 1.2 MB installation size
  * `glob`: 10 dependencies, 296 KB installation size
- Single exception to the above: the `expo-codemod` standalone package still uses `globby` because it makes use of the `gitignore: true` option (excludes gitignored files) of `globby` and passes multiple patterns.
- Get rid of all `globSync(path.join(root, pattern))` kind of calls in favor of `globSync(pattern, { cwd: root })`, as these can cause problems on Windows. ([`glob` documentation specifically asks to only use forward slashes in patterns](https://github.com/isaacs/node-glob#windows) and `path.join` can return backslashes.) Generally favor `cwd` over patterns with absolute paths.


EvanBacon:
From @brentvatne:

We needed to revert https://github.com/expo/expo-cli/commit/ae0de425ba81b227ae8ecb4eb2c10d0da40a5ede

It breaks this https://github.com/expo/expo-cli/blob/495eaa122c97b94d8918f0f55b30aa9f9fe3bc38/packages/config/src/ios/utils/Xcodeproj.ts#L14-L17

projectRoot: `/Users/brentvatne/code/myapplol`
path that is being globbed: `/Users/brentvatne/code/myapplol/ios/*/AppDelegate.m`
result: `[]`

I imagine it needs to be `/**` but I reverted because it makes me think there might be breaking changes elsewhere

fson:
@EvanBacon I'd be very careful about changing the glob to `'ios/**/AppDelegate.m`, because that will end up looking in lots of unnecessary folders, including the `Pods` folder etc. I tried running ` glob.sync('ios/**/AppDelegate.m')` in the `expo/expo` repository root and it constantly took more than 5000ms(!). (`glob.sync('ios/*/AppDelegate.m')` finished in ~5ms.)

fson:
Fixed a significant error in the original set of changes: when replacing `globSync(path.join(cwd, 'pattern'))` with `globSync('pattern', { cwd })`, it is necessary to use `absolute: true`, or else you get relative paths. I had missed this at first, but now all the calls use it in cases where an absolute path was used before.

PR #2237, opened 2020-06-05
chore: Add e2e tests for `expo export`

fson:


fson:
The test failure seems unrelated, export-test.ts passed.

PR #2247, opened 2020-06-10
fix: remove opt-in-google-play-signing

wkozyra95:
- there is only very small number of users that could use it (apps from before 2017 that didn't migrate up until now)
- there are cases were people with enabled app signing ran this command and replaced their credentials with new (invalid ones) and needed to contact google support to reset it
- alternative to this command is following google instructions and uploading new keystore to expo when done: https://docs.expo.io/distribution/app-signing/#android


fson:
I updated the description so that anyone looking at the PR can see why it was removed (without access to the Slack message).

@wkozyra95 could you also make sure that the documentation still referring to this command at https://docs.expo.io/distribution/app-signing/#android gets updated?

PR #2249, opened 2020-06-10
[expo-cli] configure expo-updates plist/xml values in expo init

esamelson:
Currently if you want to try out OTA updates in a new bare project, this is what likely happens

- expo init bare
- make release build
- expo publish
- relaunch app, OTA update doesn't load ???

This is because expo-updates depends on some native configuration values in Expo.plist/AndroidManifest.xml. `expo publish` sets these values automatically, but in the flow above, by the time the user runs `expo publish` they have already made a release build without the necessary values.

This PR adds configuring these fields with best-guess values on `expo init`. We need the SDK version and project URL; we can get the SDK version from the project config, and we can make a best guess of the URL by combining the project slug with the logged-in user's username. We also show a message warning the user they'll need to change these values if they publish under a different user (although `expo publish` will do this automatically), or letting them know if we can't configure these values at all for some reason.

Todo next:
- same thing in eject
- configure more values from app.json `updates` field in eject, if they exist

### DIFF
--- a/packages/expo-cli/src/commands/init.ts
+++ b/packages/expo-cli/src/commands/init.ts
@@ -227,10 +227,14 @@ async function action(projectDir: string, command: Command) {
     } catch (_) {}
 
     let didConfigureUpdatesProjectFiles = false;
-    const user = await UserManager.getCurrentUserAsync();
-    if (user) {
+    const username = await UserManager.getCurrentUsernameAsync();
+    if (username) {
       try {
-        await configureUpdatesProjectFilesAsync(projectPath, initialConfig as BareAppConfig, user);
+        await configureUpdatesProjectFilesAsync(
+          projectPath,
+          initialConfig as BareAppConfig,
+          username
+        );
         didConfigureUpdatesProjectFiles = true;
       } catch {}
     }
@@ -243,7 +247,7 @@ async function action(projectDir: string, command: Command) {
       workflow: 'bare',
       showPublishBeforeBuildWarning,
       didConfigureUpdatesProjectFiles,
-      username: user?.username,
+      username,
     });
     if (!podsInstalled && process.platform === 'darwin') {
       log.newLine();
@@ -274,7 +278,7 @@ function logProjectReadyAsync({
   workflow: 'managed' | 'bare';
   showPublishBeforeBuildWarning?: boolean;
   didConfigureUpdatesProjectFiles?: boolean;
-  username?: string;
+  username?: string | null;
 }) {
   log.nested(chalk.bold(`✅ Your project is ready!`));
   log.newLine();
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/init.ts
```diff
@@ -324,6 +355,54 @@ function logProjectReadyAsync({
...
+  await maybeConfigureUpdatesProjectFileAsync(expoPlistPath, appUrl, sdkVersion);
+  await maybeConfigureUpdatesProjectFileAsync(androidManifestPath, appUrl, sdkVersion);
+}
+
+async function maybeConfigureUpdatesProjectFileAsync(
+  path: string,
+  appUrl: string,
+  sdkVersion: string
+) {
+  if (fs.existsSync(path)) {
```
fson, line 395: It's best to not check for file existence before reading, because this is not concurrency-safe.

The recommended way is to simply try reading it and then handle the specific error:
```ts
let fileContents;
try {
  fileContents = fs.readFileSync('projects.txt', 'utf8');
} catch (error) {
  if (error.code === 'ENOENT') return; // File doesn't exist
  else throw error;
}
fs.writeFileSync(...);
```

esamelson:
Thank you for the super informative feedback @fson !!

brentvatne:
could we use the implementation from `@expo/config` here instead?

packages/expo-cli/src/commands/init.ts
```diff
@@ -324,6 +359,46 @@ function logProjectReadyAsync({
...
+async function configureUpdatesProjectFilesAsync(
+  projectRoot: string,
+  initialConfig: BareAppConfig,
+  username: string
+) {
+  const expoPlistPath = path.join(projectRoot, 'ios', initialConfig.name, 'Supporting/Expo.plist');
+  const androidManifestPath = path.join(projectRoot, 'android/app/src/main/AndroidManifest.xml');
+
+  const appUrl = `https://exp.host/@${username}/${initialConfig.name}`;
+  const sdkVersion = (await readConfigJsonAsync(projectRoot)).exp.sdkVersion;
```
brentvatne, line 371: we should avoid using `readConfigJsonAsync` if possible and use `getConfig` instead. if we decide to switch a template over to app.config.js/ts at some point it would then work right out of the box

esamelson:
> could we use the implementation from `@expo/config` here instead? it would be one less thing to keep up to date then

yes, good point 😄 will wait for #2275 to land and then update this PR

esamelson:
Changes applied, but one thing I don't like about the `AndroidConfig` is that the package we use to read and write the AndroidManifest doesn't preserve comments, whitespace, and in some cases, order of tags, which makes the AndroidManifest less readable (especially for a newly `init`ed project). I'm 50/50 on whether it's worth it to have slightly cleaner code on our side.

brentvatne:
hm yeah, we need to swap that package soon. let's land this and follow up with switching xml2js to xml-js, as @bbarthec has done in the [configure-splash-screen package](https://github.com/expo/expo-cli/tree/master/packages/configure-splash-screen)

brentvatne:
@cruzach - any chance you're interested in taking on switching over to xml-js in @exop/config?

cruzach:
Yeah I can handle that https://github.com/expo/expo-cli/issues/2296

bbarthec:
Be informed that `xml-js` is not the perfect solution as it's stripping away any whitespaces 😞 
I couldn't find anything better though 😢 
Maybe (just maybe though) it would benefit us (dunno how though) to fort/create our own `xml-manipulating` library 🤔 
I wonder if we can leverage some tooling like `peg.js` or even better `peg.ts` 🤔

PR #2254, opened 2020-06-12
feat(xdl): add iOS AdSupport.framework configuration

cameronmoreau:
This PR is mostly a proof of concept and to spark a discussion

The latest Expo release [included Firebase Analytics](https://docs.expo.io/versions/latest/sdk/firebase-analytics/) 🎉 but with the limitation that [Managed apps cannot access ad campaign info](https://docs.expo.io/versions/latest/sdk/firebase-analytics/#optional-enable-adsupport-in-bare-workflow) since there is no support for adding `AdSuport.framework` 😬

This solution simply injects the framework at build time (if a special flag is enabled).

Out of scope for this PR, but dynamically adding/removing frameworks based on what libs are used, could potentially reduce artifact sizes. Especially for users that don't necessarily use every module in the SDK.

## Open questions
1. is `configureAsync` in xdl the right place for this? There's a world where `turtle` calls this bit of code outside of configureAsync - but should it?
1. Speaking of turtle 🐢 this PR just mutates `project.pbxproj`. So all subsequent builds could end up with the ad frameworks attached, although not desired. 
   * `.pbxproj` could be reverted back to its final state after the build process has completed (or errored)
   * the `iosProjectDirectory` could be copied to tmp, mutated, then deleted after - similar to how the `.app` is copied (Although more work, I think I like this approach the best)

## What's next
- [ ] Introduce the `ios.enableAdSupport` config for manifests
- [ ] Bump the xdl module in `turtle`
- [ ] The docs™

## Test Plan
TBD; A simple unit tests mocking the `xcode` modules would be sufficient if the current approach is determined to be the path forward.

### DIFF
--- a/packages/xdl/src/detach/IosNSBundle.ts
+++ b/packages/xdl/src/detach/IosNSBundle.ts
@@ -1,5 +1,9 @@
 import fs from 'fs-extra';
 import path from 'path';
+import { sync } from 'glob';
+
+// @ts-ignore: No TS support
+import { project as Project } from 'xcode';
 
 import { ExpoConfig } from '@expo/config';
 import * as AssetBundle from './AssetBundle';
@@ -522,6 +526,31 @@ async function _configureGoogleServicesPlistAsync(context: AnyStandaloneContext)
   }
 }
 
+async function _configureAdSupportAsync(context: AnyStandaloneContext): Promise<void> {
+  if (context instanceof StandaloneContextUser) {
+    return;
+  }
+
+  if (!context.config.ios || !context.config.ios.enableAdSupport) {
+    return;
+  }
+
+  const buildPhaseLogger = logger.withFields({ buildPhase: 'configuring NSBundle' });
+  const { iosProjectDirectory } = IosWorkspace.getPaths(context);
+
+  const pbxprojPaths = sync(path.join(iosProjectDirectory, '*', 'project.pbxproj'));
+
+  for (const pbxprojPath of pbxprojPaths) {
+    buildPhaseLogger.info(`Adding AdSupport framework to ${pbxprojPath}`);
+
+    const project = Project(pbxprojPath);
+    await project.parse();
+    project.addFramework('AdSupport.framework');
+
+    await fs.writeFile(pbxprojPath, project.writeSync());
+  }
+}
+
 export async function configureAsync(context: AnyStandaloneContext): Promise<void> {
   const buildPhaseLogger = logger.withFields({ buildPhase: 'configuring NSBundle' });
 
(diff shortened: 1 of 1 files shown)


cameronmoreau:
Could I get some eyes on this @brentvatne @EvanBacon ?

brentvatne:
hi there! sorry we were quite busy with sdk 38 release recently, we'll have a look at this soon :)

brentvatne:
hi there! so this won't actually work because of how [turtle](http://github.com/expo/turtle) works (turtle is our library for taking expo managed apps and producing app binaries from them). on ios we precompile an "app shell" and then substitutes configuration in some places (eg: infoplist) and assets (eg: icon, splash screen, fonts, and other project assets). as far as i know we cannot add a framework at this phase in the build process, and we don't have access to the user's configuration at the time that we precompile the app. after that we sign the app and it's ready to go. this is how we're able to do our ios builds in a minute or two.

we've been working on a new version of the build service currently that will compile every app from scratch rather than use the "app shell" concept. this will enable changes like this.

cameronmoreau:
> hi there! so this won't actually work because of how [turtle](https://github.com/expo/turtle) works (turtle is our library for taking expo managed apps and producing app binaries from them). on ios we precompile an "app shell" and then substitutes configuration in some places (eg: infoplist) and assets (eg: icon, splash screen, fonts, and other project assets). as far as i know we cannot add a framework at this phase in the build process (after compilation, before signing), and we don't have access to the user's configuration at the time that we precompile the app. after that we sign the app and it's ready to go. this is how we're able to do our ios builds in a minute or two.
> 
> we've been working on a new version of the build service currently that will compile every app from scratch rather than use the "app shell" concept. this will enable changes like this.
> 
> edit: that said, if you are so motivated you can produce your own version of the app shell that includes this framework and anything else you like, and then use turtle-cli to build the app. this is fairly complicated and we don't document this process, but you could probably figure it out from looking at expo/expo, in particular this github workflow: https://github.com/expo/expo/blob/master/.github/workflows/shell-app-ios.yml

Yup! Definitely traced down and looked at a lot of the concerns you brought up. This is definitely why I proceeded with an MVP and wanted to have a discussion on the best path forward. While the process of this PR definitely needs work on how it best fits with the project, this solution ended up working for my team. I wanted to chance to PR it up to the community since it sounds like other folks have run into the same issue.

You're correct that you don't have the user's config at "app shell generation" time, but it can still be modified just like app signing. In my mind, this functionality would live near the build process where assets are injected - in this case, the `project.pbxproj` is injected/modified.

I understand this functionality is a bit complicated based on the limitations of the build service (turtle) and managed apps, but I'd really like to keep digging further - as opposed to a straight closed PR. I'm happy to keep running with the PR and package it up in a nice way that's easy for you all. But I'd love to hear about your concerns/thoughts on the "Open Questions" section.

Thanks!

PR #2275, opened 2020-06-16
[config][expo-cli] add updates config to expo eject and expo apply

esamelson:
Follow up to #2249, partial step of #1494. 

Adds Updates configuration (both the properties set in app.json, and other properties like the update URL) to the `configureAndroidProjectAsync` and `configureIOSProjectAsync` steps of `expo eject`.

This allows workflows like 

- expo eject
- make release build
- publish update
- get update in app right away

without having to set configuration options (like the URL) manually before making a release build.

Tested manually with defaults and with each of the app.json fields set manually (including owner) and verified that the resulting fields in Expo.plist and AndroidManifest.xml were correct in all cases.

- [x] TODO: add unit tests for these config methods

### DIFF
--- a/packages/config/src/android/Updates.ts
+++ b/packages/config/src/android/Updates.ts
@@ -1,0 +1,103 @@
+import { ExpoConfig } from '../Config.types';
+import { Document } from './Manifest';
+
+export function getUpdateUrl(config: ExpoConfig, username: string | null) {
+  const user = typeof config.owner === 'string' ? config.owner : username;
+  if (!user) {
+    return null;
+  }
+  return `https://exp.host/@${user}/${config.slug}`;
+}
+
+export function getSDKVersion(config: ExpoConfig) {
+  return typeof config.sdkVersion === 'string' ? config.sdkVersion : null;
+}
+
+export function getUpdatesEnabled(config: ExpoConfig) {
+  return config.updates?.enabled !== false;
+}
+
+export function getUpdatesTimeout(config: ExpoConfig) {
+  return config.updates?.fallbackToCacheTimeout ?? 0;
+}
+
+export function getUpdatesCheckOnLaunch(config: ExpoConfig) {
+  if (config.updates?.checkAutomatically === 'ON_ERROR_RECOVERY') {
+    return 'NEVER';
+  } else if (config.updates?.checkAutomatically === 'ON_LOAD') {
+    return 'ALWAYS';
+  }
+  return 'ALWAYS';
+}
+
+export async function setUpdatesConfig(
+  config: ExpoConfig,
+  manifestDocument: Document,
+  username: string | null
+) {
+  let mainApplication = manifestDocument.manifest.application.filter(
+    (e: any) => e['$']['android:name'] === '.MainApplication'
+  )[0];
+
+  addMetaDataItemToMainApplication(
+    mainApplication,
+    'expo.modules.updates.ENABLED',
+    `${getUpdatesEnabled(config)}`
+  );
+  addMetaDataItemToMainApplication(
+    mainApplication,
+    'expo.modules.updates.EXPO_UPDATES_CHECK_ON_LAUNCH',
+    getUpdatesCheckOnLaunch(config)
+  );
+  addMetaDataItemToMainApplication(
+    mainApplication,
+    'expo.modules.updates.EXPO_UPDATES_LAUNCH_WAIT_MS',
+    `${getUpdatesTimeout(config)}`
+  );
+
+  const updateUrl = getUpdateUrl(config, username);
+  if (updateUrl) {
+    addMetaDataItemToMainApplication(
+      mainApplication,
+      'expo.modules.updates.EXPO_UPDATE_URL',
+      updateUrl
+    );
+  }
+  const sdkVersion = getSDKVersion(config);
+  if (sdkVersion) {
+    addMetaDataItemToMainApplication(
+      mainApplication,
+      'expo.modules.updates.EXPO_SDK_VERSION',
+      sdkVersion
+    );
+  }
+
+  return manifestDocument;
+}
+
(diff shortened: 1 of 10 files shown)


packages/config/src/android/Updates.ts
```diff
@@ -0,0 +1,103 @@
...
+      mainApplication,
+      'expo.modules.updates.EXPO_SDK_VERSION',
+      sdkVersion
+    );
+  }
+
+  return manifestDocument;
+}
+
+function addMetaDataItemToMainApplication(
```
cruzach, line 78: nice idea to extract this out, think there are a couple other places in the `config/android/` source that could make use of this ([like here](https://github.com/expo/expo-cli/blob/master/packages/config/src/android/Branch.ts#L19-L37)) (not that you need to go through and make those changes now)

packages/config/src/android/Updates.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function getUpdatesEnabled(config: ExpoConfig) {
+  return config.updates?.enabled !== false;
+}
+
+export function getUpdatesTimeout(config: ExpoConfig) {
+  return config.updates?.fallbackToCacheTimeout ?? 0;
+}
+
+export function getUpdatesCheckOnLaunch(config: ExpoConfig) {
```
cruzach, line 24: isn't `WIFI_ONLY` one of the options as well?

packages/config/src/android/Updates.ts
```diff
@@ -0,0 +1,103 @@
+import { ExpoConfig } from '../Config.types';
+import { Document } from './Manifest';
+
+export function getUpdateUrl(config: ExpoConfig, username: string | null) {
```
cruzach, line 4: is there a way we could determine if the user is self-hosting their updates? Maybe we could add an app.json field for this, otherwise that is the one setting they would need to configure manually in AndroidManifest and Info.plist, right?

packages/config/src/android/Updates.ts
esamelson: think it's worth moving this util to `Manifest.ts`?

packages/config/src/android/Updates.ts
esamelson: it is in the new API but it has no analog in the managed config currently, so there's nothing that maps to it yet

packages/config/src/android/Updates.ts
esamelson: (also, full disclosure, I stole the idea from `Facebook.ts` :) )

packages/config/src/android/Updates.ts
esamelson: It would be great to add an app.json field or something for this -- another reason that would be nice is that you wouldn't have to provide the url every time you run `expo export`.

That said, `expo export` (same as `expo publish`) will also set the updateUrl field in the native project configs. So the only time it matters (and the developer would have to set it manually) is for builds made right after ejecting but before exporting.

packages/config/src/android/Updates.ts
cruzach: haha 😄 
Yeah I think it's worth it, but no need to replace the duplicate code in the other files' `set` methods (unless you feel very inclined to do so; I think the relevant unit tests would catch any problems)

packages/config/src/android/Updates.ts
cruzach: Okay, good to have this :
> It would be great to add an app.json field or something for this 

as a todo to improve the workflow. 
 

> the only time it matters is for builds made right after ejecting but before exporting

I also doubt many users would do this (eject, not export, build), so (i think) this is a very narrow edge case we're considering. Too small to warrant logging any warning, since it wouldn't be applicable to most folks. What do you think, is it worth us prioritizing adding the `exportUrl` key soon?

packages/expo-cli/src/commands/apply/configureAndroidProjectAsync.ts
```diff
@@ -54,6 +55,7 @@ async function modifyMainActivityJavaAsync(
 
 export default async function configureAndroidProjectAsync(projectRoot: string) {
   const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const username = await UserManager.getCurrentUsernameAsync();
```
brentvatne, line 58: this is the first property that depends on the user being signed in. i don't want to block this pr on this, but we may want to think about the following:
- should still be possible to eject without signing in (this pr maintains that)
- let people know who they are signed in as / if they are not signed in / give an opportunity to sign in
- users will need to be signed in to download their build credentials if they keep them on our credentials service
- at some point we may need users to sign in their apple account or provide some other additional info eg: around entitlements like icloud

packages/expo-cli/src/commands/apply/configureIOSProjectAsync.ts
```diff
@@ -62,6 +69,21 @@ async function modifyInfoPlistAsync(projectRoot: string, callback: (plist: any)
   await IosPlist.cleanBackupAsync(iosProjectDirectory, 'Info', false);
 }
 
+async function modifyExpoPlistAsync(projectRoot: string, callback: (plist: any) => any) {
+  const { iosProjectDirectory } = getIOSPaths(projectRoot);
+  const supportingDirectory = path.join(iosProjectDirectory, 'Supporting');
+  try {
+    await IosPlist.modifyAsync(supportingDirectory, 'Expo', callback);
+    await IosPlist.cleanBackupAsync(supportingDirectory, 'Expo', false);
```
brentvatne, line 77: maybe move this clean line to a finally block

packages/config/src/android/Updates.ts
esamelson: opened this issue to track https://github.com/expo/expo-cli/issues/2285

PR #2284, opened 2020-06-18
[xdl] Only update root client urls when sdk is latest and released

byCedric:
This prevents the root client urls being updated if an SDK version is not yet released. It avoids marking this as "latest" by default in the CLI.


byCedric:
Ok, I'll close this and work on the usage refactor. It probably takes some time, but I'll free up my monday for this!

PR #2293, opened 2020-06-19
Revert "[cli] Refactor Android keystore update test with prompts inject (#2290)"

wkozyra95:
This reverts commit d338ab2cf46ada4b30f165b6e91d1ed3fac23a40.

inject does not handle all the cases the previous version did.
- if you inject to much or to many variables test won't fail
e.g instead of
```
      prompts.inject([
        true, // Keystore source - User provided
        'test', // Keystore credentials
      ]);
```
there should be 
```
      prompts.inject([
        true, // Keystore source - User provided
        'test','test','test','test', // Keystore credentials
      ]);
```
but tests are still passing
- if you inject too many variables next test will get values from the previous one and there is no way to reset state


byCedric:
Hmm, great catch. I'll open an issue on the prompts repo for the missing reset functionality. Maybe they are also open for objects with `[question]: answer` notation, which should make it not-possible to add too much. Sorry for that!

PR #2297, opened 2020-06-22
[configure-splash-screen] Refactor and integrate with `@expo/config`

bbarthec:
# Why

`@expo/config` should autoconfigure the splash screen once `expo eject` is called instead of showing nasty warning.

# How

This whole PR is much too big and complex to be called `a nice one` 😅 
The amount of code that has been written is not even to be compared to the amount of functionality this PR provides (to the disadvantage of the the first).
Nevertheless I've decided to try to ship it in the shape it is.
Why? Because I've experimented with:
- [x] generic `SplashScreenConfig` type
- [x] platform-specific `SplashScreenConfig` types that are created dynamically via `mapped type`
- [x] JSON-friendly types equivalents that are to be exposed outside the package (e.g. for `@expo/config`) and are dynamically generated from original types via other `mapped type`
- [x] `Validator` object that has this nicely looking `addRule` method that is fully TS-supported
- [x] `cli` command separated completely from logic now (only defines CLI options and performs minimal validation and shaping raw JSON Config for further processing)
- [x] a bunch of tests (these might and should be cleaned in the future - I started to wonder how to write minimal set of tests to cover each code path 🤔  any idea here?)

configure-splash-screen-flow


packages/configure-splash-screen/src/types.ts
```diff
@@ -0,0 +1,274 @@
+import { Color } from 'color-string';
+import { SplashScreenImageResizeModeType, SplashScreenStatusBarStyleType } from './constants';
+
+/**
+ * Type describing generic (platform-agnostic) Splash Screen configuration.
+ */
+export type GenericSplashScreenConfig = {
+  backgroundColor: Color;
+
+  imagePath?: string;
```
brentvatne, line 10: to be consistent with the rest of our app.json property names with paths we should maybe remove the `Path` suffix - so just `image`.

eg: icon (https://docs.expo.io/versions/latest/config/app/#icon) and image on the current splash config (https://docs.expo.io/versions/latest/config/app/#splash)

brentvatne:
> I started to wonder how to write minimal set of tests to cover each code path 🤔 any idea here?

I think the current structure of this code lends itself very well to testing. If we test the validator/transformer functions exhaustively, then we can do some e2e tests that ensure they're wired up correctly with the CLI/external api.

One thing that we might want to do is split out the FromJsonValidator and test that independently, then just assume it works as it promises though its API within this code. We can do that in a followup PR.

packages/configure-splash-screen/src/__tests__/cli-command.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  __esModule: true,
+  default: jest.fn(),
+}));
+jest.mock('../android', () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+describe('cli-command', () => {
+  beforeEach(() => {});
```
satya164, line 15: The `beforeEach` seems unnecessary

packages/configure-splash-screen/src/android/__tests__/Colors.xml.test.ts
```diff
@@ -46,16 +46,18 @@ describe('Colors.xml', () => {
...
       const result = vol.readFileSync(filePath, 'utf-8');
       const expected = generateColorsFileContent({ backgroundColor: '#FF0000' });
       expect(result).toEqual(expected);
     });
 
     it('creates correct file for dark mode', async () => {
       await configureColorsXml(androidMainPath, {
-        backgroundColor: colorString.get('red')!,
-        darkModeBackgroundColor: colorString.get('green')!,
+        backgroundColor: colorString.get('red')!.value,
```
satya164, line 57: Is the `!` here (and in other lines) necessary?

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+
+interface Configuration {
+  platform: PlatformType;
+  android: AndroidSplashScreenJsonConfig;
+  ios: IosSplashScreenJsonConfig;
+}
+
+async function action(configuration: Configuration) {
+  const { platform, android, ios } = configuration;
+  const rootDir = path.resolve();
```
satya164, line 42: I think `process.cwd()` would be more obvious that `path.resolve()` without argument since that seems to be what we want here
```suggestion
  const rootDir = process.cwd();
```

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+function getAvailableOptions(o: object) {
+  return Object.values(o)
+    .map(v => `"${v}"`)
+    .join(' | ');
+}
+
+export default () =>
+  new Command()
+    .description(
+      'Idempotent operation that configures native splash screens using provided backgroundColor and optional .png file. Supports light and dark modes configuration. Dark mode is supported only on iOS 13+ and Android 10+.'
```
satya164, line 140: Just a suggestion: "Idempotent" might not be a familiar word to a lot of people

bbarthec:
Linking with https://github.com/expo/expo-cli/issues/1494

sjchmiela:
Oooh… thanks for reminding to review this 😅

packages/configure-splash-screen/src/validators/FromJsonValidator.ts
```diff
@@ -0,0 +1,92 @@
+import { get, set } from 'lodash';
+import { DeepRequired, IsNever, KeyofObject, OptionalPromise, JsonShape } from '../types';
+
+/**
+ * This class is responsible for validating configuration object in a form of json and produce validated object based on validating `rules` added via `addRule` method.
+ */
+export default class FromJsonValidator<From extends JsonShape<To>, To extends object> {
```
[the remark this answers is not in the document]
sjchmiela: Congratulations on implementing an object validator! I too am curious whether you considered other NPM packages. As I mentioned on one of the calls I've had pretty good experience with [`zod`](https://github.com/vriad/zod) which, I guess, does the same thing (parses an unknown object into an object of a known TS type, inferring and parsing types and optionally validates)?

While I admire the heart you've put into this and stuff you must have learned along the way I'm a little afraid maintaining this may be a burden no one will want to take a couple of months from now and this will become a zombie file _here be dragons_.

packages/config/src/ios/SplashScreen.ts
```diff
@@ -1,23 +1,33 @@
...
+  if (!splashConfig) {
     return;
   }
 
-  addWarningIOS(
-    'splash',
-    'This is the image that your app uses on the loading screen, we recommend installing and using expo-splash-screen.',
-    'https://github.com/expo/expo/blob/master/packages/expo-splash-screen/README.md'
-  );
+  await configureIosSplashScreen(projectRoot, splashConfig);
```
sjchmiela, line 32: Same as with Android — https://github.com/expo/expo-cli/pull/2297/files#r454773239

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+  'darkModeImagePath',
+  'statusBarHidden',
+  'statusBarStyle',
+  'darkModeStatusBarStyle',
+  'statusBarTranslucent',
+  'statusBarBackgroundColor',
+  'darkModeStatusBarBackgroundColor',
+] as const;
+
+type CLIOptionName = typeof AVAILABLE_OPTIONS_NAMES[number];
```
sjchmiela, line 30: Why this sophisticated `const AVAILABLE_OPTIONS_NAMES` and then this `typeof` instead of a simple enum or a union type?

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+
+type CLIOptions = Partial<Record<CLIOptionName, string>>;
```
sjchmiela, line 32: Wouldn't type of `statusBarTranslucent` be `boolean`?

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+    resolvedPlatform = (platform && validateEnumValue(platform, Platform)) || Platform.ALL;
+  } catch (e) {
+    throw new Error(`'platform': ${e.message}`);
+  }
+
+  if (!backgroundColor) {
+    throw new Error(
+      `'backgroundColor': Required option is not provided. Provide a valid color string.`
+    );
+  }
```
sjchmiela, line 83: I think recent releases of `commander.js` allow usage of `requiredOption` if that's something we may want to look into

packages/configure-splash-screen/src/cli-command.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  const resolvedStatusBarHidden =
+    statusBarHidden === undefined ? undefined : Boolean(statusBarHidden);
```
sjchmiela, line 86: Passing `undefined` in a variable called `resolved` seems icky to me. Is there a specific reason for passing `undefined` down? Does it say "do not modify anything related to this setting"?

Actually while writing this comment I had deja vu, so if we already talked about this then sorry

packages/configure-splash-screen/src/android/index.ts
```diff
@@ -1,49 +1,28 @@
...
-      darkModeStatusBarBackgroundColor,
-    }),
-    configureDrawableXml(androidMainPath, resizeMode),
-    configureStylesXml(androidMainPath, {
-      statusBarHidden,
-      statusBarStyle,
-      darkModeStatusBarStyle,
-      addStatusBarBackgroundColor: !!statusBarBackgroundColor,
-    }),
+    configureDrawables(androidMainPath, validatedConfig),
```
sjchmiela, line 21: Is there a particular reason for this change? I quite liked the idea of `configureX` functions as small, isolated, reusable commands which aligned more with the idea of passing only the required information to the method instead of the whole config object (it also makes it easier to write more tangled code). Just curious!

sjchmiela (commented):
This is a **very** clever pull request (which means both good and bad). 😅 Had some questions, apart from them it looks very solid!

packages/configure-splash-screen/src/android/index.ts
bbarthec: Ease of writing the code I think 🤔 
Instead of creating massive deconstruction od the config object and aliasing nested properties just to pass them to other function invocation I have a single point of reference in each file - the config object. I see that this is sth that might confuse the reader and I dislike it, so it should be refactored like you're suggesting 👍

packages/configure-splash-screen/src/validators/FromJsonValidator.ts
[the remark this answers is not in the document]
bbarthec: Well, it was my little experiment 😉 
I really like the `zod` idea and I'm willing to add it to our codebase somewhere. I'm not sure though if `zod` can provide validation & transformation at the same time, please take a look at following lines:
https://github.com/expo/expo-cli/pull/2297/files#diff-63c94c9e1d2966b124e8333ec934d795R55
https://github.com/expo/expo-cli/pull/2297/files#diff-63c94c9e1d2966b124e8333ec934d795R29
These represent the same think, but one in JSON and other in it's semantically-validated equivalent (`Color` is a tuple). What I'm curious about is: whether `zod` can convert for example string into some complex object/structure? 🤔 Nonetheless, IMO it's still worth to be added 👍

packages/configure-splash-screen/src/validators/__tests__/android.test.ts
```diff
@@ -0,0 +1,141 @@
...
+            style: 'unknown',
+          },
+          darkMode: {
+            backgroundColor: 'rgb(10, 10, 10)',
+            statusBar: {
+              style: 'unknown',
+            },
+          },
+        });
+      }).rejects.toThrow(
```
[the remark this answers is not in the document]
bbarthec: Nice idea, but as I've never touched snapshot matchers yet I suggest adding it in separate PR 😅

packages/configure-splash-screen/src/cli-command.ts
bbarthec: Yeah, it does, but I couldn't get this option covered by tests then 😞 Commander.js simply kills the process with exit code `1` once the option is not provided. Maybe it does not need to be tested then? 🤔 Dunno 🤷

packages/configure-splash-screen/src/android/index.ts
bbarthec: I've adjusted it a little. These functions are now expecting only parameters that they're consuming, but TS still allows us to pass some `bigger` object that satisfies the `smaller`/`subset` type 😉

packages/configure-splash-screen/src/__tests__/cli-command.test.ts
```diff
@@ -0,0 +1,138 @@
+import configureAndroid from '../android';
```
EvanBacon, line 1: We should rename all of these tests to `-test` after this PR is merged. https://github.com/expo/expo-cli/issues/2591

EvanBacon:
- Seems the XML logic isn't unified with expo/config so running expo eject/apply formats the XML files differently.
- This doesn't work as expected when it's rerun using apply. Apply should be supported because it makes using bare far better for expo users, but also having apply support makes changes like this substantially easier to test as a whole.

brentvatne:
> Seems the XML logic isn't unified with expo/config so running expo eject/apply formats the XML files differently.

@bbarthec @EvanBacon - let's chat at the next expo-cli meeting about which xml library we should be using, we currently have three (that i'm aware of? maybe more?):

1. xmldom
2. xml2js
3. xml-js

we need to consolidate on one to reduce duplicate dependencies and ensure that we're not reformatting the xml files in subtle ways when different commands are executed.

bbarthec:
Do you guys have any major concerns about this PR apart from the `xml` library? cc @brentvatne, @EvanBacon 
If no, wy not to ship it now and work on the top of it in the separate PRs?

brentvatne (approved):

PR #2303, opened 2020-06-24
[package manager] Prompt the user to use sudo to install CocoaPods

EvanBacon:
## Why 

The recommended way to install cocoapods is with `sudo gem install`, but more often than not they're installing with homebrew on mac. This PR attempts to address this issue by prompting the user to enter their password in interactive mode.

## How 

- Fix sudo-prompt integration - not tested as I don't have a win32 machine. We also often skip pod install on non-darwin machines.
- Use sudo on mac and linux - tested and works.
- `pod-install --non-interactive` will skip prompts.

### DIFF
--- a/packages/package-manager/src/PackageManager.ts
+++ b/packages/package-manager/src/PackageManager.ts
@@ -18,9 +18,8 @@ export function getPossibleProjectRoot(): string {
 }
 
 export function spawnSudoAsync(command: string): Promise<void> {
-  const packageJSON = require('../package.json');
   return new Promise((resolve, reject) => {
-    sudo.exec(command, { name: packageJSON.name }, error => {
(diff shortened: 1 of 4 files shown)


packages/package-manager/src/PackageManager.ts
```diff
@@ -17,14 +18,20 @@ export function getPossibleProjectRoot(): string {
...
+          reject(error);
+        }
+        resolve();
+      });
     });
-  });
+  } else {
+    // Attempt to use sudo to run the command on Mac and Linux.
+    // TODO(Bacon): Make a v of sudo-prompt that's win32 only for better bundle size.
+    await spawnAsync('sudo', command, spawnOptions);
```
brentvatne, line 35: we should show some message to the user to let them know why we are asking them to type in their password

PR #2306, opened 2020-06-25
[expo-cli][build:ios] Added prompt to supply bundle identifier

### LINKED ISSUE #2305: [expo-cli][build:ios] Check bundle ID immediately
If we're going to validate we should do it as soon as possible.

EvanBacon:
## Why 

- If the CLI is going to do validation and potentially bail out, it should do the validation quickly. Currently we do a couple network requests before getting to any validation.
- It's annoying to have to select "archive" / "simulator" multiple times. Almost without fail, when I finish a project I run `expo build:ios` and have it tell me I don't have a bundle identifier setup. This is annoying because then I set it and have to start the process all over again (a couple seconds but it still feels clunky). Obviously this is a nit pick but I think it can be better. - resolve #2305
- When I have a static config (`app.json` or `app.config.json`) I should just be prompted to add one there on the spot, and lose no progress. 
- Why can't Expo CLI recommend a bundle ID - I know this one is a little sketchy because there is a chance that bundle ID it might recommend could already be in use, but it's guess is almost as good as mine. I've never once opened xcode to verify.
- If I'm new to development, having some form of bundle ID validation is super helpful, even if I have to manually add it to the config later.

## How

- Move the simulator / archive question to after the config validation.
- Lazily fetch the user so I don't have to wait longer to potentially see a validation error.
- Reformat the bundle ID messages to make them look more like "expected warnings" and less like errors.
- Prompt the user for their bundle ID if it's not defined.
- When the user has a dynamic config, give a helpful copy/🍝 message then bail out.
- Suggest a bundle ID based on the formula `com.username.slug`.
- Added  's to help users know this isn't related to Expo CLI.
- Reorder the validation so it throws from hardest-to-fix to easiest-to-fix, this is important because if I'm trying to build an expokit project but I get warned to add an icon first, then I'll spend some time creating an icon, run the ⌘ again, and realize I need to rewrite my app :/
  - The new order for iOS is: ExpoKit, SDK, icons, bundle ID

### Demo

When I have an `app.config.json` without a bundle ID, and I use the suggested bundle ID:

Screen Shot 2020-06-24 at 5 11 22 PM

When I have an `app.config.js` without a bundle ID, and I use a custom bundle ID:

Screen Shot 2020-06-24 at 5 10 48 PM

Notice that it prompts for archive / simulator after the validation is complete. Much snappier, and easier to copy/pasta.

## Test Plan

- I've added a basic test to ensure non-interactive works by failing on the missing bundle Id.
- The most fragile parts imo are the modifications to the app.json but that code is well tested in `@expo/config`.


packages/expo-cli/src/commands/build/BaseBuilder.ts
```diff
@@ -1,25 +1,44 @@
...
 
 export default class BaseBuilder {
-  manifest: ExpoConfig = getConfig(this.projectDir).exp;
-  user?: User;
+  protected projectConfig: ProjectConfig;
+  manifest: ExpoConfig;
+  private user?: User;
+  private loadedUser: boolean = false;
+
+  // Lazy load the user authentication check.
```
EvanBacon, line 23: Seems like this logic isn't even used on Android anymore 🤔

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -61,21 +64,10 @@ class IOSBuilder extends BaseBuilder {
       await this.checkStatusBeforeBuild();
     }
     await this.build(publishedExpIds);
+    // TODO: Why is this invoked twice?
```
brentvatne, line 67: @byCedric wanted to do this in order to make sure people really read it

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -119,9 +111,102 @@ class IOSBuilder extends BaseBuilder {
...
+
+      const modification = await modifyConfigAsync(
+        this.projectDir,
+        { ios: { bundleIdentifier: bundleIdPrompt.bundleIdentifier } },
+        { skipSDKVersionRequirement: true }
+      );
+      if (modification.type === 'success') {
+        log.newLine();
+        // Success!
+        log(`Your  bundle identifier is now: ${bundleIdPrompt.bundleIdentifier}`);
```
brentvatne, line 175:
```suggestion
        log(`Your iOS bundle identifier is now: ${bundleIdPrompt.bundleIdentifier}`);
```

slightly prefer iOS over  - when font size is small  just looks like a weird dot

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
```diff
@@ -119,9 +111,102 @@ class IOSBuilder extends BaseBuilder {
...
+        log.newLine();
+      } else {
+        log.newLine();
+        if (modification.type === 'warn') {
+          // The project is using a dynamic config, give the user a helpful log and bail out.
+          log(log.chalk.yellow(modification.message));
+        } else {
+          log(
+            log.chalk.yellow(
+              'No Expo config was found. Please create an `app.config.json` in your project root.'
```
brentvatne, line 185:
```suggestion
              'No Expo config was found. Please create an `app.config.json` in your project root.'
```

i think this is kind of confusing - in our docs we generally don't refer to it as app.config.json. maybe we can provide some helper function that just says "Expo config" and terminal links to a fyi page about what that means

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
EvanBacon, line 67:
```suggestion
```

packages/expo-cli/src/commands/build/ios/IOSBuilder.ts
EvanBacon, line 185:
```suggestion
              'No Expo config was found. Please create an Expo config (`app.config.js` or `app.json`) in your project root.'
```

PR #2307, opened 2020-06-25
Blacklist suspicious permissions

ErikHellman:
Blacklist permissions that can be considered suspicious. Google may consider this to be bad enough to suspend an application from Google Play Store.

Fixes #8942

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -803,6 +803,10 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
       'com.sec.android.provider.badge.permission.READ',
       'com.sec.android.provider.badge.permission.WRITE',
       'com.sonyericsson.home.permission.BROADCAST_BADGE',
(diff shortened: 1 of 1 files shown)


ErikHellman:
An alternative solution to this is to remove all permissions that gets added and are not listed in `app.json`. Maybe with an exception for `INTERNET` since basically all apps use that.

brentvatne:
@byCedric is investigating this currently!

brentvatne:
@byCedric - feel free to merge this if you're OK with it

byCedric:
I'm closing this PR in favor of #2458, where the `ACCESS_BACKGROUND_LOCATION` permission and annotations are added to make sure everyone knows why these permissions are blacklisted. 

@ErikHellman Again, thanks for bringing this up. [We have changed a lot of things to prevent accidentally added permissions in the future.](https://github.com/expo/expo/issues/9499) All these changes should keep the Android permissions both transparent by internal and external developers and maintainable.

mustvlad:
@byCedric So is there a way now to remove the automatically added `ACCESS_BACKGROUND_LOCATION` permission from expo managed android builds?

PR #2315, opened 2020-06-26
fix(expo-cli): use user provided teamId, don't access appleCtx

wkozyra95:
fixes problem in credential manager when the user specifies push key manually and appleCtx is not intialized.

for distribution certificate this worked because there is `await ctx.ensureAppleCtx();`, but because adding teamId and teamName is obsolete(`newDistCert` and `newPushKey` have already teamId) I removed it. We don't need to force users to specify their credentials just for that.


byCedric:
Looks good!

Issue #2224, opened 2020-06-04
[cli] list/checkboxes broken with multiline choices


quinlanj:
# problem
inquirer breaks with a long multiline list. This is affecting users especially with credential related actions. here is a gif scrolling through a list with 200 multiline options - notice there are no cursors and it skips a huge chunk of choices.

![ezgif-6-95afd84c78d2](https://user-images.githubusercontent.com/6380927/83701709-67617c00-a5bf-11ea-9733-6741bd19cbe4.gif)

# todo

- [ ] bump inquirer once https://github.com/SBoudrias/Inquirer.js/pull/928 lands and is published

EvanBacon:
We no longer use inquirer for this.

Issue #2016, opened 2020-04-28
update react-dev-utils


EvanBacon:
fix https://github.com/expo/expo-cli/issues/1004

brentvatne:
it looks like we're using the latest version of react-dev-utils at the time of writing this - 10.2.1

Issue #1004, opened 2019-09-17
expo start hangs when opening browser, ignores all input and options


jessegranger:
## Description

expo start hangs, never opens the DevTools in the browser, ignores all input arguments.

## Expected Behavior

What actions did I perform?

`> expo init blank-template`, then choose a `Blank Template`.

In the resulting directory that is created:

`> expo start`

What did I expect?

I expect the packager to start, and for the dev tools to open in default browser (Firefox).

Or, an error, if there is some system mis-configuration.

## Observed Behavior

```
> expo start
Starting project at /xxx/blank-template
Expo DevTools is running at http://localhost:19002
Opening DevTools in the browser... (press shift-d to disable)
# expo hangs here forever, no browser is ever opened
# if you manually open a browser and go to http://localhost:19002, it never responds
# pressing shift-d has no effect (prints D in the console, but does not un-hang)
^C
Stopping packager...
Packager stopped.
# Now, the browser opens a new tab with the DevTools url but it is already down
```

Passing options like `--https`, are ignored.
```
> expo start --https
Starting project at /xxx/blank-template
Expo DevTools is running at http://localhost:19002
Opening DevTools in the browser... (press shift-d to disable)
^C
```

Same output when passing any of the options listed in `expo start --help`.
Eg, passing `--tunnel`, `--lan`, or `--localhost`, all just open on `http://localhost:19002`.
No effect from using `--dev`, `--no-dev`, `--https`, `--no-https`, `-a`, `-i`, `-w`, `-c`, `--web-only`, none of these have any effect.

## Environment

```
> expo diagnostics

  Expo CLI 3.0.10 environment info:
    System:
      OS: macOS 10.14.6
      Shell: 3.0.2 - /usr/local/bin/fish
    Binaries:
      Node: 12.10.0 - /usr/local/bin/node
      npm: 6.10.3 - /usr/local/bin/npm
    IDEs:
      Android Studio: 3.5 AI-191.8026.42.35.5791312
      Xcode: 10.3/10G8 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^34.0.1 => 34.0.4
      react: 16.8.3 => 16.8.3
      react-native: https://github.com/expo/react-native/archive/sdk-34.0.0.tar.gz => 0.59.8
    npmGlobalPackages:
      expo-cli: 3.0.10
```

Default browser is Firefox (69.0).

I tried using bash shell, I tried removing ~/.expo, neither had any effect.

## Reproducible Demo

[...]

PR #2337, opened 2020-07-02
[cli][xdl] bump inquirer

quinlanj:
# why
bump inquirer to fix some problems we were experiencing 
fixes https://github.com/expo/expo-cli/issues/2224

# breaking changes
afaik the bump from 5.2.0 to 7.3.0 doesn't affect us. The breaking changes include:

- Update to Rx.js v6
- Dropping support for Node 6

For more details, see these  inquirer [release notes](https://github.com/SBoudrias/Inquirer.js/releases?after=%40inquirer%2Fcore%400.0.8-alpha.0)

# manual tests
- ran a couple of expo cli commands that have prompts in them to do a quick sanity check

### DIFF
--- a/packages/expo-cli/package.json
+++ b/packages/expo-cli/package.json
@@ -90,7 +90,7 @@
     "glob": "7.1.6",
     "got": "^11.1.4",
     "indent-string": "4.0.0",
-    "inquirer": "5.2.0",
+    "inquirer": "7.3.0",
     "invariant": "2.2.4",
     "klaw-sync": "6.0.0",
     "leven": "^3.1.0",
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -61,7 +61,7 @@
     "hashids": "1.1.4",
     "idx": "2.4.0",
     "indent-string": "3.2.0",
-    "inquirer": "5.2.0",
+    "inquirer": "7.3.0",
     "internal-ip": "4.3.0",
     "invariant": "2.2.4",
     "js-yaml": "^3.13.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6557,6 +6557,14 @@ chalk@^4.0.0:
     ansi-styles "^4.1.0"
     supports-color "^7.1.0"
 
+chalk@^4.1.0:
+  version "4.1.0"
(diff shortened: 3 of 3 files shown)


fson:
My only concern is that the version bump is going to cause a major regression in our installation size, because `react-dev-tools` still depends on an older version of `inquirer`, so we end up depending on both.

`yarn why inquirer` shows both this:
```
=> Found "inquirer@7.0.4"
info Has been hoisted to "inquirer"
info Reasons this module exists
   - "workspace-aggregator-c7915d7a-6a15-4094-ae38-e5832b4eb142" depends on it
   - Hoisted from "_project_#eslint#inquirer"
   - Hoisted from "_project_#@expo#webpack-config#react-dev-utils#inquirer"
info Disk size without dependencies: "448KB"
info Disk size with unique dependencies: "22.8MB"
info Disk size with transitive dependencies: "24.05MB"
```
and this:
```
=> Found "expo-cli#inquirer@7.3.0"
info This module exists because "_project_#expo-cli" depends on it.
info Disk size without dependencies: "216KB"
info Disk size with unique dependencies: "22.57MB"
info Disk size with transitive dependencies: "23.82MB"
```
(the other versions are from `devDependencies` so they shouldn't have an effect).

As you can see, inquirer is ~24MB unpacked. Most of that, ~17MB of that comes from `rxjs`, and since `inquirer` bumped the `rxjs` version, we would depend on two versions of `rxjs` after this change.

Solutions:
- as a long term solution to these problems, we're in the process of moving away from `inquirer`, see https://github.com/expo/expo-cli/issues/1983. Maybe let's see if we can move the problematic lists to use `prompts` already? [`prompts.ts`](https://github.com/expo/expo-cli/blob/553eb01540554070c2e8b3894dbb4314ba2cb484/packages/expo-cli/src/prompts.ts) is a replacement for `prompts.ts` that @byCedric added that uses `prompts` instead of `inquirer`.
- alternatively, as an interim solution, we could send a PR to bump inquirer to `7.3.0` in [`react-dev-utils`](https://github.com/facebook/create-react-app/blob/2da5517689b7510ff8d8b0148ce372782cb285d7/packages/react-dev-utils/package.json#L69) and bump our `react-dev-utils` version (this is on our TODO list anyway, #2016, since it also has a fix for #1004) so that the same version is used in both.

EvanBacon:
Opened a PR to replace inquirer with prompts in create-react-app https://github.com/facebook/create-react-app/pull/10083

PR #2353, opened 2020-07-13
[expo-cli] Show longer release channel names

Foosballfan:
Make the channel column a 'big cell' to show longer channel names


byCedric:
@brentvatne Charles and I chatted about this on Slack, the reason for this is because a small column is 13 characters (15, -2x `|`). If you use something like `production-1.x.x`, it doesn't fit in that header and everything is truncated to `production-1...` which makes it fairly useless.

brentvatne:
heads up @quinlanj

Issue #2280, opened 2020-06-17
IOS build and upload commands should rely on the same env vars


cruzach:
`expo build:ios` expects the `EXPO_APPLE_PASSWORD` env variable, whereas `expo upload:ios` expects the `EXPO_APPLE_ID_PASSWORD` env variable. Unless there's a reason to have these be two different keys, it would avoid confusion to use the same one in both scenarios

Since this would be a breaking change, we can wait to fix this until we are about ready to ship a new major version. It's an easy change to make, but could cause a headache for people who are accustomed to the current workflow

1337LutZ:
> `expo build:ios` expects the `EXPO_APPLE_PASSWORD` env variable, whereas `expo upload:ios` expects the `EXPO_APPLE_ID_PASSWORD` env variable. Unless there's a reason to have these be two different keys, it would avoid confusion to use the same one in both scenarios
> 
> Since this would be a breaking change, we can wait to fix this until we are about ready to ship a new major version. It's an easy change to make, but could cause a headache for people who are accustomed to the current workflow

I vote for this as well. Now we are managing 2 different keys in 3 projects.

PR #2358, opened 2020-07-14
[expo-cli] consolidate env variables.

jkhales:
This fixes https://github.com/expo/expo-cli/issues/2280.

NB: in case of conflict we chose to prioritize the flag over the env variable.

Joint with @quinlanj 

# tests

- [x] manually tested `expo upload:ios` with flags and env var

### DIFF
--- a/packages/expo-cli/src/commands/upload/IOSUploader.ts
+++ b/packages/expo-cli/src/commands/upload/IOSUploader.ts
@@ -144,14 +144,16 @@ export default class IOSUploader extends BaseUploader {
     const appleCredsKeys = ['appleId', 'appleIdPassword'];
     const result: AppleCreds = pick(this.options, appleCredsKeys);
 
-    if (process.env.EXPO_APPLE_ID) {
-      result.appleId = process.env.EXPO_APPLE_ID;
-    }
+    const appleId = result.appleId ?? process.env.EXPO_APPLE_ID;
+    const appleIdPassword =
+      result.appleIdPassword ??
+      process.env.EXPO_APPLE_PASSWORD ??
+      process.env.EXPO_APPLE_ID_PASSWORD;
+
     if (process.env.EXPO_APPLE_ID_PASSWORD) {
-      result.appleIdPassword = process.env.EXPO_APPLE_ID_PASSWORD;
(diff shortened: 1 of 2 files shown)


jkhales:
Tests are failing but unrelated to our PR.

jkhales:
> ```
> You've provided contradictory Apple IDs. You should provide either [env var] or [flag], not both. Falling back to [whichever we fall back to].
> ```

Thanks for catching on the docs, updated!
Warnings are a nice touch.

Regarding the breaking behavior, we had discussed this and think that this is the intuitive behavior and further this happens only when a user is doing multiple things wrong so we felt comfortable moving ahead with it.

packages/expo-cli/src/commands/upload/IOSUploader.ts
```diff
@@ -144,6 +144,16 @@ export default class IOSUploader extends BaseUploader {
...
     const result: AppleCreds = pick(this.options, appleCredsKeys);
 
+    if (result.appleId && process.env.EXPO_APPLE_ID) {
+      log.warn(
+        `You've provided contradictory Apple IDs. You should provide either the EXPO_APPLE_ID env or the --apple-id flag, not both. Falling back to --apple-id.`
+      );
+    }
+    if (result.appleIdPassword && process.env.EXPO_APPLE_PASSWORD) {
+      log.warn(
+        `You've provided contradictory Apple IDs. You should provide either the EXPO_APPLE_PASSWORD env or the --apple-id-password flag, not both. Falling back to --apple-id-password.`
```
cruzach, line 154:
```suggestion
        `You've provided contradictory Apple passwords. You should provide either the EXPO_APPLE_PASSWORD env or the --apple-id-password flag, not both. Falling back to --apple-id-password.`
```

jkhales:
> just one typo fix
> 
> > this happens only when a user is doing multiple things wrong
> 
> I wouldn't say passing both the flag and the env variable is necessarily wrong, but if we are changing this then might as well add a [changelog entry](https://github.com/expo/expo-cli/blob/master/CHANGELOG.md) listing this as a bug fix so it's easily identifiable for that **extremely** rare case where someone maybe was relying on this behavior

https://github.com/expo/expo-cli/commit/b245cf5ae554645b2519e66cc6f47c251732fc5a

PR #2393, opened 2020-07-24
[expo-cli] check `when` field when inquirer is used in noninteractive mode

### LINKED ISSUE #2327: `expo fetch:android:keystore` requires a TTY, but doesn't actually take any user input
## Description

`expo fetch:android:keystore` doesn't prompt the user, but incorrectly displays a message indicating that it needs to prompt them when ran without a TTY.

The reason for this is that:

- https://github.com/expo/expo-cli/blob/f44d0d7f82d84f162ace8abe609e0b23ef750363/packages/expo-cli/src/credentials/views/AndroidKeystore.ts#L145 uses `when` to tell `prompt` that it might not need to be executed.
- https://github.com/expo/expo-cli/blob/f44d0d7f82d84f162ace8abe609e0b23ef750363/packages/expo-cli/src/prompt.ts#L13 `prompt` doesn't check `when` before bailing because the input is not interactive.

## Expected Behavior

`expo fetch:android:keystore` works without a TTY.

## Observed Behavior

```
[16:16:30] Input is required, but Expo CLI is in non-interactive mode.
Required input:
> Do you want to display the Android Keystore credentials?
```
## Environment

This will happen in any environment - see description.

## Reproducible Demo

`/dev/null | expo fetch:android:keystore > android-keystore.txt`

wkozyra95:
fixes #2327

### DIFF
--- a/packages/expo-cli/src/prompt.ts
+++ b/packages/expo-cli/src/prompt.ts
@@ -11,8 +11,24 @@ export default function prompt(
   questions: CliQuestions,
   { nonInteractiveHelp }: { nonInteractiveHelp?: string } = {}
 ) {
-  const nQuestions = Array.isArray(questions) ? questions.length : 1;
-  if (program.nonInteractive && nQuestions !== 0) {
+  questions = Array.isArray(questions) ? questions : [questions];
+  const nAllQuestions = questions.length;
+  if (program.nonInteractive) {
+    const nQuestionsToAsk = questions.filter(question => {
+      if (!('when' in question)) {
+        return true;
+      } else if (typeof question.when === 'function') {
+        // if `when` is a function it takes object containing previous answers as argument
+        // in this case we want to detect if any question will be asked, so it
+        // always will be empty object
+        return question.when({});
+      } else {
+        return question.when;
+      }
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/prompt.ts
```diff
@@ -11,8 +11,21 @@ export default function prompt(
...
-  const nQuestions = Array.isArray(questions) ? questions.length : 1;
-  if (program.nonInteractive && nQuestions !== 0) {
+  questions = Array.isArray(questions) ? questions : [questions];
+  const nAllQuestions = questions.length;
+  if (program.nonInteractive) {
+    const nQuestionsToAsk = questions.filter(question => {
+      if (!('when' in question)) {
+        return true;
+      } else if (typeof question.when === 'function') {
+        return question.when({});
```
quinlanj, line 21: It's a bit confusing for me when i read this and i need to look up `inquirer` docs to make sense of what kind of object `when` should be. It'd be helpful if you had some comments explaining that `when` should be :
- boolean 
- function: and if it is a function,  it expects to take the user's answers object, which is empty in the case of `non-interactive`

PR #2401, opened 2020-07-26
feat: use `process.env.METRO_NODE_OPTIONS` when starting Metro

aleclarson:
Instead of using `NODE_OPTIONS` from the `expo-cli` process, the Metro process overrides it with `METRO_NODE_OPTIONS`.

This has two effects:
- the Metro process can be debugged with `--inspect-brk`
- the `expo-cli` process can be debugged with `--inspect-brk` without causing an "address already in use" error in the Metro process

Fixes https://github.com/expo/expo/issues/9398

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -1796,9 +1796,12 @@ export async function startReactNativeServerAsync(
 
   if (options.reset) {
     cliOpts.push('--reset-cache');
-  } // Get custom CLI path from project package.json, but fall back to node_module path
+  }
+
+  // Get custom CLI path from project package.json, but fall back to node_module path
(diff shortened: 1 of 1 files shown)


fson (approved):
Thanks, seems good to me. 👍 

Tip: as an alternative way to make debugging Metro easier, you can also use the new `EXPO_USE_DEV_SERVER=true` environment variable, which makes Metro run in the same process instead of a subprocess. (This feature is currently in beta, but we'll make it the default soon.)

omulet:
The problem is now that the configured NODE_OPTIONS gets discarded completely. So if I previously used `NODE_OPTIONS=--max_old_space_size=4096 expo publish`, now I must use `METRO_NODE_OPTIONS=--max_old_space_size=...`. Not a big issue unless you rely on NODE_OPTIONS in multiple CI workflows.

PR #2415, opened 2020-07-31
[xdl] Add EXPO_TOKEN authentication method

byCedric:
## What I did

This integrates the `EXPO_TOKEN` authentication method, implemented as a `Bearer` authorization token.

- **Added the token retrieval in `UserSettings`**
I think this is the proper place for that. Eventually, it's an environmental setting but still set by the user. We could also move it to the `UserManager` itself, but I think using it from `UserSettings` makes more sense.

- **Added token in the `getSessionAsync` method**
When integrating the robots, I noticed this method is used in other areas to see if people are authenticated. With this, the DevTools also uses the proper username and other authentication stuff (e.g. username).

- ~~**Added to API v1 and v2**~~
~~I can't find traces of API v1 being used somewhere, but I added it using almost identical implementation. Again, the token is fetched from `UserSettings` by the `UserManager` and passed through within the (current) user. This should work fine for `ApiV2Client.clientForUser(user)` calls. For API v1, I added a call to `UserSettings` to fetch the token.~~ _**see section above**._

- **Token has priority over existing authentication session**
E.g. when a user runs this from his or her computer, we probably want to prioritize the `EXPO_TOKEN` value. When the user is already authenticated, this token is used over any existing session.

- **User session data from token isn't stored through `UserSettings`**
I think this is important to mention as well, I don't think any data fetched with the token should be stored. If users provide the token inline, this data should only be "available" during that same execution call (e.g. `EXPO_TOKEN=xxx expo publish`). It also doesn't interfere with existing sessions, falling back right to that when the token isn't set anymore.

## What needs to be done

- ~~**Add (unit) tests**~~ - Done, [see comment](https://github.com/expo/expo-cli/pull/2415#issuecomment-668082834)

- ~~**Double-check the analytics calls for EXPO_TOKEN sessions**~~ Done
Right now, when `ensureLoggedInAsync` is called, it fetches the user. If no current user is loaded yet, it will fetch the info and send out the login analytics event. I'm not sure if this is still valid when using the token. Every call using that token _and using `ensureLoggedInAsync`_ will send out the login event.

- ~~**Consider if we need to fail when using EXPO_TOKEN and `expo logout`**~~ Done
Logging out isn't that relevant here, I don't think we can even do that. As long as that token is defined in the environment, the user will still be authenticated using that token.
> As TC mentioned, this also applies to `logout`, `login`, and `register`

## See it in action

<details>
<summary><code>$ expo whoami</code> with inline and exported token</summary>
whoami example
</details>

<details>
<summary><code>$ expo build:android</code> with inline token</summary>
build android example
</details>

<details>
<summary><code>$ expo build:ios</code> with exported token</summary>
build ios example
</details>

### DIFF
--- a/packages/xdl/src/ApiV2.ts
+++ b/packages/xdl/src/ApiV2.ts
@@ -59,15 +59,20 @@ type QueryParameters = { [key: string]: string | number | boolean | null | undef
 
 type APIV2ClientOptions = {
   sessionSecret?: string;
+  authorizationToken?: string;
 };
 
 export default class ApiV2Client {
   static exponentClient: string = 'xdl';
   sessionSecret: string | null = null;
+  authorizationToken: string | null = null;
 
   static clientForUser(user?: APIV2ClientOptions | null): ApiV2Client {
-    if (user && user.sessionSecret) {
-      return new ApiV2Client({ sessionSecret: user.sessionSecret });
+    if (user && (user.authorizationToken || user.sessionSecret)) {
+      return new ApiV2Client({
+        authorizationToken: user.authorizationToken,
+        sessionSecret: user.sessionSecret,
+      });
     }
 
     return new ApiV2Client();
@@ -78,6 +83,9 @@ export default class ApiV2Client {
   }
 
   constructor(options: APIV2ClientOptions = {}) {
+    if (options.authorizationToken) {
+      this.authorizationToken = options.authorizationToken;
+    }
     if (options.sessionSecret) {
       this.sessionSecret = options.sessionSecret;
     }
@@ -194,7 +202,10 @@ export default class ApiV2Client {
       },
     };
 
-    if (this.sessionSecret) {
+    // Handle auth method, prioritizing authorization tokens before session secrets
+    if (this.authorizationToken) {
+      reqOptions.headers['Authorization'] = `Bearer ${this.authorizationToken}`;
+    } else if (this.sessionSecret) {
       reqOptions.headers['Expo-Session'] = this.sessionSecret;
     }
 
--- a/packages/xdl/src/ApiV2.ts
+++ b/packages/xdl/src/ApiV2.ts
@@ -68,11 +68,8 @@ export default class ApiV2Client {
   accessToken: string | null = null;
 
   static clientForUser(user?: APIV2ClientOptions | null): ApiV2Client {
-    if (user && (user.accessToken || user.sessionSecret)) {
-      return new ApiV2Client({
-        accessToken: user.accessToken,
-        sessionSecret: user.sessionSecret,
-      });
+    if (user) {
+      return new ApiV2Client(user);
     }
 
     return new ApiV2Client();
--- a/packages/xdl/src/ApiV2.ts
+++ b/packages/xdl/src/ApiV2.ts
@@ -59,18 +59,18 @@ type QueryParameters = { [key: string]: string | number | boolean | null | undef
 
 type APIV2ClientOptions = {
   sessionSecret?: string;
-  authorizationToken?: string;
+  accessToken?: string;
 };
 
 export default class ApiV2Client {
   static exponentClient: string = 'xdl';
   sessionSecret: string | null = null;
-  authorizationToken: string | null = null;
+  accessToken: string | null = null;
 
   static clientForUser(user?: APIV2ClientOptions | null): ApiV2Client {
-    if (user && (user.authorizationToken || user.sessionSecret)) {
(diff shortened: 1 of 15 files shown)


packages/xdl/src/ApiV2.ts
```diff
@@ -59,15 +59,20 @@ type QueryParameters = { [key: string]: string | number | boolean | null | undef
...
 
 export default class ApiV2Client {
   static exponentClient: string = 'xdl';
   sessionSecret: string | null = null;
+  authorizationToken: string | null = null;
 
   static clientForUser(user?: APIV2ClientOptions | null): ApiV2Client {
-    if (user && user.sessionSecret) {
-      return new ApiV2Client({ sessionSecret: user.sessionSecret });
+    if (user && (user.authorizationToken || user.sessionSecret)) {
```
byCedric, line 71: I'm not 100% sure why we check on the session secret here, that's already done in the constructor. So that would make these checks obsolete?

packages/xdl/src/UserSettings.ts
```diff
@@ -107,11 +107,16 @@ async function anonymousIdentifier(): Promise<string> {
   return id;
 }
 
+function userToken(): string | null {
```
fson, line 110: Maybe `accessToken` would be a more accurate name for this field? And perhaps then use the same name everywhere (i.e. `authorizationToken` -> `accessToken`.

packages/xdl/src/User.ts
```diff
@@ -195,27 +198,34 @@ export class UserManagerInstance {
...
         return null;
       }
 
       const data = await this._readUserData();
+      const token = UserSettings.userToken();
 
-      // No session, no current user. Need to login
-      if (!data || !data.sessionSecret) {
+      // No token, no session, no current user. Need to login
+      if (!token && (!data || !data.sessionSecret)) {
```
fson, line 216:
```suggestion
      if (!token && !data?.sessionSecret) {
```

packages/xdl/src/UserSettings.ts
byCedric: Done!

byCedric:
@fson I also added some unit tests for the changes I made. It's a combination of the old existing integration tests (with mocked API) and new tests for the token handling. I can remove this if it's just too much. But I'd figure adding some tests for this would be good.

packages/xdl/src/__tests__/User-test.ts
```diff
@@ -0,0 +1,192 @@
...
+    getAsync: jest.fn(),
+    postAsync: jest.fn(),
+    putAsync: jest.fn(),
+    patchAsync: jest.fn(),
+    deleteAsync: jest.fn(),
+    uploadFormDataAsync: jest.fn(),
+    _requestAsync: jest.fn(),
+  };
+
+  (ApiV2.clientForUser as jest.Mock).mockReturnValue(api);
```
byCedric, line 189: Kind of a dirty way to not-having to add this mock every time we need it. If there is a better way, please do correct me 😄

packages/xdl/src/ApiV2.ts
byCedric: Removed the property-checks 😄

packages/xdl/src/__tests__/User-test.ts
fson: Maybe `ApiV2.clientForUser as jest.MockedFunction<typeof ApiV2.clientForUser>` for more accurate typing, but otherwise seems fine 🤷

fson:
The failing test (https://app.circleci.com/pipelines/github/expo/expo-cli/3974/workflows/6dd053cc-c228-4e6b-b0c7-65b862553911/jobs/50658) runs `yarn install` inside a project, so I think it might be occasionally failing due to that taking too long... Not sure if there's a quick solution to this problem (increase the timeout? 💀), but if it this happens often we should definitely fix it (flaky tests will erode the teams' confidence in tests).

byCedric:
We might be able to improve some performance if we ship a lockfile and cache the `node_modules` from the system (e.g. `yarn cache dir`)? That should at least speed it up for a bit, but not always.

byCedric:
Ah, there we go! That works 😄

byCedric:
I refactored the `getSessionAsync` and `getCurrentUsernameAsync` methods so they work better with the devtools. I also noticed in some code `getSessionAsync` is [also used to check if people are authenticated or not](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/DevSession.ts#L44), which works better with this change.

PR #2423, opened 2020-08-05
[webpack-config] fix: handle empty favicons

jaulz:
Currently, there is no possibility to prevent the injection of favicons. This fix allows users to specify an empty favicon and consequently no favicon will be injected nor generated.


jaulz:
Hm, I'm not sure if that test error is related to my changes?

PR #2440, opened 2020-08-09
[xdl] Add robots as new supported user type

byCedric:
## What I did

This integrates the Robot users, as new user type, in the XDL. This PR was built on top of [the `EXPO_TOKEN` integration](https://github.com/expo/expo-cli/pull/2415).

- **Added helper method to find the "owner" of a project**
With the addition of robot users, we can't simply use `manifest.owner || user.username`. Robots don't have usernames and _must_ be used with the manifest owner property. I added a `getProjectOwner` method to "safely" fetch the project owner, and fail if there is none set when using robot users. I think I replaced all "unsafe" owner fetches, but this is good to double-check.

- **Added XDL User methods to only fetch users or robots**
When authenticated as a different user type, the CLI will exit with "This action only supports user type X". Useful to quickly block robot users on unsupported endpoints

- **Test all commands**
Done, see [this doc](https://www.notion.so/expo/Bot-Access-CLI-5c4d04e9034d45b9b9faeea106eaf7e4#6de4705942af4c4ab4f9632085855370).

## What still needs to be done

- ~~**Current credentials API doesn't support robots**~~
~~This is a blocker for running the `$ expo build:*` commands. We need the credentials to start the builds.~~

- ~~**Move integration to EAS**~~
~~After we finished the integration in Expo CLI, I'll work on moving it over to EAS.~~ 
[See PR #108](https://github.com/expo/eas-cli/pull/108)

# Test Plan

**To test manually**
- Create a robot user for your account (see GraphQL query below, not yet available on website)
- Create an access token for your robot user ([using this link](https://staging.expo.io/settings/access-tokens))
- Try smoke testing commands with `EXPO_TOKEN=<robot-token> expo ...` (e.g. `expo whoami`, [this doc contains all tested commands by me](https://www.notion.so/expo/Bot-Access-CLI-5c4d04e9034d45b9b9faeea106eaf7e4#6de4705942af4c4ab4f9632085855370))

<details>
<summary>Create a robot with graphql mutation</summary>
<pre>
mutation {
  robot {
    createRobotForAccount(
      accountID: "{id}"
      robotData: { name: "GLaDOS" }
      permissions: PUBLISH
    ) {
      id
      firstName
    }
  }
}
</pre>
</details>


packages/expo-cli/src/projects.ts
```diff
@@ -50,4 +51,25 @@ async function ensureProjectExistsAsync(
...
+ * This will validate if the owner field is set when using a robot account.
+ */
+function getProjectData(
+  user: User | RobotUser,
+  exp: ExpoConfig
+): Pick<ProjectData, 'accountName' | 'projectName'> {
+  if (user.kind === 'robot' && !exp.owner) {
+    throw new CommandError(
+      'ROBOT_OWNER_ERROR',
+      'The "owner" manifest property is required when using robot users.'
```
fson, line 65: Maybe add a link to https://docs.expo.io/versions/latest/config/app/#owner?

packages/expo-cli/src/projects.ts
```diff
@@ -50,4 +51,25 @@ async function ensureProjectExistsAsync(
...
+    );
+  }
+
+  return {
+    accountName: exp.owner || user.username,
+    projectName: exp.slug,
+  };
```
fson, line 72:
```suggestion
  return exp.owner || user.username;
```
I'd rename this function to `getProjectOwner(user: User | RobotUser, exp: ExpoConfig): string` and only return the project owner. Why? These "bag of fields" functions tend to grow beyond their original intended purpose: they'll be changed to return more fields and will be used to get these fields even if there's a more direct way to get them. A historical example of this:
https://github.com/expo/expo-cli/blob/cd782b4c5b580f0aaaa03229e174b7657bf26a4a/packages/xdl/src/Exp.ts#L314-L316

(Another way to view this is adhering to the "single-responsibility principle".)

Also the `exp.owner || user.username` expression is repeated in many places and it will be useful to have a focused function that can be called to do it. (That also does the check for robots.)

byCedric: Good idea, I renamed it!

packages/expo-cli/src/projects.ts
byCedric: Done!

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -202,13 +202,13 @@ export default function (program: Command) {
         }
 
         let email;
-        if (user) {
+        if (user && user.kind === 'user') {
```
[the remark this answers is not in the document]
byCedric: I did this because typescript doesn't like the alternatives 😅   With the robot user we have 2 types of users, both having the `.kind` property. It also infers the type when using this property. If you use a property that's only defined in a single type, without guarding, it will complain that it might not exist.

Screenshot 2020-08-26 at 15 58 57

Even if you use `if (user && (user as User).email)` or just `if ((user as User)?.email)` it complains later on:

Screenshot 2020-08-26 at 15 59 15

Which is why I used the `.kind` here 😄

byCedric:
After this is merged, I can move it over to EAS!

brentvatne (approved):

EvanBacon (approved):

PR #2462, opened 2020-08-14
[config] allow scheme arrays

EvanBacon:
# Why 

- Some apps will need multiple uri schemes, some also need the ability to add platform specific uri schemes.
- Turtle v1 would add the bundle ID or package name to the native build to enable Google auth compliance by default. Losing this feature would add a fair amount of confusion to future native builds. It's also harder to test how your standalone app works via your ejected app if they have different redirect URIs. 

# How

- Allow `scheme` to be an array or a string
- Add support for `ios.scheme` and `android.scheme`
- Append the `ios.bundleIdentifier` and `android.package` values to the list of native schemes for each respective platform.

# Test Plan

- I updated the unit tests to test for bundle ID and package name getting added, string vs string array, and ensuring that platform URIs don't mix.


giautm:
Hi @EvanBacon , Expo's Doctor said:
```
Error: Problem validating fields in app.json. See https://docs.expo.io/workflow/configuration/
 • Field: scheme - should be string.
```

Alberto-Pouch:
when release?

PR #2474, opened 2020-08-18
[expo-cli] Only run expo service checks from the doctor command

byCedric:
I did 2 things here:

1. This makes _absolutely_ sure the Expo services check is only executed with `$ expo doctor` and not commands like `$ expo start` or `$ expo publish`.
2. Disabled the `exp.host` check, people report on the forum that this might be intermittent issues. Eventually, I could randomly reproduce this locally.


brentvatne:
thanks for the quick turnover here :)

PR #2481, opened 2020-08-20
[expo-cli][xdl] Skip SDK version validation for eas:build

brentvatne:
You may want to run this in an app created with `npx react-native init` and you will not have an Expo SDK version in that context, so we should support that.

### DIFF
--- a/packages/expo-cli/src/credentials/context.ts
+++ b/packages/expo-cli/src/credentials/context.ts
@@ -106,7 +106,9 @@ export class Context {
     this._nonInteractive = options.nonInteractive;
 
     // Check if we are in project context by looking for a manifest
-    const status = await Doctor.validateWithoutNetworkAsync(projectDir);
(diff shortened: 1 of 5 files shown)


packages/expo-cli/src/credentials/context.ts
```diff
@@ -106,7 +106,9 @@ export class Context {
     this._nonInteractive = options.nonInteractive;
 
     // Check if we are in project context by looking for a manifest
-    const status = await Doctor.validateWithoutNetworkAsync(projectDir);
+    const status = await Doctor.validateWithoutNetworkAsync(projectDir, {
+      skipSDKVersionRequirement: true,
+    });
     if (status !== Doctor.FATAL) {
       const { exp } = getConfig(projectDir);
```
wkozyra95, line 113:
```suggestion
      const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
```

PR #2494, opened 2020-08-24
fixed #2489

### LINKED ISSUE #2489: Expo CLI - expo start from powershell, expo-cli cant find my device.
## Description

In powershell we need to escape @ symbol with quatiton mark. And avd device names starts with @ symbol.
Here I described with a simple screenshot. 

## Expected Behavior

I expected to open my android virtual device

## Observed Behavior

expo-cli said

```
emulator: ERROR: No AVD specified. Use '@foo' or '-avd foo' to launch a virtual device named 'foo'
```

## Reproducible Demo

Just run expo-cli from powershell
image

here you can see. The first one is the command executed by expo-cli  but The second command that I edited, worked successfully

c8rdev:
please test it.


brentvatne:
I tested this locally and it does not work on macOS

PR #2495, opened 2020-08-24
Generate iOS icons on eject and apply

EvanBacon:
# How

- Read icons from config on eject and apply
- Utilize image-utils to switch between global sharp-cli instance and jimp. If I'm not mistaken, this'll add a noticeable amount to the package size due to the Jimp instance.
  - This package also caches the images in `.expo` based on the image hash so reapplying is very quick since it only needs to copy the images.
  - Also support for remote URLs as icons works too -- this may not apply to other tools in the CLI.
- Resize icons to squares with white backgrounds to ensure shape and limited transparency
  - Does not fully remove the alpha channel when using with sharp, it does with Jimp though. I'm almost certain transparency cannot be added to the image regardless.
- Use `Contents.json` to reuse icons across platforms.

# Test Plan

- Wrote tests to ensure the expected amount of images are generated and the icon resolution works as expected.


packages/config/src/ios/Icons.ts
```diff
@@ -1,24 +1,209 @@
...
+    const isMarketing = platform.idiom === 'ios-marketing';
+    for (const { size, scales } of platform.sizes) {
+      for (const scale of scales) {
+        // The marketing icon is special because it makes no sense.
+        const filename = isMarketing ? 'ItunesArtwork@2x.png' : getAppleIconName(size, scale);
+        // Only create an image that hasn't already been generated.
+        if (!(filename in generatedIcons)) {
+          const iconSizePx = size * scale;
+
+          // Using this method will cache the images in `.expo` based on the properties used to generate them.
```
brentvatne, line 133: 🔥

packages/config/src/ios/utils/Xcodeproj.ts
```diff
@@ -11,11 +11,15 @@ export function getProjectName(projectRoot: string) {
   return path.basename(sourceRoot);
 }
 
-export function getSourceRoot(projectRoot: string) {
-  const paths = globSync('ios/*/AppDelegate.m', {
+export function getSourceRoot(projectRoot: string): string {
+  // Account for Swift or Objective-C
```
brentvatne, line 15: :+1:

PR #2506, opened 2020-08-26
[eject] Added support for allowBackup

EvanBacon:
# Test Plan

wrote tests

### DIFF
--- a/packages/config/src/android/AllowBackup.ts
+++ b/packages/config/src/android/AllowBackup.ts
@@ -1,0 +1,37 @@
+import { ExpoConfig } from '../Config.types';
+import { Document } from './Manifest';
+
+export function getAllowBackup(config: Pick<ExpoConfig, 'android'>): boolean {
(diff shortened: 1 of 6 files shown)


packages/config/src/android/AllowBackup.ts
```diff
@@ -0,0 +1,37 @@
+import { ExpoConfig } from '../Config.types';
+import { Document } from './Manifest';
+
+export function getAllowBackup(config: Pick<ExpoConfig, 'android'>): boolean {
+  // Defaults to true.
+  // https://docs.expo.io/versions/latest/config/app/#allowbackup
```
brentvatne, line 6: @cruzach - can you confirm that this is intentional?

cruzach: yep, bc we want to match Android's default for `android:allowBackup`- https://developer.android.com/guide/topics/manifest/application-element

PR #2509, opened 2020-08-26
[expo-cli] Fixed default bare project name for 'expo init'

### LINKED ISSUE #2472: `expo init` gives invalid default name for bare workflow
## Description

When using `expo init`, after selecting any **bare workflow** template, it prompts for project name. I'd like to leave default (`my-app`) but it doesn't pass validation.

That's because default `my-app`:

https://github.com/expo/expo-cli/blob/e7bf70aff0c05ddcbb3d885fd52bc486e03a4830/packages/expo-cli/src/commands/init.ts#L498-L504

Doesn't match regex in:
https://github.com/expo/expo-cli/blob/e7bf70aff0c05ddcbb3d885fd52bc486e03a4830/packages/expo-cli/src/commands/init.ts#L468-L472

I could fix that myself and make a PR, but I'm not sure if I should change regex to match this name or maybe change default name to match regex 🤷

In _managed workflow_, however, default `my-app` is validated properly, because its regex pattern is
https://github.com/expo/expo-cli/blob/e7bf70aff0c05ddcbb3d885fd52bc486e03a4830/packages/expo-cli/src/commands/init.ts#L458

## Expected Behavior

Default name passes the regex pattern

## Observed Behavior

```bash
~/dev/temp » expo init
? Choose a template: expo-template-bare-typescript
? What would you like to name your app? (my-app)
>> Project name can only include ASCII characters A-Z, a-z and 0-9
```

## Environment

`expo-cli@3.24.0`

## Reproducible Demo

`expo init -t bare-minimum` then when prompted for project name, just press Enter to leave default.

barthap:
Fixes #2472 

Changed default name for bare projects to match regex pattern in `expo init`.

### DIFF
--- a/packages/expo-cli/src/commands/init.ts
+++ b/packages/expo-cli/src/commands/init.ts
@@ -498,7 +498,7 @@ async function promptForBareConfig(
     ({ projectName } = await prompt({
       name: 'projectName',
       message: 'What would you like to name your app?',
-      default: 'my-app',
+      default: 'MyApp',
       filter: (name: string) => name.trim(),
       validate: (name: string) => validateProjectName(name),
     }));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This is the log of notable changes to Expo CLI and related packages.
 ### 🎉 New features
 
 - [expo-cli]: EAS Build: warn user when credentials are not git ignored ([#2482](https://github.com/expo/expo-cli/pull/2482)) by [@wkozyra95](https://github.com/wkozyra95)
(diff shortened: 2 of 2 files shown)


wkozyra95 (approved):

packages/expo-cli/src/commands/init.ts
```diff
@@ -498,7 +498,7 @@ async function promptForBareConfig(
     ({ projectName } = await prompt({
       name: 'projectName',
       message: 'What would you like to name your app?',
-      default: 'my-app',
+      default: 'MyApp',
```
EvanBacon, line 501: y tho?

barthap: See #2472

bbarthec (approved):

PR #2526, opened 2020-08-28
[dev-server] use minify in prod

EvanBacon:
Without this, bundles are ~2x larger on hello world projects.


brentvatne:
cc @fson - we might just want to do this during publish explicitly, i'm not sure if it's desirable to enable minify when running with dev=false for a local development server typically because it would slow it down a bit

PR #2531, opened 2020-08-28
[cli] Remove web push commands

EvanBacon:
Web push notifications weren't completed and these commands aren't documented anywhere. The help command is pretty hard to digest, this removes four commands from it.


packages/expo-cli/src/commands/push-creds-web.ts
```diff
@@ -0,0 +1,197 @@
...
+type VapidData = {
+  vapidSubject: string;
+  vapidPubkey?: string;
+  vapidPvtkey?: string;
+};
+
+const vapidSubjectDescription =
+  'URL or `mailto:` URL which provides a point of contact in case the push service needs to contact the message sender.';
+
+export default function (program: Command) {
```
fson, line 18: Looks like instead of removing the commands, they've currently been moved to this file? Or am I missing something?

EvanBacon: seems like an elegant way to just remove them without deleting altogether.

PR #2539, opened 2020-08-28
[xdl] pass app.json updates config to expo-updates in standalone apps

esamelson:
Fixes https://github.com/expo/expo/issues/9922, along with https://github.com/expo/expo/pull/9985

This PR passes the `updates.checkAutomatically` and `updates.fallbackToCacheTimeout` app.json values, which were previously set at runtime, into build-time configuration values for standalone apps built by turtle. These values will then be read and passed directly to expo-updates by the ExpoKit kernels.

Further, starting with SDK 39 the filenames that the ExpoKit kernel expects for the embedded manifest and bundle have changed, from `shell-app-manifest.json` and `shell-app.bundle` to `app.manifest` and `app.bundle`. This PR makes that change accordingly for building standalone apps.

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -43,22 +43,6 @@ function xmlWeirdAndroidEscape(original) {
   return replaceString(noApos, "'", "\\'");
 }
 
-function getManifestFileNameForSdkVersion(sdkVersion) {
-  if (parseSdkMajorVersion(sdkVersion) < 39) {
-    return 'shell-app-manifest.json';
-  } else {
-    return 'app.manifest';
-  }
-}
-
-function getBundleFileNameForSdkVersion(sdkVersion) {
-  if (parseSdkMajorVersion(sdkVersion) < 39) {
-    return 'shell-app.bundle';
-  } else {
-    return 'app.bundle';
-  }
-}
-
 exports.updateAndroidShellAppAsync = async function updateAndroidShellAppAsync(args) {
   let { url, sdkVersion, releaseChannel, workingDir } = args;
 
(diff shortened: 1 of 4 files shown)


packages/xdl/src/detach/IosNSBundle.ts
```diff
@@ -490,6 +506,13 @@ async function _configureShellPlistAsync(context: AnyStandaloneContext): Promise
       // enable/disable code push if the developer provided specific behavior
       shellPlist.areRemoteUpdatesEnabled = config.updates.enabled;
     }
+    if (config.updates && config.updates.hasOwnProperty('checkAutomatically')) {
```
EvanBacon, line 509:
```suggestion
    if (config.updates?.hasOwnProperty('checkAutomatically')) {
```

packages/xdl/src/detach/IosNSBundle.ts
```diff
@@ -32,6 +32,22 @@ const DEFAULT_FABRIC_KEY = '81130e95ea13cd7ed9a4f455e96214902c721c99';
 const DEFAULT_GAD_APPLICATION_ID = 'ca-app-pub-3940256099942544~1458002511';
 const KERNEL_URL = 'https://expo.io/@exponent/home';
 
+export function getManifestFileNameForSdkVersion(sdkVersion: string) {
```
EvanBacon, line 35: Why are these methods defined twice?

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -649,6 +690,42 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
       )
     );
   }
+  if (parseSdkMajorVersion(sdkVersion) >= 39 && updatesCheckAutomaticallyDisabled) {
```
EvanBacon, line 693: This seems like an expensive operation to run often

packages/xdl/src/detach/IosNSBundle.ts
esamelson: Once per platform, but I am happy to extract them to a separate util module

packages/xdl/src/detach/AndroidShellApp.js
esamelson: OK, I added a rough memoization for the majorSdkVersion value in the `runShellAppModificationsAsync` method (assuming that's the operation you're talking about here).

bbarthec (approved):

PR #2573, opened 2020-09-04
feat(webpack-config): add util function getRulesByMatchingFile

jaulz:
This PR adds a util function to retrieve rules using a single file name. So far there is only `getRulesByMatchingFiles` which expects an array and which makes it unnecessary complex to retrieve rules when dealing with one file name only.


EvanBacon:
If the method isn't used by the webpack config then it shouldn't be included, this is important for keeping bundle size concise.

jaulz:
Hm, I don't get the point. It has the advantage that it can be used by developers to adjust the config, it is cleaner and it is still internally used by `getRulesByMatchingFiles` so there is no real effect on the bundle size.

jaulz:
@EvanBacon could you please have another look at this one?

PR #2577, opened 2020-09-04
[xdl] split manifest handler out of Project

EvanBacon:
- Project is too hard to work with
- Moved manifest handler middleware methods into a new module and added basic tests for the manifest results.
- Organized the manifest middleware in a cleaner way and split functionality up so it's easier to read.


fson (approved):
Thanks! This is certainly an improvement over having all this in `Project.ts`. 👍

PR #2581, opened 2020-09-04
[cli] Add two-factor authentication to login

wschurman:
# Why

This adds CLI support for the upcoming two-factor authentication feature.

## Feature overview: 

Users are able to configure a set of devices to act as second factor authentication mechanisms. There are two kinds of devices: Authenticator TOTP and SMS. A user has the option to have a "primary" device. A user is also given backup codes during 2FA setup.

Currently there is no public way to set up 2FA since it is still in active development.

# How

See inline docblocks.

# Test Plan

Start server locally with https://github.com/expo/universe/pull/5733 patched.

Set up and confirm 2FA for my user in GraphiQL, adding a default authenticator and backup SMS.

```
export EXPO_LOCAL=1

> yarn expo login --username wschurman2 --password <password_in_plain_text_here> --otp 478045
Success. You are now logged in as wschurman2.

> yarn expo logout
Logged out

> yarn expo login
? Username/Email Address: wschurman2
? Password: [hidden]
One-time password from authenticator required.
? One-time Password or Backup Code (press enter for other options): 134243
Success. You are now logged in as wschurman2.

> yarn expo logout
Logged out

> yarn expo login
? Username/Email Address: wschurman2
? Password: [hidden]
One-time password from authenticator required.
? One-time Password or Backup Code (press enter for other options): <press enter, menu showing all devices is shown in next step>
? Select a device to send a one-time password: Phone ending in **74
? One-time Password or Backup Code (press enter to cancel): 769422
Success. You are now logged in as wschurman2.

> yarn expo login
? Username/Email Address: wschurman2
? Password: [hidden]
One-time password from authenticator required.
? One-time Password or Backup Code (press enter for other options):
? Select a device to send a one-time password: Cancel
User cancelled login
```

Also, run the new integration test.

### DIFF
--- a/packages/expo-cli/src/accounts.ts
+++ b/packages/expo-cli/src/accounts.ts
@@ -112,7 +112,7 @@ export async function login(options: CommandOptions): Promise<User> {
  */
 async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<string | null> {
   const enterMessage =
-    cancelBehavior === 'cancel' ? 'press enter to cancel' : 'press enter for other options';
+    cancelBehavior === 'cancel' ? 'press Enter to cancel' : 'press Enter for other options';
   const otpQuestion: Question = {
     type: 'input',
     name: 'otp',
@@ -120,7 +120,7 @@ async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<st
   };
 
   const { otp } = await prompt(otpQuestion);
-  if (!otp || otp === '') {
+  if (!otp) {
     return null;
   }
 
@@ -129,7 +129,7 @@ async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<st
 
 /**
  * Prompt for user to choose a backup OTP method. If selected method is SMS, a request
- * for a new  will be sent to that method. Then, prompt for the OTP, and retry the user login.
+ * for a new OTP will be sent to that method. Then, prompt for the OTP, and retry the user login.
  */
 async function _promptForBackupOTPAsync(
   username: string,
@@ -139,11 +139,11 @@ async function _promptForBackupOTPAsync(
   const nonPrimarySecondFactorDevices = secondFactorDevices.filter(device => !device.is_primary);
 
   if (nonPrimarySecondFactorDevices.length === 0) {
-    throw new Error('No other second factor devices set up for user');
+    throw new Error('No other second-factor devices set up');
   }
 
-  const hasOtherAuthenticatorSecondFactorDevice = !!nonPrimarySecondFactorDevices.find(
-    device => device.method === UserSecondFactorDeviceMethod.AUTHENTICATOR
+  const hasNonAuthenticatorSecondFactorDevice = nonPrimarySecondFactorDevices.find(
+    device => device.method !== UserSecondFactorDeviceMethod.AUTHENTICATOR
   );
 
   const smsNonPrimarySecondFactorDevices = nonPrimarySecondFactorDevices.filter(
@@ -154,11 +154,11 @@ async function _promptForBackupOTPAsync(
   const cancelChoiceSentinel = -2;
 
   const deviceChoices = smsNonPrimarySecondFactorDevices.map((device, idx) => ({
-    name: `Phone ending in ${device.sms_phone_number}`,
+    name: device.sms_phone_number!,
     value: idx,
   }));
 
-  if (hasOtherAuthenticatorSecondFactorDevice) {
+  if (hasNonAuthenticatorSecondFactorDevice) {
     deviceChoices.push({
       name: 'Authenticator',
       value: authenticatorChoiceSentinel,
@@ -173,14 +173,14 @@ async function _promptForBackupOTPAsync(
   const question: Question = {
     type: 'list',
     name: 'choice',
-    message: 'Select a second factor device:',
+    message: 'Select a second-factor device:',
     choices: deviceChoices,
     pageSize: Infinity,
   };
 
   const { choice } = await prompt(question);
   if (choice === cancelChoiceSentinel) {
-    throw new Error('User cancelled login');
+    throw new Error('Cancelled login');
   } else if (choice === authenticatorChoiceSentinel) {
     return await _promptForOTPAsync('cancel');
   }
@@ -220,13 +220,7 @@ async function _retryUsernamePasswordAuthWithOTPAsync(
     smsAutomaticallySent?: boolean;
   }
 ): Promise<User> {
-  const {
-    secondFactorDevices,
-    smsAutomaticallySent,
-  }: {
-    secondFactorDevices?: SecondFactorDevice[];
-    smsAutomaticallySent?: boolean;
-  } = metadata as any;
+  const { secondFactorDevices, smsAutomaticallySent } = metadata;
   invariant(
     secondFactorDevices !== undefined && smsAutomaticallySent !== undefined,
     'malformed OTP error metadata'
@@ -257,7 +251,7 @@ async function _retryUsernamePasswordAuthWithOTPAsync(
   }
 
   if (!otp) {
-    throw new Error('User cancelled login');
+    throw new Error('Cancelled login');
   }
 
   return await UserManager.loginAsync('user-pass', {
--- a/packages/expo-cli/src/accounts.ts
+++ b/packages/expo-cli/src/accounts.ts
@@ -7,7 +7,7 @@ import invariant from 'invariant';
 import CommandError from './CommandError';
 import log from './log';
 import prompt, { Question } from './prompt';
-import { confirmAsync } from './prompts';
+import promptNew, { confirmAsync, Question as NewQuestion, selectAsync } from './prompts';
 
 UserManager.initialize();
 
@@ -20,12 +20,12 @@ type CommandOptions = {
   };
 };
 
-enum UserSecondFactorDeviceMethod {
+export enum UserSecondFactorDeviceMethod {
   AUTHENTICATOR = 'authenticator',
   SMS = 'sms',
 }
 
-type SecondFactorDevice = {
+export type SecondFactorDevice = {
   id: string;
   method: UserSecondFactorDeviceMethod;
   sms_phone_number: string | null;
@@ -112,14 +112,16 @@ export async function login(options: CommandOptions): Promise<User> {
  */
 async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<string | null> {
   const enterMessage =
-    cancelBehavior === 'cancel' ? 'press Enter to cancel' : 'press Enter for other options';
-  const otpQuestion: Question = {
-    type: 'input',
+    cancelBehavior === 'cancel'
+      ? `press ${log.chalk.bold('Enter')} to cancel`
+      : `press ${log.chalk.bold('Enter')} for more options`;
+  const otpQuestion: NewQuestion = {
+    type: 'text',
     name: 'otp',
-    message: `One-time Password or Backup Code (${enterMessage}):`,
+    message: `One-time password or backup code (${enterMessage}):`,
   };
 
-  const { otp } = await prompt(otpQuestion);
+  const { otp } = await promptNew(otpQuestion);
   if (!otp) {
     return null;
   }
@@ -139,7 +141,10 @@ async function _promptForBackupOTPAsync(
   const nonPrimarySecondFactorDevices = secondFactorDevices.filter(device => !device.is_primary);
 
   if (nonPrimarySecondFactorDevices.length === 0) {
-    throw new Error('No other second-factor devices set up');
+    log.warn(
+      'No other second-factor devices set up. Ensure you have set up and certified a backup device.'
+    );
+    process.exit(1);
   }
 
   const hasAuthenticatorSecondFactorDevice = nonPrimarySecondFactorDevices.find(
@@ -154,38 +159,35 @@ async function _promptForBackupOTPAsync(
   const cancelChoiceSentinel = -2;
 
   const deviceChoices = smsNonPrimarySecondFactorDevices.map((device, idx) => ({
-    name: device.sms_phone_number!,
+    title: device.sms_phone_number!,
     value: idx,
   }));
 
   if (hasAuthenticatorSecondFactorDevice) {
     deviceChoices.push({
-      name: 'Authenticator',
+      title: 'Authenticator',
       value: authenticatorChoiceSentinel,
     });
   }
 
   deviceChoices.push({
-    name: 'Cancel',
+    title: 'Cancel',
     value: cancelChoiceSentinel,
   });
 
-  const question: Question = {
-    type: 'list',
-    name: 'choice',
+  const question = {
     message: 'Select a second-factor device:',
     choices: deviceChoices,
-    pageSize: Infinity,
   };
 
-  const { choice } = await prompt(question);
-  if (choice === cancelChoiceSentinel) {
-    throw new Error('Cancelled login');
-  } else if (choice === authenticatorChoiceSentinel) {
+  const selectedValue = await selectAsync(question);
+  if (selectedValue === cancelChoiceSentinel) {
+    return null;
+  } else if (selectedValue === authenticatorChoiceSentinel) {
     return await _promptForOTPAsync('cancel');
   }
 
-  const device = smsNonPrimarySecondFactorDevices[choice];
+  const device = smsNonPrimarySecondFactorDevices[selectedValue];
 
   const apiAnonymous = ApiV2.clientForUser();
   await apiAnonymous.postAsync('auth/send-sms-otp', {
@@ -212,7 +214,7 @@ async function _promptForBackupOTPAsync(
  *    we should show a picker of the SMS devices that they can have an OTP code sent to, and when
  *    the user picks one we show a prompt for the sent OTP.
  */
-async function _retryUsernamePasswordAuthWithOTPAsync(
+export async function _retryUsernamePasswordAuthWithOTPAsync(
   username: string,
   password: string,
   metadata: {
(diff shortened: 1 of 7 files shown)


packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  if (!otp || otp === '') {
+    return null;
+  }
+
+  return otp;
+}
+
+/**
+ * Prompt for user to choose a backup OTP method. If selected method is SMS, a request
+ * for a new  will be sent to that method. Then, prompt for the OTP, and retry the user login.
```
ide, line 135:
```suggestion
 * for a new OTP will be sent to that method. Then, prompt for the OTP, and retry the user login.
```
```suggestion
 * for a new  will be sent to that method. Then, prompt for the OTP, and retry the user login.
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+ */
+async function _promptForBackupOTPAsync(
+  username: string,
+  password: string,
+  secondFactorDevices: SecondFactorDevice[]
+): Promise<string | null> {
+  const nonPrimarySecondFactorDevices = secondFactorDevices.filter(device => !device.is_primary);
+
+  if (nonPrimarySecondFactorDevices.length === 0) {
+    throw new Error('No other second factor devices set up for user');
```
ide, line 145: Since these error messages are user-facing, try to phrase this to be more directed at the user e.g.
```suggestion
    throw new Error('There are no other second-factor devices set up for you');
```
```suggestion
    throw new Error('No other second factor devices set up for user');
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+
+  const smsNonPrimarySecondFactorDevices = nonPrimarySecondFactorDevices.filter(
+    device => device.method === UserSecondFactorDeviceMethod.SMS
+  );
+
+  const authenticatorChoiceSentinel = -1;
+  const cancelChoiceSentinel = -2;
+
+  const deviceChoices = smsNonPrimarySecondFactorDevices.map((device, idx) => ({
+    name: `Phone ending in ${device.sms_phone_number}`,
```
ide, line 160: This prompt looks OK but might want to change the phrasing if we change the masking format (comment is on the server-side PR).

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+
+  deviceChoices.push({
+    name: 'Cancel',
+    value: cancelChoiceSentinel,
+  });
+
+  const question: Question = {
+    type: 'list',
+    name: 'choice',
+    message: 'Select a second factor device:',
```
ide, line 179:
```suggestion
    message: 'Select a second-factor device:',
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+    choices: deviceChoices,
+    pageSize: Infinity,
+  };
+
+  const { choice } = await prompt(question);
+  if (choice === cancelChoiceSentinel) {
+    throw new Error('User cancelled login');
```
ide, line 186: This is a little more user-facing by not referring to "user in the third person:
```suggestion
    throw new Error('Cancelled login');
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+    smsAutomaticallySent?: boolean;
+  }
+): Promise<User> {
+  const {
+    secondFactorDevices,
+    smsAutomaticallySent,
+  }: {
+    secondFactorDevices?: SecondFactorDevice[];
+    smsAutomaticallySent?: boolean;
+  } = metadata as any;
```
ide, line 232: `metadata` is already typed in the param list, most of these types seem unnecessary

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+
+  const primaryDevice = secondFactorDevices.find(device => device.is_primary);
+  let otp: string | null = null;
+
+  if (smsAutomaticallySent) {
+    invariant(
+      primaryDevice,
+      'OTP should only automatically be sent when there is a primary device'
+    );
+    console.log(
```
ide, line 246: I'm not familiar with this codebase but I would double check that `console.log` is the established way to write to the terminal instead of via some other logging API

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,171 @@ export async function login(options: CommandOptions): Promise<User> {
...
+    otp = await _promptForOTPAsync('menu');
+  }
+
+  // user bailed on case 1 or 2, wants to move to case 3
+  if (!otp) {
+    otp = await _promptForBackupOTPAsync(username, password, secondFactorDevices);
+  }
+
+  if (!otp) {
+    throw new Error('User cancelled login');
```
ide, line 263:
```suggestion
    throw new Error('Cancelled login');
```
```suggestion
    throw new Error('User cancelled login');
```

ide (changes requested):
@fson Could you please review this too? We'll also need 2FA support with EAS CLI and will need to support the same type of login flows and error handling.

ide:
@EvanBacon Could you take a look at the "Test Plan" section in the PR description and see if the flow and the wording sounds good, or if you think there are some tweaks we should make to it?

packages/expo-cli/src/accounts.ts
wschurman: Ah, good call, this was leftover from refactoring.

packages/expo-cli/src/accounts.ts
wschurman: As far as I can tell via grepping it is the correct method.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  password: string,
+  metadata: {
+    secondFactorDevices?: SecondFactorDevice[];
+    smsAutomaticallySent?: boolean;
+  }
+): Promise<User> {
+  const { secondFactorDevices, smsAutomaticallySent } = metadata;
+  invariant(
+    secondFactorDevices !== undefined && smsAutomaticallySent !== undefined,
+    'malformed OTP error metadata'
```
brentvatne, line 229: should we include the stringified metadata in the message so we have more context if a user reports this happening?

brentvatne (commented):
looks good! excited to have 2fa soon

- this will need to be updated if https://github.com/expo/expo-cli/pull/2605 lands first
- we should use [`log`](https://github.com/expo/expo-cli/blob/90bab4a0c6ab8f39a5dccf71f395e0cca7b8f626/packages/expo-cli/src/log.ts#L91-L95) instead of `console.log` directly - this just makes sure that we don't end up interrupting any progress bars / spinners that might be active. if you can change other `console.log`s that were leftover in this file prior to your changes that'd be great too

ide:
Generally speaking this looks good to me and consider this a soft approval. Recusing myself from review so we don't run the risk of no single reviewer looking at this carefully. @fson please look this over (and we'll need similar code for EAS CLI); @EvanBacon please review whether the DX / workflow seems good to you (shown in the test plan) or if @brentvatne is happy with it that works for me.

packages/expo-cli/src/accounts.ts
wschurman: it could be a fairly large object but I think something along these lines is a good idea. I'll dig into how we do this in other places and update this.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+): Promise<string | null> {
+  const nonPrimarySecondFactorDevices = secondFactorDevices.filter(device => !device.is_primary);
+
+  if (nonPrimarySecondFactorDevices.length === 0) {
+    throw new Error('No other second-factor devices set up');
+  }
+
+  const hasNonAuthenticatorSecondFactorDevice = nonPrimarySecondFactorDevices.find(
+    device => device.method !== UserSecondFactorDeviceMethod.AUTHENTICATOR
+  );
```
wschurman, line 150: hmm, this logic is incorrect now. This should be `hasAuthenticatorSecondFactorDevice` and the condition should be inverted.

wschurman:
> we should use log instead of console.log directly - this just makes sure that we don't end up interrupting any progress bars / spinners that might be active. if you can change other console.logs that were leftover in this file prior to your changes that'd be great too

I'll make this change in a follow-up PR to keep the scope of this one as narrow as possible.

wschurman:
Still looking for review on this from @fson and/or @EvanBacon. The two branches for disallowing console and enforcing no lint warnings in CI were merged here because github is bad at stacked PRs and should feel bad. I'll un-merge them before squash landing and re-PR those branches and land them separately.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
   }
 }
 
-async function _usernamePasswordAuth(username?: string, password?: string): Promise<User> {
+/**
+ * Prompt for an OTP with the option to cancel the question by answering empty (pressing return key).
+ */
+async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<string | null> {
+  const enterMessage =
+    cancelBehavior === 'cancel' ? 'press Enter to cancel' : 'press Enter for other options';
```
EvanBacon, line 118:
```suggestion
    cancelBehavior === 'cancel' ? `press ${log.chalk.bold('Enter')} to cancel` : `press ${log.chalk.bold('Enter')} for other options`;
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  const otpQuestion: Question = {
+    type: 'input',
+    name: 'otp',
+    message: `One-time Password or Backup Code (${enterMessage}):`,
+  };
+
+  const { otp } = await prompt(otpQuestion);
```
EvanBacon, line 125: Use `prompts` instead of `prompt`, prompts uses a substantially smaller prompting package.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  }
+
+  return otp;
+}
+
+/**
+ * Prompt for user to choose a backup OTP method. If selected method is SMS, a request
+ * for a new OTP will be sent to that method. Then, prompt for the OTP, and retry the user login.
+ */
+async function _promptForBackupOTPAsync(
```
EvanBacon, line 137: maybe this method could be split up so the aggregation for the question and the actual interaction/api call are in two different methods, then you should be able to easily write a unit test for the aggregation.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  metadata: {
+    secondFactorDevices?: SecondFactorDevice[];
+    smsAutomaticallySent?: boolean;
+  }
+): Promise<User> {
+  const { secondFactorDevices, smsAutomaticallySent } = metadata;
+  invariant(
+    secondFactorDevices !== undefined && smsAutomaticallySent !== undefined,
+    `Malformed OTP error metadata: ${metadata}`
+  );
```
EvanBacon, line 230: if this is not an unexpected error, then be sure the error message looks actionable so a user knows what to do next.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+
+  const primaryDevice = secondFactorDevices.find(device => device.is_primary);
+  let otp: string | null = null;
+
+  if (smsAutomaticallySent) {
+    invariant(
+      primaryDevice,
+      'OTP should only automatically be sent when there is a primary device'
```
EvanBacon, line 238: Since we know the cause of the problem, perhaps we could format this more like a warning rather than an error. For example, when someone runs `expo start --config` instead of throwing an invalid type, we warn how to use flag and exit with non-zero. https://github.com/expo/expo-cli/blob/50f98f66a52cb3371aa580d9d504d465ce8558a6/packages/expo-cli/src/exp.ts#L418-L422 the lack of generic error and a stack trace helps people know that CLI isn't buggy.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+    );
+    console.log(
+      `One-time password was sent to the phone number ending in ${primaryDevice.sms_phone_number}.`
+    );
+    otp = await _promptForOTPAsync('menu');
+  }
+
+  if (primaryDevice?.method === UserSecondFactorDeviceMethod.AUTHENTICATOR) {
+    console.log(`One-time password from authenticator required.`);
```
EvanBacon, line 247:
```suggestion
    log(`One-time password from authenticator required.`);
```

This helps for deferring logs when something with higher priority is currently in use.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+    otp = await _promptForOTPAsync('menu');
+  }
+
+  // user bailed on case 1 or 2, wants to move to case 3
+  if (!otp) {
+    otp = await _promptForBackupOTPAsync(username, password, secondFactorDevices);
+  }
+
+  if (!otp) {
+    throw new Error('Cancelled login');
```
EvanBacon, line 257: as above, so below :D because the user cancelled, this is expected behavior and not an error. We should log something with default styles and exit the process.

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+/**
+ * Prompt for an OTP with the option to cancel the question by answering empty (pressing return key).
+ */
+async function _promptForOTPAsync(cancelBehavior: 'cancel' | 'menu'): Promise<string | null> {
+  const enterMessage =
+    cancelBehavior === 'cancel' ? 'press Enter to cancel' : 'press Enter for other options';
+  const otpQuestion: Question = {
+    type: 'input',
+    name: 'otp',
+    message: `One-time Password or Backup Code (${enterMessage}):`,
```
EvanBacon, line 122:
```suggestion
    message: `One-time password or backup code (${enterMessage}):`,
```

packages/expo-cli/src/accounts.ts
```diff
@@ -95,7 +110,165 @@ export async function login(options: CommandOptions): Promise<User> {
...
+ */
+async function _promptForBackupOTPAsync(
+  username: string,
+  password: string,
+  secondFactorDevices: SecondFactorDevice[]
+): Promise<string | null> {
+  const nonPrimarySecondFactorDevices = secondFactorDevices.filter(device => !device.is_primary);
+
+  if (nonPrimarySecondFactorDevices.length === 0) {
+    throw new Error('No other second-factor devices set up');
```
EvanBacon, line 145:
```suggestion
    log.warn(`You need to set up two factor authentication before you can use a OTP. ${log.chalk.dim(learnMore('<link to docs or fyi.expo>'))}`);
    process.exit(1);
```
Maybe create an FYI or link to docs about how to set up a device.

packages/expo-cli/src/accounts.ts
wschurman: This is an unexpected error. It'd be thrown when someone in the future makes an erroneous (non-backwards-compatible) change on the server, and when that happens we want to find out about it in sentry. I'm assuming throwing an invariant will go to our sentry but if there's a better way to log this let me know.

packages/expo-cli/src/accounts.ts
wschurman: This is also a constraint that is easy to express on the server but hard to express over an API. Essentially, when an SMS is automatically sent that means that there is a primary device and it is an SMS device. The invariant here is to appease TypeScript to ensure that this is non-null for use.

packages/expo-cli/src/accounts.ts
wschurman: This error occurs when the user only has one second factor device set up and chooses the option (`press Enter for more options`) in the prompt. We still want them to be shown this option though even if they only have one second factor device as they could be in a state where they have a second one that is in the process of being setup but is not yet certified/confirmed, in which case this error will help them debug their issue. I'll alter the wording a bit.

wschurman: I'll make a task to add in a link once we have documentation.

fson (changes requested):
Nice work, especially thank you for being thoughtful about different error conditions and how to handle them.

I think we need to avoid calling `process.exit()` in the login prompt functions as this can interfere with the dev server / Expo Dev Tools, which can also trigger the login prompt. Otherwise everything seems good and well tested, so consider this PR approved once `process.exit` is fixed.

packages/expo-cli/src/accounts.ts
```diff
@@ -92,7 +107,168 @@ export async function login(options: CommandOptions): Promise<User> {
...
+  }
+
+  // user bailed on case 1 or 2, wants to move to case 3
+  if (!otp) {
+    otp = await _promptForBackupOTPAsync(username, password, secondFactorDevices);
+  }
+
+  if (!otp) {
+    log.warn('Cancelled login');
+    process.exit(1);
```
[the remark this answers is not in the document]
wschurman: Sounds good! Will proceed with `CommandError`.

cc @EvanBacon for alignment as this contradicts the requested change when an error was previously thrown here: https://github.com/expo/expo-cli/pull/2581#discussion_r491070772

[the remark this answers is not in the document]
EvanBacon: I suppose you can put better styling here then https://github.com/expo/expo-cli/blob/ff0014cc5e48acb6c86eeb0de02ba54a38427ee6/packages/expo-cli/src/exp.ts#L308 really no other tools should be calling into a CLI unless they're forking a new process anyways.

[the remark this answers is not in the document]
wschurman: Looks like the styling is already the same as `log.warn` (orange in my terminal) so I think we're good to go.

fson (approved):

PR #2589, opened 2020-09-07
[cli] Use update-check for updates

EvanBacon:
- We use `update-check` in all of our other CLIs.
- Removes a bunch of dependency bloat and another XDL module.
- Simplifies the upgrade message.
- Prompts to use yarn for global installs based on `expo/package-manager`.
- Removes deprecated support, not sure how important this is but it seems like it'll be tedious to add.

Screen Shot 2020-09-07 at 12 07 50 PM


packages/expo-cli/src/update.ts
```diff
@@ -1,16 +1,37 @@
...
+// src directory and would change the directory structure of the emitted files under the build
+// directory
 const packageJSON = require('../package.json');
 
-const ModuleVersionChecker = ModuleVersion.createModuleVersionChecker(
-  packageJSON.name,
-  packageJSON.version
-);
+export default async function shouldUpdate(): Promise<void> {
+  const update = checkForUpdate(packageJSON, { interval: 1 }).catch(() => null);
```
brentvatne, line 13: the interval is in ms here, so this means we wouldn't cache the update at all. if we leave out the interval option it should default to one day, which is what we had before.

brentvatne:
aside from making sure we cache the latest version for some reasonable interval this pr lgtm

packages/expo-cli/src/update.ts
EvanBacon, line 13:
```suggestion
  const update = checkForUpdate(packageJSON).catch(() => null);
```

brentvatne:
update from previous comment after discussing this on call - would be good to know about deprecated flag too, so we can more strongly encourage an update

fson:
Vercel no longer uses `update-check` in their own CLI: https://github.com/vercel/vercel/commit/843be9658c285fa73a9461d6fa4f9e2324293f03 🤔 I wonder if they're going to keep maintaining it.

PR #2603, opened 2020-09-08
[expo-cli] EAS Build support `experimental.npmToken` in credentials.json

wkozyra95:
**[do not merge before turtle changes are deployed]**

# Why

Support for private packages 

# How

add `experimental.npmToken` field to credentials.json

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to Expo CLI and related packages.
 
 ### 🎉 New features
 
+- [expo-cli] EAS Build: add `experimental.npmToken` to `credentials.json` [#2603](https://github.com/expo/expo-cli/pull/2603)
+
 ### 🐛 Bug fixes
 
 ## [Thu, 8 Sep 2020 14:30:14 +0200](https://github.com/expo/expo-cli/commit/f0e24ee436806c109c19329c7e161fee0d2f0c81)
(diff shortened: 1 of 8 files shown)


brentvatne:
something doesn't feel quite right about having credentials.json also contain arbitrary env vars. for comparison, both netlify and vercel provide two options for setting env vars:

1) stored in config files: set in config files (vercel.json / netlify.toml) - this is useful for env vars that are not secrets
2) stored on backend: set with cli or web ui, eg: `vercel secrets add [secret-name] [secret-value]` - this is useful for values you don't want in source control

i think we should do something similar. we probably want to have env vars in `eas.json` associated with build profiles, and then run `eas build:secrets:add [name] [value]` to encrypt/save the value on our backend and then load that in the build server when running a related build job

brentvatne (changes requested):
let's chat about this more before landing it

brentvatne:
i was just discussing this with @ide and another thing we could do here so that we aren't committing to this particular api is to constrain it specifically to just setting the `NPM_TOKEN` environment variable and flag it as experimental. i think in the longer term we want some solution for env vars [like this](https://vercel.com/blog/environment-variables-ui) but we may not have time to build this at the moment. so instead, we could add something like this to `credentials.json`:

```
"experimental": { 
  "npmToken": "..."
 }
```

PR #2604, opened 2020-09-08
[cli] improve init

EvanBacon:
We have some conflicting logic across different commands which have copy/pasted features. This is an initial push to unify some of these features. This PR also implements some logic from our package `create-react-native-app`. I'll address template issues in another PR after this one lands.

- Prompts for the name before prompting for the template since there is a higher probability that the name will fail. This results in less steps for subprime flows. 
  - We did this in the past because we were attempting to prevent the usage of dashes and numbers in bare project names. We normalize the names though, and iOS + Android both support numbers and dashes so this isn't even an issue afaict.
- Custom abort message for name prompt. Fails more gracefully instead of just quitting.
- Change slug prompt to name prompt in managed, and reuse across bare/managed init.
- Skip prompting to `cd` when the project was created in place (like with `--yes`). This applies to node modules and cocoapods.
- Reuse the conflicting files check / prompt from `expo export` in `expo init`, this is a much better UX for cases when the user attempts to bootstrap a project in a directory with conflicting files.
  - We no longer validate the project directory while the user is typing the project name. Instead, if the user enters a project name that points to a folder with conflicting files, they'd get the same error as when passing a name via `expo init <name>` or `expo init --name <name>`. This is more predictable and has less cases to test and consider.
- Unify CocoaPods installation step across eject and init, this is a good example of a ton of copy/pasted code. This will bring support for `EXPO_DEBUG=false` to silence installs in the eject ⌘. 
- Fix bug where `--no-install` would tell user to `npm start` after init ⌘. Now it favors using yarn when possible.
  - side-note: we should try to remove the `--yarn` flag and default to using yarn like we do in CRNA.
    - side-side-note: we should remove the `--name` flag since it's irrelevant.
- Unify the `logNewSection` logic.
- Unify the duplicative logic across workflows at the end of the init ⌘. We shouldn't be treating the bootstrapping process different based on something fragile like template names.

### DIFF
--- a/packages/expo-cli/src/commands/utils/CreateApp.ts
+++ b/packages/expo-cli/src/commands/utils/CreateApp.ts
@@ -120,6 +120,14 @@ export function logNewSection(title: string) {
   return spinner;
 }
 
+export function getChangeDirectoryPath(projectRoot: string): string {
+  const cdPath = path.relative(process.cwd(), projectRoot);
+  if (cdPath.length <= projectRoot.length) {
+    return cdPath;
+  }
+  return projectRoot;
+}
+
 export async function installCocoaPodsAsync(projectRoot: string) {
   log.addNewLineIfNone();
   let step = logNewSection('Installing CocoaPods.');
--- a/packages/expo-cli/src/commands/utils/CreateApp.ts
+++ b/packages/expo-cli/src/commands/utils/CreateApp.ts
@@ -1,5 +1,11 @@
+import * as PackageManager from '@expo/package-manager';
+import program from 'commander';
 import fs from 'fs-extra';
+import getenv from 'getenv';
+import yaml from 'js-yaml';
+import ora from 'ora';
 import * as path from 'path';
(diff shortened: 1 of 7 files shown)


packages/expo-cli/src/commands/utils/CreateApp.ts
```diff
@@ -0,0 +1,185 @@
...
+import semver from 'semver';
+
+import log from '../../log';
+
+export function validateName(name?: string): string | true {
+  if (typeof name !== 'string' || name === '') {
+    return 'The project name can not be empty.';
+  }
+  if (!/^[a-z0-9@.\-_]+$/i.test(name)) {
+    return 'The project name can only contain URL-friendly characters.';
```
brentvatne, line 17:
```suggestion
    return 'The project name can only contain URL-friendly characters (alphanumeric and @ . -  _)';
```

maybe worth adding some additional info on what characters these are?

brentvatne (approved):

PR #2605, opened 2020-09-08
[cli] Migrate all confirm prompts to prompts package

EvanBacon:
In some places we use inquirer and others we use prompts (prompts is better bcz package size). The core issue here is that the interaction is different, which can be mildly infuriating if you were expecting to press "y" then press "enter" but the prompt completed when you just pressed "y" or vise versa. This PR aims to unify the interaction across expo-cli.

`prompts` confirm only requires a single keystroke and doesn't allow for free text to be written to the confirm prompt.


brentvatne:
thanks for doing this!

PR #2607, opened 2020-09-08
fix android circle cropping images with sharp

EvanBacon:
Run `expo apply` with sharp installed globally.
Also fixes circle cropping for jpeg images in jimp.

## Before

ic_launcher

## After

ic_launcher_round

### DIFF
--- a/packages/image-utils/src/Image.ts
+++ b/packages/image-utils/src/Image.ts
@@ -73,13 +73,9 @@ async function resizeAsync(imageOptions: ImageOptions): Promise<Buffer> {
 
     if (imageOptions.borderRadius) {
       const mask = Buffer.from(
-        `<svg><rect x="0" y="0" width="${width}" height="${height}" 
-        rx="${imageOptions.borderRadius}" ry="${imageOptions.borderRadius}" 
-        fill="${
-          backgroundColor && backgroundColor !== 'transparent' ? backgroundColor : 'none'
-        }" /></svg>`
+        `<svg><circle cx="${imageOptions.borderRadius}" cy="${imageOptions.borderRadius}" r="${imageOptions.borderRadius}" /></svg>`
       );
-      sharpBuffer.composite([{ input: mask, blend: 'dest-over' }]);
+      sharpBuffer.composite([{ input: mask, blend: 'dest-in' }]);
     }
 
     return await sharpBuffer.png().toBuffer();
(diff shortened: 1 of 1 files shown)


packages/image-utils/src/Image.ts
```diff
@@ -73,13 +79,9 @@ async function resizeAsync(imageOptions: ImageOptions): Promise<Buffer> {
...
       const mask = Buffer.from(
-        `<svg><rect x="0" y="0" width="${width}" height="${height}" 
-        rx="${imageOptions.borderRadius}" ry="${imageOptions.borderRadius}" 
-        fill="${
-          backgroundColor && backgroundColor !== 'transparent' ? backgroundColor : 'none'
-        }" /></svg>`
+        `<svg><circle cx="${imageOptions.borderRadius}" cy="${imageOptions.borderRadius}" r="${imageOptions.borderRadius}" /></svg>`
       );
-      sharpBuffer.composite([{ input: mask, blend: 'dest-over' }]);
+      sharpBuffer.composite([{ input: mask, blend: 'dest-in' }]);
```
cruzach, line 84: with these changes, when you set `foregroundImage` and `backgroundColor`, `ic_launcher_round` is missing the background color

It looks like the call to `.composite()` does away with the background set in the previous composite? Because if you remove this line, the returned image _does_ have the proper background color. 

One way to fix this is to add back the `fill` prop in the svg and then composite twice:
```
sharpBuffer.composite([
        { input: mask, blend: 'dest-over' },
        { input: mask, blend: 'dest-in' },
      ]);
```

packages/image-utils/src/Image.ts
```diff
@@ -73,13 +79,9 @@ async function resizeAsync(imageOptions: ImageOptions): Promise<Buffer> {
 
     if (imageOptions.borderRadius) {
       const mask = Buffer.from(
-        `<svg><rect x="0" y="0" width="${width}" height="${height}" 
-        rx="${imageOptions.borderRadius}" ry="${imageOptions.borderRadius}" 
-        fill="${
-          backgroundColor && backgroundColor !== 'transparent' ? backgroundColor : 'none'
-        }" /></svg>`
+        `<svg><circle cx="${imageOptions.borderRadius}" cy="${imageOptions.borderRadius}" r="${imageOptions.borderRadius}" /></svg>`
```
cruzach, line 82: why do we want to change this to a circle? with a rect we can support different rounded corners rather than just a circle

EvanBacon: because it wasn't cropping the image at all

cruzach: I think that's because of a backgroundColor change I made that broke this, though, not because of using a `rect`

cruzach: ^ problem is that background color for `roundBackgroundLayer` is always `transparent` now, so `fill` here is always `none`, so it wouldn't crop. We could default to like `#ffffff` here or something instead

PR #2611, opened 2020-09-09
Fix: Asset batch uploading

kareem-adel:
Fix: Asset batch uploading
Assets were divided into chunks of 5 for uploading, using await promise.all to upload all the chunks at once causes network and system resources issues for big number of assets, instead with this change chunks are uploaded sequentially.
Reference for issue: https://stackoverflow.com/a/30823708

### DIFF
--- a/packages/xdl/src/ProjectAssets.ts
+++ b/packages/xdl/src/ProjectAssets.ts
@@ -246,23 +246,21 @@ async function uploadAssetsAsync(projectRoot: string, assets: Asset[]) {
   const keyChunks = chunk(missing, 5);
 
   // Upload them!
-  await Promise.all(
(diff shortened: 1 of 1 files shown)


packages/xdl/src/ProjectAssets.ts
```diff
@@ -246,23 +246,21 @@ async function uploadAssetsAsync(projectRoot: string, assets: Asset[]) {
   const keyChunks = chunk(missing, 5);
 
   // Upload them!
```
EvanBacon, line 248:
```suggestion
  // Upload them in chunks of 5 to prevent network and system issues.
```

EvanBacon (approved):
Good work!

PR #2614, opened 2020-09-09
[cli] Clean up TerminalUI

EvanBacon:
- make the terminal UI shorter
- group the commands 
- match up wording across commands 
- standardize text formatting

currently

### Before

Screen Shot 2020-09-09 at 10 07 54 AM

### After

Screen Shot 2020-09-09 at 10 24 10 AM

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.ts
+++ b/packages/expo-cli/src/commands/start/TerminalUI.ts
@@ -45,37 +45,51 @@ const printHelp = (): void => {
   log.nested(`${PLATFORM_TAG} Press ${b('?')} to show a list of all available commands.`);
 };
 
+const div = chalk.dim(`|`);
+
 const printUsage = async (projectDir: string, options: Pick<StartOptions, 'webOnly'> = {}) => {
   const { dev } = await ProjectSettings.readAsync(projectDir);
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/start/TerminalUI.ts
```diff
@@ -45,37 +45,56 @@ const printHelp = (): void => {
...
-`);
+  const currentAuth = `@${username}`;
+  const currentToggle = openDevToolsAtStartup ? 'disabled' : 'enabled';
+
+  const isMac = process.platform === 'darwin';
+
+  const ui = [
+    [],
+    ['a', `open Android`],
+    ['shift+a', `select a device or emulator`],
```
fson, line 63: I wonder if we could use something like `⇧+a` here? Although I guess some fonts might not have this character.

fson (approved):
💯❤️

packages/expo-cli/src/commands/start/TerminalUI.ts
EvanBacon: I tried that but it was too small in the default terminal :(

brentvatne (approved):
pretty good

PR #2622, opened 2020-09-10
Generate types for Expo config

EvanBacon:
- Created a versioned package that has generated types based on the published schema for a given Expo version.
- This will replace the hardcoded `ExpoConfig` from `@expo/config`.
- Maybe we should have this be part of the `expo` package instead? I'm worried the publish script may cause this package to get misversioned.


packages/config-types/src/index.ts
```diff
@@ -0,0 +1,952 @@
...
+   */
+  platforms?: ('android' | 'ios' | 'web')[];
+  /**
+   * If you would like to share the source code of your app on Github, enter the URL for the repository here and it will be linked to from your Expo project page.
+   */
+  githubUrl?: string;
+  /**
+   * Locks your app to a specific orientation with portrait or landscape. Defaults to no lock. Valid values: `default`, `portrait`, `landscape`
+   */
+  orientation?: ('default' | 'portrait' | 'landscape') | null;
```
EvanBacon, line 54: this is basically an any

fson: So it can be one of `'default' | 'portrait' | 'landscape', or `undefined`, or `null` 🤔

packages/config-types/scripts/generate.ts
```diff
@@ -0,0 +1,34 @@
...
+if (parsedVersion !== 'unversioned') {
+  parsedVersion += '.0.0';
+} else {
+  parsedVersion = parsedVersion.toUpperCase();
+}
+
+(async () => {
+  const {
+    data: { data },
+  } = await axios.get(`http://exp.host/--/api/v2/project/configuration/schema/${parsedVersion}`);
```
brentvatne, line 25: it would be nice if this supported a local path as well, to make it easier to test schema changes. i'd like to be able to run `yarn generate --schema ~/universe/server/www/xdl-schemas/39.0.0-schema.json`

EvanBacon: done!

PR #2624, opened 2020-09-11
[config][eject] Added ios.entitlements

EvanBacon:
entitlements is a low-level object that people can build on top of for more dynamic control via JS. In time it could replace abstract properties like `usesIcloudStorage` which require an apple team ID


packages/xdl/src/detach/IosNSBundle.ts
```diff
@@ -168,12 +168,20 @@ async function _configureEntitlementsAsync(context: AnyStandaloneContext): Promi
       await IosPlist.createBlankAsync(supportingDirectory, entitlementsFilename);
     }
     const result = IosPlist.modifyAsync(supportingDirectory, entitlementsFilename, entitlements => {
+      if (manifest.ios?.entitlements) {
+        for (const key in manifest.ios.entitlements) {
+          if (manifest.ios.entitlements.hasOwnProperty(key)) {
+            entitlements[key] = manifest.ios.entitlements[key];
+          }
+        }
+      }
```
EvanBacon, line 177: this is a simplified version of the `ios.infoPlist` implementation.

dsokal (approved):
LGTM

PR #2625, opened 2020-09-11
Remove external config evaluation script

EvanBacon:
# Why

- Resolves https://twitter.com/notbrent/status/1304277925278019585
- Slower computers struggle to dynamically evaluate a config from an external script. 

# How

- Removed the external script for config reading in favor of parsing the string contents of the config file via `@babel/core` and evaluating it as a string.
- I also removed 2 seemingly unused deep clone / evaluation calls, this should make reading a bit faster too.
  1. It's not needed on a static config because static configs don't support JS functions.
  2. It's unified in the shared `evalConfig` method.
- Removed an unused FS existence check.

# Test Plan

- Tested locally that the contents of the config can still be updated during `expo start`. Ensured the dynamic behavior was used and `Constants.manifest.name` could be updated by reloading the app (⌘+r).


EvanBacon:
This change might break https://github.com/expo/expo-cli/pull/2220 by essentially reverting the issue they had initially by default. cc @RedMickey see test failures https://github.com/expo/expo-cli/pull/2625/checks?check_run_id=1101429349

packages/config/src/Config.ts
```diff
@@ -82,6 +85,10 @@ function getSupportedPlatforms(
  * @param options enforce criteria for a project config
  */
 export function getConfig(projectRoot: string, options: GetConfigOptions = {}): ProjectConfig {
+  if (typeof options.useDynamicEval === 'undefined' && isDynamicEvalEnabled) {
```
brentvatne, line 88: i think we should rename this to hot eval because we already call `config.js/ts` dynamic. so it's hot eval of dynamic config

packages/config/src/getConfig.ts
```diff
@@ -34,48 +29,18 @@ function readConfigFile(
...
+  // If so we should throw a more helpful error.
   throw new ConfigError(`Failed to read config at: ${configPath}`, 'INVALID_CONFIG');
 }
 
-export function getStaticConfig(configPath: string): AppJSONConfig | ExpoConfig | null {
+export function getStaticConfig(configPath: string): AppJSONConfig | ExpoConfig {
   const config = JsonFile.read(configPath, { json5: true });
   if (config) {
-    return serializeAndEvaluate(config);
+    return config as any;
```
brentvatne, line 43: https://twitter.com/benawad/status/1304985643173515275

brentvatne (approved):

packages/config/src/getConfig.ts
EvanBacon: lmfao ¯\_(ツ)_/¯

PR #2629, opened 2020-09-11
[cli] Disallow use of console

wschurman:
# Why

As mentioned by @brentvatne in https://github.com/expo/expo-cli/pull/2581#pullrequestreview-484574592, console should not be used. This should cut down on future rounds of review as it is enforceable by lint.

# How

To prevent future devs from making the same mistake, this PR disallows console from being used and replaces all uses of it with the `log` equivalents. Note that there are still a few places that it is needed (outside src and in the actual log code), but I feel that the tradeoff of having eslint disable exceptions is worth it for the lessened developer attention cost.

# Test Plan

`yarn lint`

### DIFF
--- a/packages/expo-cli/.eslintrc.js
+++ b/packages/expo-cli/.eslintrc.js
@@ -1,5 +1,5 @@
 module.exports = {
   rules: {
-    'no-console': 'warn',
+    'no-console': 'error',
   },
 };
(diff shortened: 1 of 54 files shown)


brentvatne (approved):
awesome 👏 
thanks @wschurman!

brentvatne:
didn't notice the base branch 🤦  sorry

PR #2632, opened 2020-09-12
Use subprocess to populate dynamic config cache, add function to clear it

brentvatne:
This approach makes a bit easier to understand (`@babel/register` is a black box for me) and has the bonus of making it easy for us to expose functionality to let users clear the config cache without restarting expo-cli.

Note that the base branch is `@evanbacon/config/add-support-for-toggling-config-evaluation-modes`

### DIFF
--- a/packages/config/src/getConfig.ts
+++ b/packages/config/src/getConfig.ts
@@ -4,10 +4,17 @@ import { spawnSync } from 'child_process';
 import { AppJSONConfig, ConfigContext, ExpoConfig } from './Config.types';
 import { ConfigError, errorFromJSON } from './Errors';
 import { serializeAndEvaluate } from './Serialize';
-import { DynamicConfigResults, evalConfig } from './evalConfig';
+import { DynamicConfigResults } from './evalConfig';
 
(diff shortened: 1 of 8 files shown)


packages/config/src/getConfig.ts
```diff
@@ -4,26 +4,37 @@ import { spawnSync } from 'child_process';
...
+}
+
 // We cannot use async config resolution right now because Next.js doesn't support async configs.
 // If they don't add support for async Webpack configs then we may need to pull support for Next.js.
 function readConfigFile(configFile: string, context: ConfigContext): null | DynamicConfigResults {
   try {
     if (context.useDynamicEval) {
       return spawnAndEvalConfig(configFile, context);
+    } else if (!context.useDynamicEval && _cachedConfig === null) {
+      _cachedConfig = spawnAndEvalConfig(configFile, context);
```
EvanBacon, line 27: If someone wants to reuse a stale config they can always just limit how often they invoke getConfig. Moving that logic into getConfig seems like it’ll break stuff.

PR #2635, opened 2020-09-13
Improve JSON error formatting

EvanBacon:
# Why

- the duplicate message is annoying.
- it's not clear what file the JSON error is referring to.
- empty JSON files are a more controlled issue that's easy to diagnose, they don't need an entire trace.
- the stack trace has a bunch of compiled mumbo jumbo like `Function.`, `Object.`, `<anonymous>`.
- dim path components and strip parentheses.

# Empty JSON

## Before

Screen Shot 2020-09-13 at 11 21 59 PM

## After

Screen Shot 2020-09-13 at 11 22 08 PM

# Syntax error

## Before

Screen Shot 2020-09-13 at 11 20 24 PM

## After

Screen Shot 2020-09-13 at 11 20 35 PM


fson:
The stacktraces! 😍

These new stack traces are beautiful. If we want to make other stack traces we show (e.g. Metro/Webpack errors, device logs) look the same, can we reuse this code?

packages/xdl/src/project/Doctor.ts
```diff
@@ -435,24 +435,12 @@ async function validateAsync(
...
-      'doctor-config-json-not-read'
-    );
-    return ERROR;
-  }
+  const { exp, pkg } = getConfig(projectRoot, {
+    strict: true,
+    skipSDKVersionRequirement,
+  });
+
+  ProjectUtils.clearNotification(projectRoot, 'doctor-config-json-not-read');
```
fson, line 443: So this just throws now if config can't be read? Seems better than wrapping the original error with "could not load config json..."

But I wonder if throwing will cause an `UnhandledPromiseRejectionWarning` here:
https://github.com/expo/expo-cli/blob/2c51e2f7b549ef6e2b3cb86f721959ad01ae6c5d/packages/xdl/src/project/ManifestHandler.ts#L120

packages/expo-cli/src/exp.ts
```diff
@@ -321,6 +331,58 @@ Command.prototype.asyncAction = function (asyncFn: Action, skipUpdateCheck: bool
...
+function getStringBetweenParens(value: string): string {
+  const regExp = /\(([^)]+)\)/;
+  const matches = regExp.exec(value);
+  if (matches && matches?.length > 1) {
+    return matches[1];
+  }
+  return value;
+}
+
+function focusLastPathComponent(value: string): string {
```
fson, line 343: 93028954-d2d3d480-f617-11ea-822c-6ad25ce10493

Maybe the beginning of the path should be dimmed for lines that don't have a function name too? It seems that in this screenshot they're not?

EvanBacon: Here's what that looks like:
Screen Shot 2020-09-14 at 6 58 18 PM

I think it's a little more jarring because the solid stack trace now has a gap in it.

packages/xdl/src/project/Doctor.ts
EvanBacon: Good call, I'm now catching the error in the middleware and logging it like before

packages/expo-cli/src/exp.ts
fson: Can we put something there to fill the gap? E.g. Chrome Dev Tools shows `(anonymous function)` instead of the name, if there's no function name for the stack frame, I think.

PR #2641, opened 2020-09-14
[expo-cli] add --config flag to credentials:manager

wkozyra95:
# Why

Support credentials manager for people that have multiple projects on a single codebase. For those cases, all credentials management had to be done via build command.

# How

Support config flag. I couldn't use logic from `exp.ts` because that code requires a valid project and in the case of `credentials:manager` command it's optional

### DIFF
--- a/packages/expo-cli/src/commands/credentials.ts
+++ b/packages/expo-cli/src/commands/credentials.ts
@@ -1,4 +1,7 @@
+import * as ConfigUtils from '@expo/config';
 import { CommanderStatic } from 'commander';
+import fs from 'fs-extra';
+import path from 'path';
 
 import { Context, runCredentialsManagerStandalone } from '../credentials';
 import {
@@ -6,9 +9,11 @@ import {
   SelectIosExperience,
   SelectPlatform,
 } from '../credentials/views/Select';
+import log from '../log';
 
 type Options = {
   platform?: 'android' | 'ios';
+  config?: string;
   parent?: {
     nonInteractive: boolean;
   };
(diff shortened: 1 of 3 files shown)


quinlanj (approved):
The code looks good to me. 

Regarding the use case: 
IIUC the user has separate configs (ie) `app-1.json` and `app-2.json` in the same project directory and they want to specify which one to use. Is this because people have published two separate expo apps within the same project directory?

packages/expo-cli/src/commands/credentials.ts
```diff
@@ -20,8 +25,23 @@ export default function (program: CommanderStatic) {
     .description('Manage your credentials')
     .helpGroup('credentials')
     .option('-p --platform <platform>', 'Platform: [android|ios]', /^(android|ios)$/i)
+    .option('--config [file]', 'Specify a path to app.json or app.config.js')
```
EvanBacon, line 28: we should move the `exp.ts` logic into a method and attempt to reuse it since this is a lot of stylistic logic. Perhaps we could add a way to skip project validation?

packages/expo-cli/src/commands/credentials.ts
```diff
@@ -20,8 +25,23 @@ export default function (program: CommanderStatic) {
...
+        log(
+          log.chalk.green(`  expo credentials:manager --config ${log.chalk.cyan(`<app-config>`)}`)
+        );
+      }
+      if (options.config) {
+        const configPath = path.resolve(projectDir, options.config);
+        if (!(await fs.pathExists(configPath))) {
+          throw new Error(`File ${configPath} does not exist`);
+        }
+        ConfigUtils.setCustomConfigPath(projectDir, configPath);
```
EvanBacon, line 43: iirc this particular method is somewhat fragile and shouldn't be used in more than one place.

EvanBacon (changes requested):
♻️ reuse existing

packages/expo-cli/src/commands/credentials.ts
wkozyra95: Sure, skip flag would be probably best solution.

I decided to go this way because it's a bit isolated use cases and we will probably remove that command when web gui will be ready, so i didn't want modify code that handles other command. I don't have strong opinion on that either way.

wkozyra95:
Replaced `validateWithoutNetworkAsync` with try catch round get config because `validateWithoutNetworkAsync` logs error

PR #2643, opened 2020-09-14
Transpile expo/config in expo-optimize

EvanBacon:
Now that the external script is gone we can transpile expo/config in expo-optimize which should drastically reduce npx time. Probably wanna wait to see if we need to revert the config stuff before merging this.

Tested locally and it works as expected when the `require.resolve`s are removed.


brentvatne (approved):
nice!

PR #2645, opened 2020-09-15
[config][eject] Added android.metadata

EvanBacon:
# Why

`meta-data` fields in the first `application` of the `AndroidManifest.xml` is easy to predict and required for a bunch of API related features.

# How

Make API syncing modules convert API related config to `android.metadata` which is added to the manifest all at once. This reduces the amount of XML testing we need to do and also adds support for lots of other third-party APIs. The presence of certain keys causes apply to remove the keys if they no longer exist in the config.

## Considerations

Having a MetaData module means syncing will always clear out the existing meta-data keys in the AndroidManifest. Essentially, users will need to add custom new `meta-data` items to the Expo config object or give up using the syncing functionality. 

I think this is a fair trade off because the `android.metadata` array is easier to use and understand than the XML.

# Test Plan

- Refactored the tests to ensure the metadata array is modified properly -- easier and cleaner than testing the XML like we were doing.
- Locally tested that adding `android.manifest: [{name: 'custom', value: true}]` and running `expo eject` generated a new meta tag in the project's `AndroidManifest.xml`
- Tested that adding `facebookAppId` to the config and ejecting, generated a meta tag, then removing it deleted the meta tag again.


esamelson (commented):
Could we make it so that if android.metadata is undefined, syncing doesn't clear out the existing tags? Would be nice to let people manage it themselves if they want, but still use expo apply for other things.

EvanBacon:
@esamelson I think that's hard to ensure since other things internally will apply changes to the metadata object.

esamelson:
Could we maybe log a warning about this then, at least the first time you run expo apply? If you guys think it's annoying and unnecessary then never mind, but I think it might be confusing & frustrating to people who are used to managing their own AndroidManifest that they need to copy every customization into app.json in order to keep it. (AFAIK this is not the case with iOS Info.plist, correct?)

EvanBacon:
@esamelson people can always just _not_ use the command. It's not very reasonable to assume a tool can be used to generate config with idempotency and consider user modifications.

EvanBacon:
Currently, I'm ok with sticking to the AndroidManifest modifications now that the typings for manifest, and the config mods have been created.

PR #2648, opened 2020-09-15
[expo-cli][eject] warn about Constants.manifest and assetBundlePatterns

cruzach:
- We should warn about moving from `Constants.manifest` -> `Updates.manifest` when using the bare workflow
- Assets are bundled into bare workflow binaries differently (don't rely on `assetBundlePatterns`), so we should point to these new docs when you eject

screenshot of the output:
image

### DIFF
--- a/packages/expo-cli/src/commands/eject/Eject.ts
+++ b/packages/expo-cli/src/commands/eject/Eject.ts
@@ -122,7 +122,9 @@ export async function ejectAsync(projectRoot: string, options?: EjectAsyncOption
     log.nested(
       `- 📁 The property ${chalk.bold(
         `'assetBundlePatterns'`
-      )} is not used in the bare-workflow. ${log.chalk.dim(learnMore('https://docs.expo.io/bare/updating-your-app/#embedding-assets'))}`
+      )} does not have the same effect in the bare workflow. ${log.chalk.dim(
+        learnMore('https://docs.expo.io/bare/updating-your-app/#embedding-assets')
+      )}`
     );
   }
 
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -277,6 +285,11 @@ async function ensureConfigAsync(
   return { exp, pkg };
 }
 
+function usesAssetBundlePatterns(projectRoot: string): boolean {
+  const { exp } = getConfig(projectRoot);
```
EvanBacon, line 289: this is really expensive

EvanBacon:
In general, consider how the logs look when running `expo eject` on an ejected project. The subsequent runs should be super quick and nonintrusive while they sync up the project config with the native code.

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -627,23 +640,16 @@ async function warnIfDependenciesRequireAdditionalSetupAsync(
     skipSDKVersionRequirement: true,
   });
 
-  const extraSetupPath = projectHasModule('expo/requiresExtraSetup.json', projectRoot, exp);
```
EvanBacon, line 630: Ensure the following works without throwing:
- in a managed project delete the node_modules (add the sdkVersion to the app.json)
- `expo eject --no-install`

EvanBacon (changes requested):
feedback

packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -118,6 +118,14 @@ export async function ejectAsync(projectRoot: string, options?: EjectAsyncOption
     )}`
   );
 
+  if (usesAssetBundlePatterns(projectRoot)) {
+    log.nested(
+      `- 📁 In the bare workflow, ${chalk.yellow(
+        `${'assetBundlePatterns'}`
+      )} no longer affects your builds. To learn about assets in the bare workflow, read here: https://docs.expo.io/bare/updating-your-app/#embedding-assets`
```
EvanBacon, line 125:
```suggestion
      `- 📁 The property ${chalk.bold(
        `'assetBundlePatterns'`
      )} is not used in the bare-workflow. ${learnMore('https://docs.expo.io/bare/updating-your-app/#embedding-assets')}`
```

- We use bold instead of color changes for this kind of thing since it works for more terminal styles. 
- Use the `learnMore` link for keeping links as concise as possible.
- "no longer affects" kinda sounds like it's been deprecated.

packages/expo-cli/src/commands/eject/Eject.ts
cruzach: yeah it works, although if you use `app.config.js` instead of app.json it doesn't:
```
Failed to read config at: /path/to/app.config.js
```

but that's also the case on master

packages/expo-cli/src/commands/eject/Eject.ts
cruzach: changed so that we grab config once, at the top of `ejectAsync`

packages/expo-cli/src/commands/eject/Eject.ts
EvanBacon: good find, opened a PR to fix

EvanBacon (approved):
good work!

PR #2649, opened 2020-09-15
Log about upload after build:ios completes

### LINKED ISSUE #1916: Log info about publish:ios after running build:ios successfully

EvanBacon:
resolve #1916 

After `expo build:ios` completes successfully, log about using `expo upload:ios`


brentvatne (approved):

brentvatne (changes requested):
- we shouldn't show this if the build type was simulator
- we also shouldn't show this on any OS except macOS

PR #2653, opened 2020-09-16
remove extra .gitignores

EvanBacon:
reducing clutter


.gitignore
```diff
@@ -14,4 +16,33 @@ node_modules
 *.tsbuildinfo
 .history
 
+**/*/build
+
+# Logs
+logs
```
fson, line 22: Does this have an effect on https://github.com/expo/expo-cli/tree/f88f72faa5845e12827d24534c8e171be97e92b7/packages/xdl/src/logs ? And does some package actually use this path to store logs?

.gitignore
```diff
@@ -14,4 +16,33 @@ node_modules
...
+*.seed
+
+# Directory for instrumented libs generated by jscoverage/JSCover
+lib-cov
+
+# Coverage directory used by tools like istanbul
 coverage/
+
+# Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
+.grunt
```
fson, line 37: I believe this can be removed as we're not using Grunt.

.gitignore
```diff
@@ -14,4 +16,33 @@ node_modules
...
+
+# node-waf configuration
+.lock-wscript
+
+# Compiled binary addons (http://nodejs.org/api/addons.html)
+build/Release
```
fson, line 43: This could already be covered by `packages/*/build/`?

.gitignore
EvanBacon: it doesn't but I made the path a bit more specific anyways since it seems like it could cause issues in the future.

PR #2658, opened 2020-09-16
deprecate password argument of login

EvanBacon:
resolve https://github.com/expo/expo-cli/issues/126


brentvatne (approved):

brentvatne:
probably worth searching docs to see if we recommend using --password or -p anywhere

EvanBacon:
According to this doc https://docs.expo.io/guides/setting-up-continuous-integration/ the `--pasword` argument can securely be used in CI (which was the original intent of #126). The other aspects of the issue are supporting `EXPO_CLI_PASSWORD` and blacklisting it from the manifest middleware, which we already do.

PR #2661, opened 2020-09-17
Fix: Next warning Expected Document Component Html was not rendered

oste:
Since next 9.5.3 the html element in _document.js should be Html instead. This error is rendered:

```
Expected Document Component Html was not rendered. Make sure you render them in your custom `_document`
See more info here https://err.sh/next.js/missing-document-component
```

See also https://github.com/vercel/next.js/discussions/16908


EvanBacon (approved):
LGTM thanks

PR #2662, opened 2020-09-17
[expo-cli] Allow manually provided credentials when keytool is missing

wkozyra95:
# Why

When keytool was missing update process was stopped before the user had a chance to select the option to manually provide credentials.

# How

Move `keytool` check to later stage

# Test plan

Remove `keytool` and run `expo build:android -c`

### DIFF
--- a/packages/expo-cli/src/credentials/views/AndroidKeystore.ts
+++ b/packages/expo-cli/src/credentials/views/AndroidKeystore.ts
@@ -1,5 +1,6 @@
 import { AndroidCredentials } from '@expo/xdl';
 import chalk from 'chalk';
+import commandExists from 'command-exists';
 import fs from 'fs-extra';
 import omit from 'lodash/omit';
 import os from 'os';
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/credentials/views/AndroidKeystore.ts
```diff
@@ -13,24 +14,48 @@ import { askForUserProvided } from '../actions/promptForCredentials';
 import { Context, IView } from '../context';
 import { Keystore, keystoreSchema } from '../credentials';
 
+interface UpdateKeystoreOptions {
+  optionalKeystoreGeneration: boolean;
```
quinlanj, line 18:
```suggestion
  bestEffortKeystoreGeneration: boolean; // don't error if user doesn't have the binaries to generate a Keystore 
```

quinlanj (changes requested):
IIUC, we are passing in additional options into the `UpdateKeystore` constructor because we expect different behaviours in `credentials:manager` vs `build:android`. 

In the event the user doesn't have the `keytool` binary, we want:
- `credentials:manager` to throw, because the user is specifically asking for client side generation of Keystore
- `build:android` to return null, because we will generate a Keystore for the user on the Expo servers.

wkozyra95:
@quinlanj 
> In the event the user doesn't have the keytool binary, we want:
 credentials:manager to throw, because the user is specifically asking for client side generation of Keystore
 build:android to return null, because we will generate a Keystore for the user on the Expo servers.

If I did not miss anything this code has the same exact behaviour as with your changes. The only difference is that I'm throwing an error when cli attempts to generate Keystore and with your changes, it's happening based on `userHasKeytoolCommand` value.

`bestEffortKeystoreGeneration` is a better name, so I'll update that

quinlanj:
@wkozyra95 ok sounds good to me, just wanted to make sure we were not missing anything!

PR #2664, opened 2020-09-17
[@expo/json-file] add option to create directory on write

wkozyra95:
# Why

Currently, there are two options two ensure directory is created,
- run `mkdir` on start of the application
- wrap entire JsonFIle api

# How

add option `ensureDirOnWrite`

# Test plan

Run `setAsync` when file does not exists


packages/json-file/src/JsonFile.ts
```diff
@@ -209,6 +211,9 @@ async function writeAsync<TJSONObject extends JSONObject>(
   object: TJSONObject,
   options?: Options<TJSONObject>
 ): Promise<TJSONObject> {
+  if (options?.ensureDirOnWrite) {
+    await mkdirp(path.dirname(file));
```
EvanBacon, line 215: `ensureDir`?

PR #2667, opened 2020-09-18
[expo-cli] fix prompt helpers

wkozyra95:
# Why

The current implementation of prompts helpers does not work with list of questions, but types suggest that


EvanBacon (approved):
I see, thanks for catching this.

PR #2684, opened 2020-09-22
[xdl] Remove unused methods from Project module

EvanBacon:
- Remove deprecated methods
- Delete seemingly unused functions and types from `Project` module.

### DIFF
--- a/packages/expo-cli/src/commands/utils/PublishUtils.ts
+++ b/packages/expo-cli/src/commands/utils/PublishUtils.ts
@@ -1,4 +1,4 @@
-import { getConfig } from '@expo/config';
+import { ExpoConfig, getConfig } from '@expo/config';
 import { ApiV2, Project, UserManager } from '@expo/xdl';
 import ora from 'ora';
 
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/commands/utils/PublishUtils.ts
```diff
@@ -63,6 +63,21 @@ export type PublicationDetail = {
...
 const VERSION = 2;
 
+async function getSlugAsync({
+  projectRoot,
+  exp,
+}: {
+  projectRoot: string;
+  exp: Pick<ExpoConfig, 'slug'>;
+}): Promise<string> {
+  if (exp.slug) {
```
fson, line 73: This check (and the whole function?) now seems redundant, because at some point we made it so that a `slug` is always returned from `getConfig`:
https://github.com/expo/expo-cli/blob/4aef8157645bed019ecc65d3ec216e59fedf6337/packages/config-types/src/index.ts#L17-L20

fson (approved):
Seems good! 🎉 

Just see the note about `getSlugAsync`, maybe it can be removed too.

packages/expo-cli/src/commands/utils/PublishUtils.ts
EvanBacon: nice

PR #2694, opened 2020-09-23
[config] Created XML module

EvanBacon:
- split functionality out of manifest and into a new XML module.
- strongly type manifest and resources.
- functionally there should be no changes.


packages/config/src/android/Resources.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export async function getProjectResourcePathAsync(
+  projectDir: string,
+  { kind = 'values', name }: { kind?: string; name: string }
+): Promise<string | null> {
+  try {
+    const shellPath = path.join(projectDir, 'android');
+    if ((await fs.stat(shellPath)).isDirectory()) {
+      const stylesPath = path.join(shellPath, `app/src/main/res/${kind}/${name}.xml`);
```
brentvatne, line 24: `resourcePath` instead of `stylesPath`?

brentvatne (commented):
i didn't review the code deeply but the overall idea here of typing manifest and resources is excellent, and refactoring out code that is more generally related to xml than either of those specifically is :+1:

PR #2695, opened 2020-09-23
[config] Created Paths module

EvanBacon:
- Unify path resolution for android files 
- Support Kotlin MainActivities by failing gracefully

### DIFF
--- a/packages/config/src/android/NavigationBar.ts
+++ b/packages/config/src/android/NavigationBar.ts
@@ -28,7 +28,7 @@ export async function setNavigationBarConfig(config: ExpoConfig, projectDirector
   const colorsPath = await getProjectColorsXMLPathAsync(projectDirectory);
 
   let stylesJSON = await readXMLAsync({ path: stylesPath });
-  let colorsJSON = await readColorsXMLAsync(colorsPath);
+  let colorsJSON = await readColorsXMLAsync({ path: colorsPath });
 
   if (immersiveMode) {
     // Immersive mode needs to be set programatically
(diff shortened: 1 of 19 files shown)


packages/config/src/android/Paths.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export async function getResourceXMLAsync(
+  projectDir: string,
+  { kind = 'values', name }: { kind?: string; name: string }
+): Promise<string> {
+  const shellPath = path.join(projectDir, 'android');
+  try {
+    if ((await fs.stat(shellPath)).isDirectory()) {
+      const stylesPath = path.join(shellPath, `app/src/main/res/${kind}/${name}.xml`);
+      //   await fs.ensureFile(stylesPath);
```
brentvatne, line 57:
```suggestion
```

packages/config/src/android/Paths.ts
```diff
@@ -0,0 +1,67 @@
...
+      if ((await fs.stat(stylesPath)).isFile()) {
+        return stylesPath;
+      }
+    }
+  } catch {}
+
+  throw new Error(
+    `Could not find android/app/src/main/res/${kind}/${name}.xml because the android project folder is missing.`
```
brentvatne, line 65: this isn't necessarily true. `await fs.stat(stylesPath)` can throw and we would end up here

packages/config/src/android/NavigationBar.ts
```diff
@@ -26,9 +26,6 @@ export async function setNavigationBarConfig(config: ExpoConfig, projectDirector
 
   const stylesPath = await getProjectStylesXMLPathAsync(projectDirectory);
   const colorsPath = await getProjectColorsXMLPathAsync(projectDirectory);
-  if (!colorsPath || !stylesPath) {
```
brentvatne, line 29: if `getProjectColorsXMLPathAsync` throws, maybe the developer won't get the most useful error message here. better than no error message when we silently fail like before by just returning false

brentvatne (commented):
nice cleanup work.

something to think about in this broader config project is how we handle errors at any point in the process config pipeline. what does the error message look like? if it's a custom config plugin, what do we expect the developer to do when unable to apply the config due to some issue in the project? how can the developer debug it?

Issue #2586, opened 2020-09-05
[configure-splash-screen] Incorrect call for SplashScreen.show in MainActivity


rodrigorm:
I have ejected my projected and the `configure-splash-screen` generate the following invalid code:

```
SplashScreen.show(this, SplashScreenImageResizeMode.CONTAIN, false);
```

The correct code should be:

```
SplashScreen.show(this, SplashScreenImageResizeMode.CONTAIN, ReactRootView.class, false);
```

I am using:

expo: 38.0.3
expo-cli: 3.26.2
expo-splash-screen: 0.3.1

byCedric:
Hi @rodrigorm! Sorry for the delay, and thanks for writing in. I can confirm this is happening when ejecting on SDK 39 as well. I marked it as bug, we will get to this asap! (cc @EvanBacon)

For context, this is the exact error on Android:

```
<path>/background-location-bare/android/app/src/main/java/com/bycedric/backgroundlocation/MainActivity.java:19: error: no suitable method found for show(MainActivity,SplashScreenImageResizeMode,boolean)
    SplashScreen.show(this, SplashScreenImageResizeMode.CONTAIN, false);
                ^
    method SplashScreen.show(Activity,SplashScreenViewProvider,Class<? extends ViewGroup>,boolean,Function0<Unit>,Function1<? super String,Unit>) is not applicable
      (actual and formal argument lists differ in length)
    method SplashScreen.show(Activity,SplashScreenViewProvider,Class<? extends ViewGroup>,boolean,Function0<Unit>) is not applicable
      (actual and formal argument lists differ in length)
    method SplashScreen.show(Activity,SplashScreenViewProvider,Class<? extends ViewGroup>,boolean) is not applicable
      (actual and formal argument lists differ in length)
    method SplashScreen.show(Activity,SplashScreenImageResizeMode,Class<? extends ViewGroup>,boolean,SplashScreenViewProvider,Function0<Unit>,Function1<? super String,Unit>) is not applicable
      (actual and formal argument lists differ in length)
    method SplashScreen.show(Activity,SplashScreenImageResizeMode,Class<? extends ViewGroup>,boolean,SplashScreenViewProvider,Function0<Unit>) is not applicable
      (actual and formal argument lists differ in length)
    method SplashScreen.show(Activity,SplashScreenImageResizeMode,Class<? extends ViewGroup>,boolean,SplashScreenViewProvider) is not applicable
      (actual and formal argument lists differ in length)
[...]

PR #2698, opened 2020-09-24
[configure-splash-screen]<feat>: Make Android configuration conform to the new native API

bbarthec:
Recent change in `expo-splash-screen` https://github.com/expo/expo/commit/2770008129aef794dcdf5fcdaa35ef5056b6ff29#diff-70eefa11c7e96de30d02e5d069488fe2R35-R40
changed the native Android API.

This PR accommodates this native API change.

Resolves https://github.com/expo/expo-cli/issues/2586

#### All published versions since https://github.com/expo/expo/commit/2196cebae55bda181ccceadec809942f51ee9e39 needs this adjustment:
- [x] `0.6.x`
- [x] `0.7.x@next`

# TODOs after merge:
- [ ] publish new version of `@expo/configure-splash-screen`
- [ ] bump dependency version in `expo-splash-screen` package on `expo/expo#sdk-39` branch and publish with patch version for `0.6.x`
- [ ] bump dependency version in `expo-splash-screen` package on `expo/expo#master` branch and publish with patch version for `0.7.x@next`

### DIFF
--- a/packages/configure-splash-screen/src/android/MainActivity.ts
+++ b/packages/configure-splash-screen/src/android/MainActivity.ts
@@ -44,6 +44,27 @@ export default async function configureMainActivity(
     'utf-8'
   );
   const { state: newFileContent } = new StateManager<string, boolean>(fileContent)
+    // importing ReactRootView
+    .applyAction(content => {
+      const [succeeded, newContent] = replace(content, {
+        replacePattern: /^import com\.facebook\.react\.ReactRootView.*?$/m,
+        replaceContent: `import com.facebook.react.ReactRootView${LE}`,
+      });
+      return [newContent, 'replacedReactRootViewImports', succeeded];
+    })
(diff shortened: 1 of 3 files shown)


sjchmiela (approved):
The logic seems a bit too magicy to me (especially the fact that the first `.applyAction` could be replaced by `if (content.includes()` in the would-be-only second action), but as I understand it the results should be correct and as I'm not a major contributor to `expo-cli` I won't argue to change this. 👍

bbarthec:
`@expo/configure-splash-screen` is not designed to be backward compatible and there are at least two risky scenarios that might not be working fully:
- when the user was using the previous version of this command, upgraded and want to run this command again - I've tried to generalise RegExps to cover it though
- this command would not be working with `@expo/config` if the targeted project is not using `expo-splash-screen@0.{6,7}.x` package - solution for this is to tell the people to upgrade to the newest `expo-splash-screen` version

PR #2705, opened 2020-09-25
Fix windows build compatibility

byCedric:
When I test stuff on Windows, I need to manually disable these scripts. This adds support for Windows when running `$ yarn build` 😄

## How

- Replaced `rm -rf` with `rimraf`
- Prefixed `NODE_ENV` with `cross-env NODE_ENV`
- Didn't touch `&&`, these "magically" work on Windows (thanks Yarn 😄)


EvanBacon (approved):
ok

PR #2706, opened 2020-09-25
[config] ios splash screen plugin

EvanBacon:
- implement parts of configure-splash-screen for iOS in a way that works with the Expo config 

Added the following props (mostly undocumented) just for testing purposes:
```json
{
  "ios": {
    "splash": {
      "image": "",
      "backgroundColor": "",
      "tabletImage": "",
      "tabletBackgroundColor": "",
      "dark": {
        "image": "",
        "backgroundColor": "",
        "tabletImage": "",
        "tabletBackgroundColor": "",
      },
    }
  }
}
```

# Test Plan

- Added unit tests
- Added a simple E2E test
- Test no-image workflow
- Test light/dark/mobile/tablet -- different colors and images


packages/prebuild-config/src/plugins/unversioned/expo-splash-screen/withIosSplashInfoPlist.ts
```diff
@@ -33,7 +33,9 @@ export function setSplashInfoPlist(
   debug(`isDarkModeEnabled: `, isDarkModeEnabled);
 
   if (isDarkModeEnabled) {
-    const existing = IOSConfig.UserInterfaceStyle.getUserInterfaceStyle(config);
+    // IOSConfig.UserInterfaceStyle.getUserInterfaceStyle(config);
+    // Determine if the user manually defined the userInterfaceStyle incorrectly
+    const existing = config.ios?.userInterfaceStyle ?? config.userInterfaceStyle;
```
brentvatne, line 38: i think it's preferable to keep `IOSConfig.UserInterfaceStyle.getUserInterfaceStyle` unless we change the default value to `automatic`

brentvatne (approved):

packages/prebuild-config/src/plugins/unversioned/expo-splash-screen/withIosSplashInfoPlist.ts
EvanBacon: no because this will always warn the user that they misconfigured their project, we should determine the user' intent (attempting to setup dark mode) and ensure dark mode works, unless specified otherwise manually.

PR #2726, opened 2020-09-28
feat(expo-cli): Add 'validate-dependencies' command

NGMarmaduke:
# What

Adds `expo validate-dependencies`.

This command will terminate with a non-zero exit code if a dependency version is
incompatible with the projects SDK version.

# Why

It is designed to be used with automated dependency managers such as Dependabot
as part of the CI process to verify if native dependencies will be compatible
with the current Expo SDK version.

We were merging dependency updates to later finding they were causing warnings 
when running `expo start` which then meant reverting the dependency updates.
Instead with `validate-dependencies` we can fail the build on CI and prevent the PR
being merged in the first place.

# Test plan

- [ ] Automated test (e2e test in place, I have made changes to the start command however the test coverage for this command looks patchy - the e2e tests are skipped and fail on master. Would appreciate guidance from the maintainers about testing)
- [ ] Create a new expo bare minimal app with expo init
- [ ] Run `expo start` and `expo validate-dependencies` expect both to run without warnings
- [ ] Update a dependency to be incompatible 
- [ ] Run `expo start` and `expo validate-dependencies` expect both to run with stderr warnings and the exit code of `expo validate-dependencies` to be "1"

### DIFF
--- a/packages/expo-cli/e2e/__tests__/validate-dependencies-test.ts
+++ b/packages/expo-cli/e2e/__tests__/validate-dependencies-test.ts
@@ -1,0 +1,52 @@
+import spawnAsync from '@expo/spawn-async';
+import fs from 'fs-extra';
+import path from 'path';
+import temporary from 'tempy';
+
+import { createMinimalProjectAsync, EXPO_CLI, minimumNativePkgJson } from '../TestUtils';
+
+const tempDir = temporary.directory();
+
+beforeAll(async () => {
+  jest.setTimeout(30000);
+  await fs.mkdirp(tempDir);
+});
+
+function executeCommand(cwd: string, args: string[]) {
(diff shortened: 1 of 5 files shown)


brentvatne:
thanks for the contribution! i would prefer this as a flag for `expo doctor`. `expo doctor --only dependencies`. we want to avoid a proliferation of commands at the top level because it makes it very difficult to read through help.

NGMarmaduke:
@brentvatne Makes total sense - a question about the doctor comand; at the moment it always returns a 0 exit code, would you be open for to it returning a non-zero exit code if issues are detected? Using the enum value returned from `validateWithNetworkAsync` would work, as 0 is used to mean no issues.

brentvatne:
> at the moment it always returns a 0 exit code, would you be open for to it returning a non-zero exit code if issues are detected?

yup this would be desirable

PR #2729, opened 2020-09-29
[expo-cli] Make expo install work in any React Native app

brentvatne:
Fixes https://github.com/expo/expo/issues/10424

### DIFF
--- a/packages/expo-cli/src/commands/install.ts
+++ b/packages/expo-cli/src/commands/install.ts
@@ -35,21 +35,33 @@ async function installAsync(packages: string[], options: PackageManager.CreateFo
     log,
   });
 
-  // This ends up being confusing for people. If they're using expo install,
-  // let's just install the deps such that they work in the client even if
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/install.ts
```diff
@@ -35,21 +35,33 @@ async function installAsync(packages: string[], options: PackageManager.CreateFo
...
+  // in package.json, just fall through to npm/yarn.
   //
-  // if (workflow === 'bare') {
-  //   return await packageManager.addAsync(...packages);
-  // }
+  if (!pkg.dependencies['expo']) {
+    return await packageManager.addAsync(...packages);
+  }
+
+  if (!exp.sdkVersion) {
```
brentvatne, line 47: this has the side effect of improving the messaging in the weird edge case where maybe somebody 1) has expo pkg in pkg.json 2) has node_modules directory but 3) somehow we can't resolve the sdk version

PR #2739, opened 2020-09-30
[configure-splash-screen] do not export color-string types

wkozyra95:
# Why

When adding `@expo/config` to `eas-cli` `color-string` types are missing.

# Test plan

Didn't test if it fixes the issue

### DIFF
--- a/packages/configure-splash-screen/src/SplashScreenConfig.ts
+++ b/packages/configure-splash-screen/src/SplashScreenConfig.ts
@@ -1,6 +1,6 @@
-import { Color } from 'color-string';
-
 import { SplashScreenImageResizeModeType, SplashScreenStatusBarStyleType } from './constants';
+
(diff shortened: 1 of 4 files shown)


bbarthec (changes requested):
It's wrong IMO ... 😞 
`@types` are dev time only things. The problem might be happening, because the package exposes the types somehow and it shouldn't 🤔

wkozyra95:
> It's wrong IMO ... disappointed
> `@types` are dev time only things. The problem might be happening, because the package exposes the types somehow and it shouldn't thinking

I disagree with that as a general statement, but in this case it's just one type so yes it's easier to make sure it's not reexported

btw we have some other packages with `@types` in dependencies

fson (approved):

bbarthec (approved):

PR #2740, opened 2020-09-30
[xdl] Configure splash screen for standalone apps according to @expo/config

sjchmiela:
I haven't tested it yet, but if I understand the API correctly this should fix https://github.com/expo/expo/issues/10382. I've used `@expo/config` as it exposed easier API for SplashScreen modification than `@expo/configure-splash-screen` (a single method to which I can pass manifest and path both I already have defined there).

I have also added `@expo/config` warnings integration to `runShellAppModificationsAsync`.

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -1,3 +1,4 @@
+import { AndroidConfig } from '@expo/config';
 import fs from 'fs-extra';
 import { sync as globSync } from 'glob';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


bbarthec:
I'm afraid that without testing it no-go 😞 I'm going to work through this standalone building process, so I'll eventually see if that is working.

sjchmiela:
I also wouldn't merge this without verifying if it works. Not knowing whether this is a right approach to fixing the issue but wanting to help I decided to post the PR to ask for your opinion.

packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1030,6 +1031,11 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
     );
   }
 
+  await AndroidConfig.SplashScreen.setSplashScreenAsync(
```
EvanBacon, line 1034: if I'm not mistaken, we disabled this.

PR #2744, opened 2020-10-01
[config] Fix relative module resolution for config files

nokutana:
I have a number of different app configs and was receiving a `Failed to read config` / `MODULE_NOT_FOUND` error on `expo start` since they import shared config variables. This sets `configFile` as the `filename` argument to `require-from-string` so relative modules can be resolved properly


EvanBacon (approved):
LGTM, good work, good test.

Issue #2591, opened 2020-09-07
Rename all of tests from `.test` to `-test`


EvanBacon:
After https://github.com/expo/expo-cli/pull/2297 is merged, we'll want to unify our test naming convention

EvanBacon:
fixed

PR #2749, opened 2020-10-02
[configure-splash-screen]<chore>: Update test file names to follow naming convention

chernandez7:
### TODO
---
- [ ] Confirm XML naming convention
- [ ] Make sure tests are passing on CircleCI

### Description
--- 
As mentioned in issue #2591 and PR #2297, we want to update tests from `.test` to `-test`.

Only question I have is relating to the `.xml` files. Should those also follow the convention mentioned above or stay as `xml.test`?

Also, it's my first PR towards Expo so if I missed any conventions to follow feel free to let me know.

### Test Plan
---
I don't believe much testing would be needed since it's just renaming files.

I ran the tests on the package but some were failing and it seems like they're related to my environment or even OS. But would like some confirmation on if that's true or not.

```json
Ex. 1:
  -     "image": "/assets/background.png",
  +     "image": "E:\\assets\\background.png",

Ex. 2:
    -   "image": "/assets/background.png",
    +   "image": "E:\\assets\\background.png",

Ex. 3:
 EEXIST: file already exists, mkdir 'E:\app\android\app\src\main\res\drawable'

      at createError (../../node_modules/memfs/lib/volume.js:121:17)
      at Volume.mkdirBase (../../node_modules/memfs/lib/volume.js:1527:19)
      at Immediate.<anonymous> (../../node_modules/memfs/lib/volume.js:681:39)
```


EvanBacon:
You can ensure it works on your machine by running the tests on master.

bbarthec (approved):
This LGTM 👍 
Tests are fired and are passing ✅

PR #2763, opened 2020-10-05
Update Android uninstalling copy

### LINKED ISSUE #2756: Use "Expo client" instead of "Expo"
```
Uninstalling Expo from Android device.
Installing Expo on device
```

we probably do this elsewhere too

brooksbecton:
Fixes #2756

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -426,7 +426,7 @@ export async function isDeviceBootedAsync({
 }
 
 export async function uninstallExpoAsync(device: Device): Promise<string | undefined> {
-  Logger.global.info('Uninstalling Expo from Android device.');
+  Logger.global.info('Uninstalling Expo client from Android device.');
 
   // we need to check if its installed, else we might bump into "Failure [DELETE_FAILED_INTERNAL_ERROR]"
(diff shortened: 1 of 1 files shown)


brentvatne:
thanks! we should also update the installing copy:

> Installing Expo on device

brooksbecton:
May have found two more instances

One says, "_Opens your app in Expo on a connected Android device_"
Wasn't sure if these should say something like "_Opens your app the Expo client on a connected Android device_"

https://github.com/expo/expo-cli/blob/0d6eba36f29ef86e03e09ba394df1ebce3074b37/packages/expo-cli/src/urlOpts.ts#L21
https://github.com/expo/expo-cli/blob/0d6eba36f29ef86e03e09ba394df1ebce3074b37/packages/expo-cli/src/urlOpts.ts#L24

brentvatne (approved):

brentvatne:
> One says, "Opens your app in Expo on a connected Android device"
> Wasn't sure if these should say something like "Opens your app the Expo client on a connected Android device"

i think the latter is indeed better wording! follow-up pr welcome! thanks for the help

brooksbecton:
Would you mind adding a `hacktoberfest-accepted` label to this PR?
If not, I understand.

PR #2772, opened 2020-10-08
[xdl] fix: use ExpoFcmMessagingService in shell apps

cruzach:
see https://github.com/expo/expo/pull/10608


cruzach:
closing because we are opting to handle this without the need to create a scoped messaging service

Issue #2711, opened 2020-09-25
ENOENT: no such file or directory, scandir  when ejecting


rusagent:
## Description

When using expo eject i always get this error: NOENT: no such file or directory, scandir 

## Expected Behavior

expo eject

## Observed Behavior

```
⚠️  iOS config synced with warnings that should be fixed:
- splash: Unable to automatically configure splash screen. Please refer to the expo-splash-screen README for more information: https://github.com/expo/expo/tree/master/packages/expo-splash-screen
ENOENT: no such file or directory, scandir 'G:\repos\NewBelled\beLLed\android\app\src\main\java\G:\repos\NewBelled\beLLed\android\app\src\main\java\com\belled'
Error: ENOENT: no such file or directory, scandir 'G:\repos\NewBelled\beLLed\android\app\src\main\java\G:\repos\NewBelled\beLLed\android\app\src\main\java\com\belled'
    at Object.readdirSync (fs.js:955:3)
    at Object.renamePackageOnDisk (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\node_modules\@expo\config\src\android\Package.ts:78:24)
    at configureAndroidProjectAsync (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\src\commands\apply\configureAndroidProjectAsync.ts:119:31)
    at configureAndroidStepAsync (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\src\commands\eject\Eject.ts:218:3)
    at Object.ejectAsync (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\src\commands\eject\Eject.ts:82:3)
    at action (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\src\commands\eject.ts:48:5)
    at Command.<anonymous> (C:\Users\flori\AppData\Roaming\npm\node_modules\expo-cli\src\exp.ts:302:7)
```

## Environment

  Expo CLI 3.27.10 environment info:
    System:
      OS: Windows 10 10.0.18363
    Binaries:
      Node: 12.18.3 - G:\Program Files\nodejs\node.EXE
      Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.14.6 - G:\Program Files\nodejs\npm.CMD
    SDKs:
      Android SDK:
        API Levels: 28, 29, 30
        Build Tools: 28.0.3, 29.0.2, 30.0.2
[...]

PR #2775, opened 2020-10-09
Add windows tests for expo cli only

byCedric:
This would catch issues like #2711.

It adds windows as OS in the matrix build but also overwrites the `package`'s to test. Windows is really slow compared to the other OSes, and because we have E2E in `expo-cli` I think this is the only one that should run for now. 

As a side note, I removed the `CI: true` env var ([because it's set by default](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables)) and refactored the list to a yaml multiline array.

Also, we can't reuse the cache from the build step. It looks like caching are based on Linux/MacOS based paths, [which fails on windows](https://github.com/expo/expo-cli/runs/1232944843?check_suite_focus=true#step:2:7). So I skipped this and only scoped the windows tests to `expo-cli`'s `e2e` tests, with only building the `expo-cli`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,27 +109,22 @@ jobs:
 
   windows:
     runs-on: windows-latest
-    needs: build
     strategy:
       fail-fast: false
       matrix:
-        node: ['10', '12']
-        package:
-          - expo-cli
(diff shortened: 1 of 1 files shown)


EvanBacon:
We now support windows tests

Issue #2782, opened 2020-10-13
Cannot use env data in dynamic configuration


RedMickey:
Hello,
I am a maintainer of [React Native Tools](https://github.com/microsoft/vscode-react-native) VS Code extension. The extension is used for debugging React Native (including Expo) applications.

## Description

We received an [issue](https://github.com/microsoft/vscode-react-native/issues/1436) with usage of `env` variables in dynamic configuration. Then we tried to reproduce it on a plain Expo project running it from the terminal and we also got an error. In the project we use the [babel-plugin-dotenv-import](https://www.npmjs.com/package/babel-plugin-dotenv-import) package to access env variables.

We have already faced a similar problem before and proposed a fix for it in [this PR](https://github.com/expo/expo-cli/pull/2220).

## Expected Behavior

It would be great to have an opportunity to use env variables in dynamic configuration as it worked earlier.

## Observed Behavior

When we start a project from the terminal using the `expo start`, we get the following error:
![Screen Shot 2020-10-13 at 08 53 27](https://user-images.githubusercontent.com/33267199/95821130-0d717f80-0d32-11eb-9002-b357aa69c4e1.png)

We found out that this problem arose since `@expo/xdl@58.0.4` package after the https://github.com/expo/expo-cli/commit/cf63ad081df676bf5592bac2244259a03eaa7fe7 commit. We consider that the problem could be in the config file compilation [here](https://github.com/expo/expo-cli/blob/master/packages/config/src/evalConfig.ts#L24-L36). We think so, because according to the original error it cannot use `@env` module:
```
'MODULE_NOT_FOUND'

[...]

PR #2786, opened 2020-10-13
eval config with local babel.config

EvanBacon:
- fixes https://github.com/expo/expo-cli/issues/2782 but seems like a risky change to make


EvanBacon:
closing in favor of a secondary config file approach

PR #2791, opened 2020-10-14
[xdl, schemer]<fix>: Fixed reading schema with $ref keywords

bbarthec:
- [xdl, schemer] Added package for resolving $ref keywords in JSON schema objects
- [xdl] Added better tests for assets detection

# Why

I was working with the `UNVERSIONED` app in `expo/expo` and observed splash image is not loading 🤔
I started investigating and it turned out that the manifest that is served from the server is no longer  containing proper paths for local assets (only for UNVERSIONED SDK).
I've dove into the xdl code and with the power of a bunch `console.log`s observed that there is a change in the JSON schema format (cominig from https://github.com/expo/universe/pull/5904).
Mentioned PR introduced `$ref`s in the schema and these are not handled by any code in the `expo-cli` repo 😞 (there's a difference between validation & traversing the schema and the `ajv` nor `json-schema-traverse` is not resolving `$ref`s ([see the section at the bottom from the ajv FAQ](https://ajv.js.org/faq.html) or [`json-schema-traverse` README](https://www.npmjs.com/package/json-schema-traverse)).
I've searched the net for some lib that would handle json schema ref resolution and found https://www.npmjs.com/package/json-schema-traverse.

I've additionally detected that there is one other place in the `schemer` package that is also traversing non-resolved json schema, but only at the top level so it seems safe to leave it as is.

# TODO
- [ ] publish `xdl` & `expo-cli`

# Testing
- [x] created another test for smoke testing assets detection from the schema object (added a few more fields and `UNVERSIONED` SDK test case)
- [x] launched the `sandbox` app from the `expo/expo` repo with `UNVERSIONED` SDK via local `expo-cli` containing this change and the splash image was shown as expected 💪

### DIFF
--- a/packages/schemer/src/index.ts
+++ b/packages/schemer/src/index.ts
@@ -1,6 +1,6 @@
 import Ajv from 'ajv';
 import fs from 'fs';
-import { default as JsonSchemaRefParser } from 'json-schema-ref-parser';
+import schemaDerefSync from 'json-schema-deref-sync';
 import traverse from 'json-schema-traverse';
 import get from 'lodash/get';
 import path from 'path';
@@ -49,7 +49,7 @@ export default class Schemer {
     };
 
     this.ajv = new Ajv(this.options);
-    this.schema = schema;
+    this.schema = schemaDerefSync(schema);
     this.rootDir = this.options.rootDir || __dirname;
     this.manualValidationErrors = [];
   }
@@ -149,10 +149,7 @@ export default class Schemer {
 
   async _validateAssetsAsync(data: any) {
     const assets: AssetField[] = [];
-
-    const schema = await JsonSchemaRefParser.dereference(this.schema);
-
-    traverse(schema, { allKeys: true }, (subSchema, jsonPointer, a, b, c, d, property) => {
+    traverse(this.schema, { allKeys: true }, (subSchema, jsonPointer, a, b, c, d, property) => {
       if (property && subSchema.meta && subSchema.meta.asset) {
         const fieldPath = schemaPointerToFieldPath(jsonPointer);
         assets.push({
--- a/packages/schemer/src/index.ts
+++ b/packages/schemer/src/index.ts
@@ -240,10 +240,6 @@ export default class Schemer {
     }
   }
 
-  /**
-   * This function cannot resolve '$ref' in the schema object
-   * so it works only for non-referenced fields/sub-schemas
-   */
   async validateProperty(fieldPath: string, data: any) {
     const subSchema = fieldPathToSchema(this.schema, fieldPath);
     this.ajv.validate(subSchema, data);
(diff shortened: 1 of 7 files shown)


packages/xdl/src/project/ExpSchema.ts
```diff
@@ -44,8 +46,11 @@ export async function getSchemaAsync(sdkVersion: string): Promise<Schema> {
  * @param sdkVersion
  */
 export async function getAssetSchemasAsync(sdkVersion: string): Promise<string[]> {
-  const schema = await getSchemaAsync(sdkVersion);
+  const rawSchema = await getSchemaAsync(sdkVersion);
   const assetSchemas: string[] = [];
+
+  const schema = await JsonSchemaRefParser.dereference(rawSchema);
```
ide, line 52: Move this into getSchemaAsync so we don't accidentally use schemas with `$ref`

packages/schemer/src/index.ts
```diff
@@ -148,7 +149,10 @@ export default class Schemer {
 
   async _validateAssetsAsync(data: any) {
     const assets: AssetField[] = [];
-    traverse(this.schema, { allKeys: true }, (subSchema, jsonPointer, a, b, c, d, property) => {
+
+    const schema = await JsonSchemaRefParser.dereference(this.schema);
```
ide, line 153: Dereference the schema in the constructor so that this issue is fixed for all fields

ide (changes requested):

packages/xdl/package.json
```diff
@@ -66,6 +66,8 @@
     "internal-ip": "4.3.0",
     "invariant": "2.2.4",
     "is-reachable": "^4.0.0",
+    "json-schema-ref-parser": "^9.0.6",
+    "json-schema-traverse": "0.5.0",
```
EvanBacon, line 70: seems like this should all live in expo/schemer

EvanBacon (changes requested):
AFAICT, the whole PR should just be something like this:
```js
import RefParser from '@apidevtools/json-schema-ref-parser';

export async function getSchemaAsync(sdkVersion: string): Promise<Schema> {
  const json = await _getSchemaJSONAsync(sdkVersion);
  return await RefParser.dereference(json.schema);
}
```

bbarthec:
@EvanBacon, yeah you're mostly right 👍

EvanBacon (approved):
lgtm

EvanBacon:
Just ensure the tests are passing

packages/xdl/src/project/ExpSchema.ts
```diff
@@ -35,7 +36,8 @@ export async function validateAsync(projectRoot: string) {
 
 export async function getSchemaAsync(sdkVersion: string): Promise<Schema> {
```
ide, line 37: Every instantiation of a Schemer instance gets its schema from `getSchemaAsync` so we'll end up dereferencing the schemas twice. We should decide whether we want Schemer to require dereferenced schemas or if it will handle dereferencing.

bbarthec: As there're other code paths that originate in `getSchemaAsync` method and do not go through the `Schemer` object I suggest having the dereferencing code in the `getSchemaAsync` only.

ide (approved):

PR #2800, opened 2020-10-15
[xdl] Support beta versions when EXPO_BETA=1

brentvatne:
This PR changes `Versions.newestReleasedSdkVersionAsync()`:

- Currently a "released" version is defined as being a version that has a `releaseNoteUrl` field attached to it on the versions endpoint
- For beta releases, we will add a boolean `beta` field.
- We only want to return beta releases as the newest released version when the user opts in to it.
- Users are able to opt-in by setting `EXPO_BETA=1` in their environment.

This small change has the following intended impacts:

- `EXPO_BETA=1 expo upgrade` will now be able to upgrade to a beta version
- `EXPO_BETA=1 expo client:install:[ios/android]` outside of a project directory will install the beta version of the client

### DIFF
--- a/packages/expo-cli/src/commands/eject.ts
+++ b/packages/expo-cli/src/commands/eject.ts
@@ -36,14 +36,8 @@ async function action(
 
   // Set EXPO_VIEW_DIR to universe/exponent to pull expo view code locally instead of from S3 for ExpoKit
   if (Versions.lteSdkVersion(exp, '36.0.0')) {
-    // Don't show a warning if we haven't released SDK 37 yet
-    const latestReleasedVersion = await Versions.newestReleasedSdkVersionAsync();
-    if (Versions.lteSdkVersion({ sdkVersion: latestReleasedVersion.version }, '36.0.0')) {
+    if (options.force || (await userWantsToEjectWithoutUpgradingAsync())) {
       await LegacyEject.ejectAsync(projectDir, options as LegacyEject.EjectAsyncOptions);
-    } else {
-      if (options.force || (await userWantsToEjectWithoutUpgradingAsync())) {
-        await LegacyEject.ejectAsync(projectDir, options as LegacyEject.EjectAsyncOptions);
-      }
     }
   } else {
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/eject.ts
```diff
@@ -61,6 +55,7 @@ export default function (program: Command) {
       'Create Xcode and Android Studio projects for your app. Use this if you need to add custom native functionality.'
     )
     .helpGroup('eject')
+    .option('--force', 'Skip legacy eject warnings.')
```
EvanBacon, line 58: I think this is not the best use of `--force`, a better use that comes to mind is the ability to skip the git warning, the ability to always generate a new set of native folders, or the ability to force the yarn install to clear before running.

packages/xdl/src/Versions.ts
```diff
@@ -146,7 +148,10 @@ export async function newestReleasedSdkVersionAsync(): Promise<{
   let result = null;
   let highestMajorVersion = '0.0.0';
   for (const [version, data] of Object.entries(sdkVersions)) {
-    if (semver.major(version) > semver.major(highestMajorVersion) && data.releaseNoteUrl) {
+    if (
+      semver.major(version) > semver.major(highestMajorVersion) &&
+      (data.releaseNoteUrl || (data.beta && getenv.boolish('EXPO_BETA', false)))
```
EvanBacon, line 153: maybe put these in named variables or add comments.

packages/expo-cli/src/commands/eject.ts
brentvatne: this already existed before, we're using `options.force` above for the purpose described here. the description of the command was incorrect before: https://github.com/expo/expo-cli/commit/d7ef4593de17826456bf2b63a0b043feaa070ef3#diff-3fd4eed71d65dfed8c0a6b88b86a23c9c05327ae44ab43fef489e50cc3d7c6e9L69-L72 and i removed it a bit too hastily. i think we should repurpose it at some point to do what you're describing - it probably makes sense to drop it when we no longer support sdk 36 (and therefore no longer support "legacy eject").

PR #2802, opened 2020-10-16
[xdl][expo-cli] add the "target" warning for export

cruzach:
## Why

we warn if you run `expo publish` in a bare project without the `--target` flag. So we should do the same for `expo export` 

to keep things in one place, i moved the `logBareWorkflowWarnings` code from `expo-cli/publish.ts` to `xdl/Project.ts`, since that's where we do the relevant publishing/exporting work. This also prevents us from calling `getConfig` (which is expensive) more times than we need to

Tested in a bare project:
- `expo export` logs warning ✅ 
- `expo publish` logs warning ✅ 
- `expo export --target bare` does not log warning ✅ 
- `expo publish --target bare` does not log warning ✅ 
- remove `expo` from dependencies in package.json, does not log warning ✅ 

Tested in a managed project:
- never logs warning✅

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -388,7 +388,6 @@ export async function exportForAppHosting(
 
   logger.global.info('Finished saving JS Bundles.');
 
-  // save the assets
   const { assets } = await exportAssetsAsync({
     projectRoot,
     exp,
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -355,8 +355,15 @@ export async function exportForAppHosting(
     publishOptions?: PublishOptions;
   } = {}
 ): Promise<void> {
-  const defaultTarget = getDefaultTarget(projectRoot);
-  const target = options.publishOptions?.target ?? defaultTarget;
+  // Get project config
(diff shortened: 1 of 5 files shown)


packages/xdl/src/Project.ts
```diff
@@ -2049,3 +2058,32 @@ export async function stopAsync(projectDir: string): Promise<void> {
...
+ * it will run in Expo client in development even. We should revisit this with
+ * dev client, and possibly also by excluding SDK version for bare
+ * expo-updates usage in the future (and then surfacing this as an error in
+ * the Expo client app instead)
+ *
+ * Related: https://github.com/expo/expo/issues/9517
+ *
+ * @param pkg package.json
+ */
+export function logBareWorkflowWarnings(pkg: PackageJSONConfig) {
```
EvanBacon, line 2075: no new code in Project.ts

EvanBacon (changes requested):
see comment

packages/xdl/src/Project.ts
cruzach: ah okay, didn't know that 👍

EvanBacon:
Is this still relevant now that selecting workflows are deprecated (at least for builds).

PR #2804, opened 2020-10-17
[xdl][cli] Add support for --latest flag in client:install:x

brentvatne:
This PR makes it possible to run:

- `expo client:install:ios --latest`
- `expo client:install:android --latest`

This is useful to skip any prompts and just grab and install the latest client. These are convenient commands we can copy/paste in a variety of situations without providing any additional instructions around the cwd of the terminal, eg: `EXPO_BETA=1 expo client:install:ios --latest`.

A drive-by improvement I made here was to include the client version in the "Installing Expo client" message.

Screen Shot 2020-10-16 at 6 16 02 PM

### DIFF
--- a/packages/expo-cli/src/commands/client/index.ts
+++ b/packages/expo-cli/src/commands/client/index.ts
@@ -289,22 +289,46 @@ export default function (program: Command) {
   program
     .command('client:install:ios')
     .description('Install the Expo client for iOS on the simulator')
+    .option(
+      '--latest',
+      'Ignore the local project version and install the latest available version of Expo client'
+    )
     .helpGroup('client')
-    .asyncAction(async () => {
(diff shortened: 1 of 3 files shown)


EvanBacon (approved):
seems fine

packages/expo-cli/src/commands/client/index.ts
```diff
@@ -370,23 +403,48 @@ export default function (program: Command) {
   program
     .command('client:install:android')
     .description('Install the Expo client for Android on a connected device or emulator')
+    .option(
+      '--latest',
+      'Ignore the local project version and install the latest available version of Expo client'
```
EvanBacon, line 408:
```suggestion
      'Install the latest version of Expo client, ignoring the project\'s SDK version'
```
Maybe lead with the actionable info.

PR #2810, opened 2020-10-19
Upgrade @expo/configure-splash-screen version used in @expo/config

fson:
Update to `0.3.0`. I need this in `eas-cli` because the type definition is broken in `0.2.1`.

It seems that this version number doesn't get updated automatically during releases now that `configure-splash-screen` is in `unlinked-packages`, but maybe that's OK.

Not sure how to best test this change, but CHANGELOG didn't show anything alarming.

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -35,7 +35,7 @@
     "@babel/core": "7.9.0",
     "@expo/babel-preset-cli": "0.2.18",
     "@expo/config-types": "^40.0.0-beta.1",
-    "@expo/configure-splash-screen": "0.2.1",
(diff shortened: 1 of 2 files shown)


brentvatne:
thanks @fson!

the testing process for this is to run eject in a managed app with a splash screen set, using the latest sdk version. so you can init a project, ensure splash screen is configured, then eject using your local copy of expo-cli with this change. when you build the apps for ios and android you should see the splash screen that you set for both. i like to use a splash screen image copied over from https://buildicon.netlify.app/ to be totally sure that everything is being applied properly (because the default splash is the same between managed and bare apps)

PR #2812, opened 2020-10-19
[config] only serialize the expo object

EvanBacon:
- moved from https://github.com/expo/expo-cli/pull/2789
- only serialize the expo object if present, this allows for a more dynamic config to be added in `app.config.js`.


EvanBacon:
moving modifiers into the expo object

Issue #2803, opened 2020-10-17
Logs in terminal are in json


shahidcodes:
## Description
After the server is started the logs in the terminal are appearing in JSON format. The logs are so much (file count logs at the first start) that it sometimes freezes VSCode.

## Expected Behavior
It should print human-readable logs than JSON.

## Observed Behavior
Here is the sample log - 
```
{
  name: 'expo',
  hostname: 'HackerOne',
  pid: 13236,
  type: 'project',
  project: 'D:\\projects\\react-native\\popway-consumer-react-native',
  level: 30,
  tag: 'metro',
  msg: '{"buildID":"1","type":"bundle_transform_progressed","transformedFileCount":1898,"totalFileCount":1941}',
  time: 2020-10-17T00:40:54.573Z,
  id: '6a1d5dd0-1011-11eb-b4ed-33f605e898d7',
  v: 0
}
```

## Environment

```
 Expo CLI 3.28.0 environment info:
    System:
      OS: Windows 10 10.0.18362
    Binaries:
      Node: 12.14.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.10 - C:\Program Files\nodejs\yarn.CMD
      npm: 6.13.4 - C:\Program Files\nodejs\npm.CMD   
    IDEs:
      Android Studio: Version  3.5.0.0 AI-191.8026.42.35.5791312  
    npmPackages:
      expo: ^39.0.3 => 39.0.3
      react: 16.13.1 => 16.13.1
      react-dom: 16.13.1 => 16.13.1
      react-native: https://github.com/expo/react-native/archive/sdk-39.0.0.tar.gz => 0.63.2
      react-native-web: ~0.13.7 => 0.13.13
    Expo Workflow: managed
```

## Reproducible Demo
NA

brentvatne:
> The logs are so much (file count logs at the first start) that it sometimes freezes VSCode.

can you elaborate on why this is running in vscode? are you using some extension? keep in mind that we do not work on the vscode extension, we only explicitly support running `expo-cli` through your terminal

shahidcodes:
@brentvatne I am running it inside vscode terminal. But the issue is the same in the `Windows cmd prompt` or `PowerShell`.

brentvatne:
does this happen in a brand new project created with `expo init` as well?

shahidcodes:
yes, I just created a test project, it's the same.
[...]

PR #2818, opened 2020-10-20
[xdl] Rename debug to expo raw log to avoid collision

byCedric:
This environment var might collide with the [`debug()` package](https://www.npmjs.com/package/debug), if people use it. It's fairly popular so I would imagine this might be the cause of issues like #2803.

Screenshot 2020-10-20 at 20 12 35


byCedric:
Like I mentioned to Brent, I really like the way the debug package can filter certain types of logging. Would be interesting if we can integrate it in CLI to improve error feedback for users and improve requesting information when triaging.

brentvatne (approved):

PR #2832, opened 2020-10-26
[config] add android notification config

cruzach:
# why

With `expo-notifications`, you can modify the notification icon and accent color on Android (iOS forces you to use your app icon), so we should handle that configuration whenever someone `eject`s. 

adds the following to androidmanifest.xml (if applicable):
```
<meta-data
      android:name="expo.modules.notifications.default_notification_icon"
      android:resource="@drawable/notification_icon" />
<meta-data
      android:name="expo.modules.notifications.default_notification_color"
      android:resource="@color/notification_icon_color" />
``` 

and the following to `colors.xml`:
```
<color name="notification_icon_color">#MYCOLOR</color>
```
# test plan

added tests, they pass 🎉 and tested it locally. confirmed rerunning it over and over doesn't cause any duplication (there's a test for that too)


packages/config/src/android/Manifest.ts
```diff
@@ -173,21 +173,23 @@ export function getMainActivity(manifest: Document): ManifestActivity | null {
 export function addMetaDataItemToMainApplication(
   mainApplication: any,
   itemName: string,
-  itemValue: string
+  itemValue: string,
+  itemType: 'resource' | 'value' = 'value'
```
cruzach, line 177: wasn't sure whether or not to make this an `options` object instead, to allow for more customization down the line, but `resource` and `value` are the only two options: https://developer.android.com/guide/topics/manifest/meta-data-element

packages/config/src/android/__tests__/NotificationsManifest-test.ts
```diff
@@ -0,0 +1,75 @@
...
+const notificationConfig: ExpoConfig = {
+  name: 'lol',
+  slug: 'lol',
+  notification: {
+    icon: '/app/assets/notificationIcon.png',
+    color: '#00ff00',
+  },
+};
+
+describe('Applies proper Android Notification configuration to AndroidManifest.xml', () => {
```
cruzach, line 24: put this in a separate file because mocking `fs` was causing some issues for `Manifest. readAndroidManifestAsync`

packages/config-plugins/src/android/Manifest.ts
```diff
@@ -180,18 +180,21 @@ export function getMainActivity(androidManifest: AndroidManifest): ManifestActiv
 export function addMetaDataItemToMainApplication(
   mainApplication: ManifestApplication,
   itemName: string,
-  itemValue: string
+  itemValue: string,
+  itemType: 'resource' | 'value' = 'value'
```
EvanBacon, line 184: Probably want this to be more agnostic, maybe a different method could implement this, then `addMetaDataItemToMainApplication` calls into that method so the changes aren't breaking.

packages/config-plugins/src/android/__tests__/Notifications-test.ts
```diff
@@ -0,0 +1,101 @@
...
+  getNotificationColor,
+  getNotificationIcon,
+  NOTIFICATION_ICON_COLOR,
+  setNotificationIconAsync,
+  setNotificationIconColorAsync,
+} from '../Notifications';
+import { SAMPLE_COLORS_XML } from './fixtures/icon';
+
+jest.mock('fs');
+jest.mock('@expo/image-utils', () => ({
```
EvanBacon, line 16: this is auto mocked, see other tests

packages/config-plugins/src/android/__tests__/Notifications-test.ts
```diff
@@ -0,0 +1,101 @@
...
+  generateImageAsync(_: any, { src }) {
+    const fs = require('fs');
+    return { source: fs.readFileSync(src) };
+  },
+}));
+
+const actualFs = jest.requireActual('fs') as typeof fs;
```
EvanBacon, line 23:
```suggestion
const fsReal = jest.requireActual('fs') as typeof fs;
```
see other tests

packages/config-plugins/src/android/Notifications.ts
```diff
@@ -0,0 +1,145 @@
...
+        await fs.writeFile(path.resolve(dpiFolderPath, NOTIFICATION_ICON + '.png'), resizedIcon);
+      } catch (e) {
+        throw new Error('Encountered an issue resizing Android notification icon: ' + e);
+      }
+    })
+  );
+}
+
+async function removeNotificationIconImageFilesAsync(projectRoot: string) {
+  Promise.all(
```
EvanBacon, line 138:
```suggestion
 await Promise.all(
```

probably want to write a test for this

packages/config-plugins/src/android/Notifications.ts
```diff
@@ -0,0 +1,145 @@
...
+    const colorItemToAdd = buildResourceItem({ name: NOTIFICATION_ICON_COLOR, value: color });
+    colorsJson = Colors.setColorItem(colorItemToAdd, colorsJson);
+  } else {
+    colorsJson = Colors.removeColorItem(NOTIFICATION_ICON_COLOR, colorsJson);
+  }
+  await writeXMLAsync({ path: colorsXmlPath, xml: colorsJson });
+}
+
+async function writeNotificationIconImageFilesAsync(icon: string, projectRoot: string) {
+  Promise.all(
```
EvanBacon, line 109:
```suggestion
  await Promise.all(
```
test

packages/config-plugins/src/android/Notifications.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export function getNotificationColor(config: ExpoConfig) {
+  return config.notification?.color || null;
+}
+
+/**
+ * Applies configuration for expo-notifications, including
+ * the notification icon and notification color.
+ */
+export async function setNotificationIconAsync(config: ExpoConfig, projectRoot: string) {
```
EvanBacon, line 56: Is this used anywhere?

packages/config-plugins/src/android/Notifications.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (icon) {
+    await writeNotificationIconImageFilesAsync(icon, projectRoot);
+  } else {
+    await removeNotificationIconImageFilesAsync(projectRoot);
+  }
+}
+
+export async function setNotificationConfigAsync(
+  config: ExpoConfig,
+  manifestDocument: AndroidManifest
```
EvanBacon, line 67:
```suggestion
  manifest: AndroidManifest
```

see other modules

packages/config-plugins/src/android/Notifications.ts
cruzach: woops, it's supposed to be! created the appropriate plug in

packages/config-plugins/src/android/Manifest.ts
cruzach: since `itemType` defaults to `value` if it's not provided (this is what all the other instances of `addMetaDataItemToMainApplication` expect), this change shouldn't be breaking

you mean to make the `itemType` argument more agnostic? eg:

```suggestion
  itemType?: string
): ManifestApplication {
  itemType = itemType ?? 'value';
```

packages/config-plugins/src/android/__tests__/NotificationsManifest-test.ts
```diff
@@ -0,0 +1,66 @@
...
+    let androidManifestJson = await readAndroidManifestAsync(sampleManifestPath);
+
+    androidManifestJson = await setNotificationConfigAsync(notificationConfig, androidManifestJson);
+    // Run this twice to ensure copies don't get added.
+    androidManifestJson = await setNotificationConfigAsync(notificationConfig, androidManifestJson);
+
+    const mainApplication = getMainApplication(androidManifestJson);
+
+    const notificationIcon = mainApplication['meta-data'].filter(
+      e => e['$']['android:name'] === META_DATA_NOTIFICATION_ICON
```
EvanBacon, line 35:
```suggestion
      e => e.$['android:name'] === META_DATA_NOTIFICATION_ICON
```

PR #2833, opened 2020-10-27
[expo-cli] Fix eas gradle script not working when used with react-native-config

satya164:
Why
===
Currently, when a user has `react-native-config` configured, our configuration for signing config doesn't work. The resulting app gets signed with debug key instead of release key. I tried to debug but couldn't exactly figure out why it happens. Just adding the following in the `react-native-config`'s gradle script produces this behaviour (which shouldn't affect anything?):

```gradle
tasks.whenTaskAdded { task ->
    // Nothing here
}
```

The commit changes the logic to do the configuration on `whenTaskAdded` hook rather than on `afterEvaluate` hook which somehow fixes the issues, probably because then our logic executes before dotenv's script instead of after like previously. Though to be honest, this is result of many trial and errors.

Test plan
=========
I've tested by building the app both with and without the changes, with and without `react-native-config` and verified that the signing key is not the debug key when we're building release build. To test, I used the following command:

```sh
./gradlew assembleRelease
keytool -list -printcert -jarfile app/build/outputs/apk/release/app-release.apk
```

Screen Shot 2020-10-27 at 08 56 26

### DIFF
--- a/packages/config/src/android/EasBuildGradleScript.ts
+++ b/packages/config/src/android/EasBuildGradleScript.ts
@@ -17,18 +17,27 @@ android {
   }
 }
 
-project.afterEvaluate {
-  android.signingConfigs.release { config ->
-    def debug = gradle.startParameter.taskNames.any { it.toLowerCase().contains('debug') }
+def isEasBuildConfigured = false
 
-    if (debug) {
-      return
-    }
+tasks.whenTaskAdded {
+  def debug = gradle.startParameter.taskNames.any { it.toLowerCase().contains('debug') }
 
+  if (debug) {
+    return
+  }
+
+  // We only need to configure EAS build once
+  if (isEasBuildConfigured) {
+    return
+  }
+
+  isEasBuildConfigured = true;
+
+  android.signingConfigs.release {
     def credentialsJson = rootProject.file("../credentials.json");
 
     if (credentialsJson.exists()) {
-      if (config.storeFile && System.getenv("EAS_BUILD") != "true") {
+      if (storeFile && System.getenv("EAS_BUILD") != "true") {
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eas-build/build/templates/gradleContent.ts
```diff
@@ -17,18 +17,27 @@ android {
...
 
-project.afterEvaluate {
-  android.signingConfigs.release { config ->
-    def debug = gradle.startParameter.taskNames.any { it.toLowerCase().contains('debug') }
+def isEasBuildConfigured = false
 
-    if (debug) {
-      return
-    }
+tasks.whenTaskAdded {
```
wkozyra95, line 22: Did you identified how react-native-config affected our script and what was the actual reason for signing with debug keystore?

- Did it stop script not to execute at all?
- Did it override values set by our script
- Did it assign release keystore (to debug one) before our script had a chance to do that
- sth else?

satya164: Like I mentioned in the issue, this is the minimum code which triggers the issues:

```gradle
tasks​.​whenTaskAdded​ { task ​->​
    ​//​ Nothing here​
}
```

Like delete everything in it's Gradle config, place this much and the app is signed with debug key. That doesn't leave me with much to go on.

All of our code is still executed after the tasks are added. I'm afraid that without deep knowledge of how the Gradle system works, I don't have an answer to those questions.

wkozyra95 (approved):
We will need to look into that in the future to make sure it's handling all the cases but for now, it should be enough.

btw I moved this code to @expo/config and added one small change there.

dsokal (approved):

PR #2834, opened 2020-10-27
[xdl] Clean up UrlUtils

EvanBacon:
# Why

I want to add the custom client support but it's hard cuz this code is unruly and untested.

### DIFF
--- a/packages/xdl/src/UrlUtils.ts
+++ b/packages/xdl/src/UrlUtils.ts
@@ -211,10 +211,7 @@ function assertValidOptions(opts: Partial<URLOptions>): URLOptions {
     strict: joi.boolean(),
     minify: joi.boolean(),
     https: joi.boolean().optional(),
-    urlRandomness: joi
-      .string()
-      .optional()
-      .allow(null),
+    urlRandomness: joi.string().optional().allow(null),
   });
 
   const { error } = schema.validate(opts);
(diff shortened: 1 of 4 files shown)


packages/xdl/src/UrlUtils.ts
```diff
@@ -11,24 +13,37 @@ import XDLError from './XDLError';
...
 }
 
 export async function constructManifestUrlAsync(
   projectRoot: string,
-  opts?: any,
+  opts?: Partial<URLOptions>,
   requestHostname?: string
 ) {
-  return constructUrlAsync(projectRoot, opts, false, requestHostname);
+  return await constructUrlAsync(projectRoot, opts ?? {}, false, requestHostname);
```
brentvatne, line 39: probably not needed to include `?? {}` here since `constructUrlAsync` supports opts being `null` as well - could extend it to accept `undefined` w/o changing behavior. but doesn't really matter

packages/xdl/src/UrlUtils.ts
```diff
@@ -134,14 +171,15 @@ export async function constructBundleQueryParamsAsync(projectRoot: string, opts:
   if (usesAssetPluginsQueryParam) {
     // Use an absolute path here so that we can not worry about symlinks/relative requires
     const pluginModule = resolveModule('expo/tools/hashAssetFiles', projectRoot, exp);
-    queryParams += `&assetPlugin=${encodeURIComponent(pluginModule)}`;
+    // queryParams += `&assetPlugin=${encodeURIComponent(pluginModule)}`;
```
brentvatne, line 174:
```suggestion
```

brentvatne (approved):

PR #2837, opened 2020-10-27
[xdl] fix: expo-notifications android standalone configuration

cruzach:
## First and second commits

Apply changes to the `notification_icon_color`, which (paired with THIS PR will fix https://github.com/expo/expo/issues/10469). Defaults to white.

## Third commit

We currently leave the `shell_notification_icon` drawable assets unchanged unless the user specifies one. With the changes introduced in the third commit, we get rid of our Expo icon assets no matter what (no more standalone apps with Expo's notification icon 👍) _and_ we fallback to using the app icon if no `notification.icon` was provided. This isn't recommended since the notification icon really should be all white with a transparent background, but that's well documented as of https://github.com/expo/expo/pull/10828 and defaulting to something is better than nothing (worst case scenario, it will just show a blank square as the icon).

### DIFF
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -1056,6 +1056,15 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
     isRunningInUserContext
   );
 
+  // Notifications
+  if (manifest.notification.color) {
+    await regexFileAsync(
+      '"notification_icon_color">#005eff',
+      `"notification_icon_color">${manifest.notification.color}`,
+      path.join(shellPath, 'app', 'src', 'main', 'res', 'values', 'colors.xml')
+    );
+  }
+
   // Splash Background
   if (backgroundImages && backgroundImages.length > 0) {
     // Delete the placeholder images
--- a/packages/xdl/src/detach/AndroidShellApp.js
+++ b/packages/xdl/src/detach/AndroidShellApp.js
@@ -1057,13 +1057,11 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
   );
 
   // Notifications
-  if (manifest.notification.color) {
-    await regexFileAsync(
-      '"notification_icon_color">#005eff',
-      `"notification_icon_color">${manifest.notification.color}`,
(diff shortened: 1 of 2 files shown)


packages/xdl/src/detach/AndroidShellApp.js
```diff
@@ -1056,6 +1056,13 @@ export async function runShellAppModificationsAsync(context, sdkVersion, buildMo
     isRunningInUserContext
   );
 
+  // Notifications
+  await regexFileAsync(
+    '"notification_icon_color">#005eff',
```
sjchmiela, line 1061: This seems a bit dangerous to me — there's no information in the original manifest that this should be kept in sync with this pattern. Would we be able to write a regex that catches anything between? Like…

```suggestion
    /"notification_icon_color">.*?</s,
```
? (not sure about the `/s` flag, I used https://regex101.com and it allowed for matching tags spanning across many lines. this may be overengineering, but the original comment is still relevant I think)

sjchmiela (approved):
Looks good to me! Thank you for taking care of this!

One thing I would maybe add were small comments on what each part of the code does even if it's clear — one day maybe we'd be able to remove/split/extract stuff out of XDL and these comments may help in this task and in understanding what goes wrong if it does.

packages/xdl/src/detach/AndroidShellApp.js
cruzach: very good point, I'll change this to matching anything in between

We follow the same pattern (relying on the original manifest **not** changing) when it comes to [setting the `splashscreen_background` color as well](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/AndroidShellApp.js#L783-L795), do you think we should make the same changes there? in a separate PR, of course

summerkiflain:
@cruzach after adding `color` as suggested [here](https://github.com/expo/expo/issues/10469#issuecomment-714746004) local notification's color is showing fine but push notification icon's color is gone now and its showing default color like it was for local notification before adding `color` key, I have tried adding `color` key in push notifications sent via `expo-server-sdk` but its not working, any idea?, I have tested on `Samsung A31`, `Android 10`, `SDK 39`, `StandAlone version`.

Issue #2107, opened 2020-05-12
Strongly suggest using a good terminal for Windows


brentvatne:
cmd.exe is not a good tool for development. They should use Powershell or Bash via WSL or something instead. If we can detect WSL, we should also warn and suggest using WSL2

Fahad-Mahmood:
I have created a pre-start script that logs for windows: 'Please avoid using CMD for development. Use PowerShell or Bash via WSL';

`
let os = require('os');

// check for windows OS
if (os.platform() === 'win32') {
  console.log('Please avoid using cmd.exe for development. Instead use PowerShell or Bash via WSL');
}
`

brentvatne:
nice! can you open a pull request?

brentvatne:
you could add it to https://github.com/expo/expo-cli/blob/master/packages/expo-cli/bin/expo.js

Fahad-Mahmood:
right now I have added it in scripts folder with the file name: "check-windows.js";
package.json looks like this with prestart: 
````
"scripts": {
    "format": "prettier --write .",
    "build": "lerna run prepare --stream",
    "postinstall": "expo-yarn-workspaces check-workspace-dependencies",
    "publish": "echo \"This script is deprecated. Run \\\"node ./scripts/publish.js\\\" instead.\"; exit 1",
    "prestart": "node ./scripts/check-windows.js",
    "start": "lerna --ignore \"@expo/{dev-tools,next-adapter,electron-adapter}\" --ignore expo-optimize --ignore pod-install --ignore uri-scheme run watch --parallel",
    "tsc": "echo 'You are trying to run \"tsc\" in the workspace root. Run it from an individual package instead.' && exit 1",
    "lint": "eslint . --ext js,ts"
  },
````

Fahad-Mahmood:
should it work for expo? not for expo-cli development.

Fahad-Mahmood:
added script in the mentioned file and created pull request here: https://github.com/expo/expo-cli/pull/2164

Fahad-Mahmood:
Did you check PR? What should be done to improve it?

PR #2838, opened 2020-10-28
[expo-cli]<feat>: Add cmd.exe detection on windows & warn about it

bbarthec:
Resolves https://github.com/expo/expo-cli/issues/2107

I've only succeeded in checking whether the `expo-cli` is run from `cmd.exe`.

## HOW

I've found a package that seems to be quite popular and behaves similar to `ps aux`: https://www.npmjs.com/package/ps-list.
I'm using it to find the parent process of the current running program taking advantage of NodeJS API `process.ppid`.

The problem with this approach is that on Windows we're querying the processes table and that is taking some time.
I've measured the timings on VMWare Fusion running Windows 10 and checking for `cmd.exe` takes `~200ms`.

## PREVIEW
Screenshot 2020-10-28 at 13 20 09

## ABOUT `WSL`

Unfortunately, I've failed to find any reliable way to distinguish between WSL v1 & v2.
There's a package that can detect running the script on WSL though (https://www.npmjs.com/package/is-wsl), but once we're in WSL we don't really need to warn about it 😅 

Although there's a way to check WSL version:
```
PS C:\Users\Bartłomiej Bukowski> wsl -l -v
  NAME            STATE           VERSION
* Ubuntu-20.04    Running         1 <<- this is version
```
it only runs in PowerShell/CMD, so we're unable to check from inside WSL what version we're running using this technique.


packages/expo-cli/bin/expo.js
```diff
@@ -17,25 +17,51 @@ var supportedVersions =
...
+    const psList = require('ps-list');
+    await (async () => {
+      const usersProcesses = await psList({ all: false });
+      // find parent process name
+      const shellProcess = usersProcesses.find(({ pid }) => pid === process.ppid) || {};
+      if ((shellProcess.name || '').toLowerCase().includes('cmd.exe')) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          yellow(
+            'WARNING: Please avoid using cmd.exe for development. Instead use PowerShell or Bash via WSL.\n'
```
brentvatne, line 55:
```suggestion
            'WARNING: We recommend using PowerShell or Bash via WSL for development with expo-cli on Windows. You may encounter issues using cmd.exe.\n'
```

brentvatne:
> The problem with this approach is that on Windows we're querying the processes table and that is taking some time.
I've measured the timings on VMWare Fusion running Windows 10 and checking for cmd.exe takes ~200ms.

ah, that's too bad. some ideas:

- maybe we can make this an asynchronous warning rather than blocking
- we could cache it per pid so it's only slow the first time you run it in a terminal (seems like a lot of work for a small feature like this)
- we could move it to `expo doctor` or `expo diagnostics`

bbarthec:
@brentvatne, I've adjusted the warning, detection process (removed the `ps-list` dependency, bc it is shipped with `fastlist` binary for windows) and scoped the detection into `doctor` command.

brentvatne (approved):

packages/expo-cli/bin/expo.js
```diff
@@ -41,4 +41,4 @@ if (major > 15) {
   process.exit(1);
 }
 
-require('../build/exp.js').run('expo');
+require('../build/exp.js').run('expo');
```
brentvatne, line 44:
```suggestion
require('../build/exp.js').run('expo');

```

PR #2840, opened 2020-10-28
[configure-splash-screen] Use proper bin paths to files

byCedric:
I noticed these weren't matching after I linked it with `yarn link`. Maybe we should include this folder in the `yarn workspace` so installs are easier?

### DIFF
--- a/unlinked-packages/configure-splash-screen/package.json
+++ b/unlinked-packages/configure-splash-screen/package.json
@@ -3,8 +3,8 @@
   "version": "0.3.0",
   "description": "Supplementary module for 'expo-splash-screen' providing cli configuration command",
   "bin": {
-    "configure-splash-screen": "./build/src/index-cli.js",
-    "expo-splash-screen": "./build/src/index-cli.js"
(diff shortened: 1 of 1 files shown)


bbarthec (approved):

brentvatne:
> Maybe we should include this folder in the yarn workspace so installs are easier?

i tried to explain the motivation for it living in the unlinked-packages directory in the readme in that directory:

# unlinked-packages

Packages in this directory are published independently of other packages in the repository and should not be symlinked in development. They may be depended on by packages in `packages/` but should be consciously and carefully versioned. It's likely that libraries here should be moved somewhere else or need to be fundamentally re-designed, eg: configure-splash-screen lives here because it only supports one very specific version of expo-splash-screen at any given time, but it should probably support multiple versions of expo-splash-screen.

The publish script used for `packages` will not publish packages from this directory. They should be published manually when needed, and updates to other libraries that depend on them should be updated manually.

<hr />

let me know if you have any questions about that. it's certainly not ideal, just the state of the world today

Issue #2675, opened 2020-09-21
Ejecting to bare workflow in Windows 10 issues with package name/path


popeating:
## Description

When ejecting to bare workflow (using Windows10 and Powershell) the resulting Android project do not create correctly the path and files based on package name choosen while ejecting; also package and reference in MainActivity.java, MainApplication.java, generated/BasePackageList.java are incomplete

## Expected Behavior

Ejecting should create path, files, package and reference correctly

## Observed Behavior

First of all i created a new expo project with:
```expo init eject_app```
and everything went with no error:
![image](https://user-images.githubusercontent.com/10193354/93821658-23be7b00-fc5f-11ea-9917-78a915d1d2a3.png)

Then i entered the app folder and tested (successfully) the app in expo with Android emulator
![image](https://user-images.githubusercontent.com/10193354/93821837-7ac45000-fc5f-11ea-9158-1a2e7465d763.png)

At this point i ejected the app using:
```expo eject```
and used com.popland.eject_app as package name and com.popland.ejectapp as iOS Bundle identifier
![image](https://user-images.githubusercontent.com/10193354/93822387-72b8e000-fc60-11ea-9f38-1bdd6ab925eb.png)

i got a warning about MainActivity (that's because MainActivity was not created where it should)
if i check 
```eject_app\android\app\src\main\java\com```
there is ejectapp folder that contains

- MainActivity.java

- MainApplication.java 

- generated\BasePackageList.java

![image](https://user-images.githubusercontent.com/10193354/93823203-bc55fa80-fc61-11ea-9f5f-fc53c5a9ebab.png)

and popland (which contains the empty folder eject_app)
![image](https://user-images.githubusercontent.com/10193354/93823284-d98ac900-fc61-11ea-9a97-c406ca5827c2.png)

if i inspect the above java file the package name is wrong (but coherent to the path where the files are created) and the import reference is wrong too (line 14)
![image](https://user-images.githubusercontent.com/10193354/93823473-2bcbea00-fc62-11ea-8b08-215e3e004161.png)

now, when i run 
```yarn android```
the app fail to compile with the following error (repeated more times)

```
C:\Users\lorenzo\eject_app\android\app\src\main\java\com\ejectapp\MainApplication.java:39: error: cannot find symbol
      return BuildConfig.DEBUG;
             ^
  symbol: variable BuildConfig
```
[...]

PR #2841, opened 2020-10-28
[config] Force ios entitlement paths to be in posix

byCedric:
This is a fix that doesn't do a lot currently, since we disabled ejecting iOS on windows. But this, in combination [with this `xcode` fork](https://github.com/byCedric/cordova-node-xcode/tree/fix/windows-issues), actually fixes all differences in the iOS files we get from ejecting the same app on both platforms.

I'll try to push the fork to upstream, [but there isn't a lot of activity in the repo](https://github.com/apache/cordova-node-xcode/issues/116). So we probably need to fork and republish instead.

This would fix (partially, the fork changes arent in here yet) #2675 if we enable ejecting ios on windows again.

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -45,6 +45,7 @@
     "require-from-string": "^2.0.2",
     "resolve-from": "^5.0.0",
     "semver": "^7.1.3",
+    "slash": "^3.0.0",
     "slugify": "^1.3.4",
     "xcode": "^2.1.0",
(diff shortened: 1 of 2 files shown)


byCedric:
See this PR what happens with this change + fork when I re-eject on windows after removing the `ios` folder and without pod-install ofc. https://github.com/byCedric/expo-eject-ios-windows/pull/1/files

PR #2848, opened 2020-11-03
[config] Updated Android Facebook module to better accommodate plugins

EvanBacon:
- `Document` -> `AndroidManifest`
- Support a proper `assert` function
- Create a typed object instead of parsing a string (async)
- Split `applyFacebookAppIdString` out

### DIFF
--- a/packages/config/src/android/Facebook.ts
+++ b/packages/config/src/android/Facebook.ts
@@ -1,18 +1,44 @@
-import { Parser } from 'xml2js';
-
 import { ExpoConfig } from '../Config.types';
+import { assert } from '../Errors';
 import {
   addMetaDataItemToMainApplication,
-  Document,
+  AndroidManifest,
   getMainApplication,
+  ManifestActivity,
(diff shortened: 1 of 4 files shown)


packages/config/src/android/Facebook.ts
```diff
@@ -1,18 +1,44 @@
...
   removeMetaDataItemFromMainApplication,
 } from './Manifest';
-import { readResourcesXMLAsync, ResourceItemXML } from './Resources';
+import { buildResourceItem, readResourcesXMLAsync, ResourceXML } from './Resources';
 import { getProjectStringsXMLPathAsync, removeStringItem, setStringItem } from './Strings';
 import { writeXMLAsync } from './XML';
 
-const facebookSchemeActivity = (scheme: string) => `
-<activity
+function buildXMLItem({
```
brentvatne, line 15: i imagine we might want to pull these utilities out somewhere more generic in the future

brentvatne (approved):

PR #2850, opened 2020-11-03
[cli] Use exact @expo/eas-build-job version, update it, and fix related TS errors

brentvatne:
@sstur pointed out that he was experiencing a new error today when running a build with EAS Build preview:

Screen Shot 2020-11-03 at 10 38 11 AM

He noticed that `BuildType` was no longer exported from `@expo/eas-build-job`, I looked into this and it seems that the reason we didn't catch this locally is that we use the semver `^` matcher for the `@expo/eas-build-job` dep in expo-cli and so unless we manually update the lockfile every time we publish a new version of it, it's possible that we won't catch errors. Users installing via npm were getting `0.1.2` and in the expo-cli repo we were getting `0.1.1`.

So, I think for now it is reasonable for us to use a fixed version of `@expo/eas-build-job` for each version of expo-cli.

@wkozyra95 - I left some comments here for your consideration, but I'm going to move ahead with deploying this to unblock people previewing EAS Build

### DIFF
--- a/packages/expo-cli/src/commands/eas-build/build/builders/AndroidBuilder.ts
+++ b/packages/expo-cli/src/commands/eas-build/build/builders/AndroidBuilder.ts
@@ -1,5 +1,5 @@
 import { AndroidConfig } from '@expo/config';
-import { Android, BuildType, Job, Platform, sanitizeJob } from '@expo/eas-build-job';
+import { Android, Job, Platform, sanitizeJob, Workflow } from '@expo/eas-build-job';
 import path from 'path';
 
 import CommandError from '../../../../CommandError';
@@ -11,7 +11,6 @@ import {
   AndroidGenericBuildProfile,
   AndroidManagedBuildProfile,
   CredentialsSource,
-  Workflow,
 } from '../../../../easJson';
 import { gitAddAsync, gitRootDirectory } from '../../../../git';
 import { Builder, BuilderContext } from '../../types';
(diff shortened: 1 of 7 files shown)


packages/expo-cli/src/commands/eas-build/build/builders/AndroidBuilder.ts
```diff
@@ -183,9 +182,7 @@ class AndroidBuilder implements Builder<Platform.Android> {
   ): Promise<Partial<Android.ManagedJob>> {
     return {
       ...(await this.prepareJobCommonAsync(archiveUrl)),
-      type: BuildType.Managed,
-      packageJson: { example: 'packageJson' },
```
brentvatne, line 187: `packageJson` and `manifest` weren't included in the `ManagedJob` type and so i removed them

packages/expo-cli/src/easJson.ts
```diff
@@ -1,19 +1,10 @@
-import { Platform } from '@expo/eas-build-job';
+import { Platform, Workflow } from '@expo/eas-build-job';
 import Joi from '@hapi/joi';
 import fs from 'fs-extra';
 import path from 'path';
 
 // TODO(wkozyra95): move it to @expo/config or to separate package
 
-// Workflow is representing different value than BuildType from @expo/eas-build-job
```
brentvatne, line 8: is this still true now that it's called [`Workflow` in `@expo/eas-build-job`](https://github.com/expo/turtle-v2/blob/6ffc8aeab05927476e82c5871abdac06896deb5b/src/libs/eas-build-job/src/constants.ts#L1-L4)? if so, we should maybe differentiate between the type names. for the sake of fixing builds right now i assumed they're the same, but we restore this workflow enum later and do whatever needs to be done

wkozyra95: it's no longer true, originally when we added those types we assumed that client build process will be different from standalone and with dev client, it's almost the same

Issue #2845, opened 2020-11-02
Android Keystore: error parsing PKCS#12 with DSA encrypted certificate


kratoskp:
## Description

The following is already set in the environment:

EXPO_ANDROID_KEYSTORE_FILE=/path/to/keystore.jks
EXPO_ANDROID_KEYSTORE_ALIAS=“alias”
EXPO_ANDROID_KEY_PASSWORD=“password”
EXPO_ANDROID_KEYSTORE_PASSWORD=“password”

This is the command that I have executed locally:

expo build:android --non-interactive --release-channel build-33 --keystore-path=$EXPO_ANDROID_KEYSTORE_FILE --keystore-alias=$EXPO_ANDROID_KEYSTORE_ALIAS

## Expected Behavior

Build come out fine

## Observed Behavior

```
[11:34:10] Checking if there is a build in progress...

[11:34:10] Accessing credentials for kratoskp in project easysocial
[11:34:18] An unknown error occurred.
```

## Environment

Expo CLI 3.28.2 environment info:
    System:
      OS: macOS 10.15.7
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 14.4.0 - /usr/local/bin/node
      Yarn: 1.22.5 - /usr/local/bin/yarn
      npm: 6.14.8 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.9.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.1, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    IDEs:
      Android Studio: 4.0 AI-193.6911.18.40.6626763
      Xcode: 12.1/12A7403 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^39.0.0 => 39.0.3 
      react: 16.13.1 => 16.13.1 
      react-native: https://github.com/expo/react-native/archive/sdk-39.0.3.tar.gz => 0.63.2 
      react-navigation: ^4.4.0 => 4.4.2 
    npmGlobalPackages:
      expo-cli: 3.28.2
    Expo Workflow: managed

kratoskp:
Seems like there's a problem when uploading the keystore through the credentials manager and hitting this error:

ApiV2Error: An unknown error occurred.
[...]

PR #2854, opened 2020-11-04
[pkcs12] add fingerprint support for unparseable x509 certs

quinlanj:
# why
fixes https://github.com/expo/expo-cli/issues/2845

`node-forge` only supports x509 certificates that is RSA encrypted. This is causing problems because some of our users are passing in DSA encrypted x509s. 

Fortunately, if `node-forge` errors when parsing the x509 certificate, it'll pass back its asn1 representation instead (see [here](https://github.com/digitalbazaar/forge/blob/1887cfce43a8f5ca9cb5c256168cf12ce1715ecf/lib/pkcs12.js#L679)). We can use this asn1 value to compute the certificate fingerprints instead.

This PR does the following:
- adds `getX509Asn1ByFriendlyName`: extract the asn1 representation of the x509 instead of trying to parse it into a certificate object
- adds `getAsn1Hash`: computes the hash of an asn1 object
- adds more tests, re-organizes current tests


packages/pkcs12/src/index.ts
```diff
@@ -43,11 +39,40 @@ export function getX509CertificateByFriendlyName(
   if (!bags || bags.length === 0) {
     return null;
   }
-  const certificate = bags[0].cert;
-  if (!certificate) {
+  return extractx509CertificateFromBag(bags[0]);
+}
+
+function extractx509CertificateFromBag(bag: forge.pkcs12.Bag): forge.pki.Certificate {
```
ide, line 45:
```suggestion
function extractX509CertificateFromBag(bag: forge.pkcs12.Bag): forge.pki.Certificate {
```

packages/pkcs12/src/index.ts
```diff
@@ -43,11 +39,40 @@ export function getX509CertificateByFriendlyName(
...
+  const { cert, asn1 } = bag;
+  if (!cert && asn1) {
+    // if asn1 is present but certificate isnt, the certificate type was unknown
+    // github.com/digitalbazaar/forge/blob/1887cfce43a8f5ca9cb5c256168cf12ce1715ecf/lib/pkcs12.js#L703
+    throw new Error('PKCS12: unknown x509 certificate type');
```
ide, line 50:
```suggestion
    throw new Error('PKCS12: unknown X.509 certificate type');
```

packages/pkcs12/src/__tests__/index-test.ts
```diff
@@ -39,7 +41,29 @@ const uppercaseAliasKeystoreP12 = {
...
+/**
+ * Generate DSA certificate:
+ * openssl dsaparam -out params.pem 3072
+ * openssl gendsa -out key.pem params.pem
+ * openssl req -new -key key.pem -out req.pem
+ * openssl x509 -req -in req.pem -signkey key.pem -out certificate.cer
+ *
+ * Package into a PKCS#12 keystore:
+ * keytool -import -alias test -file certificate.cer -keystore dsapkcs.keystore -storetype pkcs12
+ *
```
ide, line 54: How are the fingerprints calculated?

packages/pkcs12/src/__tests__/index-test.ts
```diff
@@ -87,6 +111,79 @@ describe('reading PKCS#12 files', () => {
     expect(sha1Fingerprint).toEqual(expectedSha1Fingerprint);
     expect(sha256Fingerprint).toEqual(expectedSha256Fingerprint);
   });
+});
+describe('computing fingerprints from the asn1 representation of a x509 certificate', () => {
+  it('computes fingerprints of unsupported certificates using their asn1 values in p12 keystores', async () => {
```
ide, line 116:
```suggestion
  it.each([
    ['supported RSA certificates', dsaKeystoreP12],
    ['unsupported RSA certificates', keystoreP12],
  ])('computes fingerprints of %s using their ASN.1 values in PKCS #12 keystores', async (_description, keystoreP12) => {
```

ide (changes requested):

packages/pkcs12/src/__tests__/index-test.ts
quinlanj: fingerprints are calculated by getting the asn.1 representation of the certificate, then hashing it.

ide (approved):

PR #2856, opened 2020-11-05
[xdl] fix updates ON_ERROR_RECOVERY setting for SDK 39 Android apps.

esamelson:
This PR fixes the updates.checkAutomatically `ON_ERROR_RECOVERY` behavior for SDK 39 standalone Android apps. Currently this setting is not applied properly.

I just found this change in my local working directory -- I think I discovered this issue (and found the fix) during QA, but managed to forget to actually push and PR the change 😮 🤦 

TODO:
- release new xdl with this change
- update xdl in turtle and redeploy Android turtle to production


brentvatne (approved):
ah, of course this happens in a .js file since manifest is untyped here 😆

PR #2862, opened 2020-11-06
chore: remove unused deps, RE NPM Advisory 1556

Nytelife26:


brentvatne:
thanks for the pr! can you please explain the context on this further?

Nytelife26:
Sure! NPM Advisory 1556 is an issue with `node-fetch` resulting in a DoS vulnerability. This is carried through to expo's next-adapter package through the use of a deprecated (and unused) library, `@types/next`. Therefore, removing it does not require a rewrite of any other parts of the package, and resolves the security issue.

I hope that helps :) @brentvatne

packages/next-adapter/package.json
```diff
@@ -60,7 +60,6 @@
   "dependencies": {
     "@expo/package-manager": "0.0.33",
     "@expo/webpack-config": "0.12.43",
-    "@types/next": "^8.0.6",
```
EvanBacon, line 63: We export next.js types from this package

brentvatne: it looks like in next@9 this becomes unnecessary https://www.npmjs.com/package/@types/next

Nytelife26: It is unnecessary, hence why I said it's deprecated, not to mention it isn't used anywhere else in the library.
Although thank you for being critical rather than just blindly merging, that's always good to see.

PR #2863, opened 2020-11-07
[config] add getPublicExpoConfig method and tests

esamelson:
First piece of `Constants.manifest` support in bare workflow. Later we'll be adding a script to `expo-constants` that evaluates and generates and app config to include in bare apps for `Constants.manifest`.

This PR adds a method to `@expo/config` which that script can use to get the subset of the app config intended for public consumption (e.g. the data that would be served with the manifest from expo-cli or www).

From reading the publish code in XDL, it looks like the pieces of the app config that are deleted *for the purpose of not being visible in the public manifest* are the `hooks`, `ios.config`, and `android.config` fields. (Many other fields are changed, of course, but these seem to be the sensitive ones.) Therefore, the current implementation of this method simply deletes those fields.

Added some unit tests to ensure there are no accidental regressions here, and that other fields aren't touched.

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -15,7 +15,6 @@ import {
   Platform,
   ProjectConfig,
   ProjectTarget,
-  PublicExpoConfig,
   WriteConfigOptions,
 } from './Config.types';
 import { ConfigError } from './Errors';
@@ -75,6 +74,8 @@ export function getConfigWithMods(projectRoot: string, options?: GetConfigOption
  * If a function is exported from the `app.config.js` then a partial config will be passed as an argument.
  * The partial config is composed from any existing app.json, and certain fields from the `package.json` like name and description.
  *
+ * If options.omitPrivateExpoConfig is true, the Expo config will include only public-facing portions.
+ * The resulting config should be suitable for hosting or embedding in a publicly readable location.
  *
  * **Example**
  * ```js
@@ -109,7 +110,7 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}): 
   );
 
   function fillAndReturnConfig(config: SplitConfigs, dynamicConfigObjectType: string | null) {
-    return {
+    const configWithDefaultValues = {
       ...ensureConfigHasDefaultValues(
         projectRoot,
         config.expo,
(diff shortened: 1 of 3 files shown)


brentvatne (approved):

packages/config/src/Config.ts
```diff
@@ -149,6 +150,28 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}):
...
+/**
+ * Evaluate and return the public-facing portions of the Expo config for a project.
+ * The resulting config should be suitable for hosting or embedding in a publicly readable location.
+ *
+ * @param projectRoot the root folder containing all of your application code
+ * @param options enforce criteria for a project config
+ */
+export function getPublicExpoConfig(
+  projectRoot: string,
+  options: GetConfigOptions = {}
```
EvanBacon, line 162: Maybe this should be an option in `GetConfigOptions`?

packages/config/src/Config.ts
```diff
@@ -121,6 +123,20 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}):
       dynamicConfigPath: paths.dynamicConfigPath,
       staticConfigPath: paths.staticConfigPath,
     };
+
+    if (options.omitPrivateExpoConfig) {
```
EvanBacon, line 127:
```suggestion
    if (options.isPublicManifest) {
```
Maybe something like this indicating what the config is being used for.

packages/config/src/Config.ts
```diff
@@ -74,6 +74,8 @@ export function getConfigWithMods(projectRoot: string, options?: GetConfigOption
  * If a function is exported from the `app.config.js` then a partial config will be passed as an argument.
  * The partial config is composed from any existing app.json, and certain fields from the `package.json` like name and description.
  *
+ * If options.omitPrivateExpoConfig is true, the Expo config will include only public-facing portions.
```
EvanBacon, line 77:
```suggestion
 * If options.omitPrivateExpoConfig is true, the Expo config will include only public-facing options (omitting private keys).
```

packages/config/src/Config.ts
esamelson: would like to avoid the word "manifest" because we're trying to start being more intentional about separating this from the app config. Would `isPublicConfig` or `isPublicExpoConfig` be ok with you?

EvanBacon (approved):
lgtm

jvgaeta:
This is causing my integration with react-native-maps to be broken. In the expo mapview docs, you have to set the field android.config.googleMaps.apiKey. Eas is using expo config and this pr causes that key to be filtered out, making the build of my app unusable.

PR #2864, opened 2020-11-07
[config][xdl] migrate project/publish to getPublicExpoConfig

esamelson:
Follow-up to https://github.com/expo/expo-cli/pull/2863 -- not necessary for `Constants.manifest` support but just to clean up the codebase a little bit.

Since we essentially use the public expo config during `expo publish` and `expo export`, this PR makes those commands use the actual `getPublicExpoConfig` method for this rather than duplicating its logic. Hopefully this will help catch issues and keep the public config in sync between publish/export and `Constants.manifest`.

Tested this by publishing NCL (which does have private config fields) with and without this change and then diffing the resulting manifests served by www. Ensured that no fields changed other than the ones we'd expect due to the time difference between publishes. (Definitely open to suggestions for how to test this more thoroughly, also.)

### DIFF
--- a/packages/xdl/src/Project.ts
+++ b/packages/xdl/src/Project.ts
@@ -4,6 +4,7 @@ import {
   ExpoConfig,
   getConfig,
   getDefaultTarget,
+  getPublicExpoConfig,
   Hook,
   HookArguments,
   HookType,
@@ -388,7 +389,7 @@ export async function exportForAppHosting(
   // save the assets
   // Get project config
   const publishOptions = options.publishOptions || {};
-  const { exp, pkg } = await _getPublishExpConfigAsync(projectRoot, publishOptions);
+  const { exp, pkg, hooks } = await _getPublishExpConfigAsync(projectRoot, publishOptions);
   const { assets } = await exportAssetsAsync({
     projectRoot,
     exp,
@@ -416,8 +417,6 @@ export async function exportForAppHosting(
   }
 
   // Delete keys that are normally deleted in the publish process
-  const { hooks } = exp;
-  delete exp.hooks;
   const validPostExportHooks: LoadedHook[] = prepareHooks(hooks, 'postExport', projectRoot, exp);
 
   // Add assetUrl to manifest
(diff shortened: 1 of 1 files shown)


packages/config/src/Config.types.ts
```diff
@@ -65,7 +65,7 @@ export type PublicExpoConfig = Omit<ExpoConfig, 'hooks' | 'ios' | 'android'> & {
   android?: Omit<ExpoConfig['android'], 'config'>;
 };
 
-export type ExpoAppManifest = ExpoConfig & {
```
brentvatne, line 68: 👌

brentvatne: later on in this type:

```
ios?: { publishSourceMapPath?: string } & ExpoConfig['ios'];
```

should we update that too?

```
ios?: { publishSourceMapPath?: string } & PublicExpoConfig['ios'];
```

same for android

packages/xdl/src/Project.ts
```diff
@@ -416,8 +417,6 @@ export async function exportForAppHosting(
   }
 
   // Delete keys that are normally deleted in the publish process
-  const { hooks } = exp;
```
brentvatne, line 419: can remove the comment on the line above this too

brentvatne:
while reading this over, I noticed that the `_getPublishExpConfigAsync` function has a surprising side effect - we mutate the `options` object after validating the release channel if it is falsey. this seems unexpected, we should probably do that in a separate function (perhaps in the caller of `_getPublishExpConfigAsync` or a fn like `ensureValidReleaseChannel`). this could be a nice little cleanup follow up task.

one small nit - we take `exp` returned from `_getPublishExpConfigAsync`, which is type `ExpoAppManifest`, and pass it in to `prepareHooks`, where the expected type is `ExpoConfig`. maybe the `exp` argument to `prepareHooks` should be `ExpoPublicConfig` instead.

esamelson:
Thanks, nice catches. The fix for the `ios` and `android` types on `ExpoAppManifest` forced a couple other changes and also made clear that the `Omit` type was not working as I intended for those fields on `PublicExpoConfig` -- it needs the types to be non-nullable.

Agree that mutating the options object in a separate function would be a good follow-up task -- I'll take care of that (or open a ticket for it) after this lands.

brentvatne (approved):

PR #2868, opened 2020-11-10
[config] Generated gradle files

EvanBacon:
- [docs](https://github.com/expo/expo-cli/blob/@evanbacon/config/android-generated-gradle-files/packages/config-plugins/README.md#generated-gradle-files)

Instead of modifying the gradle files directly, we should use generated gradle files which the project could ignore (probably shouldn't though). These files would be rewritten every time `expo eject` is run to prevent needing to clean up old files. This would provide expo-camera on android a system for safely appending the custom camera maven package to an Android project.

We currently require the following gradle file modifications:

- `expo-camera` https://www.npmjs.com/package/expo-camera#configure-for-android - a maven package must be added to the project `build.gradle`.
- `react-native-lottie` https://github.com/react-native-community/lottie-react-native (these could be resolved with a PR that adds auto linking support)
- google-services (`android.googleServicesFile`) - requires classpath be added to the project `build.gradle` and `apply plugin` be added to `app/build.gradle`. We should migrate this to the new system.

## Implementation

~~Both `build.gradle` files~~ Just the project build.gradle has an import applied to the bottom of the file (bottom seems most powerful). The import references a generated file in `android/.expo/*-build.gradle`. These files are generated on each run of `expo eject` so their state can be easily predicted. The files cannot be easily git ignored because a team member would need to run `expo eject` after pulling the project which is a future goal, but right now more inconvenient.

The project `build.gradle` is generated without indentation and uses comment tags (`@begin`, `@end`, `@plugin`) to predict where code should be added, or removed. 

A convenience plugin `Gradle.withProjectGradleMavenPackage` is provided specifically for the `expo-camera` case, it can be used across packages for reliable maven additions:

```ts
const config = withProjectGradleMavenPackage({ /* Expo config... */ }, {
  filePath: "../node_modules/expo-camera/android/maven",
  mavenPluginId: "expo-camera",
});
```

Running this generates the following:

`android/.expo/project-build.gradle`
```groovy
// GENERATED FILE DO NOT MODIFY. REGENERATE VIA EXPO CLI (expo eject)
// This file is used for Expo config plugins to safely modify a project's `./build.gradle` file,
// if this file ever causes build errors, simply delete it and run `expo eject` again.

allprojects {
repositories {
// @begin allprojects-repositories
maven{url "$rootDir/../node_modules/expo-camera/android/maven"} // @plugin expo-camera
// @end allprojects-repositories
}
}
```

the following is added by default to all projects:
`android/build.gradle`
```groovy
// long unmodified build.gradle...

apply from: "$rootDir/.expo/project-build.gradle"
```

Can confirm this works as expected for configuring the `expo-camera` package via a plugin.


EvanBacon:
Closing for now.

Issue #2873, opened 2020-11-11
Ejected CRNA blank project fails to build with react-native-community cli


EvanBacon:
## Description

Projects bootstrapped with `npx create-react-native-app -t blank` (blank template) do not build after being ejected. This is because the blank template does not have an `main` field in the `package.json`, so eject is skipping the step to generate an `index.js`.

```js
import { registerRootComponent } from 'expo';

import App from './App';

// registerRootComponent calls AppRegistry.registerComponent('main', () => App);
// It also ensures that whether you load the app in the Expo client or in a native build,
// the environment is set up appropriately
registerRootComponent(App);
```

## Environment

```

  Expo CLI 3.28.5 environment info:
    System:
      OS: macOS 10.15.4
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 12.18.1 - ~/.nvm/versions/node/v12.18.1/bin/node
      Yarn: 1.22.4 - ~/.nvm/versions/node/v12.18.1/bin/yarn
      npm: 6.14.5 - ~/.nvm/versions/node/v12.18.1/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.9.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.1, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
      Android SDK:
        API Levels: 21, 22, 23, 24, 25, 26, 27, 28, 29, 30
        Build Tools: 26.0.3, 28.0.3, 29.0.2, 29.0.3
        System Images: android-22 | Intel x86 Atom_64, android-26 | Google APIs Intel x86 Atom_64, android-30 | Google APIs Intel x86 Atom, android-30 | Google Play Intel x86 Atom
        Android NDK: 19.2.5345600
    IDEs:
      Android Studio: 4.0 AI-193.6911.18.40.6514223
      Xcode: 12.1/12A7403 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~39.0.0 => 39.0.4 
      react: 16.13.1 => 16.13.1 
      react-dom: ~16.13.0 => 16.13.1 
      react-native: ~0.63.3 => 0.63.3 
      react-native-web: 0.13.13 => 0.13.13 
    npmGlobalPackages:
      expo-cli: 3.28.5
    Expo Workflow: bare
```

## Reproducible Demo

[...]

PR #2874, opened 2020-11-11
[expo-cli][eject] Resolve main fields to determine if an index.js should be generated

EvanBacon:
- fix https://github.com/expo/expo-cli/issues/2873
- Ensure that an `index.js` is generated in projects that don't define a main field in their `package.json` if the `./index.js` (and platform extended variations) aren't present.
- Ensure that an `index.js` is not generated if the `package.json` defines a main field that resolves to an existing file, unless that file is the expo managed workflow default (`expo/AppEntry` + variations).

### DIFF
--- a/packages/expo-cli/src/commands/eject/Eject.ts
+++ b/packages/expo-cli/src/commands/eject/Eject.ts
@@ -514,7 +514,7 @@ async function updatePackageJSONAsync({
 }
 
 export function resolveBareEntryFile(projectRoot: string, main: any) {
-  // expo app entry is disallowed for bare projects.
+  // expo app entry is not needed for bare projects.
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -512,6 +513,17 @@ async function updatePackageJSONAsync({
   };
 }
 
+export function resolveBareEntryFile(projectRoot: string, main: any) {
+  // expo app entry is disallowed for bare projects.
```
brentvatne, line 517:
```suggestion
  // expo app entry is not needed for bare projects.
```

brentvatne (approved):

PR #2883, opened 2020-11-13
[xdl]<fix>: Remove default splash screen image on Android

bbarthec:
When a user does not provide a splash image we shouldn't show any image.
Ensured the splash screen image from Expo Client is not propagated to standalones.

Resolves https://forums.expo.io/t/how-to-remove-default-expo-splash-screen/44789/2

# Test plan

I've jumped back to `sdk-39` branch, build local standalone simulator app for Android.

# TODOs

- [x] test it on staging (publish `xdl` `-rc` version on npm & update in `turtle` on staging)
- [ ] publish `xdl` with this change
- [x] ~(back-porting to SDK 39) rebuild `sdk-39` Android shell tarball using this newer xdl version~ _edit: this is not needed as `turtle` is running this piece of code_
- [ ] update `@expo/xdl` version in `turtle` and ship to production

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to Expo CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- [xdl] On Android remove default template splash image when no splash image is specified in the app manifest ([#2883](https://github.com/expo/expo-cli/pull/2883))
+
 ## [Mon, 9 Nov 2020 13:44:59 -0800](https://github.com/expo/expo-cli/commit/d5b8759b32d5a7747067a969728d9ba732926824)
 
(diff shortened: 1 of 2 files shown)


brentvatne:
this wouldn't require deploying a new shell app tarball would it? since it's just an xdl change for code that runs against the tarball at build time, i believe.

bbarthec:
@brentvatne, correct, as I've described in the `TODO`s we have to rebuild `SDK-39` shel tarball and deploy it with this change 😉

brentvatne:
@bbarthec - i think we are not on the same page 😆  i was thinking that it *would not be needed* to rebuild a new shell tarball for this change. maybe let's follow up on slack to clarify, just ping me when we're both around :)

Issue #2685, opened 2020-09-22
Make upgrade work without an app.json


EvanBacon:
Running `expo upgrade` in a project without an `app.json` throws an error:
```
Your git working tree is clean
To revert the changes after this command completes, you can run the following:
  git clean --force && git reset --hard

✔ You are currently using SDK 38.0.0. Would you like to update to the latest version, 39.0.0? … no

✔ Choose a SDK version to upgrade to: › 39.0.0

✔ Would you like to upgrade the Expo app in the iOS simulator? … no

✔ Would you like to upgrade the Expo app in the Android emulator? … no

✔ Installed expo@^39.0.0
✖ Unable to validate configuration.
⠋ Updating your app.json to account for breaking changes (if applicable)...Project at path /Users/evanbacon/Documents/GitHub/examples/with-facebook-auth does not contain a valid Expo config
ConfigError: Project at path /Users/evanbacon/Documents/GitHub/examples/with-facebook-auth does not contain a valid Expo config
    at readConfigJson (/Users/evanbacon/.nvm/versions/node/v12.18.1/lib/node_modules/expo-cli/node_modules/@expo/config/src/Config.ts:175:13)
    at Object.readConfigJsonAsync (/Users/evanbacon/.nvm/versions/node/v12.18.1/lib/node_modules/expo-cli/node_modules/@expo/config/src/Config.ts:211:10)
    at makeBreakingChangesToConfigAsync (/Users/evanbacon/.nvm/versions/node/v12.18.1/lib/node_modules/expo-cli/src/commands/upgrade.ts:167:44)
    at upgradeAsync (/Users/evanbacon/.nvm/versions/node/v12.18.1/lib/node_modules/expo-cli/src/commands/upgrade.ts:503:9)
with-facebook-auth % 
```

brooksbecton:
https://github.com/expo/expo-cli/compare/master...brooksbecton:Make-upgrade-work-without-app.json

Gave this a shot, but not sure if I'm on the right track.

PR #2888, opened 2020-11-16
[cli][upgrade] support projects without a config

EvanBacon:
- fix https://github.com/expo/expo-cli/issues/2685 -- this will make upgrading expo/examples much easier.
- bit of an aha moment, projects without configs don't need any kind of upgrading, so certain steps can just be skipped.
- tested by running `expo upgrade` in `examples/with-facebook-auth`


brentvatne:
nice

Issue #2801, opened 2020-10-16
configure-splash-screen is not working on iOS (bare project).


hehex9:
## Description

**@expo/configure-splash-screen** (using `npx expo-splash-screen`) does not work on **iOS**.

```
npx expo-splash-screen -b white -i ./android/app/src/main/res/mipmap-hdpi/ic_launcher.png -p ios
```

due to the unuseful error message (`Cannot read property 'undefined' of undefined`), I tried

```diff
// node_modules/@expo/configure-splash-screen/build/index-cli.js

async function runAsync() {
    try {
        await cli_command_1.default().parseAsync(process.argv);
    }
    catch (e) {
-        console.error(`\n${e.message}\n`);
+        console.error(e);
        process.exit(1);
    }
}
```

then rerun the command

```
TypeError: Cannot read property 'undefined' of undefined
    at pbxProject.getPBXVariantGroupByKey (/demo/node_modules/@expo/configure-splash-screen/node_modules/xcode/lib/pbxProject.js:1932:56)
    at pbxProject.addFile (/demo/node_modules/@expo/configure-splash-screen/node_modules/xcode/lib/pbxProject.js:2047:19)
    at Object.addStoryboardFileToProject (/demo/node_modules/@expo/configure-splash-screen/build/xcode/index.js:9:29)
    at updatePbxProject (/demo/node_modules/@expo/configure-splash-screen/build/ios/Storyboard.js:24:17)
    at Object.configureStoryboard [as default] (/demo/node_modules/@expo/configure-splash-screen/build/ios/Storyboard.js:148:11)
    at async Promise.all (index 3)
    at async Object.configureIos [as default] (/demo/node_modules/@expo/configure-splash-screen/build/ios/index.js:16:5)
    at async action (/demo/node_modules/@expo/configure-splash-screen/build/cli-command.js:19:13)
    at async Command.<anonymous> (/demo/node_modules/@expo/configure-splash-screen/build/cli-command.js:102:5)
    at async Promise.all (index 0)
```

[calling pbxProject.addFile with `undefined` group key](https://github.com/expo/expo-cli/blob/master/unlinked-packages/configure-splash-screen/src/xcode/index.ts#L13) might be the cause of this problem, since we don't have any `PBXVariantGroup` in [bare project's project.pbxproj](https://github.com/facebook/react-native/blob/0.63-stable/template/ios/HelloWorld.xcodeproj/project.pbxproj).

## Expected Behavior

no error

## Observed Behavior

output

[...]

PR #2896, opened 2020-11-17
[configure-splash-screen] guard .pbxproj for bare workflow

hehex9:
a workaround for #2801


bbarthec:
Closing in favour of #2945

PR #2901, opened 2020-11-17
Replace process.exits with errors

EvanBacon:
- opt towards throwing more than process.exit to make code easier to test.
- created SilentError which logs nothing when thrown in expo-cli but attempts to throw relevant error info when caught outside of the typical expo-cli flow.
- use `CommandError` more in favor of `log.error()` or `log(chalk.red())`


packages/expo-cli/src/commands/push-creds.ts
```diff
@@ -33,10 +34,9 @@ export default function (program: Command) {
 
       const fcmCredentials = await ctx.android.fetchFcmKey(experienceName);
       if (fcmCredentials?.fcmApiKey) {
-        log(`FCM Api Key: ${fcmCredentials?.fcmApiKey}`);
+        log(`FCM API key: ${fcmCredentials?.fcmApiKey}`);
```
brentvatne, line 37: nice lil clean up

brentvatne (approved):
nice cleanup work :+1:

PR #2902, opened 2020-11-17
Remove es6-error in favor of Error

EvanBacon:
- Doesn't appear as though `es6-error` is adding any value. We have three copies of it in the repo.

### DIFF
--- a/packages/expo-cli/src/commands/build/BuildError.ts
+++ b/packages/expo-cli/src/commands/build/BuildError.ts
@@ -1,6 +1,4 @@
-import ExtendableError from 'es6-error';
-
-export default class BuildError extends ExtendableError {
+export default class BuildError extends Error {
   name = 'BuildError';
 
   constructor(public message: string) {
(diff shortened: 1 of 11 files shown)


packages/expo-cli/src/commands/build/BuildError.ts
```diff
@@ -1,9 +1,8 @@
...
-
-export default class BuildError extends ExtendableError {
+export default class BuildError extends Error {
   name = 'BuildError';
 
   constructor(public message: string) {
     super();
   }
 }
+BuildError.prototype.name = BuildError.name;
```
fson, line 8: You could skip this because `name` is set using a property initializer above (see line 2). Maybe property initializer would be a good way to do this across the board – makes it super clear what the value will be.

In built-in `Error` the name property seems to be non-enumerable, so to match that we’d have the use `defineProperty` instead, though. But I’m not too concerned about that, a regular property should be fine.

PR #2910, opened 2020-11-18
fix: Propogate private config from app.json to manifest in build process with the public URL flag

ShaneMcNamara:
As described here: https://github.com/expo/expo/issues/10252
Private config values in app.json, such as Google Maps API Key, are not propagated properly through the build process when the --public-url flag is set.

@brentvatne

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -117,6 +117,15 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}): 
       staticConfigPath: paths.staticConfigPath,
     };
 
+    if (options.isPrivateConfig) {
+      return {
+        exp: {
+          hooks: configWithDefaultValues.exp.hooks,
+          ios: configWithDefaultValues.exp.ios?.config,
+          android: configWithDefaultValues.exp.android?.config,
+        },
+      };
+    }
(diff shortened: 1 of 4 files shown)


ShaneMcNamara:
Oops, looks like I have some work to do.

quinlanj:
wait why did u close this @ShaneMcNamara

ShaneMcNamara:
> wait why did u close this @ShaneMcNamara

It is failing the checks because I was returning incorrect structure:

```
Error: @expo/config: src/Config.ts(167,5): error TS2322: Type '{ mods: ModConfig; dynamicConfigObjectType: string | null; rootConfig: AppJSONConfig; dynamicConfigPath: string | null; staticConfigPath: string | null; exp: ExpoConfig; pkg: Record<...>; } | { ...; }' is not assignable to type 'ProjectConfig'.
```

brentvatne:
you can keep pr open and push another commit when you resolve that :)

ShaneMcNamara:
@quinlanj  - I think I'll create a new method, because I don't want to return the full `ExpoConfig`. Thoughts?

brentvatne:
@ShaneMcNamara - i think that is fine

Issue #2909, opened 2020-11-17
Error while creating Expo build through managed work flow


lraghavantrinet:
## Description

I use Expo's managed workflow to create builds. Am currently on Expo 38. Expo cli version: 3.28.5.
When I create the build, I get this error:
Build queued...Cannot read property 'status' of undefined

## Expected Behavior

No Error should be thrown and the download link for the ipa/apk should be shown.

## Observed Behavior

An error is thrown and download link is not displayed.
```
Build queued...Cannot read property 'status' of undefined
TypeError: Cannot read property 'status' of undefined
    at IOSBuilder.wait (/usr/local/lib/node_modules/expo-cli/src/commands/build/BaseBuilder.ts:304:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at IOSBuilder.build (/usr/local/lib/node_modules/expo-cli/src/commands/build/BaseBuilder.ts:381:28)
    at IOSBuilder.run (/usr/local/lib/node_modules/expo-cli/src/commands/build/ios/IOSBuilder.ts:68:5)
    at IOSBuilder.command (/usr/local/lib/node_modules/expo-cli/src/commands/build/BaseBuilder.ts:39:7)
    at Command.<anonymous> (/usr/local/lib/node_modules/expo-cli/src/exp.ts:332:7)
```

## Environment

 Expo CLI 3.28.5 environment info:
    System:
      OS: macOS 10.15.7
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 12.13.0 - /usr/local/bin/node
      Yarn: 1.22.4 - /usr/local/bin/yarn
      npm: 6.14.5 - /usr/local/bin/npm
    Managers:
      CocoaPods: 1.9.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 13.7, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
      Android SDK:
        API Levels: 29
        Build Tools: 29.0.2
        System Images: android-25 | Google Play Intel x86 Atom, android-26 | Google APIs Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.5 AI-191.8026.42.35.5977832
      Xcode: 11.7/11E801a - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #2912, opened 2020-11-18
Handle undefined build job more gracefully

EvanBacon:
- related https://github.com/expo/expo-cli/issues/2909


dsokal (changes requested):
~With your changes, Expo CLI will throw an error if the build doesn't finish after 3 minutes (5 * 30 seconds). This is not expected behavior. Maybe just surround the switch with `if (job)`. This should do the trick.~

dsokal (approved):
Never mind. It'll only fail when the build job record is still not created after 3 minutes (which is a weird case).

wkozyra95 (approved):

Issue #2012, opened 2020-04-27
Web favicon images are blurry


EvanBacon:
maybe the order of the .ico is being generated wrong, making the 16x16 the main image

EvanBacon:
Fixed in 0.12.2

oguzgelal:
Hey @EvanBacon, favicons still seems to be blurry. Here's the icon I'm providing in app.json, and how it's turns out after deploy:

![wk4](https://user-images.githubusercontent.com/2817993/81487677-32843280-9268-11ea-8ed4-406e3bfd4c52.png)

<img width="68" alt="Screen Shot 2020-05-10 at 2 46 12 AM" src="https://user-images.githubusercontent.com/2817993/81487699-7840fb00-9268-11ea-9a16-7d98be46615e.png">

andordavoti:
I have the same issue with all my projects

barnu5:
Still having this issue as well

PR #2914, opened 2020-11-18
[image-utils] Fix blurry web favicon

MrHertal:
Web favicon is blurry, it has been reported in https://github.com/expo/expo-cli/issues/2012, which is closed but problem is still here.

I checked code and notice an error when icons are resized.
Jimp [resize method](https://github.com/oliver-moran/jimp/tree/master/packages/plugin-resize#usage) is mutable, so images were all 16x16.

### DIFF
--- a/packages/image-utils/src/jimp.ts
+++ b/packages/image-utils/src/jimp.ts
@@ -19,6 +19,7 @@ export async function resizeBufferAsync(buffer: Buffer, sizes: number[]): Promis
   return Promise.all(
     sizes.map(async size => {
       // Parse the buffer each time to prevent mutable copies.
+      // Parse the buffer each time to prevent mutable copies.
       const jimpImage = await Jimp.read(buffer);
       const mime = jimpImage.getMIME();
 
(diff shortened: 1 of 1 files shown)


packages/image-utils/src/jimp.ts
```diff
@@ -16,9 +16,14 @@ type JimpGlobalOptions = Omit<SharpGlobalOptions, 'input'> & {
 };
 
 export async function resizeBufferAsync(buffer: Buffer, sizes: number[]): Promise<Buffer[]> {
-  const jimpImage = await Jimp.read(buffer);
-  const mime = jimpImage.getMIME();
-  return Promise.all(sizes.map(size => jimpImage.resize(size, size).getBufferAsync(mime)));
+  return Promise.all(
+    sizes.map(async size => {
+      const jimpImage = await Jimp.read(buffer);
```
EvanBacon, line 21:
```suggestion
      // Parse the buffer each time to prevent mutable copies.
      const jimpImage = await Jimp.read(buffer);
```

EvanBacon (approved):
Looks fine, thanks!

PR #2918, opened 2020-11-18
[package-manager] Fix repo update

kevinnguy:
When running `npx pod-install` throws an error, `stderr` return as an empty string, which does not run `pod repo update`.   We don't know if `stderr` throws because of `pod repo update` so we'll assume that we need to repo update by running `pod install --repo-update`. If that fails, then we can throw the error.

### DIFF
--- a/packages/package-manager/src/CocoaPodsPackageManager.ts
+++ b/packages/package-manager/src/CocoaPodsPackageManager.ts
@@ -160,27 +160,13 @@ export class CocoaPodsPackageManager implements PackageManager {
     });
   }
 
-  private async _installAsync(shouldUpdate: boolean = true): Promise<void> {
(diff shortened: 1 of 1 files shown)


EvanBacon:
Seems like if `pod install` throws any kind of error (like syntax error in the podfile) it'll respond by running `pod repo update` now 🤔 that doesn't seem optimal.

kevinnguy:
> if pod install throws any kind of error (like syntax error in the podfile) it'll respond by running pod repo update

That is true but the current implementation doesn't work even if there is an error because `stderr` is returning an empty string. I can look into why stderr is returning an empty string in the first place.

EvanBacon:
closing, feel free to reopen with more changes

PR #2938, opened 2020-11-24
Update --dev-client for new requirements

brentvatne:
- ~Should work without having the `expo` package installed.~ We will require that developers either set an SDK version or install the `expo` package (as a regular dependency or dev dependency) to use expo-dev-launcher.
- Should deep link into the app using the url format: `myscheme://expo-development-client/?url=url-encoded-path-to-app-manifest` ([described on Notion](https://www.notion.so/expo/Deep-Linking-URL-Format-for-Development-Clients-b721472abba0426ea1e747c3718aaf4a)).

### TODO

- [x] Write tests
- [x] Follow up on other TODOs included in the diff

### DIFF
--- a/packages/dev-tools/__mocks__/@expo/xdl.js
+++ b/packages/dev-tools/__mocks__/@expo/xdl.js
@@ -1,6 +1,9 @@
 const xdl = jest.genMockFromModule('@expo/xdl');
 
 xdl.UrlUtils = {
+  constructDeepLinkAsync(projectDir) {
+    return this.constructManifestUrlAsync(projectDir);
+  },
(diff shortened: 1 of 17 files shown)


tcdavis:
Basic behaviour all looks great to me, thanks!
- Would be nice to add a line saying "You will need to rebuild your development client if you choose to add a common scheme." if --dev-client cannot find a common scheme and suggests the user adds it with npx uri-sheme
- Users need the address of the bundler to use the dev client on a simulator.  The URL the QR code represents is not very useful,  other than knowing what the QR code encodes.  I'd propose just replacing the displayed URL with the bundler URL.

PR #2958, opened 2020-11-26
Enable notifications by default

EvanBacon:
- ejected projects have notifications by default so our template should come with notifications enabled by default too. Without this iTunes warns about a binary issue: `ITMS-90078: Missing Push Notification Entitlement`
- wasn't sure if we should do development or production by default but it seems we use development in the Expo client, and Xcode defaults to development.


brentvatne (approved):
xcode will automatically set it to production when you do a release archive, so this should be sufficient

PR #2962, opened 2020-11-27
[config-types] feat: add android.applicationId

bbarthec:
Decouple `android.applicationId` from `android.package` as these properties could be provided separately.

Followup of https://github.com/expo/universe/pull/6487


bbarthec:
@EvanBacon, as we're migrating to EAS there's no point in adding `android.applicationId` to the old build system and related items  👀

PR #2965, opened 2020-11-27
Created withRunOnce

EvanBacon:
- When we start migrating users from unversioned config plugins to versioned config plugins, we'll need a way to prevent running plugins more than once.
- Initially a user will eject, see a deprecation warning about an unversioned feature, maybe resolve the issue by adding a static plugin for a package like `expo-facebook`,  then run `expo eject` again. 
  - Without a way to track that `expo-facebook` has already been run, expo-cli would just run the unversioned legacy plugin again.
  - Now our versioned plugins can import the `History` interface, or the `withRunOnce` plugin to prevent unversioned plugins from running.
    - Note: User plugins run before Expo built-in plugins.
- Uses the new `_internal.pluginHistory` to track plugin history.

# Test Plan

- Wrote some unit tests


packages/config-plugins/src/utils/history.ts
```diff
@@ -0,0 +1,36 @@
...
+  name: string;
+  version: string;
+  platform?: ModPlatform;
+};
+
+export function getHistoryItem(
+  config: Pick<ExpoConfig, 'extra'>,
+  name: string
+): PluginHistoryItem | null {
+  return config.extra?._pluginHistory?.[name] ?? null;
```
brentvatne, line 15: `extra` is meant for users to add additional fields to make available to their app at runtime. we should move this to a new field like `_internal`

brentvatne: or maybe `.expo` if this is intended to be persistent across runs

PR #2967, opened 2020-11-27
[cli] Make it possible to run expo upgrade with beta release prior to setting flag

brentvatne:
This is useful so we can easily test standalone app builds on existing apps without needing to manually update deps or make the beta release publicly available through documented mechanisms.

Screen Shot 2020-11-27 at 2 14 18 PM

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -12,6 +12,7 @@ import ora from 'ora';
 import semver from 'semver';
 import terminalLink from 'terminal-link';
 
+import CommandError from '../CommandError';
 import log from '../log';
 import { confirmAsync, selectAsync } from '../prompts';
 import { findProjectRootAsync } from './utils/ProjectUtils';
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -465,21 +466,32 @@ export async function upgradeAsync(
...
+        (val, acc) => Math.min(val, acc),
+        sdkVersionNumbers[0]
+      );
+      const maxSdkVersion = sdkVersionNumbers.reduce(
+        (val, acc) => Math.max(val, acc),
+        sdkVersionNumbers[0]
+      );
+      throw new CommandError(
+        `You provided the target SDK version value of ${targetSdkVersionString}, which does not seem to exist.\n` +
+          `Valid SDK versions are in the range of ${minSdkVersion}.0.0 to ${maxSdkVersion}.0.0.`
```
brentvatne, line 493: printing a range here because providing a list of SDK versions makes the error hard to read and will be even harder to read over time. might want to generalize to other places where we do the same thing

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -704,9 +719,15 @@ ${chalk.gray(`${upgradeHelperUrl}?from=${previousReactNativeVersion}&to=${newRea
     );
     log(chalk.bold(targetSdkVersion.releaseNoteUrl));
   } else {
-    log.gray(
-      `Unable to find release notes for ${targetSdkVersionString}, please try to find them on https://blog.expo.io to learn more about other potentially important upgrade steps and breaking changes.`
-    );
+    if (getenv.boolish('EXPO_BETA', false)) {
+      log.gray(
+        `Release notes are not available for beta releases. Please refer to the CHANGELOG: https://github.com/expo/expo/blob/master/CHANGELOG.md.`
```
brentvatne, line 724: provide changelog rather than release notes when upgrading to beta release

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -465,21 +466,32 @@ export async function upgradeAsync(
...
+    // return the beta version, even with the EXPO_BETA flag set.
+    if (getenv.boolish('EXPO_BETA', false)) {
+      const allSdkVersions = await Versions.sdkVersionsAsync();
+      targetSdkVersion = allSdkVersions[targetSdkVersionString];
+    }
 
-      if (!answer) {
-        return;
-      }
+    // If we still don't have a version, even after searching through unreleased versions
```
brentvatne, line 478: this changes existing behavior on `expo upgrade` for unknown sdk versions. these would not actually work anyways, so just actually bail out instead

PR #3014, opened 2020-12-10
Remove last call to intercomHash endpoint

ide:
Why: This endpoint is otherwise unused; remove the last caller so we can remove it from www.

How: I replaced the endpoint with the `auth/userInfo` endpoint, which is still used and suffices for this test.

Test Plan: Ran `yarn integration-tests` in XDL after removing `.skip` from `UserSessions-test.ts`.


fson (approved):
👍

PR #3020, opened 2020-12-11
[expo-cli] Remove fs-extra from @expo/config and @expo/json-file

fson:
`fs-extra` is used in quite a lot of packages and therefore it's easy to end up many versions of it. We have previously tried to avoid this by making all our packages use the same version (#2190), but with packages spread across many repos and new packages added all the time, this is a losing battle. For example, `eas-cli` pulls in 7 copies of `fs-extra`:
```
% du -h $(dirname $(realpath $(which eas)))/../node_modules | egrep fs-extra$
228K	{...}/eas-cli/bin/../node_modules/@react-native-community/cli-platform-android/node_modules/fs-extra
220K	{...}/eas-cli/bin/../node_modules/fs-extra
228K	{...}/eas-cli/bin/../node_modules/@oclif/errors/node_modules/fs-extra
220K	{...}/eas-cli/bin/../node_modules/@expo/json-file/node_modules/fs-extra
220K	{...}/eas-cli/bin/../node_modules/@expo/config/node_modules/fs-extra
220K	{...}/eas-cli/bin/../node_modules/@expo/image-utils/node_modules/fs-extra
220K	{...}/eas-cli/bin/../node_modules/@expo/config-plugins/node_modules/fs-extra
```
**This PR removes support for Node.js v10 LTS** and makes the `@expo/config` and `@expo/json-file` packages use the built-in `fs.promises` APIs instead of the `fs-extra` package. (Node.js v10 is not supported because using these APIs in v10 results in a warning `ExperimentalWarning: The fs.promises API is experimental` printed in the console.)


CHANGELOG.md
```diff
@@ -6,10 +6,10 @@ This is the log of notable changes to Expo CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- [expo-cli] Remove support for Node.js v10
```
EvanBacon, line 9:
```suggestion
- [expo-cli] Drop support for Node.js v10
```

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ This is the log of notable changes to Expo CLI and related packages.
 
 - [cli][xdl] Clear versions cache when running expo upgrade, to be safe
 - [config] fix mod serialization ([#3008](https://github.com/expo/expo-cli/issues/3008))
-- [config-plugins] use env variable for debug when _internal isn't defined ([#3011](https://github.com/expo/expo-cli/issues/3011))
+- [config-plugins] use env variable for debug when \_internal isn't defined ([#3011](https://github.com/expo/expo-cli/issues/3011))
```
EvanBacon, line 21:
```suggestion
- [config-plugins] use env variable for debug when `_internal` isn't defined ([#3011](https://github.com/expo/expo-cli/issues/3011))
```

CHANGELOG.md
```diff
@@ -150,7 +150,7 @@ This is the log of notable changes to Expo CLI and related packages.
 
 ### 🎉 New features
 
-- [config] fill _internal object ([#2968](https://github.com/expo/expo-cli/issues/2968))
+- [config] fill \_internal object ([#2968](https://github.com/expo/expo-cli/issues/2968))
```
EvanBacon, line 153:
```suggestion
- [config] fill `_internal` object ([#2968](https://github.com/expo/expo-cli/issues/2968))
```

packages/json-file/src/JsonFile.ts
```diff
@@ -212,7 +212,7 @@ async function writeAsync<TJSONObject extends JSONObject>(
   options?: Options<TJSONObject>
 ): Promise<TJSONObject> {
   if (options?.ensureDir) {
-    await mkdirp(path.dirname(file));
+    await fs.promises.mkdir(path.dirname(file), { recursive: true });
```
EvanBacon, line 215: `ensureDir` ?

EvanBacon (approved):
Looks fine

EvanBacon (changes requested):
actually looks like the tests are failing cuz this change

packages/json-file/src/JsonFile.ts
fson: Nah, that's an alias that only exists in `fs-extra`. I want [`fs.promises.mkdir`](https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options), so `fs-extra` is no longer needed.

fson:
Yeah, I forgot to change one `await fsExtra.readFile(path)` call, which is why the test was failing. `fs.readFile` takes a callback, `fs.promises.readFile` returns a promise.

EvanBacon:
I guess we should probably wait until the end of node 10 maintenance on May 1st to prevent getting a bunch of issues regarding our partial LTS support.

brentvatne:
Agree with @EvanBacon that we should hold off on dropping Node 10 support until it's no longer in LTS. Right now our docs indicate that we support LTS versions of Node, and I think our instructions would be complicated quite a bit by needing to clarify further which specific LTS versions.

fson:
Here's the [Node.js Release Schedule](https://github.com/nodejs/Release/blob/bd541f09877bb4ff4350c4ab05691a904d67611e/README.md#release-schedule):

| Release  | Status              | Codename    |Initial Release | Active LTS Start | Maintenance Start | End-of-life               |
| :--:     | :---:               | :---:       | :---:          | :---:            | :---:             | :---:                     |
| 10.x | **Maintenance**     | Dubnium | 2018-04-24     | 2018-10-30       | 2020-05-19        | 2021-04-30                |
| 12.x | **Maintenance**     | Erbium  | 2019-04-23     | 2019-10-21       | 2020-11-30        | 2022-04-30                |
| 14.x | **Active LTS**      | Fermium | 2020-04-21     | 2020-10-27       | 2021-10-19        | April 2023                |
| 15.x     | **Current**         |             | 2020-10-20     | -                | 2021-04-01        | 2021-06-01                |
| 16.x     | **Pending**         |             | 2021-04-20     | 2021-10-26       | 2022-10-18        | 2024-04-30                |

I agree supporting Node 10 until 30th of April 2021 is the correct approach. And we should then decide if we want to continue with the current schedule after that or if we want to perhaps change it to only include *current* and *active LTS*.

fson:
Rebased and updated now.

Issue #2192, opened 2020-05-27
expo upload:ios iTMSTransporter error


jonsamp:
## Description

I am attempting to use `expo upload:ios` to upload my latest build to testflight.

## Expected Behavior

I expect this to download my latest build on Expo's servers, then upload it to testflight.

## Observed Behavior

The build downloads fine, but then there's an error with `iTMSTransporter`. Here's the full output after running `expo upload:ios`: 

(I have REDACTED some parts for privacy)

```
Downloading build from https://exp-shell-app-assets.s3.us-west-1.amazonaws.com/ios/REDACTED...baa281427504-archive.ipa
Downloading [================================================================================================================================================] 100% 0.0s
Resolving the ITC team ID...
ITC team ID is 2131102
Ensuring the app exists on App Store Connect, this may take a while...

+----------------+---------------------+
|     Summary for produce 2.141.0      |
+----------------+---------------------+
| app_identifier | com.REDACTED |
| app_name       | Gray One            |
| username       | REDACTED   |
| language       | English             |
| sku            | REDACTED          |
| platform       | ios                 |
| skip_itc       | false               |
| skip_devcenter | false               |
+----------------+---------------------+

[22:46:40]: [DevCenter] App 'com.REDACTED' already exists, nothing to do on the Dev Center
[22:46:42]: App 'com.REDACTED' already exists (REDACTED), nothing to do on App Store Connect
Uploading the app to Testflight, hold tight...
[22:46:44]: Login to App Store Connect (REDACTED)
[22:46:54]: Login successful
[22:46:55]: Ready to upload new build to TestFlight (App: REDACTED)...
[22:46:57]: [iTMSTransporter] -Djava.ext.dirs=/usr/local/itms/java/lib/ext is not supported.  Use -classpath instead.

[22:46:57]: [iTMSTransporter] Error: Could not create the Java Virtual Machine.

[22:46:57]: [iTMSTransporter] Error: A fatal exception has occurred. Program will exit.

[22:46:57]: iTunes Transporter output above ^
[22:46:57]: The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.
[22:46:57]: Going to upload updated app to App Store Connect
[22:46:57]: This might take a few minutes. Please don't interrupt the script.
[22:46:57]: [iTMSTransporter] -Djava.ext.dirs=/usr/local/itms/java/lib/ext is not supported.  Use -classpath instead.

[...]

PR #3030, opened 2020-12-15
Mark expo upload:ios as unsupported

EvanBacon:
# Why

`upload:ios` is a light wrapper around fastlane deliver and transporter app, over time it's become buggy and now it mostly just fails and tells people to use transporter app instead. `upload:ios` also only works on MacOS computers.
 
With the creation of `eas submit` which works on all computers, users now have a couple really good and stable options. We should deprecate `expo upload:ios` and warn users to switch over to EAS. This pattern is similar to other popular React tools like Next.js which uses "next-cli" to start the app, and "now" to build and distribute the app. Removing `upload:ios` gets us closer to pulling Fastlane out of `expo-cli`, which will result in a 30% bundle size decrease, and fix installation issues for many windows and linux users.

# How

Add a notice when running `expo upload:ios`

Screen Shot 2020-12-15 at 2 49 31 PM

Also add a notice to the `--help` prompt

Screen Shot 2020-12-15 at 3 05 13 PM


fson:
"Deprecated" is different from "removed".
> Deprecated refers to something that should not be used anymore and may be removed in the future.
> But deprecated things are still around and still exist.

So this PR doesn't really "deprecate" the command, but removes it and the wording should be changed. However, to be clear, I fully support this change – `eas submit` is a big improvement over `expo upload:ios` and will provide a much better experience.

brentvatne:
Linking to this comment for more context for people who stumble on this PR: https://github.com/expo/expo-cli/issues/2192#issuecomment-745626708

brentvatne:
Looks good to me, agree with @fson that we should change deprecated to "Removed" or something to that effect

packages/expo-cli/src/commands/upload.ts
```diff
@@ -59,7 +54,9 @@ export default function (program: Command) {
   program
     .command('upload:ios [path]')
     .alias('ui')
-    .description('macOS only: Upload an iOS binary to Apple. An alternative to Transporter.app')
+    .description(
+      `${chalk.yellow('Deprecated:')} Use ${chalk.bold('eas submit')} or Transporter app instead.`
```
brentvatne, line 58:
```suggestion
      `${chalk.yellow('Unsupported:')} Use ${chalk.bold('eas submit')} or Transporter app instead.`
```

packages/expo-cli/src/commands/upload.ts
```diff
@@ -101,38 +98,22 @@ export default function (program: Command) {
...
-            TerminalLink.transporterAppLink()
-          )} instead of the ${chalk.bold(
-            'expo upload:ios'
-          )} command if you have any trouble with it.`
-        );
-
-        throw err;
-      }
+      log.newLine();
+      log(chalk.yellow('expo upload:ios is deprecated'));
```
brentvatne, line 108:
```suggestion
      log(chalk.yellow('expo upload:ios is no longer supported'));
```

brentvatne (approved):

PR #3036, opened 2020-12-18
Do not override pageExtensions but extend it

jfrolich:
This causes a problem when other page extensions need to be configured


EvanBacon:
Closing in favor of https://github.com/expo/expo-cli/commit/69965f25e402a3e7d8e019761a103ac79db6640c

PR #3045, opened 2020-12-24
[eject] auto apply optional managed plugins

EvanBacon:
When ejecting a project with expo SDK packages, we'll automatically apply a hardcoded list of package's plugins if they exist. This will create configuration parity with managed apps.

When a plugin can be auto-configured, it'll be removed from the list of plugins with additional setup.


brentvatne:
this seems very reasonable. we should get these plugins published as part of the expo packages soon so we can try this out. we should probably also point to those versions of the packages with plugins in `bundledNativeModules.json` so we can instruct people to run `expo upgrade` again in their project before building on eas build to ensure they're using the correct versions

PR #3056, opened 2021-01-06
Move withPlugins into config-plugins

EvanBacon:
- In the past we split config-plugins out of config to reduce package size of `expo/config`
- Now that static plugins exist we need config plugin logic in the `expo/config` package
- To support unversioned plugins (#2977) we need the plugin resolver logic to be accessible in `expo/config-plugins`
- This PR refactors the plugin module resolution to `expo/config-plugins` and makes `expo/config` depend on `expo/config-plugins` again 🙃


packages/config-plugins/src/utils/plugin-resolver.ts
```diff
@@ -2,14 +2,12 @@ import findUp from 'find-up';
...
-import { fileExists } from '../Modules';
-import { resolveConfigPluginExport } from '../evalConfig';
-
-export type StaticPlugin<T = any> = [string | ConfigPlugin<T>, T];
+import { ConfigPlugin, StaticPlugin } from '../Plugin.types';
+import { assert, PluginError } from './errors';
+import { fileExists } from './modules';
 
 // Default plugin entry file name.
+// :DO_NOT_CHANGE_OR_YOU_WILL_BE_FIRED:
```
brentvatne, line 10: 👀

brentvatne (approved):

PR #3057, opened 2021-01-07
Create prebuild command

EvanBacon:
# Why

To support `eas build` in managed workflow projects as a drop-in replacement for `expo build`, we need a command to generate all the native code required for building the iOS and Android app, along with modifying the native code according to the needs of the individual packages installed in the user's project. 
`expo eject` provides this functionality but it specifically works to aid bare workflow users who need to make modifications to their native code manually. The concept of "re-running eject" is also confusing. 

The `expo prebuild` command runs a subset of features from the `eject` command and requires less user intervention.
prebuild assumes the user hasn't modified their native `/ios` or `/android` code manually, this enables the command to safely be rerun without undoing a user's work.

# How

- Removed git warning for prebuild.
- Split the "Next Steps" logs out of the eject command and only run them after ejecting (removed for prebuild).
- prebuild always deletes the native ios and android folder (currently no option to disable this since it's internal only).
- ~~prebuild will also have a system for regenerating the managed files Info.plist, build.gradle, settings.gradle instead of mutating the existing copies~~.


packages/expo-cli/src/commands/eject/platformOptions.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function platformsFromPlatform(platform?: string): ModPlatform[] {
+  if (!platform) {
+    return getDefaultPlatforms();
+  }
+  switch (platform) {
+    case 'ios':
+      if (process.platform === 'win32') {
+        log.warn('Ejecting on windows is unsupported locally on windows, use eas build instead');
```
brentvatne, line 21:
```suggestion
        log.warn('Ejecting is unsupported locally on windows, use eas build instead');
```

packages/expo-cli/src/commands/prebuild.ts
```diff
@@ -0,0 +1,47 @@
...
+    ...options,
+    platforms: platformsFromPlatform(platform),
+  } as Eject.EjectAsyncOptions);
+}
+
+export default function (program: Command) {
+  program
+    .command('prebuild [path]')
+    .description(
+      `Create native iOS and Android project files before building natively. ${chalk.dim(
```
brentvatne, line 35: let's mark this as experimental so it's clear that people shouldn't depend on it yet

brentvatne (approved):

packages/expo-cli/src/commands/prebuild.ts
EvanBacon, line 35:
```suggestion
      `Experimental: Create native iOS and Android project files before building natively. ${chalk.dim(
```

brentvatne (approved):

PR #3061, opened 2021-01-07
[expo-cli] Validate project owner when publishing

byCedric:
Right now, people who are using robots without the `owner` field see the "Robots are not supported for this API call" error. Which isn't descriptive about the missing `owner` field. This adds validation with descriptive error messages for all the 3 `expo publish:*` commands we have.

I also added the `Owner: ...` field when publishing as robot, but maybe we should include it whenever the owner isn't the same as the user's username?

publish with robot | publish with user
--- | ---
screenshot_2021-01-07_at_16 20 23 |  screenshot_2021-01-07_at_16 20 55

### DIFF
--- a/packages/expo-cli/src/commands/publish.ts
+++ b/packages/expo-cli/src/commands/publish.ts
@@ -309,7 +309,7 @@ export function logBareWorkflowWarnings(pkg: PackageJSONConfig) {
   );
 }
 
-export default function(program: Command) {
+export default function (program: Command) {
   program
     .command('publish [path]')
     .alias('p')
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/publish.ts
```diff
@@ -46,6 +57,9 @@ export async function action(
   }
   log(`- Release channel: ${log.chalk.bold(options.releaseChannel)}`);
   log(`- Workflow: ${log.chalk.bold(target.replace(/\b\w/g, l => l.toUpperCase()))}`);
+  if (user.kind === 'robot') {
```
byCedric, line 60: As mentioned, we could also either show this "all the time" or `if (user.username !== owner)`. That means it's always visible for robot users, but also visible for normal users whenever `owner` is defined in the manifest.

byCedric: alternatively, we could also render something like `App: ...` or `Project: ...` with `@<owner>/<name>`

fson: I like the second suggestion, with `Project: @<owner>/<name>`. This should perhaps be the first bullet point?

fson (approved):

PR #3063, opened 2021-01-08
Don't expose Expo environment variables publicly

EvanBacon:
We currently expose the environment variables matching `EXPO_` and `REACT_NATIVE_` in development. This is a security risk and we have other more preferred methods for passing data to the client via the manifest `extras` field. 

- On web, this feature didn't quite work as expected, it only added to `process.env` and not `Constants.manifest.env`. It's now completely disabled as web presents a larger risk.
- On native, this feature will be disabled starting in SDK 41

# Test Plan

- Lower the minimum SDK requirement to a released version like `40.0.0`
- Add an environment variable starting with `EXPO_` like `expo EXPO_SOMN=true`
- Run a project with `console.log(process.env, Constants.manifest.env)` in it
- The logs should no longer show these env variables exposed

### DIFF
--- a/packages/webpack-config/src/plugins/ExpoDefinePlugin.ts
+++ b/packages/webpack-config/src/plugins/ExpoDefinePlugin.ts
@@ -1,5 +1,6 @@
 import { ExpoConfig } from '@expo/config';
 import { boolish } from 'getenv';
+import semver from 'semver';
 import { DefinePlugin as OriginalDefinePlugin } from 'webpack';
 
 import { getConfig, getMode, getPublicPaths } from '../env';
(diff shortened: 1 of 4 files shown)


packages/webpack-config/src/plugins/ExpoDefinePlugin.ts
```diff
@@ -58,7 +58,7 @@ export function createClientEnvironment(
   const environment = getMode({ mode });
   const __DEV__ = environment !== 'production';
 
-  const ENV_VAR_REGEX = /^(EXPO_|REACT_NATIVE_|CI$)/i;
+  const ENV_VAR_REGEX = /^(CI$)/i;
```
brentvatne, line 61: how do we ensure that this webpack-config change will only be rolled out with sdk 41?

PR #3074, opened 2021-01-11
[expo-cli] add --experimental-bundle flag to 'export' command

jkhales:
For EAS, we are separating the creation of the bundles from their publication.

`expo export` does a lot of other stuff with the assumption that the user will be hosting a personal, non-EAS, *static* manifest and asset server.

`expo export --eas` strips away all of the extraneous work in `expo export` and introduces a standardized format to save bundles in for preparation use by `eas publish`.

```
dist
├── assets
│   └── *
├── bundles
│   ├── android.js
│   └── ios.js
└── metadata.json
```

Where `metadata.json` has type:

```
type Metadata = {
  version: number;
  bundler: 'metro';
  fileMetadata: {
    [key in Platform]: {
      bundle: string; 
      assets: { [key in 'path' | 'ext']: string }[]; 
    };
  };
}
```

This allows for easy expansion of the bundlers we support (which is the key reason for the separation of the bundling and the publishing). The versioning allows us to accommodate dramatic format changes.

Test Plan:
run:
`expo export --experimental-bundle` and confirm it creates the new file structure.
`expo export --public-url test.com` and confirm it creates the old file structure.


packages/expo-cli/src/commands/bundle.ts
```diff
@@ -0,0 +1,139 @@
+import { getConfig, getDefaultTarget, ProjectTarget } from '@expo/config';
+import { buildPublishBundlesAsync, shouldUseDevServer } from '@expo/xdl/build/Project';
+import { saveAssetsAsync } from '@expo/xdl/build/ProjectAssets';
```
fson, line 3: Curious about why these are imported from `@expo/xdl/build/<module>` rather than
```
import { Project, ProjectAssets } from '@expo/xdl';
```

packages/expo-cli/src/commands/bundle.ts
```diff
@@ -0,0 +1,139 @@
...
+import { Command } from 'commander';
+import fs from 'fs-extra';
+import { uniqBy } from 'lodash';
```
fson, line 6:
```suggestion
import uniqBy from 'lodash/uniqBy';
```

packages/expo-cli/src/commands/bundle.ts
```diff
@@ -0,0 +1,139 @@
...
+  quiet: boolean;
+  target?: ProjectTarget;
+  force: boolean;
+};
+type PlatformMetadata = { bundle: string; assets: { [key in 'path' | 'ext']: string }[] };
+type FileMetadata = {
+  [key in Platform]: PlatformMetadata;
+};
+type Metadata = {
+  version: number;
```
fson, line 27:
```suggestion
  version: 0;
```
This is the type for metadata v0 anyway? Metadata v1 object might have different fields not included here.

packages/expo-cli/src/commands/bundle.ts
```diff
@@ -0,0 +1,139 @@
...
+
+const bundlePlatforms = [Platform.Android, Platform.iOS];
+
+type Options = {
+  outputDir: string;
+  quiet: boolean;
+  target?: ProjectTarget;
+  force: boolean;
+};
+type PlatformMetadata = { bundle: string; assets: { [key in 'path' | 'ext']: string }[] };
```
fson, line 22:
```suggestion
type PlatformMetadata = { bundle: string; assets: { path: string; ext: string; }[] };
```

fson (commented):
Thanks, this looks good! Left a few comments/questions below.

packages/expo-cli/src/commands/bundle.ts
jkhales: great point.

fson (approved):
:shipit:

EvanBacon (changes requested):
- if this is supposed to be part of EAS, it should maybe be in EAS?
- if this shares a majority of features with export, then it should share code too. 
- I'm not sure why we need two CLI commands for generating the static output of a project, this should really be consolidated.
- we want to limit the amount of commands in the expo-cli since there are already a ton and the `--help` prompt is too large.

brentvatne (changes requested):
i agree with @EvanBacon that we should hide this from `--help` for now. we may want to consolidate `expo export` into this command and put it behind a flag, ping me on slack and we can chat about that more

jkhales:
> * if this is supposed to be part of EAS, it should maybe be in EAS?

There is a big push to seperate the bundler from EAS. @fson, @ide, and @ccheever have suggested going even further in the future and allowing arbitrary bundlers (not included in `expo-cli`) with an expo plugin.

> * if this shares a majority of features with export, then it should share code too.

Agreed. I made sure to reuse the relevant code from `expo export`. See, for example, the reuse of `buildPublishBundlesAsync`: https://github.com/expo/expo-cli/pull/3074/files#diff-3b46ddce3f17b5ba410e409f3404d47ddbb44c0a45fa865ff8f0cf5404190ed7R71 

> * I'm not sure why we need two CLI commands for generating the static output of a project, this should really be consolidated.

I initially started writing this as a flag for `expo export`. I stopped as `expo export` does more than just create a bundle (for example it modifies the plist) and its output is tailored to the assumption that the user will be hosting the update from a static server.

Note: "hosting an update from a static server" is different from "generating the static output":

Hosting on a static server means we create the manifest as well as the bundles and that the client's plist looks for the manifest along a different route `*/android-index.json` or `*/ios-index.json`.

Creating just the `static output` means the bundles and the assets. But we also wanted to add some metadata about the assets (their extension) as that is required for the way we store assets in EAS (thus the metadata.json file). I also added versioning and the bundler name to this file in preparation for issues with multi-bundler support.

Combining these commands is possible, but because the output is rather different it seems confusing for the user and from the maintenance perspective the code becomes littered with if/else statements.

jkhales:
> i agree with @EvanBacon that we should hide this from `--help` for now. we may want to consolidate `expo export` into this command and put it behind a flag (or the other way around), ping me on slack and we can chat about that more

cool, pinged you on slack!

packages/xdl/src/Project.ts
```diff
@@ -354,8 +376,10 @@ export async function exportForAppHosting(
     dumpAssetmap?: boolean;
     dumpSourcemap?: boolean;
     publishOptions?: PublishOptions;
-  } = {}
+  } = {},
+  eas: boolean
 ): Promise<void> {
+  const absoluteOutputDir = path.resolve(process.cwd(), outputDir);
```
jkhales, line 382: This used to be called right before `ExportAppAsync`. I brought it inside and replaced the relevant `outputDir` with `absoluteOutputDir`.

packages/xdl/src/Project.ts
```diff
@@ -1586,7 +1651,7 @@ export async function startReactNativeServerAsync({
       _logPackagerOutput(projectRoot, 'error', data);
     }
   });
-  const exitPromise = new Promise((resolve, reject) => {
+  const exitPromise = new Promise<void>((resolve, reject) => {
```
jkhales, line 1654: (Unrelated typescript error that I tacked onto this PR)

packages/xdl/src/Project.ts
```diff
@@ -511,51 +510,117 @@ export async function exportForAppHosting(
...
-    androidBundle,
-    androidSourceMap,
-    androidManifest,
-    projectRoot,
-    log: (msg: any) => {
-      logger.global.info({ quiet: true }, msg);
-    },
-  };
+  // Skip the hooks and manifest creation if building for EAS.
+  if (!eas) {
```
jkhales, line 519: Everything in this `if` statement is copied over from before. Because `if (options.dumpSourcemap)` got shifted above the "hooks" block it looks like more change than it is.

jkhales:
@EvanBacon @brentvatne  changed from a new command to a flag!

ide:
The flag name looks fine for a temporary option but ideally the exported format would be largely compatible with any host.

old:
`expo publish`: publish to classic Expo updates
`expo export`: command for self-hosting classic Expo updates

new:
`expo export --eas`: there is no difference between Expo-hosting and self-hosting; there is just hosting. command for exporting an update that can be uploaded to EAS or to any other compatible server. EAS-specific details are the responsibility of eas-cli.

jkhales:
> The flag name looks fine for a temporary option but ideally the exported format would be largely compatible with any host.
> 
> old:
> `expo publish`: publish to classic Expo updates
> `expo export`: command for self-hosting classic Expo updates
> 
> new:
> `expo export --eas`: there is no difference between Expo-hosting and self-hosting; there is just hosting. command for exporting an update that can be uploaded to EAS or to any other compatible server. EAS-specific details are the responsibility of eas-cli.

Good point, talking with @jonsamp about this a bit and maybe `--bundle` would work? 😅

ide:
tbh I think the right path is to work towards making it the default to export the EAS-compatible update format. The old format just needs to stick around to not break existing workflows. Don't need to get there right away but the new format works for all conformant hosts - EAS and otherwise - so there eventually won't be a need for old "expo export" behavior except for compat reasons.

packages/expo-cli/src/commands/export.ts
```diff
@@ -230,5 +234,6 @@ export default function (program: Command) {
       []
     )
     .option('--max-workers [num]', 'Maximum number of tasks to allow Metro to spawn.')
+    .option('--eas', 'export bundles for use with EAS updates.')
```
EvanBacon, line 237: If I'm not mistaken this just exports bundles from Metro, has nothing to do with EAS.

packages/xdl/src/Project.ts
```diff
@@ -28,6 +28,7 @@ import freeportAsync from 'freeport-async';
 import fs from 'fs-extra';
 import getenv from 'getenv';
 import HashIds from 'hashids';
+import invariant from 'invariant';
```
EvanBacon, line 31: Use built-in `assert`

packages/xdl/src/Project.ts
```diff
@@ -119,6 +120,18 @@ type Release = {
 
 type ProjectStatus = 'running' | 'ill' | 'exited';
 
+type BundlePlatform = 'android' | 'ios';
+const bundlePlatforms: BundlePlatform[] = ['android', 'ios'];
```
EvanBacon, line 124: Group the types together

packages/xdl/src/ProjectAssets.ts
```diff
@@ -190,18 +192,20 @@ export async function publishAssetsAsync(
...
-  projectRoot,
-  exp,
-  hostedUrl,
-  assetPath,
-  outputDir,
-  bundles,
-}: ExportAssetsOptions) {
+export async function exportAssetsAsync(
+  { projectRoot, exp, hostedUrl, assetPath, outputDir, bundles }: ExportAssetsOptions,
+  eas?: boolean
```
EvanBacon, line 197: Why would this arbitrary setting not be included with the others?

EvanBacon (approved):
looks good but at least add a test plan to the PR description

PR #3098, opened 2021-01-20
[expo-cli] add option to assign created push key to current project

cruzach:
fixes https://github.com/expo/expo/issues/11656

After generating a new push key, we should give the option to assign that key to the current project context, if there is one. Otherwise, it can be confusing that you need to create the key, _then_ assign it

### DIFF
--- a/packages/expo-cli/src/credentials/views/IosPushCredentials.ts
+++ b/packages/expo-cli/src/credentials/views/IosPushCredentials.ts
@@ -39,10 +39,6 @@ export class CreateIosPush implements IView {
     displayIosUserCredentials(pushKey);
     log();
 
-    if (ctx.hasProjectContext) {
-      await this.assignToCurrentProject(ctx, pushKey);
-    }
-
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/credentials/views/IosPushCredentials.ts
```diff
@@ -38,6 +38,11 @@ export class CreateIosPush implements IView {
     log('Successfully created Push Notification Key\n');
     displayIosUserCredentials(pushKey);
     log();
+
```
wkozyra95, line 41: This action is called from other places e.g. as part of the setup before build where this behaviour would not be correct.

This was intended to simple CRUD action that results in the creation of a new push key, if the intention is that ensuring valid push key is assigned to the app then logic for that should be different and is already present in build command.

If it's unclear that this command won't assign dist cert to the app then maybe it's better to add some warnings. If you want to modify this action without affecting build then create new action that is using `CreateIosPush` class.
```
async open(ctx: Context): Promise<> {
  await runCredentialsManager(ctx, new CreateIosPush(accountName));
  // some code you want to add
}
```
and replace it here https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/credentials/views/Select.ts#L99

cruzach: ah I see, thanks for catching that

Changed it and created a new class that extends `CreateIosPush` instead - https://github.com/expo/expo-cli/pull/3098/commits/dcae9ca0983a5b9fc7c136fbbc056f9d79c56088

quinlanj (approved):
hm, this does seem like the better user experience. Thanks for this PR @cruzach. I'll work on changing this in eas too

wkozyra95 (approved):

PR #3121, opened 2021-01-25
Check for optional `devDependencies`

benallfree:
Noticed this bug when trying to eject without `devDependencies` defined in package.json.


EvanBacon (commented):
fix the lint plz

benallfree:
@EvanBacon sorry bout that

PR #3135, opened 2021-01-27
remove validator packages from expo-cli

EvanBacon:
- `validator` is a package that [continues to grow](https://packagephobia.com/result?p=validator) larger and we barely use it.
- We have two copies of the package which is about 1Mb of extra code that is downloaded in two different requests.
- Remove it in favor of two methods that accomplish the same tasks (isURL, isUUID).
- Write unit tests for the URL method since it's a bit more confusing than the UUID method.


packages/expo-cli/src/commands/utils/isUUID.ts
```diff
@@ -0,0 +1,13 @@
+const uuid = {
+  3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+  4: /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
+  5: /^[0-9A-F]{8}-[0-9A-F]{4}-5[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
+  all: /^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+};
+
+export function isUUID(str: string, version: keyof typeof uuid = 'all'): boolean {
```
fson, line 8:
```suggestion
export function isUUID(str: string): boolean {
```
I believe the `all` regex is all we need for now.

fson (approved):

brentvatne (approved):

PR #3141, opened 2021-01-27
Drop nodeModulesPath support

EvanBacon:
- Update config https://github.com/expo/universe/pull/6889
- Delete all instances of nodeModulesPath in favor of regular module resolution
- Update tests


packages/config-types/src/ExpoConfig.ts
```diff
@@ -429,7 +427,7 @@ export interface IOS {
    */
   splash?: {
     /**
-     * Local path to a XIB file as the loading screen. It overrides other loading screen options. Note: This will only be used in the standalone app (i.e., after you build the app). It will not be used in Expo Go.
+     * Local path to a XIB file as the loading screen. It overrides other loading screen options. Note: This will only be used in the standalone app (i.e., after you build the app). It will not be used in the Expo Go.
```
brentvatne, line 430: "the Expo Go" 🤔 

maybe we need to fix schema

packages/expo-cli/src/commands/install.ts
```diff
@@ -72,17 +71,17 @@ async function installAsync(packages: string[], options: PackageManager.CreateFo
   }
 
   // This shouldn't be invoked because `findProjectRootAsync` will throw if node_modules are missing.
-  if (!fs.existsSync(path.join(exp.nodeModulesPath || projectRoot, 'node_modules'))) {
+  // Every React project should have react installed...
+  if (!resolveFrom.silent(projectRoot, 'react')) {
```
brentvatne, line 75: reasonable

packages/expo-optimize/src/assets.ts
```diff
@@ -17,6 +17,14 @@ import { basename, join, parse, relative } from 'path';
 import prettyBytes from 'pretty-bytes';
 import temporary from 'tempy';
 
+function fileExists(file: string): boolean {
```
brentvatne, line 20: why the different implementations here and in `fileExistsAsync`? i understand the code between expo-cli utils and expo-optimize isn't shared, but probably no need for an async version to exist and this sync version is easier to read/understand imo.

brentvatne (approved):
😭 `nodeModulesPath` is the worst and i hope nothing ends up coming up when we release this because i really don't want it to exist anymore

PR #3143, opened 2021-01-28
[webpack-config] [expo-pwa] Handles publicPath consisting of protocol + host

pckilgore:
Implements joinUrlPath helper to create correctly formed urls for PWAs using an absolute url as a public path, such as `https://www.example.com`.

--- original notes ---

Expo PWA and its helpers are currently generating malformed urls when using a a fully formed protocol + host as a publicPath.

Example:
image

I've traced this to the use of `path.join` breaking with urls.

This is quite a common pattern so before doing anything substantial I wanted to solicit suggestions as to approach:
image

^^ I swear there were more instances of this, perhapse with another var instead of publicPath?

I would propose something like the helper in the PR, but IIR `URL` is a pain in the ass to work with due to many edge cases where it throws (see, for example, my bad helper here).  I'd be happy to use another lib, just wanted to check in before I looked any deeper at alternatives.

The attached shitty PR is just because I know some folks prefer code > prose.

Please feel free to reach out on #expo-draftbit slack channel, happy to assist synchronously on this as it is a bit of a blocker for us right now.

Thanks 🙏


packages/webpack-config/src/plugins/__tests__/joinPaths-test.js
```diff
@@ -0,0 +1,27 @@
...
+it(`joins full url publicPath`, () => {
+  const got = joinPublicPath(publicPath, filePath);
+  expect(got).toBe(want);
+  expect(got).not.toBe(bad);
+});
+
+const spy = jest.spyOn(path, 'join');
+
+it(`joins non-urls with path.join`, () => {
+  const got = joinPublicPath('/pwa', filePath);
```
EvanBacon, line 24: Maybe also test against `''` or `'/'`

pckilgore:
I found a few more cases hiding behing a different var name
image

packages/webpack-config/src/plugins/__tests__/joinPaths-test.js
pckilgore: 9b57858

packages/pwa/src/index.ts
```diff
@@ -1,11 +1,26 @@
...
+/**
+ * Joins a url protocol + host to path segments, falls back to path.join
+ * if result is not a valid url.
+ */
+export function joinUrlPath(publicPath: string, ...toJoin: string[]): string {
+  const segments = path.join(...toJoin);
+  try {
+    // Throws if publicPath is not a valid protocol+host
+    return new URL(segments, publicPath).href;
+  } catch (_) {
```
EvanBacon, line 19:
```suggestion
  } catch {
```

EvanBacon (approved):
LGTM

Issue #3062, opened 2021-01-07
Not authorized error when running project in development with owner field


brentvatne:
If a project has an owner field and your current user is not part of that team/org, you cannot start the project for development. This is a problem in https://github.com/expo/expo-cli/issues/2436 - we want to publish it under `react-navigation`, but we don't want to force all users who want to run the example to be a part of the team/org.

Issue #238, opened 2018-12-05
Automatically switch to offline mode when manifest can't be signed


fson:
When not in offline mode, Expo CLI currently needs to make a request to `exp.host` to to be able to sign a manifest. If making the request fails (e.g. no internet connection), we should automatically fall back to offline mode and notify the user about it.

> Your internet connection has been lost, so Expo CLI has switched to offline mode. For security reasons there are several limitations in offline mode, read more at: https://docs.expo.io/versions/v31.0.0/workflow/expo-cli-offline-mode_TBD

PR #3148, opened 2021-01-28
[xdl] Handle errors when trying to sign the manifest

fson:
When the `ManifestHandler` tries to sign the manifest (requested by the client), automatically handle typical errors and fall back to offline mode:
- The project has an `owner` specified, but you lack permissions to that account -> fall back to offline mode! Fixes https://github.com/expo/expo-cli/issues/3062.
  Screen Shot 2021-01-28 at 17 02 17

- You don't have an internet connection, which results in `Error: getaddrinfo ENOTFOUND exp.host` -> fall back to offline mode! Fixes https://github.com/expo/expo-cli/issues/238.
  Screen Shot 2021-01-28 at 16 59 05

### DIFF
--- a/packages/xdl/src/project/ManifestHandler.ts
+++ b/packages/xdl/src/project/ManifestHandler.ts
@@ -251,7 +251,12 @@ export async function getManifestResponseAsync({
       manifestString = await getManifestStringAsync(manifest, hostInfo.host, acceptSignature);
     } else if (error.code === 'ENOTFOUND') {
       // Got a DNS error, i.e. can't access exp.host, warn and enable offline mode.
-      addSigningDisabledWarning(projectRoot, 'Not connected to internet.');
+      addSigningDisabledWarning(
+        projectRoot,
+        `Could not reach Expo servers, please check if you can access ${
+          error.hostname || 'exp.host'
+        }.`
+      );
       Config.offline = true;
       manifestString = await getManifestStringAsync(manifest, hostInfo.host, acceptSignature);
     } else {
--- a/packages/xdl/src/project/ManifestHandler.ts
+++ b/packages/xdl/src/project/ManifestHandler.ts
@@ -1,5 +1,6 @@
 import { ExpoAppManifest, ExpoConfig, getConfig } from '@expo/config';
 import { JSONObject } from '@expo/json-file';
+import chalk from 'chalk';
 import express from 'express';
 import http from 'http';
 import os from 'os';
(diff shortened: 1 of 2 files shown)


packages/xdl/src/project/ManifestHandler.ts
```diff
@@ -240,6 +266,21 @@ export async function getManifestResponseAsync({
...
 
+const addSigningDisabledWarning = (() => {
+  let seen = false;
+  return (projectRoot: string, reason: string) => {
+    if (!seen) {
+      seen = true;
+      ProjectUtils.logWarning(
+        projectRoot,
+        'expo',
+        reason + '\n' + `Falling back to offline mode.`,
```
brentvatne, line 277:
```suggestion
        reason + '\nFalling back to offline mode.`,
```

packages/xdl/src/project/ManifestHandler.ts
```diff
@@ -231,7 +233,31 @@ export async function getManifestResponseAsync({
...
+    manifestString = await getManifestStringAsync(manifest, hostInfo.host, acceptSignature);
+  } catch (error) {
+    if (error.code === 'UNAUTHORIZED_ERROR' && manifest.owner) {
+      // Don't have permissions for siging, warn and enable offline mode.
+      addSigningDisabledWarning(
+        projectRoot,
+        `This project belongs to ${chalk.bold(
+          `@${manifest.owner}`
+        )} and you have not been granted the appropriate permissions.\n` +
+          `Please request access from an admin of @${manifest.owner} or change the \"owner\" field to an account you belong to.\n` +
```
brentvatne, line 247: github warnings here are accurate re: escape character

brentvatne (approved):
looks good, i think this is a nice solution to our owner field problems. just need to fix the warnings raised by linter then ship it

packages/xdl/src/project/ManifestHandler.ts
```diff
@@ -231,7 +233,31 @@ export async function getManifestResponseAsync({
...
+          learnMore('https://docs.expo.io/versions/latest/config/app/#owner')
+      );
+      Config.offline = true;
+      manifestString = await getManifestStringAsync(manifest, hostInfo.host, acceptSignature);
+    } else if (error.code === 'ENOTFOUND') {
+      // Got a DNS error, i.e. can't access exp.host, warn and enable offline mode.
+      addSigningDisabledWarning(projectRoot, 'Not connected to internet.');
```
byCedric, line 254: For users in countries where you might need a VPN, this could be confusing. They could have access to "the internet" but not `exp.host`. 

- Maybe rewording this to "Could not reach the Expo servers" might be better?
- Or even better, add tips for the user to manually test/fix connection (e.g. "Could not reach Expo servers, please check if you can access exp.host")

fson: Great point, I'll update the wording here.

Issue #2879, opened 2020-11-13
Eject hanging on installing CocoaPods when sudo is needed


brentvatne:
When trying to eject after setting up a fresh macOS machine, I ended up in this state:

> ⠙ CocoaPods CLI not found in your PATH, installing it now.Your password might be needed to install CocoaPods CLI: https://guides.cocoapods.org/using/getting-started.html#installation

![Screen Shot 2020-11-12 at 6 49 09 PM](https://user-images.githubusercontent.com/90494/99022989-82de9480-2518-11eb-9d05-68eca9e13f32.png)

I didn't see any prompt to continue, but tried entering my password anyhow and hitting return and this unblocked it.

![Screen Shot 2020-11-12 at 6 55 14 PM](https://user-images.githubusercontent.com/90494/99023030-9d187280-2518-11eb-82f5-a466b8c5d438.png)

brentvatne:
Also - `sudo gem install cocoapods` no longer works with the Ruby version that ships with macOS, or at least that's my experience here

cloggo:
The installation will continue if you actually enter the password even though there is no prompt that asking you to enter the password.

brentvatne:
@cloggo - yup! i noted that above:

> I didn't see any prompt to continue, but tried entering my password anyhow and hitting return and this unblocked it.

ismaelbarry:
> @cloggo - yup! i noted that above:
> 
> > I didn't see any prompt to continue, but tried entering my password anyhow and hitting return and this unblocked it.

Exactly, which password should be entered here as it doesn't specify?

brentvatne:
@ismaelbarry - your operating system user. the same as if you ran `sudo <some-command>`

ismaelbarry:
> @ismaelbarry - your operating system user. the same as if you ran `sudo <some-command>`

Gotcha! Thanks, I was able to get it to work.

ghacosta:
this happened to me too on macOS Catalina and using latest SDK and expo-cli at the moment. I've solved this too by entering my SO password and hitting return/enter key

PR #3150, opened 2021-01-28
Always install eject pods in non interactive mode

EvanBacon:
- resolve https://github.com/expo/expo-cli/issues/2879 by skipping the sudo prompt. There is too much going on in the single process to attempt a prompt, so we'll just skip it and warn the user to install CocoaPods CLI and install the pods manually `npx pod-install`. 
- Improved the errors being thrown by `CocoaPodsPackageManager`, you can now see more input on why pod install failed:
```
⚠️  Unable to install the CocoaPods CLI. Continuing with project sync, you can install CocoaPods CLI afterwards.
Failed to install CocoaPods with Homebrew. Please install CocoaPods CLI manually and try again.
└─ Cause: CocoaPodsError: Homebrew installation appeared to succeed but CocoaPods CLI not found in PATH and unable to link.
└─ Cause: Error: brew exited with non-zero code: 1
```

### DIFF
--- a/packages/expo-cli/src/commands/utils/CreateApp.ts
+++ b/packages/expo-cli/src/commands/utils/CreateApp.ts
@@ -1,6 +1,5 @@
 import * as PackageManager from '@expo/package-manager';
 import chalk from 'chalk';
-import program from 'commander';
 import fs from 'fs-extra';
 import getenv from 'getenv';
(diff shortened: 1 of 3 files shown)


brentvatne (approved):
reasonable choice, i agree that prompting for sudo might be a bit much in the middle of all of this output

PR #3153, opened 2021-01-29
simplify open in editor util

EvanBacon:
- Moved from #3146
- Drop FileSystem module from xdl
- Remove edge case where editor is opened twice on macOS


packages/xdl/src/FileSystem.ts
```diff
@@ -1,65 +0,0 @@
-import * as osascript from '@expo/osascript';
-import spawnAsync from '@expo/spawn-async';
-import { execSync } from 'child_process';
-
-import XDLError from './XDLError';
-
```
fson, line 6: Good riddance 😄

fson (approved):

PR #3154, opened 2021-01-29
Split up Exp module

EvanBacon:
- Moved from #3146
- Move minipass and pacote to expo-cli out of xdl
- Move ngrok code to ngrok module
- Move extract template into expo-cli


packages/expo-cli/src/commands/eject/Eject.ts
```diff
@@ -687,7 +687,7 @@ async function cloneNativeDirectoriesAsync({
   let copiedPaths: string[] = [];
   let skippedPaths: string[] = [];
   try {
-    await Exp.extractTemplateAppAsync(templateSpec, tempDir, exp);
+    await extractTemplateAppAsync(templateSpec, tempDir, exp);
```
fson, line 690: Excellent idea, there's no need for this to be in `xdl` since it's only used here. It was a relic from the times when both `exp` and XDE depended on xdl and shared this functionality 😅

fson (approved):
Looks good, thanks for all the good housekeeping this week!

PR #3164, opened 2021-02-02
[xdl] split up Project module

EvanBacon:
# Why

A simple refactor to split up the XDL `Project` module into smaller more disconnected methods that can be pruned and deprecated if needed. 

# How

- Split various methods from Project into `project/*`. Ideally these methods would be split into more aptly named folders like build, publish, export, etc. but that can be done another time.
- Deprecated the seemingly unused `setOptionsAsync()`
- Deprecated `Project.currentStatus` in favor of a better named and located method `ProjectSettings.getCurrentStatusAsync()`

# Test Plan

- [x] start
- [x] stop
- [x] start web
- [x] publish
- [x] export
- sparse unit tests should continue working


packages/xdl/src/project/publishAsync.ts
```diff
@@ -0,0 +1,459 @@
...
+} from './startLegacyReactNativeServerAsync';
+
+const MINIMUM_BUNDLE_SIZE = 500;
+
+type PackagerOptions = {
+  dev: boolean;
+  minify: boolean;
+};
+
+export async function buildPublishBundlesAsync(
```
fson, line 47: This is used by both export and publish, so I think it would be better to put it in a separate file (`buildPublishBundlesAsync.ts`).

packages/xdl/src/project/publishAsync.ts
```diff
@@ -0,0 +1,459 @@
...
+  formData.append('androidBundle', androidBundle, 'androidBundle');
+  formData.append('options', JSON.stringify(options));
+
+  const user = await UserManager.ensureLoggedInAsync();
+  const api = ApiV2.clientForUser(user);
+
+  return await api.uploadFormDataAsync('publish/new', formData);
+}
+
+async function _handleKernelPublishedAsync({
```
fson, line 415: Looking forward to getting this removed 🤞 

https://github.com/expo/expo/issues/9086

packages/xdl/src/project/startLegacyReactNativeServerAsync.ts
```diff
@@ -0,0 +1,345 @@
...
+  projectRoot,
+  options = {},
+  exp = getConfig(projectRoot).exp,
+  verbose = true,
+}: {
+  projectRoot: string;
+  options: StartOptions;
+  exp?: ExpoConfig;
+  verbose?: boolean;
+}): Promise<void> {
```
fson, line 129: Nice to see how much code we will be able to remove when we no longer support this way to run the server!

fson (approved):
Excellent, this will make things more manageable and ease cleaning things up further.

packages/xdl/src/project/mergeAppDistributions.ts
```diff
@@ -0,0 +1,108 @@
...
+type SelfHostedIndex = ExpoAppManifest & {
+  dependencies: string[];
+};
+
+function isSelfHostedIndex(obj: any): obj is SelfHostedIndex {
+  return !!obj.sdkVersion;
+}
+
+// Takes multiple exported apps in sourceDirs and coalesces them to one app in outputDir
+export async function mergeAppDistributions(
```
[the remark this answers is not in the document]
fson: Just saw you already mentioned this in the description, feel free to ignore this comment! 😄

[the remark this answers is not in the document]
EvanBacon: can't, too much shared code with publish which dev-tools calls into

PR #3165, opened 2021-02-02
[xdl] Remove windows 7z1602-extra binary

EvanBacon:
# Why 

This binary (3.3mb) is unused on mac and linux machines, it's also appears to only be used for extracting the Android client app and extracting `androidExpoViewUrl` when legacy detaching. 

# How

Delete `7z1602-extra` windows binary and use the fallback node tar package to extract incoming requests.

# Test Plan

- TBD


fson (approved):
Makes sense, thanks 👍

PR #3170, opened 2021-02-02
Unify node version

EvanBacon:
Apparently we mostly work against node 12 types. Fixed a few issues, but mostly unified `types/node` version so we can easily upgrade / downgrade across the repo.


packages/expo-cli/src/commands/start/TerminalUI.ts
```diff
@@ -122,15 +122,19 @@ export const printServerInfo = async (
 export const startAsync = async (projectRoot: string, options: StartOptions) => {
   const { stdin } = process;
   const startWaitingForCommand = () => {
-    stdin.setRawMode(true);
+    if (stdin.setRawMode) {
```
fson, line 125: I think things won't work correctly if we can't call `stdin.setRawMode` (I think this might happen when `stdin` isn't a TTY, e.g. if it's a piped from some other process, in other words "non-interactive mode"). Probably best to bail out from the whole thing in that case, e.g.:
```ts
if (!stdin.setRawMode) {
  log.warn('Non-interactive terminal, keyboard commands are disabled.);
  return;
}
```
I think we'd normally not hit this case because we check the interactivity earlier and put it in the `command.nonInteractive` property, but better to be safe than sorry!

fson (changes requested):

brentvatne (approved):

PR #3176, opened 2021-02-02
[config-plugins] Use the `export * as ns` syntax

fson:
Follow up to https://github.com/expo/expo-cli/pull/3161: Changed the code to use the `export * as ns` syntax introduced in TypeScript 3.8.

Spotted this comment in the file:
```
// We can change this to export * as X with TypeScript 3.8+
// https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#export-star-as-namespace-syntax
// .. but we should wait for this to be the default VSCode version.
```
This is now applicable because 3.8 shipped months ago and 4.1 is the current stable version.

### DIFF
--- a/packages/config-plugins/src/ios/index.ts
+++ b/packages/config-plugins/src/ios/index.ts
@@ -1,57 +1,25 @@
-import * as AdMob from './AdMob';
-import * as Branch from './Branch';
-import * as BundleIdentifier from './BundleIdentifier';
-import * as CustomInfoPlistEntries from './CustomInfoPlistEntries';
-import * as DeviceFamily from './DeviceFamily';
-import * as Entitlements from './Entitlements';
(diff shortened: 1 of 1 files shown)


fson:
ESLint is failing: https://github.com/expo/expo-cli/pull/3176/checks?check_run_id=1813935257

Seems like `eslint-plugin-import` doesn't work correctly with this syntax: https://github.com/benmosher/eslint-plugin-import/issues/1883

fson:
Tried upgrading to the latest version of `eslint-config-universe`, which bumped `eslint-plugin-import` to `2.22.1` and `@babel/eslint-parser` to `7.12.1` (currently their latest versions), but it didn't help.

Worth revisiting when ESLint supports this.

PR #3178, opened 2021-02-02
[configure-splash-screen] Remove @react-native-community/cli-tools

EvanBacon:
# Why

- `@react-native-community/cli-tools` is bloated and imports a lot of code, we only use it for one function. 
- Removing the android counter part cut the bundle size in half.

# How

- Copied code over from `expo/config-plugins`

# Test Plan

- copied code into expo-cli, ejected and built.


bbarthec (approved):
Looks good to me 👍 
I want to ask for some CHANGELOG entry describing this change and I'm fine with it.

PR #3182, opened 2021-02-03
[xdl] Allow dev client apps to be launched in the iOS simulator

fson:
This is the "v0" version of the dev client simulator flow for iOS. In this version you can open an app in its dev client in the iOS simulator, provided that the dev client has already been built and installed to the device. The v1 version will also build the app and install it on the device.

### Why is this in XDL?

We're currently splitting up XDL and making it lighter. However, the existing simulator functionality is in included in the `Simulator` module of XDL. It depends on many XDL modules, including `UrlUtils`, `Versions`, `Logger`, `UserSettings`, etc. making it difficult to extract to a separate package until these modules have been moved out. Additionally, Expo Dev Tools depends on the `Simulator` module, so it can't be moved to `expo-cli` itself, as of now.

For this reason, I've added the dev client support in the existing `Simulator` module in XDL. Refactoring the `Simulator` module should be done in a separate PR. Additonally, the new `BundleIdentifier` module should be moved to a different package and parts of it potentially shared between `expo-cli` and `eas-cli`.

Screen Shot 2021-02-04 at 17 04 42
Screen Shot 2021-02-04 at 17 05 50

### DIFF
--- a/packages/xdl/src/BundleIdentifier.ts
+++ b/packages/xdl/src/BundleIdentifier.ts
@@ -1,3 +1,15 @@
+/**
+ * BundleIdentifier.ts
+ *
+ * NOTE:
+ * The code in this module originates from eas-cli and the canonical version of
+ * it is in
+ * https://github.com/expo/eas-cli/blob/6a0a9bbaaad03b053b5930f7d14bde35b4d0a9f0/packages/eas-cli/src/build/ios/bundleIdentifer.ts#L36
+ * Any changes to this code should be applied to eas-cli as well!
+ *
+ * TODO: move the code for configuring and validating the bundle identifier
+ * to a shared package that can be used for both eas-cli and expo-cli.
+ */
 import { ExpoConfig, getConfigFilePaths, getProjectConfigDescription } from '@expo/config';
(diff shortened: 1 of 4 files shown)


packages/xdl/src/Simulator.ts
```diff
@@ -603,6 +614,20 @@ export async function openUrlInSimulatorSafeAsync({
   };
 }
 
+async function ensureDevClientInstalledAsync(
```
EvanBacon, line 617:
```suggestion
async function assertDevClientInstalledAsync(
```

packages/xdl/src/BundleIdentifier.ts
```diff
@@ -0,0 +1,132 @@
+import { ExpoConfig, getConfigFilePaths, getProjectConfigDescription } from '@expo/config';
```
brentvatne, line 1: let's add a big doc block here to point to the canonical version of this code in eas-cli, so it's clearly documented that this is a copy/paste fork of it

brentvatne (approved):

EvanBacon (approved):

PR #3191, opened 2021-02-04
Refactor log module for better mocking capabilities

wschurman:
# Why

Mocking `log` and verifying results is a common way to unit test the CLI. Until this PR, the only way to mock the `log` module was to do the following due to the way it was set up (custom properties on a function instance, which was copied from expo-cli):
```typescript
const mockLog = {	
  __esModule: true, // this property makes it work	
  default: jest.fn(toLog => toLog),	
};	
(mockLog.default as any).newLine = jest.fn();	
jest.mock('../../../../log', () => mockLog);
```

Or to just not use the default `log` export and use one of the `nested` or other properties, which didn't require the weird esmodule mock thing. This is the strategy that I took with the `OTP` testing. This was bad since it required changing the behavior of the actual code just to make testing easier.

This PR makes it a better default export (class) for cleaner mocking. The corresponding `eas-cli` change was https://github.com/expo/eas-cli/pull/223.

# How

This should be reviewed as a single commit, but the process for generating it is as follows (making use of TypeScript refactoring):
1. Make the default export of `log.ts` an object instead of a function, and create a property on that object called `log`. Find/replace all `log(` with `log.log(`.
2. TypeScript refactor the name of the object to `Log` (uppercase)
3. Convert object to class with static methods, fix `tsc`.

# Test Plan

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/expo-cli/src/__tests__/accounts-test.ts
+++ b/packages/expo-cli/src/__tests__/accounts-test.ts
@@ -2,7 +2,7 @@ import { ApiV2 } from '@expo/xdl';
 
 import { _retryUsernamePasswordAuthWithOTPAsync, UserSecondFactorDeviceMethod } from '../accounts';
 import { jester } from '../credentials/__tests__/fixtures/mocks-constants';
-import log from '../log';
+import Log from '../log';
 import prompt, { selectAsync } from '../prompts';
 import { mockExpoXDL } from './mock-utils';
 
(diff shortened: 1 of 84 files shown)


quinlanj (approved):
didnt you have this exact pr several days ago?

wschurman:
> didnt you have this exact pr several days ago?

Yep, but in `eas-cli`.

PR #3194, opened 2021-02-05
[xdl] Add Dev Client simulator flow v0 for Android devices

fson:
Similar to https://github.com/expo/expo-cli/pull/3182, but for Android.


packages/config-plugins/src/android/Package.ts
```diff
@@ -146,3 +146,14 @@ export function setPackageInAndroidManifest(
 
   return androidManifest;
 }
+
+export async function getApplicationIdAsync(projectDir: string): Promise<string | null> {
```
EvanBacon, line 150: add tests for this

EvanBacon (approved):
Seems fine but a test plan would be helpful.

PR #3199, opened 2021-02-08
Update web index.html height style

joamartico:
# Why

Fixed the screen height in the web when the statusBar is transparent. Before this change, in the web was displayed an extra white space in the bottom of the screen.

# Test Plan

It works perfectly with react-navigation bottom tabs and header, and also with SafeArea component:
image

Landscape mode:
image

In iPhone 6/7/8 works normally:
image

### DIFF
--- a/packages/webpack-config/web-default/index.html
+++ b/packages/webpack-config/web-default/index.html
@@ -40,7 +40,7 @@
         scroll-behavior: smooth;
         /* Prevent text size change on orientation change https://gist.github.com/tfausak/2222823#file-ios-8-web-app-html-L138 */
         -webkit-text-size-adjust: 100%;
-        height: 100%;
+        height: calc(100% + env(safe-area-inset-top));
       }
 
       body {


packages/webpack-config/web-default/index.html
```diff
@@ -40,7 +40,7 @@
         scroll-behavior: smooth;
         /* Prevent text size change on orientation change https://gist.github.com/tfausak/2222823#file-ios-8-web-app-html-L138 */
         -webkit-text-size-adjust: 100%;
-        height: 100%;
+        height: calc(100% + env(safe-area-inset-top));
```
EvanBacon, line 43: Hmm, interesting idea. Have you tested this behavior with react-native-safe-area-context (react-navigation), landscape mode on iOS (resizing header bar)? Could you please add a "test plan" to your PR description:

```
# Test Plan

...
```

joamartico: Yes it works! I added screenshots to the test plan showing the behavior with react-navigation bottom tabs and header, with SafeArea component and with landscape mode.

EvanBacon (approved):
I've tested this on iOS Safari, iOS PWA (with status bar and without), Chrome Android, Android PWA, and desktop Chrome. It appears to work reliably. I think this mostly effects iOS PWAs with translucent status bars (fixing a known bug where the height didn't resize properly).

PR #3203, opened 2021-02-09
[pwa-manifest] add `crossorigin="use-credentials"` attribute

simoneb:
# Why

If the PWA is served with basic authentication, it will result in a 401 when getting the manifest file unless this attribute is included.

# How

Add `crossorigin="use-credentials"` attribute to the <link> element for the manifest file

See explanation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin#example_webmanifest_with_credentials).

# Test Plan

Serve the generated PWA with basic auth on and see a 401 being returned without this change.

### DIFF
--- a/packages/webpack-config/src/plugins/PwaManifestWebpackPlugin.ts
+++ b/packages/webpack-config/src/plugins/PwaManifestWebpackPlugin.ts
@@ -77,6 +77,7 @@ export default class PwaManifestWebpackPlugin extends JsonWebpackPlugin {
                 attributes: {
                   rel: this.rel,
                   href: joinUrlPath(this.pwaOptions.publicPath, this.pwaOptions.path),
+                  crossorigin: 'use-credentials',
                 },
               });
 


EvanBacon:
What are the drawbacks to using this? If it causes breaking changes to existing projects then it might be better to just manually [customize the manifest tag](https://docs.expo.io/guides/progressive-web-apps/#manifestjson).

simoneb:
> What are the drawbacks to using this? If it causes breaking changes to existing projects then it might be better to just manually [customize the manifest tag](https://docs.expo.io/guides/progressive-web-apps/#manifestjson).

Doing manual customization is an option too I guess, but it would require ejecting Web. I have not done it so I'm not sure exactly what that implies.

On your question about what the drawbacks are, there doesn't seem to be any, thought there are many ongoing discussions about why manifest links have the credentials off by default. What we know for sure, which is documented, is that if requesting the manifest file requires credentials that attribute must be there, with that specific value.

I did some research to understand why this behavior of a manifest link not sending credentials by default exists and here's what I found.

[This issue](https://github.com/w3c/manifest/issues/535) describes it quite well. Other `<link rel="...">` except manifest send credentials, and the reason is that the intention in the spec was to make it "secure by default" and to match the behavior of `fetch`. But then, because in `fetch` this was causing more troubles than not, [`fetch` was changed to send credentials by default](https://github.com/w3c/manifest/issues/535#issuecomment-457398011), while manifest remained as it was. 

Finally, [this pull request](https://github.com/GoogleChromeLabs/pwa-wp/pull/371) mentions that enabling it shouldn't make any (negative) difference, and it cites a few sources. 

For completeness, [create-react-app doesn't include that by default](https://github.com/facebook/create-react-app/blob/master/packages/cra-template/template/public/index.html#L17), but on the other hand CRA makes it much easier to customize the index.html file.

To summarize, I believe this change wouldn't have any negative impact, but if you think it may be a possible breaking change I guess we can stick to the option of ejecting Web, but I guess I would prefer to avoid ejecting and just having the ability to customize this.

simoneb:
This doesn't seem to be going forward. Closing

PR #3207, opened 2021-02-10
Persist dev tools preference

EvanBacon:
# Why

~~The devtools UI is not as well maintained as the CLI and it can be jarring for users to have this popup start from a CLI command every time. We'd like to disable opening it by default but also don't want to interrupt users by introducing a new change.~~

The proposed change will persist the user choice each time it's read. In the future, we can swap this to false, and most users will see no interruption.

# Test Plan

- `rm -rf ~/.expo/state.json && expo start` should still open dev tools.
- `shift+d` should change the state.
- restarting expo-cli should use the persisted state.


brentvatne (approved):
note for other people reading this - we discussed migrating state.json in such a way that it wouldn't change any existing user behavior but decided not to go down that route and add more code to xdl. users can find how to open it using the interactive prompt

brentvatne (approved):

PR #3209, opened 2021-02-10
[expo cli] redesign startup logs

EvanBacon:
# Why

Attempt to make startup logs more readable, and tie them closer to the other logs in the CLI

# How

- "Expo DevTools" -> "Developer tools"
- Group developer tools commands (d, shift+d) with other commands under the QR code. This is because the QR code often shifts all the logs up a lot, many users could end up missing them.
- Remove random colors.
- Move QR URL to under the QR and add some message.

## Before

Screen Shot 2021-02-09 at 7 23 02 PM

## After

Screen Shot 2021-02-09 at 7 13 38 PM

# Test Plan

- `shift+d` to toggle the developer tools

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.ts
+++ b/packages/expo-cli/src/commands/start/TerminalUI.ts
@@ -36,21 +36,23 @@ type StartOptions = {
 };
 
 const printHelp = (): void => {
-  Log.newLine();
-  Log.nested(`Press ${b('?')} to show a list of all available commands.`);
+  logCommandsTable([[], ['?', 'show all commands']]);
 };
 
 const div = chalk.dim(`│`);
 
-const printUsage = async (projectDir: string, options: Pick<StartOptions, 'webOnly'> = {}) => {
-  const { dev } = await ProjectSettings.readAsync(projectDir);
+const printUsageAsync = async (
+  projectRoot: string,
+  options: Pick<StartOptions, 'webOnly'> = {}
+) => {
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/start/TerminalUI.ts
```diff
@@ -87,38 +108,39 @@ const printUsage = async (projectDir: string, options: Pick<StartOptions, 'webOn
...
-  const iosInfo = process.platform === 'darwin' ? `, or ${b('i')} for iOS simulator` : '';
-  const webInfo = `${b`w`} to run on ${u`w`}eb`;
-  Log.nested(wrapAnsi(u('To run the app, choose one of:'), wrapLength));
-
-  // TODO: if dev client, chanege this message!
-  Log.nested(item(`Scan the QR code above with the Expo app (Android) or the Camera app (iOS).`));
+  Log.nested(item(`Waiting on ${u(url)}`));
+  // Log.newLine();
+  // TODO: if dev client, change this message!
+  Log.nested(item(`Scan the QR code above with the Expo app (Android) or the Camera app (iOS)`));
```
brentvatne, line 129:
```suggestion
  Log.nested(item(`Scan the QR code above with Expo Go (Android) or the system Camera app (iOS)`));
```

brentvatne:
i like this change! i think it makes the interactive prompt more discoverable and shifts actionable content to below the qr code

PR #3239, opened 2021-02-18
Add a command to build and run on Android

fson:
# Why

This is going to be the v1 of the emulator flow for Android dev client. The command allows building, installing and launching the dev client app on Android.
 
# How

Launch with `expod run --platform android`.

TODO:
- [x] Hide the command from `--help`

# Test Plan

- Create an app
    npx crna -t with-development-client my-dev-client-app
    cd my-dev-client-app
- Uncomment this line in `android/gradle.properties`: `# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8`
- Build and run it
    expod run:android

### DIFF
--- a/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
+++ b/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
@@ -21,6 +21,10 @@ export default async function buildAndroidClientAsync(
   if (!device) {
     return;
   }
+  const bootedDevice = await Android.attemptToStartEmulatorOrAssertAsync(device);
+  if (!bootedDevice) {
+    return;
+  }
 
   const spinner = ora('Building app ').start();
 
--- a/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
+++ b/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
@@ -4,14 +4,20 @@ import { Android } from '@expo/xdl';
 import ora from 'ora';
 import path from 'path';
 
-type Options = object;
+type Options = {
+  buildVariant: string;
+};
+
+function getGradleTask(buildVariant: string): string {
+  return `install${buildVariant.charAt(0).toUpperCase()}${buildVariant.slice(1)}`;
+}
 
 export default async function buildAndroidClientAsync(
   projectRoot: string,
-  _options: Options
+  options: Options
 ): Promise<void> {
   const devices = await Android.getAllAvailableDevicesAsync();
-  const device = await Android.promptForDeviceAsync(devices);
+  const device = devices.length > 1 ? await Android.promptForDeviceAsync(devices) : devices[0];
   if (!device) {
     return;
   }
@@ -24,7 +30,10 @@ export default async function buildAndroidClientAsync(
     process.platform === 'win32' ? 'gradlew.bat' : 'gradlew'
   );
 
-  await spawnAsync(gradlew, ['installRelease'], { cwd: androidProjectPath, stdio: 'inherit' });
+  await spawnAsync(gradlew, [getGradleTask(options.buildVariant)], {
+    cwd: androidProjectPath,
+    stdio: 'inherit',
+  });
 
   spinner.text = 'Starting the development client...';
   await Android.openProjectAsync({
--- a/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
+++ b/packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
@@ -1,0 +1,37 @@
+import { AndroidConfig } from '@expo/config-plugins';
+import spawnAsync from '@expo/spawn-async';
+import { Android } from '@expo/xdl';
+import ora from 'ora';
+import path from 'path';
+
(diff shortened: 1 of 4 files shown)


packages/expo-cli/src/commands/run/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      },
+    ],
+  });
+  return platform;
+}
+
+export default function (program: Command) {
+  program
+    .command('run [path]')
+    .helpGroup('experimental')
```
EvanBacon, line 48:
```suggestion
    .helpGroup('internal')
```
This will hide the command from `--help`

packages/expo-cli/src/commands/run/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    .description('Build a development client and run it in on a device.')
+    .option('-p --platform <platform>', 'Platform: [android|ios]', /^(android|ios)$/i)
+    .option('--build-variant [name]', '(Android) build variant', 'release')
```
EvanBacon, line 51: hmm if this is a cross platform command, then we should probably attempt to make the arguments cross-platform too. If that's not feasible, then we should split ios and android build commands up.

packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
```diff
@@ -0,0 +1,50 @@
...
+  const device = devices.length > 1 ? await Android.promptForDeviceAsync(devices) : devices[0];
+  if (!device) {
+    return;
+  }
+  const bootedDevice = await Android.attemptToStartEmulatorOrAssertAsync(device);
+  if (!bootedDevice) {
+    return;
+  }
+
+  const spinner = ora('Building app ').start();
```
EvanBacon, line 29: build logs?

packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const androidProjectPath = await AndroidConfig.Paths.getProjectPathOrThrowAsync(projectRoot);
```
EvanBacon, line 31: Maybe add a check to use prebuild here?
```ts
  const androidProjectPath = await AndroidConfig.Paths.getProjectPathOrThrowAsync(projectRoot).catch(() => null);
  // If the project doesn't have native code, prebuild it...
  if (!androidProjectPath) {
    await prebuildAsync(projectRoot, {
      install: true,
      platforms: ['android'],
    } as EjectAsyncOptions);
  }
```

packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
```diff
@@ -0,0 +1,50 @@
...
+import ora from 'ora';
+import path from 'path';
+
+type Options = {
+  buildVariant: string;
+};
+
+function getGradleTask(buildVariant: string): string {
+  return `install${buildVariant.charAt(0).toUpperCase()}${buildVariant.slice(1)}`;
+}
```
EvanBacon, line 13: Add more types and make it explicitly clear what's going on here. Also iirc, people often use the `app:` prefix to target just the module? If there's reasoning to do/not do that, could you add it in a comment.

packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
```diff
@@ -0,0 +1,50 @@
...
+    process.platform === 'win32' ? 'gradlew.bat' : 'gradlew'
+  );
+
+  await spawnAsync(gradlew, [getGradleTask(options.buildVariant)], {
+    cwd: androidProjectPath,
+    stdio: 'inherit',
+  });
+
+  spinner.text = 'Starting the development client...';
+  await Android.openProjectAsync({
```
EvanBacon, line 43: wrap in a try/catch and fail the spinner

packages/expo-cli/src/commands/run/index.ts
fson: The intent was to match the convention used by `eas build`, where the same command builds iOS and Android apps: `eas build --platform ios`, `eas build --platform android` or even `eas build --platform all`.

It would be possible to do `expo run:ios` and `expo run:android`, but this (specifying the platform in the command name) is a pattern we've been trying to avoid in new code.

Cross-plaform arguments: Are you suggesting a new argument that maps to build variant name on Android and build configuration name on iOS?

packages/expo-cli/src/commands/run/buildAndroidClientAsync.ts
fson: Looks like this interferes with them, so I'm going to remove the spinner and use `Log.log` instead.

packages/expo-cli/src/commands/run/index.ts
fson: Renamed to `run:android` for now.  ¯\\\_(ツ)\_/¯

PR #3252, opened 2021-02-20
[xdl] Update node-forge

brentvatne:
# Why

Our current version of node-forge is flagged by npm as a high risk security vulnerability due to prototype pollution. In reality, this is not actually impacting us at all due to how we use the library, but npm will still warn users when they install expo-cli. node-forge@0.10.0 is the first version that fixes this (it's also the latest version at the time of writing). 

Advisory link: https://npmjs.com/advisories/1561 

# How

Bump the version. As far as I can tell we are not impacted by breaking changes.

# Test Plan

I'm not familiar with how to test this code.

### DIFF
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -113,7 +113,7 @@
     "@types/hapi__joi": "^17.1.2",
     "@types/hashids": "^1.0.30",
     "@types/ncp": "^2.0.1",
-    "@types/node-forge": "^0.8.6",
+    "@types/node-forge": "^0.9.7",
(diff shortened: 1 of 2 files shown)


dsokal (approved):
1. Looks good, thanks!
2. I also upgraded the `@types/node-forge`. I let myself push this change to your branch 🙈 
3. There are unit tests for the code that uses `node-forge` - https://github.com/expo/expo-cli/blob/master/packages/xdl/src/__tests__/detach/PKCS12Utils-test.ts - so if they're still passing (and they are!) we're good to merge this :)

Issue #3116, opened 2021-01-23
`expo eject` not working with Expo 40


Nostrus:
## Description

I recently upgraded from Expo 38 to 40, and then tried to do an eject. It needed some extra Gradle configuration (because of memory issues mainly). Managed to get it running eventually, but my previously working app crashed immediately on startup. 

So I reverted all changes, tried again, at one point reinstalled `node_modules`, and couldn't get `expo eject` working anymore. 

## Environment

```
  Expo CLI 4.0.17 environment info:
    System:
      OS: macOS 10.15.7
      Shell: 5.7.1 - /bin/zsh
    Binaries:
      Node: 14.7.0 - /usr/local/bin/node
      Yarn: 1.22.4 - /usr/local/bin/yarn
      npm: 6.14.11 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.9.1 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.3, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
      Android SDK:
        API Levels: 27, 28, 29
        Build Tools: 28.0.3, 29.0.2
        System Images: android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 4.1 AI-201.8743.12.41.6953283
      Xcode: 12.3/12C33 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^40.0.0 => 40.0.1
      react: 16.13.1 => 16.13.1
      react-dom: 16.13.1 => 16.13.1
      react-native: ~0.63.4 => 0.63.4
      react-native-web: ~0.13.12 => 0.13.18
    npmGlobalPackages:
      expo-cli: 4.0.17
    Expo Workflow: bare
```

## Expected Behavior

App working after `expo eject`.

## Observed Behavior

App crashes immediately on startup or `expo eject` breaks. 

It fails at the `iOS config syncing` step with the below error:

```
/usr/local/lib/node_modules/expo-cli/node_modules/@expo/configure-splash-screen/src/ios/pbxproj.ts:45
        throw new Error(`.pbxproj file parsing issue: ${err.message}.`);
              ^
Error: .pbxproj file parsing issue: undefined.
[...]

PR #3256, opened 2021-02-24
[eject] Added prompt to clear malformed projects

EvanBacon:
# Why

- help prevent basic errors like this https://github.com/expo/expo-cli/issues/3116
- Sometimes a user will run `expo eject` then git revert the native changes, effectively leaving the native ios, and android folders in place. 

# How

- Check for some core native files in both ios and android on `expo eject`. Prompt to properly clear the malformed projects before prebuilding the project. 
  - This prompt runs after the git prompt.
  - In non interactive mode, the malformed projects are cleared automatically.

# Test Plan

- in a managed project: `expo eject` -- ensure the malformed prompt isn't shown for standard managed projects.
- in a bare project:
  - `rm -rf ios/yourapp/AppDelegate.m`
    - `expo eject` -- should prompt to clear the native ios folder
      - pressing n should proceed then fail.
      - pressing enter or y will clear the ios folder then proceed correctly.
  - `rm -rf ios/yourapp/AppDelegate.m android/build.gradle`
    - `expo eject` -- should prompt to clear both native folders
      - pressing n should proceed then fail.
      - pressing enter or y will clear the ios and android folders then proceed correctly.

### DIFF
--- a/packages/expo-cli/src/commands/eject/clearNativeFolder.ts
+++ b/packages/expo-cli/src/commands/eject/clearNativeFolder.ts
@@ -1,7 +1,16 @@
-import fs from 'fs-extra';
-import path from 'path';
+import { AndroidConfig, IOSConfig } from '@expo/config-plugins';
+import chalk from 'chalk';
+import program from 'commander';
+import * as fs from 'fs-extra';
+import * as path from 'path';
 
+import Log from '../../log';
+import { confirmAsync } from '../../prompts';
 import * as CreateApp from '../utils/CreateApp';
+
+async function directoryExistsAsync(file: string): Promise<boolean> {
+  return (await fs.stat(file).catch(() => null))?.isDirectory() ?? false;
+}
 
 export async function clearNativeFolder(projectRoot: string, folders: string[]) {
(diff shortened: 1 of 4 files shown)


packages/expo-cli/src/commands/eject/clearNativeFolder.ts
```diff
@@ -13,3 +22,78 @@ export async function clearNativeFolder(projectRoot: string, folders: string[])
...
+      ? `The ${displayPlatforms[0]} and ${displayPlatforms[1]} projects are malformed`
+      : `The ${displayPlatforms[0]} project is malformed`;
+
+  if (
+    // If the process is non-interactive, default to clearing the malformed native project.
+    // This would only happen on re-running eject.
+    program.nonInteractive ||
+    // Prompt to clear the native folders.
+    (await confirmAsync({
+      message: `${message}, would you like to clear before prebuilding?`,
```
brentvatne, line 90:
```suggestion
      message: `${message}, would you like to clear the project files and reinitialize them?`,
```

brentvatne (approved):

PR #3257, opened 2021-02-24
[prebuild] Skip overwriting modules that are symlinked

EvanBacon:
# Why

- Running `expo prebuild` in `expo/apps/native-component-list` doesn't work because it overwrites the `expo-updates` package with a version that isn't symlinked, then it runs the build command which effectively breaks the build. 
- This PR checks to ensure each package that's about to be overwritten is symlinked to ensure custom versions of `'react', 'react-native-unimodules', 'react-native', 'expo-updates'` can still be used.
- Removes the weird logic added to only switch RN versions when the installed version has `expo` in the name. 

# Test Plan

- Added unit tests for the new method.
- Running `expo prebuild` in NCL logs:
  - `› Using symlinked react-native-unimodules, react-native, expo-updates instead of recommended version(s).`
  - The `package.json` also is not modified.
- Running `expo prebuild` in a new `expo init` project logs nothing new, and overwrites all of the recommended packages.

### DIFF
--- a/packages/expo-cli/package.json
+++ b/packages/expo-cli/package.json
@@ -89,6 +89,7 @@
     "dateformat": "3.0.3",
     "env-editor": "^0.4.1",
     "envinfo": "7.5.0",
+    "find-up": "^5.0.0",
     "find-yarn-workspace-root": "~2.0.0",
     "form-data": "^2.3.2",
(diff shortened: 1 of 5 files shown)


brentvatne (approved):
clever

PR #3259, opened 2021-02-24
[cli] Newline after full log output and skip newline before help output

brentvatne:
# Why

The newline before the help prompt on a clear screen was bothering me a bit. One solution is to remove the newline. Another alternative would be to add a newline after it too. I opted for the former, but could be convinced. Also, added a newline after the full help output because it looked a bit off to me without it.

## Before

(you probably need to click on these to see)

group2
group

# Test Plan

Run `expo start` and play with terminal UI to re-create the scenarios from screenshots.


EvanBacon (approved):
heads up, this effects a lot of TerminalUI commands.

PR #3268, opened 2021-03-04
Remove unused detach code from XDL and rename the package

fson:
# Why

A large portion of `@expo/xdl` was no longer used in Expo CLI and Expo Dev Tools, but only by a few packages in the `expo/expo` and `expo/turtle` repos. For these packages, a backwards compatible copy of XDL has been moved to https://github.com/expo/xdl.

The remaining package in this repo has been renamed to `xdl` (the old name of the XDL package we used before) and the Turtle specific code has been removed from it. The plan is to split it further and no longer have a monolithic library like XDL as it exists today. This will make Expo CLI easier to maintain and make Turtle v1 no longer dependent on Expo CLI release processes.

# How

- Renamed @expo/xdl to xdl
- Removed unused modules from the xdl interface and the src/detach folder

# Test Plan

Relying on the TypeScript compiler to tell if any module is missing. To make this more reliable, all the remaining code in XDL is now in TS.


EvanBacon:
Did you validate that the internal commands `expo bundle-assets` and `expo prepare-detached-build` still function?

fson:
I'm not sure how to reliably validate that.

However, the only change I made to `bundle-assets` was this minimal change to make it compile as TypeScript:
```diff
-  export async function bundleAssetsAsync(projectDir, args) {
+  export async function bundleAssetsAsync(projectDir: string, args: any) {
```

As for `prepare-detached-build`, I also removed a if-else branch that only runs in service context (Turtle), as this version will only be used in the CLI:
```diff
- async function prepareDetachedBuildIosAsync(projectDir, args) {
- async function prepareDetachedBuildIosAsync(projectDir: string, args: any) {
   const config = await readNullableConfigJsonAsync(projectDir);
   if (config && config.exp.name !== SERVICE_CONTEXT_PROJECT_NAME) {
     return prepareDetachedUserContextIosAsync(projectDir, config.exp, args);
   } else {
-    return prepareDetachedServiceContextIosAsync(projectDir, args);
+    throw new Error('Service context is not supported in this version of XDL');
   }
 }
```
(I could have left the `prepareDetachedServiceContextIosAsync` in place but removing it fixed a bunch of TS errors that otherwise would have required more extensive changes to fix. Now all of the remaining code in `Detach` is also TS, making it easier to catch problems when refactoring it.)

fson:
I removed the changes that might affect `prepare-detached-build` for now. The plan is to move `prepare-detached-build` to `expotools` and after that we can remove the code used by `prepare-detached-build` from this library.

PR #3271, opened 2021-03-05
Add prompt to recommend users upgrade before ejecting

AdamJNavarro:
# Why

To make for a smoother dev experience when migrating to the Bare workflow, we should recommend projects be running the latest SDK version before ejecting.
# How

Added a string comparison to detect if the project is on the latest sdk version and if it is not, trigger a prompt.

# Test Plan

Ran locally against a project running SDK 36, 37, 39 and 40.

### DIFF
--- a/packages/expo-cli/src/commands/eject.ts
+++ b/packages/expo-cli/src/commands/eject.ts
@@ -13,7 +13,7 @@ import { learnMore } from './utils/TerminalLink';
 
 async function userWantsToEjectWithoutUpgradingAsync() {
   const answer = await confirmAsync({
-    message: `We recommend upgrading to the latest SDK version before ejecting. SDK 37 introduces support for OTA updates and notifications in ejected projects, and includes many features that make ejecting your project easier. Would you like to continue ejecting anyways?`,
+    message: `We recommend upgrading to the latest SDK version before ejecting. It will be more difficult to upgrade your app after ejecting because you will also be responsible for native iOS and Android related upgrade steps. Continue ejecting?`,
   });
 
   return answer;
@@ -30,7 +30,9 @@ export async function actionAsync(
   }
 ) {
   const { exp } = getConfig(projectDir);
-
+  const currentSdkVersionString = exp.sdkVersion!;
+  const latestSdkVersion = await Versions.newestReleasedSdkVersionAsync();
+  const latestSdkVersionString = latestSdkVersion.version;
   if (options.npm) {
     options.packageManager = 'npm';
   }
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eject.ts
```diff
@@ -13,7 +13,7 @@ import { learnMore } from './utils/TerminalLink';
 
 async function userWantsToEjectWithoutUpgradingAsync() {
   const answer = await confirmAsync({
-    message: `We recommend upgrading to the latest SDK version before ejecting. SDK 37 introduces support for OTA updates and notifications in ejected projects, and includes many features that make ejecting your project easier. Would you like to continue ejecting anyways?`,
+    message: `We recommend upgrading to the latest SDK version before ejecting. It will be more difficult to upgrade your app after ejecting because you will also be responsible for native iOS and Android related upgrade steps. Continue ejecting?`,
```
brentvatne, line 16:
```suggestion
    message: `We recommend upgrading to the latest SDK version before proceeding. It will be more difficult to upgrade your app after ejecting because you will also be responsible for native iOS and Android related upgrade steps. Continue ejecting?`,
```

packages/expo-cli/src/commands/eject.ts
```diff
@@ -43,11 +45,21 @@ export async function actionAsync(
...
-    } as EjectAsyncOptions);
+    if (currentSdkVersionString !== latestSdkVersionString) {
+      if (await userWantsToEjectWithoutUpgradingAsync()) {
+        Log.debug('Eject Mode: Latest');
+        await ejectAsync(projectDir, {
+          ...options,
+          platforms: platformsFromPlatform(platform),
+        } as EjectAsyncOptions);
+      }
+    } else {
```
brentvatne, line 56: no need for else branch here. should be more like:

```
if (currentSdkVersionString !== latestSdkVersionString) {
  if (!(await userWantsToEjectWithoutUpgradingAsync()) {
    // abort
  }
}

// eject
```

packages/expo-cli/src/commands/eject.ts
```diff
@@ -13,7 +13,7 @@ import { learnMore } from './utils/TerminalLink';
 
 async function userWantsToEjectWithoutUpgradingAsync() {
```
EvanBacon, line 14: I don't think I fully understand this feature. Many users eject because their project no longer runs in the client and they want to run it without needing to upgrade.

brentvatne: yeah i think in those cases they can ignore it. we wouldn't force them to upgrade but rather strongly encourage them to do so

EvanBacon:
Eject is now deprecated and prebuild is versioned.

Issue #2378, opened 2020-07-20
Prompt users whether they want to init a git project if they are already within a git project


goofiw:
## Description

Expo init command initializes a git repo in the target folder.  When using a mono repo, this ends up being a submodule which is undesired behavior.

## Expected Behavior

Expo init should not initialize a git repo.

## Observed Behavior

I thought I had been commiting and adding files by using git commands in the root directory, and running a clean on submodules permanently deleted all the work I had done in the project. 😢 

## Environment

expo 3.22.1
osx 10.15.6
yarn (with workspaces)

brentvatne:
perhaps we should adjust this to prompt the user whether they want to init a git repo if already inside a git repo. otherwise i think this is a reasonable default to keep

goofiw:
@brentvatne does expo need the git history for some commands/file watchers?

brentvatne:
@goofiw - nope, but most developers use git and this is a convenience for people. additionally, some packages installed through npm depend on git, so it's required to have it on your machine to use expo anyhow

PiXHULA:
Hi! 
I actually did this and ended up with a nested git repo. Could anyone be so kind to explain like I'm five on how to remove the git (I've removed the .git files coming with the "expo init" - command). IE make the folder created just being a part of the git created with my backend project? Thanks!

EvanBacon:
rm -rf .git in the new project folder to delete the git project @PiXHULA

PiXHULA:
@EvanBacon Cheers! I guess you'd have to copy the content of .gitignore to the original one aswell?

saideepesh000:
@brentvatne, @EvanBacon, Hey I am working on this.

kbrandwijk:
See: https://github.com/expo/expo-cli/pull/4121

PR #3272, opened 2021-03-05
feat: Prompt user for git init if in existing git work tree

caseyli:
# Why

This PR addresses Issue #2378.

When using `expo init`, the expo-cli automatically runs `git init` inside the project folder. However, if the user is running the command line tool within an existing Git Working Tree, then `git init` has the effect of creating a submodule. This is generally undesired behaviour. In the situation of running the command line tool within an existing Git Working Tree, we should give the user the option to avoid running `git init`.

# How

Checking whether or not the user was in an existing Git Working tree was actually done already. I just needed to use the result of that check to bring up a prompt to the user. The prompt asks 'y/n' and appropriately runs `git init` or not based on the answer.

# Test Plan

The tests were very manual. To reproduce

- Test on existing git working tree
  1. Create an empty folder, and run `git init`.
  2. Within that folder, run `expo init`
  3. It should prompt you whether or not to run 'git init'.
  4. Check the resulting folder for a `.git` folder.
- Test on fresh folder 
  1. Create an empty folder that is not within a git repository
  2. Within that folder, run `expo init`
  3. It should NOT prompt you whether or not to run 'git init'.
  4. Check the resulting folder for a `.git` folder. It should be there.


packages/expo-cli/src/commands/init.ts
```diff
@@ -362,19 +362,43 @@ export async function initGitRepoAsync(
...
 
-  // not in git tree, so let's init
+  let runGitInit = true;
+
+  if (insideGitTree) {
+    const { answer } = await prompts({
+      type: 'text',
+      name: 'answer',
+      message:
+        "You appear to be creating an Expo app within an existing Git work tree. Do you still want to run 'git init' within your new project folder? [y/n]",
```
EvanBacon, line 387:
```suggestion
        "Project folder is being created inside of an existing Git project. Skip creating another Git tree for the project folder?",
```

packages/expo-cli/src/commands/init.ts
```diff
@@ -362,19 +362,43 @@ export async function initGitRepoAsync(
...
       !flags.silent && Log.warn('Unable to initialize git repo. `git` not in PATH.');
       return false;
     }
   }
 
-  // not in git tree, so let's init
+  let runGitInit = true;
+
+  if (insideGitTree) {
+    const { answer } = await prompts({
```
EvanBacon, line 383: Use helper method `confirmAsync`, see example usage across the repo.

packages/expo-cli/src/commands/init.ts
```diff
@@ -362,19 +362,43 @@ export async function initGitRepoAsync(
...
+    insideGitTree = resultPromise.stdout.trim() === 'true';
   } catch (e) {
     if (e.errno === 'ENOENT') {
       !flags.silent && Log.warn('Unable to initialize git repo. `git` not in PATH.');
       return false;
     }
   }
 
-  // not in git tree, so let's init
+  let runGitInit = true;
```
EvanBacon, line 380:
```suggestion
  let shouldInitializeGitTree = true;
```

packages/expo-cli/src/commands/init.ts
```diff
@@ -362,19 +362,43 @@ export async function initGitRepoAsync(
...
+      initial: 'n',
+    });
+
+    runGitInit = answer.trim() === 'y';
+  }
+
+  if (!runGitInit) {
+    !flags.silent &&
+      insideGitTree &&
+      Log.log('New project is already inside of a git repo, skipping git init.');
```
EvanBacon, line 397: No need for this log because the prompt message/answer will act as a record of operations.

packages/expo-cli/src/commands/init.ts
```diff
@@ -362,19 +362,43 @@ export async function initGitRepoAsync(
...
     }
   }
 
-  // not in git tree, so let's init
+  let runGitInit = true;
+
+  if (insideGitTree) {
+    const { answer } = await prompts({
+      type: 'text',
+      name: 'answer',
```
EvanBacon, line 385: Ensure this works as expected when running `expo init --non-interactive`

caseyli:
@EvanBacon thanks for the review! Will address those changes!

packages/expo-cli/src/commands/init.ts
caseyli: After the change to `confirmAsync`, I tested with `expo init --non-interactive` and it works by defaulting to skipping when within a Git tree, and not skipping otherwise.

brentvatne:
Screen Shot 2021-05-07 at 5 52 06 PM

this prompt position feels a bit awkward. i think we should prompt for this earlier on in the process - maybe before you choose a template even. i think often people will do this without realizing the directory that they're in and that will be a good way to remind them.

it would also be nice to provide a bit more context in the message

```bash
╭─~/code/workspacey/apps ‹master›
╰─$ expod init third-app            
Your project will be initialized inside an existing git repository (/Users/brentvatne/code/workspacey).
? Skip initializing a git repository for third-app? (Y/n)

✔ Choose a template: › blank                 a minimal app as clean as an empty canvas
✔ Downloaded and extracted project files.
🧶 Using Yarn to install packages. Pass --npm to use npm instead.
✔ Installed JavaScript dependencies.

✅ Your project is ready!

To run your project, navigate to the directory and run one of the following yarn commands.

- cd third-app
- yarn start # you can open iOS, Android, or web from here, or run them directly with the commands below.
- yarn android
- yarn ios
```
 
the repo path can be determined with `git rev-parse --show-toplevel`

with these changes i'd be happy to merge this! i think this is a nice feature

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
...
 
-  // not in git tree, so let's init
+  let shouldSkipInitializeGitTree = false;
+
+  if (insideGitTree) {
+    Log.log(
+      `Your project will be initialized inside an existing git repository (${process.cwd()})`
+    );
+    shouldSkipInitializeGitTree =
+      program.nonInteractive ||
```
caseyli, line 394: I figured it would be best to default to `shouldSkipInitializeGitTree = true` for non-interactive mode.

My reasoning is that:

1. It seems like most of the comments would like this as the desired behaviour anyway...
2. ...and for those who don't want this behaviour, it's easy enough for them to `git init` themselves afterwards. Not to mention, if they don't realize this behaviour, they'll very quickly realize it the first time they run a git command within the project root.

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
...
+    const resultPromise = await spawnAsync('git', ['rev-parse', '--is-inside-work-tree'], {
       cwd: root,
     });
-    !flags.silent && Log.log('New project is already inside of a git repo, skipping git init.');
+
+    insideGitTree = resultPromise.stdout.trim() === 'true';
   } catch (e) {
     if (e.errno === 'ENOENT') {
-      !flags.silent && Log.warn('Unable to initialize git repo. `git` not in PATH.');
+      !flags.silent && Log.warn('Unable to check if within existing git repo. `git` not in PATH.');
```
brentvatne, line 382: i think this error message isn't so meaningful for people. it is technically the correct thing based on the function name, but if i saw this warning when initializing a project it would make a lot less sense to me than "Unable to initialize git repo"

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
...
       return false;
     }
   }
 
-  // not in git tree, so let's init
+  let shouldSkipInitializeGitTree = false;
+
+  if (insideGitTree) {
+    Log.log(
+      `Your project will be initialized inside an existing git repository (${process.cwd()})`
```
brentvatne, line 391: this won't point to the root of the git repo. you need to run `git rev-parse --show-toplevel` to get the root of the git repo

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
...
+    );
+    shouldSkipInitializeGitTree =
```
brentvatne, line 393: i would just do

```js
if (program.nonInteractive) {
  return false;
}

return await confirmAsync( ..... );
```

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
   }
 }
 
-export async function initGitRepoAsync(
+async function checkSkipInitializeGitTree(
   root: string,
-  flags: { silent: boolean; commit: boolean } = { silent: false, commit: true }
+  cdPath: string,
```
brentvatne, line 368: should use another name here that is more relevant to the the use of this string in this function. it's the path you want to check.

packages/expo-cli/src/commands/init.ts
```diff
@@ -357,24 +363,47 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
   }
 }
 
-export async function initGitRepoAsync(
+async function checkSkipInitializeGitTree(
```
brentvatne, line 366: i would split this function out into two pieces and then move it into the action function

```js
let shouldSkipInitializeGitTree = false;
try {
  if (await isInsideGitRepo(cdPath)) {
    // code log "your project will be initialized within a ... blah blah"
    shouldSkipInitializeGitTree = // code to prompt for skipping
  }
} catch(e) {
  if (e.code === 'GIT_NOT_INSTALLED') { // throw an error with this code from isInsideGitRepo
    // `git` not found in PATH, skipping initializing a git repository for project.
    shouldSkipInitializeGitTree = true;
  } else {
    // unexpected exception, rethrow
    throw e;
  }
}
```

brentvatne: the problem with the `checkSkipInitializeGitTree` is it is doing too much. we really just want a function that checks if we're in git repo or not, and then we can prompt from action function. it's often cleaner to have prompts in the action than chase them around between different helper methods

brentvatne (changes requested):

packages/expo-cli/src/commands/init.ts
caseyli: @brentvatne thanks for the suggestions. Working on the changes right now. That being said, I'm running into a little snag and wondering if you have any recommendations. 

So it looks like we're importing @types/node for some of our TS definitions, and specifically `Error`.

However, that definition of `Error` does not include the `code` property: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53034

So I'm mulling over alternatives:

- Create a custom Error and throw that?
- Find a way around the TS definition?
- Other solutions?

Curious if you have any thoughts or recommendations? Thanks!

brentvatne: we define our own error classes in a couple places here:

- https://github.com/expo/expo-cli/blob/522b989a1c5bf6cb88ecdd7330dd50ae4fc99857/packages/expo-cli/src/CommandError.ts
- https://github.com/expo/expo-cli/blob/522b989a1c5bf6cb88ecdd7330dd50ae4fc99857/packages/json-file/src/JsonFileError.ts

we could do the same sort of thing for this error

caseyli: Thanks @brentvatne - much appreciated 🙏. I'll take a look.

EvanBacon:
Related https://github.com/expo/expo-cli/pull/4121

EvanBacon:
expo init is now deprecated in favor of create-expo-app

PR #3278, opened 2021-03-09
[expo-cli] skip updating dependencies in package.json in `expo prebuild`

wkozyra95:
# Why

We need a way to preserve react-native version when building on EAS

# How

- add param to `expo prebuild`
- use comma separated list because variadic params are only supported in newer commander and updating commander version would break a lot of hacks we are using in cli

# Test Plan

`expo prebuild`
add tests

### DIFF
--- a/packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts
+++ b/packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts
@@ -44,7 +44,7 @@ export async function createNativeProjectsFromTemplateAsync({
   pkg: PackageJSONConfig;
   tempDir: string;
   platforms: ModPlatform[];
-  skipDependencyUpdate: string[];
+  skipDependencyUpdate?: string[];
 }): Promise<
   { hasNewProjectFiles: boolean; needsPodInstall: boolean } & DependenciesModificationResults
 > {
--- a/packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts
+++ b/packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts
@@ -37,12 +37,14 @@ export async function createNativeProjectsFromTemplateAsync({
   pkg,
   tempDir,
   platforms,
+  skipDependencyUpdate,
 }: {
   projectRoot: string;
   exp: ExpoConfig;
   pkg: PackageJSONConfig;
   tempDir: string;
   platforms: ModPlatform[];
+  skipDependencyUpdate: string[];
 }): Promise<
   { hasNewProjectFiles: boolean; needsPodInstall: boolean } & DependenciesModificationResults
 > {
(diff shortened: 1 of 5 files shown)


brentvatne (approved):

brentvatne:
could you add test coverage for this?

brentvatne:
let's also wait for @EvanBacon to approve this before merging

packages/expo-cli/src/commands/eject/updatePackageJson.ts
```diff
@@ -19,10 +19,12 @@ export async function updatePackageJSONAsync({
   projectRoot,
   tempDir,
   pkg,
+  skipDependencyUpdate,
 }: {
   projectRoot: string;
   tempDir: string;
   pkg: PackageJSONConfig;
+  skipDependencyUpdate: string[];
```
EvanBacon, line 27:
```suggestion
  skipDependencyUpdate?: string[];
```

packages/expo-cli/src/commands/eject/updatePackageJson.ts
```diff
@@ -72,10 +79,12 @@ function updatePackageJSONDependencies({
...
```
EvanBacon, line 87:
```suggestion
  skipDependencyUpdate?: string[];
```

packages/expo-cli/src/commands/prebuild.ts
```diff
@@ -46,5 +49,9 @@ export default function (program: Command) {
     .option('--no-install', 'Skip installing npm packages and CocoaPods.')
     .option('--npm', 'Use npm to install dependencies. (default when Yarn is not installed)')
     .option('-p, --platform [platform]', 'Platforms to sync: ios, android, all. Default: all')
+    .option(
+      '--skip-dependency-update [dependencies]',
```
EvanBacon, line 53:
```suggestion
      '--skip-dependency-update <dependencies>',
```

EvanBacon: i.e. it cannot be optional if it's defined.

packages/expo-cli/src/commands/prebuild.ts
```diff
@@ -46,5 +49,9 @@ export default function (program: Command) {
...
+      'Preserves versions of listed packages in package.json (comma separated list)'
```
EvanBacon, line 54: Do we use comma separated anywhere else? Does anyone have a preference towards collected values (`--skip-dependency-update react-native --skip-dependency-update expo-updates`)?

wkozyra95: > Do we use comma separated anywhere else?

No

> Does anyone have a preference towards collected values

It's the first thing I tried, but something like that is not supported in version of commander we are currently using and based on previous attempts updating commander version is not that easy.

packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts
```diff
@@ -37,12 +37,14 @@ export async function createNativeProjectsFromTemplateAsync({
...
   tempDir,
   platforms,
+  skipDependencyUpdate,
 }: {
   projectRoot: string;
   exp: ExpoConfig;
   pkg: PackageJSONConfig;
   tempDir: string;
   platforms: ModPlatform[];
+  skipDependencyUpdate: string[];
```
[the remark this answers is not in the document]
wkozyra95: sure I can add that, but with the current code I'm passing `skipDependencyUpdate ?? []` at a certain point so the rest of the code does does not need to handle optional values

EvanBacon (approved):
Lgtm

PR #3282, opened 2021-03-10
[Metro] Created a custom metro babel transformer

EvanBacon:
# Why

Most projects may never need a modified babel configuration, but they have the "ejected" babel.config.js file by default. This PR aims to simplify the minimum requirements needed to properly run an Expo project:

```diff
App.js
package.json
- babel.config.js
```

The package `expo` ships with `babel-preset-expo`, this preset extends the default RN transformer and adds support for icons, decorators, web, and a few other features used in the Expo SDK. We already do this in `expo/webpack-config`, `babel.config.js` isn't already defined in the root we fall back on `babel-preset-expo`, this is akin to other web tooling (and `react-native`).

# How

- Created a custom transformer file in the default `expo/metro-config`, this transformer utilizes a feature I merged into Metro almost 2 years ago https://github.com/facebook/metro/pull/460 for this exact purpose. 
- [ ] Ensure this doesn't break other tools like RNSVG or [vue-native](https://vue-native.io/docs/installation.html)

# Test Plan

- `expo init`
- delete `babel.config.js`
- Modify the `App.js`:
```tsx
import React from "react";
import { View, Text } from "react-native";

function deco(wrapped) {
  return wrapped;
}

export default class Foo extends React.Component {
  @deco
  render() {
    return (
      <View
        style={{
          backgroundColor: "#4630eb",
          flex: 1,
          justifyContent: "center",
          alignItems: "center",
        }}
      >
        <Text style={{ color: "white", fontWeight: "bold", fontSize: 24 }}>
          Expo CLI
        </Text>
      </View>
    );
  }
}
```
- Run `expo start -c`, and open on iOS or Android.

Without this PR, you'd get:
```log
Failed building JavaScript bundle.
SyntaxError: /Users/evanbacon/Documents/GitHub/lab/yolo42/App.js: Support for the experimental syntax 'decorators-legacy' isn't currently enabled (13:3):

  11 | class Foo extends React.Component {
  12 | 
> 13 |   @deco
     |   ^
  14 |   render() {
  15 |     return ( <View
  16 |       style={{
```

With this PR, the app compiles as expected. This is because babel-preset-expo ships with decorator support for some reason. You can also run this with `expo web` to see that web already supports the fallback.

**Overwriting**

To test that overwriting the default config works:

- create a new `babel.config.js`
```js
module.exports = function(api) {
    api.cache(true);
    return {
      presets: [],
    };
};  
```
- `expo start -c` open on iOS, the decorator should fail to transpile.
  - Running with `EXPO_DEBUG` will print the babel config path that's in use.


packages/metro-config/src/metro-expo-babel-transformer.ts
```diff
@@ -0,0 +1,69 @@
...
+function getCacheKey(): string {
+  const key = crypto.createHash('md5');
+  cacheKeyParts.forEach(part => key.update(part));
+  return key.digest('hex');
+}
+
+module.exports = {
+  getCacheKey,
+  transform,
+};
```
fson, line 69: Looks like this module is not used for anything: https://github.com/expo/expo-cli/blob/e2a3469ebe5351ee0d4dd34da0ad53ebafd45720/packages/metro-config/transformer.js

Could you delete it to avoid these getting mixed up?

PR #3293, opened 2021-03-17
Added debug simctl logger

EvanBacon:
# Why

- https://exponent-internal.slack.com/archives/C5ERY0TAR/p1615864760007800

# How

- Automatically attach logger when an app is launched from the terminal. Only attach when using devClient.

# Test Plan

- `expo start --dev-client --ios`
```tsx
import * as React from "react";
import { View, Text } from "react-native";
import { usePermissions, CAMERA } from 'expo-permissions'

export default function App() {
  // Force a crash
  usePermissions(CAMERA, { ask: true })
  return (
    <View />
  );
}

``` 

- `expo start --ios` no extra logs should be shown.


lukmccall (approved):
The code looks good 💪

PR #3297, opened 2021-03-17
[expo-cli] Change template project default branch from master to main

fson:
# Why

[New repositiories on GitHub use `main` as the default branch name](https://github.com/github/renaming#new-repositories-use-main-as-the-default-branch-name)

# How

Added `git branch -M main` after the initial commit in `expo init`.

# Test Plan

Enabled the previously disabled end-to-end test (now with `--no-install` so it won't time out like before, which was the original reason for disabling it, IIRC). Added an assertion that checks that the branch name after running `expo init` is correct.

Additionally, tested running `expo init` manually and checked the branch name using `git status` once finished.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,6 +71,7 @@ jobs:
         # run: cd packages/${{ matrix.package }} && yarn test --coverage
         env:
           CI: true
+          EXPO_DEBUG: true
       # - name: Get Test Flag Name
       #   id: cov-flag-name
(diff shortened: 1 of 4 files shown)


brentvatne (approved):

brentvatne:
webpack failure seems unrelated but we should investigate the node 10 failure before merging

fson:
The tests were failing because initializing the git repo didn't work in CI. This was the problem:
```
Author identity unknown

*** Please tell me who you are.

Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.
```
I'm setting this in the test case now, so it should work.

We didn't previously test if `git init` worked and that feature is set up to ignore any errors.

PR #3323, opened 2021-03-23
[cli] Added expo config command

EvanBacon:
# Why

- help debug projects, it's also common for tools that have a config to provide a command that prints the config like jest, and tsc

# How

- Added `expo config` with flags:
  - `type <type>`
    - `prebuild` show the modded config used in eject and prebuild (this doesn't eval the mods so the native projects are unchanged)
    - `public` shows the config that's published publicly in the app.
  - `full` shows the entire project config including the package.json, static and dynamic config paths, etc.
- Sort the config so it pushes all objects to the end of the list, this helps to prevent property crowding.

# Test Plan

- Run `expo config` in a project
- Run `expo config --type prebuild` in a bare and managed project
- Run `expo config --type public` deletes `hooks`, `_internal`, `ios.config`, and `android.config`
- Run `expo config --type invalid` throws an error
- Run `expo config --type` throws an error
- Run `expo config --full` prints all info (in NCL) you should see:
```
dynamicConfigObjectType: 'function',
dynamicConfigPath: '/Users/evanbacon/Documents/GitHub/expo/apps/native-component-list/app.config.js',
staticConfigPath: '/Users/evanbacon/Documents/GitHub/expo/apps/native-component-list/app.json',
```


packages/expo-cli/src/commands/config/config.ts
```diff
@@ -0,0 +1,46 @@
...
+  Log.log();
+}
+
+export default function (program: Command) {
+  program
+    .command('config [path]')
+    .description('Show the project config')
+    .helpGroup('info')
+    .option('--modded', 'Print the modded config used in prebuild')
+    .option('--manifest', 'Print the public config used for publishing')
```
ide, line 43: Manifests and configs are separate concepts. A published app config is different than the manifest.

packages/expo-cli/src/commands/eject/configureProjectAsync.ts
```diff
@@ -93,37 +98,77 @@ export default async function configureManagedProjectAsync({
...
-      package: packageName,
+      package: config.android!.package,
     });
   }
 
+  return { exp: config, ...rest };
+}
+
+export function logConfig(config: ExpoConfig | ProjectConfig) {
+  const isObjStr = (str: string): boolean => !!str.match(/^\w+: {/g);
```
ide, line 123:
```suggestion
  const isObjStr = (str: string): boolean => str.test(/^\w+: {/g);
```

PR #3319, opened 2021-03-23
fix: Fixes a bug where an IntentFilter would be removed by the Scheme plugin

fallaciousreasoning:
# Why

Bug #3316

Currently, it is impossible to setup linking/deep linking in a managed expo project on Android, because https schemes are always removed by the scheme plugin.

# How

I was trying to add linking to my Android app, but the generated AndroidManifest was incorrect. The bug is that the `Scheme` plugin removes all intent filters with a scheme before adding it's own custom schemes. The solution to this is to run the intentFilters plugin after the scheme plugin.

# Test Plan

1. Clone the test project https://github.com/fallaciousreasoning/expo-broken-schemes
2. `npm install`
3. Run `expo prebuild --no-install` and inspect the generated `AndroidManifest.xml` in `android/app/src/main`.

If using a current version of `expo-cli` the generated output will be missing a `<data android:scheme="https"/>` element. If using a version with this fix it should include `<data android:scheme="https" android:host="example.com" android:pathPrefix="/"/>`.

Broken `AndroidManifest.xml`
```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example">
  <uses-permission android:name="android.permission.INTERNET"/>
  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
  <uses-permission android:name="android.permission.READ_PHONE_STATE"/>
  <uses-permission android:name="android.permission.USE_FINGERPRINT"/>
  <uses-permission android:name="android.permission.USE_BIOMETRIC"/>
  <uses-permission android:name="android.permission.VIBRATE"/>
  <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS"/>
  <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
  <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
  <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
  <uses-permission android:name="android.permission.CAMERA"/>
  <uses-permission android:name="android.permission.RECORD_AUDIO"/>
  <uses-permission android:name="android.permission.READ_CONTACTS"/>
  <uses-permission android:name="android.permission.WRITE_CONTACTS"/>
  <uses-permission android:name="android.permission.READ_CALENDAR"/>
  <uses-permission android:name="android.permission.WRITE_CALENDAR"/>
[...]

PR #3324, opened 2021-03-23
Prevent removing existing schemes from AndroidManifest.xml

EvanBacon:
# Why

- Related PR https://github.com/expo/expo-cli/pull/3319

This plugin used to remove the unused schemes but this is unpredictable because other plugins could add schemes.
The only way to reliably remove schemes from the project is to nuke the file and regenerate the code (`expo prebuild --clean`).
Regardless, having extra schemes isn't a fatal issue and therefore a tolerable compromise is to just add new schemes that aren't currently present.

# Test Plan

- `expo prebuild -p android --no-install` in a project with a scheme defined.
- Remove the scheme and rerun `expo prebuild -p android --no-install` the scheme will no longer be removed.
- Manually add a new scheme to the `AndroidManifest.xml` (emulates an unknown plugin), rerun `expo prebuild -p android --no-install` the scheme will not have the manually defined scheme removed.

### DIFF
--- a/packages/config-plugins/src/android/Scheme.ts
+++ b/packages/config-plugins/src/android/Scheme.ts
@@ -23,20 +23,23 @@ export function getScheme(config: { scheme?: string | string[] }): string[] {
   return [];
 }
 
+// This plugin used to remove the unused schemes but this is unpredictable because other plugins could add schemes.
+// The only way to reliably remove schemes from the project is to nuke the file and regenerate the code (`expo prebuild --clean`).
+// Regardless, having extra schemes isn't a fatal issue and therefore a tolerable compromise is to just add new schemes that aren't currently present.
 export function setScheme(
   config: Pick<ExpoConfig, 'scheme' | 'android'>,
   androidManifest: AndroidManifest
 ) {
-  const scheme = [
+  const schemes = [
     ...getScheme(config),
     // @ts-ignore: TODO: android.scheme is an unreleased -- harder to add to turtle v1.
     ...getScheme(config.android ?? {}),
   ];
   // Add the package name to the list of schemes for easier Google auth and parity with Turtle v1.
   if (config.android?.package) {
-    scheme.push(config.android.package);
+    schemes.push(config.android.package);
   }
-  if (scheme.length === 0) {
+  if (schemes.length === 0) {
     return androidManifest;
   }
 
@@ -48,14 +51,15 @@ export function setScheme(
     return androidManifest;
   }
 
-  // Get the current schemes and remove them.
+  // Get the current schemes and remove them from the list of schemes to add.
   const currentSchemes = getSchemesFromManifest(androidManifest);
   for (const uri of currentSchemes) {
-    androidManifest = removeScheme(uri, androidManifest);
+    const index = schemes.indexOf(uri);
+    if (index > -1) schemes.splice(index, 1);
   }
 
-  // Now add all the new schemes.
-  for (const uri of scheme) {
+  // Now add all of the remaining schemes.
+  for (const uri of schemes) {
     androidManifest = appendScheme(uri, androidManifest);
   }
 


packages/config-plugins/src/android/Scheme.ts
```diff
@@ -48,14 +51,15 @@ export function setScheme(
     return androidManifest;
   }
 
-  // Get the current schemes and remove them.
+  // Get the current schemes and remove them from the list of schemes to add.
   const currentSchemes = getSchemesFromManifest(androidManifest);
   for (const uri of currentSchemes) {
```
fallaciousreasoning, line 56: So, I'm not an expert in the code base, but if I'm reading this right this is basically: "Don't add schemes that are already in the manifest". This means that existing schemes won't get updated (I don't even know if schemes can be updated, so this may be a non-issue).

What if we removed all schemes that the scheme plugin is generating and then regenerated them. I think it would have a similar effect, with the added bonus that we could update existing schemes (if that's even a thing that's possible).

```suggestion
  for (const uri of schemes) {
      androidManifest = removeScheme(uri, androidManifest);
```

fallaciousreasoning: So instead of `Don't add existing schemes` we would just update schemes the `Scheme` plugin knows about. In addition, it would be interesting to see if we could handle this deduplication login in the `IntentFilters` plugin (which I'm pretty sure this depends on).

EvanBacon: In the **future** when a critical mass of packages in the ecosystem have their own plugin, we can support two different workflows:

1. bare workflow: you prebuild/eject once and make custom modifications to your `ios`, and `android` folders. Of course, most plugins aren't very destructive and can allow for users to re-run prebuild, but this is just a convenient side-effect and YMMV.
2. a new "custom" managed workflow: you `prebuild` your app as many times as you want and use it to add and customize native features using a unified TS/JS/JSON interface, in this workflow, the `ios`, and `android` folders are treated as disposable generated folders which can have files reset between prebuilds (this is when existing schemes would be removed) think of it kinda like a bundler for native projects. This workflow was designed mainly for more robust managed workflow cloud builds with [eas build](https://docs.expo.io/eas/) (to supersede `expo build`) so rerunning prebuild locally today isn't efficient enough to create a very pleasant developer experience. 

With this in mind, plugins should only add and modify content, they should not try to clean up modifications made in previous runs. This rule allows plugins with overlapping functionality to cooperate.

For example; if you have a couple authentication plugins which append custom schemes, they can create a combined list of schemes. But if just one of those plugins attempts to remove schemes that it assumes got added in a previous run, it could potentially break the other plugins and cause unexpected behavior.

fallaciousreasoning: We're actually doing something somewhere in between the two options (possibly closer to 2) in a project at the moment. We have one project but based on some config files and environment variables, we can build either of two similar apps.

Unfortunately, we have to keep the generated Android folder around in order to configure our signing key, so we don't purely do option two, which would be much more comfortable.

Sorry for the off topic ramble - but it's pretty cool what you can do with expo :)

> For example; if you have a couple authentication plugins which append custom schemes, they can create a combined list of schemes. But if just one of those plugins attempts to remove schemes that it assumes got added in a previous run, it could potentially break the other plugins and cause unexpected behavior.

I completely agree with this. However, my concern (which may not be valid) is that the Scheme plugin won't update the manifest at all, if you change something in the config. 

Say, for example, the expected output of the plugin this run is
```xml
        <data android:scheme="com.example" some-other:thing="stuff" />
```

but last run it had generated
```xml
        <data android:scheme="com.example" />
```

The manifest wouldn't be updated at all. In this case (it would see that the scheme already exists on the Manifest, and skip generating it). Perhaps this is desired behavior (perhaps it was the user who removed the `some-other:thing` attribute), but I'd argue that as this plugin is responsible for generating the `com.example` scheme, it's safe for the plugin to update it.

This behavior would be in between the old behavior (nuke all schemes) and  the behavior in the PR (don't touch any schemes that already exist). Instead the behavior would be to nuke all schemes that we are going to generate (ensuring they are up to date). This wouldn't affect schemes generated by other plugins, unless they were using the same scheme as something that the scheme plugin is generating (in which case I suspect there will be other problems).

Sorry for the wall of test 😆

EvanBacon: Thanks for the insight, it's cool to hear that you're using a similar approach to what we've been working towards! The `withScheme` plugin's job is to take `scheme` from the Expo config and ensure the Android app conforms to them. I think keeping the plugin functionality simple (i.e. no extra attributes unless they're defined in the ExpoConfig schema). 

To customize the schemes more, you could create a new mod plugin that adds more custom functionality by passing it custom properties in the app.config.js:

`app.config.js`
```ts
export default {
  plugins: [
    [
      withCustomSchemes, [
      { scheme: 'com.example', 'some-other:thing': 'stuff' },
      ] 
    ]
  ]
}
```

`withCustomSchemes.js`
```js
const {withAndroidManifest} = require('@expo/config-plugins');
module.exports = (config, customSchemes) => {
  return withAndroidManifest(config, config => {
    // Modify config.modResults as an AndroidManifest
    return config;
  }) 
}
```

PR #3334, opened 2021-03-26
[xdl] remove gulp and build with tsc

fson:
# Why

The build setup for XDL has been different from all other packages in this repo:
- using Gulp
- using Babel instead of `tsc`
- relying on the use of the `lazy` option of `@babel/plugin-transform-modules-commonjs` to force all modules to load lazily

To make this project easier to maintain and make optimizations to in the future, this PR changes the build setup of XDL to the more standard way of using `tsc` to build TypeScript and removes the unnecessary Gulp dependency. Also the final non-TypeScript module in XDL is converted to TypeScript.

# How

- Switching to Babel was not trivial because of circular dependencies that were tolerated when using the lazy loading with Babel.
- To tackle the circular dependency issues I introduced the `internal.ts` source file that loads all modules of XDL and is used for all internal imports of XDL modules within XDL. (More about this technique here: [How to fix nasty circular dependency issues once and for all in JavaScript & TypeScript](https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de)
- Converted `packages/xdl/src/detach/IosCodeSigning.js` to TS after removing all functions in it that were no longer used anywhere
- `scripts/updateCaches.js` is now a plain JS script (previously `gulp caches`)

# Test Plan

- Make sure all automated tests pass
- Manually test using Expo CLI to develop a project


packages/xdl/src/webpack-utils/WebpackCompiler.ts
```diff
@@ -93,7 +93,7 @@ export function printPreviewNotice(projectRoot: string, showInDevtools: boolean)
   );
 }
 
-export default function createWebpackCompiler({
+export function createWebpackCompiler({
```
EvanBacon, line 96: seems like a lot of this import refactor can be done in another PR without needing a review.

packages/xdl/package.json
```diff
@@ -12,11 +12,11 @@
   ],
   "scripts": {
     "start": "yarn run prepare && yarn run watch",
-    "build": "tsc --emitDeclarationOnly && gulp build",
+    "build": "tsc",
```
EvanBacon, line 15: We use [babel + lazy loading in expo-cli](https://github.com/expo/expo-cli/blob/7c40ae0a6ad0c59c707c9c167395c621c1adab32/packages/expo-cli/package.json#L13), have you validated that the approach hasn't increased execution time?

fson: I did profile the time it takes to import each module imported in `internal.ts`. The only module stands out is `Android.ts`, but the reason is that it's the first module that imports `@expo/config`.
Screen Shot 2021-03-26 at 15 53 29

`@expo/config` is really slow to import, but since practically any command requires it, lazy loading would only move the slowdown to a different point in the command lifecycle.
Screen Shot 2021-04-01 at 12 01 15

packages/xdl/src/webpack-utils/WebpackCompiler.ts
fson: I renamed this export so that I can re-export it in `internal` like this:
```ts
export * as WebpackCompiler from './webpack-utils/WebpackCompiler';
```
(I could have re-exported individual functions, but I didn't want to add `printInstructions`, `printSuccessMessages` etc. to `internal` without the `WebpackCompiler` namespace because otherwise it would be harder to see where they come from. I hope that makes sense.)

fson:
I've converted the imports in tests, fixed the import/export order (`Project.ts` needs to be loaded after the modules it re-exports), and did a final check to ensure no function/constant remains undefined after loading, using this snippet in the repl:
```js
var xdl = require('xdl');
Object.keys(xdl).forEach(mod => {
  Object.keys(xdl[mod]).forEach(field => {
    if (xdl[mod][field] == null) {
      console.log(mod, field);
    }
  });
});
```

packages/xdl/package.json
fson: Almost all of the `@expo/config` slowdown seems to come from loading `@expo/config-plugins`, which is really heavy:
Screen Shot 2021-04-01 at 17 56 26

PR #3338, opened 2021-03-26
Run built APK on device

EvanBacon:
# Why

- `run:android` currently just opens the Expo Go app after it finishes building (and on an unrelated device no less). This PR adds the ability to install, and open an arbitrary APK, then open it. 
- Also abstracted some functionality out of `Android` since it's getting large. 
- Added a generalized caching system for reading Android device properties.

# Test Plan

- `expo run:android` should work as expected, building an APK then installing _that_ APK and opening it on the selected device.

### DIFF
--- a/packages/expo-cli/src/commands/run/android/runAndroid.ts
+++ b/packages/expo-cli/src/commands/run/android/runAndroid.ts
@@ -1,16 +1,26 @@
 import { AndroidConfig } from '@expo/config-plugins';
+import fs from 'fs';
+import path from 'path';
 import { Android } from 'xdl';
 
 import CommandError from '../../../CommandError';
 import Log from '../../../log';
-import { ora } from '../../../utils/ora';
 import { prebuildAsync } from '../../eject/prebuildAsync';
-import * as Gradle from './Gradle';
 import { resolveDeviceAsync } from './resolveDeviceAsync';
+import { spawnGradleAsync } from './spawnGradleAsync';
 
 type Options = {
-  buildVariant: string;
+  variant: string;
   device?: boolean | string;
+};
+
+export type AndroidRunOptions = Omit<Options, 'device'> & {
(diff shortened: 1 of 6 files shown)


packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -26,28 +36,88 @@ async function resolveAndroidProjectPathAsync(projectRoot: string): Promise<stri
...
 
   const androidProjectPath = await resolveAndroidProjectPathAsync(projectRoot);
 
-  await Gradle.spawnGradleTaskAsync({ androidProjectPath, buildVariant: options.buildVariant });
+  await spawnGradleAsync({ androidProjectPath, variant: options.variant });
+
+  const apkFile = await getInstallApkNameAsync(props.device, props);
+  Log.debug(`Installing: ${apkFile}`);
+  const binaryPath = path.join(props.apkVariantDirectory, apkFile);
+  await Android.installOnDeviceAsync(props.device, { binaryPath });
```
fson, line 89: If we're going to use `adb install` to get the APK installed on the device, maybe switch to a different Gradle task that doesn't install the APK? (Now we run `gradlew installDebug`/`gradlew installRelease` (https://developer.android.com/studio/build/building-cmdline#build_apk) to build and install the app, followed by `adb install`, which also installs the app.)

packages/xdl/src/Android.ts
```diff
@@ -295,9 +292,35 @@ export async function getAdbOutputAsync(args: string[]): Promise<string> {
     _isAdbOwner = alreadyRunning === false;
   }
 
+  if (isDebug()) {
+    Logger.global.info([adb, ...args].join(' '));
+  }
```
fson, line 297:
```suggestion
Logger.global.debug([adb, ...args].join(' '));
```

PR #3342, opened 2021-03-30
Add migration from @react-native-community/async-storage

fson:
# Why

`@react-native-community/async-storage` has been renamed to `@react-native-async-storage/async-storage`. The old `@react-native-community/async-storage` package will not work with SDK 41. 

# How

- Added the codemod `sdk41-async-storage`.
- Added a check in `expo doctor`.

# Test Plan

- Unit tests for codemod.
- Manually tested the doctor command on a project with/without this dependency.

Screen Shot 2021-03-30 at 18 10 20

### DIFF
--- a/packages/xdl/src/project/Doctor.ts
+++ b/packages/xdl/src/project/Doctor.ts
@@ -301,7 +301,10 @@ async function _validateReactNativeVersionAsync(
     }
     ProjectUtils.clearNotification(projectRoot, 'doctor-no-react-native-in-package-json');
 
-    if (pkg.dependencies?.['@react-native-community/async-storage']) {
+    if (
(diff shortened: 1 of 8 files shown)


packages/xdl/src/project/Doctor.ts
```diff
@@ -301,6 +301,18 @@ async function _validateReactNativeVersionAsync(
     }
     ProjectUtils.clearNotification(projectRoot, 'doctor-no-react-native-in-package-json');
 
+    if (pkg.dependencies?.['@react-native-community/async-storage']) {
```
brentvatne, line 304: this should only apply to sdk 41 projects

brentvatne (changes requested):
need to gate async storage pkg check to sdk 41, otherwise lgtm

PR #3343, opened 2021-03-31
[config-plugins] Created unversioned react-native-maps plugin

EvanBacon:
# Why

- auto configure react-native-maps upon ejecting.

# How

- created an unversioned plugin for `react-native-maps` namespace which:
  - adds `pod 'react-native-google-maps', path: '../node_modules/react-native-maps'` to the podfile if `ios.config.googleMapsApiKey` is defined.
  - adds the following blocks to the AppDelegate.m when Google Maps is enabled:
```objc
// @generated begin react-native-maps-import - expo prebuild (DO NOT MODIFY) sync-f2f83125c99c0d74b42a2612947510c4e08c423a
#if __has_include(<GoogleMaps/GoogleMaps.h>)
#import <GoogleMaps/GoogleMaps.h>
#endif
// @generated end react-native-maps-import

...

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{

// @generated begin react-native-maps-init - expo prebuild (DO NOT MODIFY) sync-153e3eb24b7da99b846df4d93dd1d33da90e7c22
#if __has_include(<GoogleMaps/GoogleMaps.h>)
  [GMSServices provideAPIKey:@"API_KEY"];
#endif
// @generated end react-native-maps-init
```
These blocks are special because they can be removed or updated between prebuilds (plugin anti pattern). The `react-native-maps-init` block is matched against RN and Unimodules projects using the following (tested) regex `/(?:(self\.|_)(\w+)\s?=\s?\[\[UMModuleRegistryAdapter alloc\])|(?:RCTBridge\s?\*\s?(\w+)\s?=\s?\[\[RCTBridge alloc\])/g`

- The Android permissions `android.permission.ACCESS_COARSE_LOCATION` and `android.permission.ACCESS_FINE_LOCATION` are added
- The iOS permission `NSLocationWhenInUseUsageDescription` is added (no customization options)
- The AndroidManifest tags `com.google.android.geo.API_KEY` and `org.apache.http.legacy` are added, this was previously done directly but the plugin has now been moved into the unversioned `react-native-maps` plugin (meaning it can be disabled by a future plugin in the `react-native-maps` package).
- Finally, `GMSApiKey` is added to the Info.plist (not sure why though).

# Test Plan

- Extensive regex testing is done in the unit tests
- In a new project
- Install `yarn add react-native-maps`
- Sync `expod prebuild`
- Build `expod run:ios --bundler` -- this works as expected for Apple maps
- Define a value in `ios.config.googleMapsApiKey` 
- Sync `expod prebuild`
- Install `npx pod-install`
- Build `expod run:ios --bundler` -- this enables Google Maps on iOS
- `yarn remove react-native-maps`
- Sync `expod prebuild` -- this removes the related CocoaPods and AppDelegate code safely.

### DIFF
--- a/packages/config-plugins/src/utils/generateCode.ts
+++ b/packages/config-plugins/src/utils/generateCode.ts
@@ -1,0 +1,102 @@
+/**
+ * Get line indexes for the generated section of a file.
+ *
+ * @param src
+ */
+import crypto from 'crypto';
+
+function getGeneratedSectionIndexes(
+  src: string,
+  tag: string
+): { contents: string[]; start: number; end: number } {
+  const contents = src.split('\n');
+  const start = contents.findIndex(line => line.includes(`@generated begin ${tag}`));
+  const end = contents.findIndex(line => line.includes(`@generated end ${tag}`));
+
+  return { contents, start, end };
+}
+
+export type MergeResults = {
+  contents: string;
+  didClear: boolean;
+  didMerge: boolean;
+};
+
+/**
+ * Merge the contents of two files together and add a generated header.
+ *
+ * @param targetContents contents of the existing file
+ * @param sourceContents contents of the extra file
+ */
+export function mergeContents(
+  targetContents: string,
+  sourceContents: string,
+  tag: string,
+  anchor: string | RegExp,
+  offset: number,
+  comment: string
+): MergeResults {
+  const header = createGeneratedHeaderComment(sourceContents, tag, comment);
+  if (!targetContents.includes(header)) {
+    // Ensure the old generated contents are removed.
+    const sanitizedTarget = removeGeneratedContents(targetContents, tag);
+    return {
+      contents: addLines(sanitizedTarget ?? targetContents, anchor, offset, [
+        header,
+        ...sourceContents.split('\n'),
+        `${comment} @generated end ${tag}`,
+      ]),
+      didMerge: true,
+      didClear: !!sanitizedTarget,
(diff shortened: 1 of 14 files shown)


packages/config-plugins/src/utils/generateCode.ts
```diff
@@ -0,0 +1,123 @@
...
+  tag: string,
+  comment: string
+): string {
+  const hashKey = createHash(contents);
+
+  // Everything after the `${tag} ` is unversioned and can be freely modified without breaking changes.
+  return `${comment} @generated begin ${tag} - expo prebuild (DO NOT MODIFY) ${hashKey}`;
+}
+
+export function createHash(gitIgnore: string): string {
```
brentvatne, line 119: `gitIgnore`?

packages/config-plugins/src/utils/generateCode.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return lines.join('\n');
+}
+
+/**
+ * Removes the generated section from a file, returns null when nothing can be removed.
+ * This sways heavily towards not removing lines unless it's certain that modifications were not made to the gitignore manually.
+ *
+ * @param src
```
brentvatne, line 95:
```suggestion
 * @param src
 * @param tag
```

brentvatne (approved):

PR #3356, opened 2021-04-04
remove node 10 support and assert wrapper

EvanBacon:
# Why

- Node 10 dies in 27 days.

# How

- Remove transpilation for node 10
- Remove assert TS wrapper


fson:
Also need to update tested Node versions in `.github/workflows/test.yml` and add a CHANGELOG entry.

Once you've landed this I can update https://github.com/expo/expo-cli/pull/3020, which removes fs-extra.

PR #3359, opened 2021-04-07
Added getUserProjectIdAsync method

EvanBacon:
# Why

- Required for bare projects to embed the `id` property for AuthSession and Notifications.
- If the `expo-constants` package has to depend on `xdl` it'll be pretty unfortunate to download, so we should investigate splitting this functionality out of xdl.

### DIFF
--- a/packages/xdl/src/User.ts
+++ b/packages/xdl/src/User.ts
@@ -297,12 +297,12 @@ export class UserManagerInstance {
    * @param manifest
    * @returns
    */
-  async getUserProjectIdAsync(manifest: ExpoConfig): Promise<string> {
-    const username = await this.getIdUsernameAsync(manifest);
+  async getProjectCurrentFullNameAsync(manifest: ExpoConfig): Promise<string> {
+    const username = await this.getProjectAccountNameAsync(manifest);
     return `@${username}/${manifest.slug}`;
   }
 
-  async getIdUsernameAsync(manifest: ExpoConfig): Promise<string> {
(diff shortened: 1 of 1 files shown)


packages/xdl/src/User.ts
```diff
@@ -289,6 +290,33 @@ export class UserManagerInstance {
...
   }
 
+  /**
+   * Used in expo-constants to generate the `id` property statically for an app in custom managed workflow.
+   * This `id` is used for legacy Expo services AuthSession proxy and Expo notifications device ID.
+   *
+   * @param manifest
+   * @returns
+   */
+  async getUserProjectIdAsync(manifest: ExpoConfig): Promise<string> {
```
ide, line 300:
```suggestion
  async getProjectCurrentFullNameAsync(manifest: ExpoConfig): Promise<string> {
```

packages/xdl/src/User.ts
```diff
@@ -289,6 +290,33 @@ export class UserManagerInstance {
...
+    const username = await this.getIdUsernameAsync(manifest);
+    return `@${username}/${manifest.slug}`;
+  }
+
+  async getIdUsernameAsync(manifest: ExpoConfig): Promise<string> {
```
ide, line 305:
```suggestion
  async getProjectAccountNameAsync(manifest: ExpoConfig): Promise<string> {
```

brentvatne (approved):

ide (approved):

PR #3360, opened 2021-04-07
[xdl] Delete Api module

EvanBacon:
# Why

Seems like it's no longer needed. We need to split the api code out of xdl for use in expo-constants, I noticed that this module isn't used anymore, so deleting it first.


fson (approved):
LGTM 🚀

PR #3361, opened 2021-04-07
Created @expo/api package

EvanBacon:
# Why

- Resolve ENG-2819
- Split modules out of xdl and into a new internal package `@expo/api`. This will enable us to authenticate and sign manifests from `npx expo` without needing to fully depend on xdl which contains deprecated ExpoKit logic.
- Decouples user auth from the rest of xdl.
- Might make sense to combine `Config.offline` with `ConnectionStatus`. 
- Maybe split up `Config` and only keep relevant parts in `expo/api`.
- `Config` -> `SessionSettings` to fit the other settings naming conventions.
- Type all input/output for network requests.

# How

- pending https://github.com/expo/expo-cli/pull/3360

# Test Plan

- Ensure all tests continue to work.
- Split up e2e and unit tests from expo-cli runs.
- Manually run `expo login`, `expo whoami`, `expo logout`, `expo publish` (app with assets), `expo start`, `expo publish:history`.


packages/api/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "!**/__integration_tests__/**"
+  ],
+  "dependencies": {
+    "@expo/json-file": "^8.2.28",
+    "fs-extra": "9.0.0",
+    "uuid": "3.3.2",
+    "internal-ip": "4.3.0",
+    "lodash": "^4.17.19",
+    "getenv": "^1.0.0",
+    "concat-stream": "1.6.2",
```
fson, line 39:
```suggestion
```
No longer used with the `Api` module gone.

packages/api/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "axios": "0.21.1",
+    "analytics-node": "3.5.0",
+    "form-data": "^2.3.2"
+  },
+  "devDependencies": {
+    "@types/fs-extra": "^9.0.1",
+    "@types/uuid": "^3.4.4",
+    "@types/form-data": "^2.2.0",
+    "@types/analytics-node": "^3.1.1",
+    "@types/concat-stream": "^1.6.0"
```
fson, line 49:
```suggestion
    "@types/analytics-node": "^3.1.1"
```
No longer used with the `Api` module gone.

packages/api/src/Analytics.ts
```diff
@@ -1,7 +1,7 @@
 import Segment from 'analytics-node';
```
fson, line 1: Not sure if `Analytics` should be in the `@expo/api` package. I think it makes sense to keep libraries centered around a single purpose. Analytics doesn't interact with the expo.io API and also doesn't have strong interdependencies with this library. Maybe throw it into a separate package or move to `expo-cli`?

packages/expo-cli/src/accounts.ts
```diff
@@ -1,7 +1,7 @@
+import { ApiV2, User, UserManager } from '@expo/api';
 import chalk from 'chalk';
 import program from 'commander';
 import openBrowser from 'react-dev-utils/openBrowser';
-import { ApiV2, User, UserManager } from 'xdl';
 import { ApiV2Error } from 'xdl/build/ApiV2';
```
fson, line 5:
```suggestion
```
❗Need to remove this and the `instanceof` check here: https://github.com/expo/expo-cli/blob/4a9cc924f1dd16fe42a3a1ae343b515fba7c435e/packages/expo-cli/src/accounts.ts#L325

fson (changes requested):

packages/api/src/Analytics.ts
EvanBacon: Can't put it in expo-cli because expo-cli already deps expo/api. Makes sense to split it out but that might be a bit heavy for the single module, prolly want to audit the analytics and see if we can reduce the surface area. We need the api package pretty quickly to get notifications and auth-session working for 41 so it seems reasonable it leave Analytics in api for now and move it later.

fson: Related https://github.com/expo/expo-cli/pull/3372#discussion_r609405158

EvanBacon:
stale

byCedric (commented):
LGTM

PR #3372, opened 2021-04-08
[xdl][expo-cli] Adds dau event to unified project

kgc00:
# Why

Send event type 'action' to the 'Unified' Amplitude project to understand who is actively developing.

# How

Added a second analytics client instance configured to send events to the 'unified' segment source.  Used this client to send events to the 'Unified' project.  Added a call inside `startAsync` to send a `start` action to the unified project.

Analytics instance requires an identified user or it will not send events 
- updated this logic to pull global (analytics) variables into class instances
- added a call in the `startAsync` command to load the user and call `analytics.identify` with the value

# Test Plan

Ran 
- `yarn test` ✅  
- `yarn expo start`✅  : confirmed the event gets sent to the unified project


wkozyra95 (commented):
The naming of the event seems to be inconsistent with names used in other places, make sure to contact Jon Liak before making any changes. For context this docs for eas analytics are here https://docs.google.com/spreadsheets/d/1h0sxk-0uztkQuowssn9jTp8D4hfc5NO1d-9qCxBpe-8/edit?pli=1#gid=407807964 if there is doc like that for expo-cli please send me link it(here or on slack).

The one possibly important note is that we will be migrating from segment and use amplitude directly. eas-cli already does that.

packages/xdl/src/Analytics.ts
```diff
@@ -87,6 +90,15 @@ export class AnalyticsClient {
   }
 }
 
-const defaultClient = new AnalyticsClient();
+class XDLClient extends AnalyticsClient {
```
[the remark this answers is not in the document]
kgc00: So the goal is to _continue_ sending all events to the current XDL segment source, but add an additional segment source for specific kinds of events.  the snippet you linked only writes to one source, so it would not work for this requirement.

packages/xdl/src/start/startAsync.ts
```diff
@@ -49,6 +49,13 @@ export async function startAsync(
   verbose: boolean = true
 ): Promise<ExpoConfig> {
   assertValidProjectRoot(projectRoot);
+  Analytics.unifiedClient.logEvent('dau', {
```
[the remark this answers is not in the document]
kgc00: I checked with Jon and this is how he wants it formatted. I think the reason is that the definition of what a dau is can change over time.  It can include events in other codebases and we want them all listed as a single 'dau' event.

packages/xdl/src/Analytics.ts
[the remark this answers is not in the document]
kgc00: did not pull the second instance out into a new file because i'd have to export the analytics client class & potentially those global variables... It's simpler to just embed it in xdl/analytics, but can make the change if you feel strongly

packages/xdl/src/Analytics.ts
```diff
@@ -16,6 +16,9 @@ export class AnalyticsClient {
   private segmentNodeInstance: Segment | undefined;
   private version: string | undefined;
   private userIdentifyCalled: boolean = false;
+  public get Version() {
+    return this.version;
+  }
```
wkozyra95, line 21: it's not used anywhere

packages/xdl/src/Analytics.ts
```diff
@@ -87,6 +90,9 @@ export class AnalyticsClient {
   }
 }
 
-const defaultClient = new AnalyticsClient();
+const AnalyticsUnified = new AnalyticsClient();
+AnalyticsUnified.setSegmentNodeKey('u4e9dmCiNpwIZTXuyZPOJE7KjCMowdx5');
```
wkozyra95, line 94: maybe lets keep setSegmentNodeKey and setVersionName calls at the same place

wkozyra95 (approved):

packages/xdl/src/Analytics.ts
kgc00: thought you asked to pull them in, sry! i agree they should be the same place

animeeples:
Updated event with `action` lgtm!

wkozyra95 (approved):

packages/xdl/src/Analytics.ts
```diff
@@ -43,10 +47,10 @@ export class AnalyticsClient {
   }
 
   public logEvent(name: string, properties: any = {}) {
-    if (this.segmentNodeInstance && _userId) {
+    if (this.segmentNodeInstance && this._userId) {
       this.ensureUserIdentified();
```
wkozyra95, line 51:
```suggestion
```
If you are calling  setUserProperties for every instance, then this line is not necessary anymore

packages/xdl/src/Analytics.ts
```diff
@@ -9,17 +9,21 @@ const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
...
 
-let _userId: string | undefined;
-let _userTraits: any;
-
 export class AnalyticsClient {
+  private _userId: string | undefined;
+  private _userTraits: any;
   private segmentNodeInstance: Segment | undefined;
   private version: string | undefined;
   private userIdentifyCalled: boolean = false;
```
wkozyra95, line 17:
```suggestion
```

packages/xdl/src/Analytics.ts
```diff
@@ -55,10 +59,10 @@ export class AnalyticsClient {
   }
 
   private ensureUserIdentified() {
```
wkozyra95, line 61: either
- move content of this function to setUserProperties
- pass userId and userTraits as argument (and maybe rename this function to identify or identifyUser)

wkozyra95 (approved):

kgc00:
converted this to draft because i noticed that the userId (regardless of this PR or not) is generally not getting set during the expo start command's execution.  this means that we skip sending the analytics event because the user is not identified.  making a new pr to address this which will have to be merged before i merge this.  

(also, thanks for the feedback Wojciech, I'll be pulling those suggestions in as well)

kgc00:
There is some inconsistency around when the `currentUser` is available which made this PR a bit tricky.  The `currentUser` needs to be available so we can run `analytics.identify` with that information.  I ended up settling on the lowest lift method of calling `getUserAsync` in exactly the place where we need that info (the `startAsync` command). 

I feel that perhaps we should read the user from disk somewhere closer to the application entry point and store that result in a `UserManager.currentUser` property to make that state more widely and easily available.  It would be nice to bootstrap services (like analytics) that need that info in one place close to the app entrypoint.

kgc00:
forgot to add tests- will add tests and update back to pr

packages/expo-cli/src/exp.ts
```diff
@@ -30,6 +30,7 @@ import {
   ProjectUtils,
   UserManager,
 } from 'xdl';
+import UnifiedAnalytics from 'xdl/build/UnifiedAnalytics';
```
byCedric, line 33: I think you should be able to just import it from `xdl`. Using `xdl/build/UnifiedAnalytics` might be risky, since it leaks package-specific structure into the `expo-cli`.

packages/xdl/src/Analytics.ts
```diff
@@ -9,13 +9,19 @@ const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
...
-let _userTraits: any;
-
 export class AnalyticsClient {
+  private _userId: string | undefined;
+  private _userTraits: any;
   private segmentNodeInstance: Segment | undefined;
   private version: string | undefined;
-  private userIdentifyCalled: boolean = false;
+
+  public get Verion() {
```
byCedric, line 18: I think you intended `Version` 😬  I think we also prefer lowercase, if that's conflicting with the private `version` class property, I think it's ok to use `_` prefix (although that's not the perfect solution)

packages/xdl/src/User.ts
```diff
@@ -84,6 +85,10 @@ export class UserManagerInstance {
   _getSessionLock = new Semaphore();
   _interactiveAuthenticationCallbackAsync?: () => Promise<User>;
 
+  constructor() {
+    this.getCachedUserDataAsync = this.getCachedUserDataAsync.bind(this);
```
byCedric, line 89: This was required for the `profileMethod` function right? Since it breaks the binding context. Might be worth the check out other/better solutions.

packages/xdl/src/Analytics.ts
kgc00: thanks for pointing that out!

I copied the styling from `.packages/expo-cli/build/credentials/context.js` and used `_property` for private and `property` for public getters.

packages/xdl/src/User.ts
kgc00: I will just remove it since it may cause issues.  The profiling comes out to <1ms so I don't think it needs to be in there.

kgc00:
Ok, so I think this is good to go.  Thanks for the reviews!

I ended up leaving the default analytics client to function as it always has, however, I fixed the inconsistent event firing for the new Unified client.  During the start command, we now read the user settings from the cache so we avoid an API call.  The only concern I have is that some of this information, such as `userId` or `username`, is optional in the `UserSettings` object, whereas grabbing it from the API guarantees properties.  I feel like that info should be available 99-100% of the time, but if anyone knows any edge-cases I would love to hear about them.

packages/xdl/src/Analytics.ts
```diff
@@ -9,59 +9,59 @@ const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
...
-let _userTraits: any;
-
 export class AnalyticsClient {
-  private segmentNodeInstance: Segment | undefined;
-  private version: string | undefined;
-  private userIdentifyCalled: boolean = false;
+  private _userId: string | undefined;
+  private _userTraits: any;
+  private _segmentNodeInstance: Segment | undefined;
+  private _version: string | undefined;
```
wkozyra95, line 16: in typescript with private/public specifiers using _ is unnecessary (unless you need getter with the same name, so maybe remove _ where possible. Some code in xdl is still written this way but it's mostly because it was originally written in js

packages/xdl/src/Analytics.ts
```diff
@@ -9,59 +9,59 @@ const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
...
+    if (this._segmentNodeInstance) {
+      this._segmentNodeInstance.identify({
+        userId: this._userId,
+        traits: this._userTraits,
+        context: this.getContext(),
+      });
+    }
   }
 
   public setVersionName(version: string) {
```
wkozyra95, line 50: if you have public getter and setter then just make the field public and remove both better and setter

packages/xdl/src/start/startAsync.ts
```diff
@@ -50,6 +52,32 @@ export async function startAsync(
...
+        userData.userId, // userId is used as the identifier in the other codebases (www/website) running unified analytics so we want to keep using it on the cli as well to avoid double counting users
+        {
+          userId: userData.userId,
+          currentConnection: userData?.currentConnection,
+          username: userData?.username,
+          userType: '', // not available without hitting api
+        }
+      );
+    }
+  }
```
wkozyra95, line 71: Every new event will still have the same problem, it's ok as tmp fix, but the solution should be more generic

you could move it to function in `packages/xdl/src/UnifiedAnalytics.ts` and call that function from `packages/expo-cli/src/exp.ts` after setVersionName

kgc00: since we only need the start event for now I'm keeping it at this level, but do plan to pull it up if we have more events

packages/xdl/src/Analytics.ts
kgc00: the idea was to keep a consistent naming convention within the class.  Since the getters / private field names were colliding i decided on `_` for private and no prefix for public.  As you suggested I removed the prefix where there was not a collision.

packages/xdl/src/Analytics.ts
kgc00: I removed the setter and moved the setting logic into an `initializeClient` method and kept the getter.

PR #3376, opened 2021-04-08
Added `currentFullName` to the public config

EvanBacon:
# Why

- https://github.com/expo/universe/pull/7311
- Required for auth-session and notifications. This will append `currentFullName` to the manifest when `getConfig` is run with `isPublicConfig: true` (what `expo-constants` does). 
- I've moved the UserSettings logic into `expo/config` for now, we can move it out later if needed.
- Deletes changes introduced in https://github.com/expo/expo-cli/pull/3359

# Test Plan

- Added unit tests.
- I've verified that the value `currentFullName` will be overwritten by the server so this shouldn't introduce a breaking change.
- Unless we also add this value to `expo.id`, the auth session and notifications modules will need to be updated to use `currentFullName`.

### DIFF
--- a/packages/config/src/__tests__/ConfigParsing-test.ts
+++ b/packages/config/src/__tests__/ConfigParsing-test.ts
@@ -96,6 +96,7 @@ describe(getConfig, () => {
   // - generated `.expo` object is created and the language hint is added
   describe('language support', () => {
     beforeEach(() => {
+      delete process.env.EXPO_DEBUG;
       const projectRoot = 'js';
       setCustomConfigPath(projectRoot, undefined);
     });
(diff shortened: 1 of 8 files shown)


packages/config/src/__tests__/ConfigParsing-test.ts
```diff
@@ -96,6 +96,7 @@ describe(getConfig, () => {
   // - generated `.expo` object is created and the language hint is added
   describe('language support', () => {
     beforeEach(() => {
+      delete process.env.EXPO_DEBUG;
```
brentvatne, line 99: we should probably do this before every test in the expo-cli repo, maybe in global setup. i noticed that we set EXPO_DEBUG on ci but people may not have that set locally, which is i guess what you ran into here too

brentvatne: i guess in general process.env should just have no contents unless we explicitly set them for specific tests

brentvatne (approved):

PR #3377, opened 2021-04-09
[next-adapter] update RNW peer, remove deprecated types

Simek:
# Why

Currently `@expo/next-adapter` lists the outdated version of `react-native-web` in the `peerDependencies`.

Also the package uses the deprecated and outdate `@types/next` package:

Screenshot 2021-04-09 at 12 25 55

# How

This PR bumps the `react-native-web` version in  `@expo/next-adapter` dependencies to match the one used in Expo environment - `0.13.12`:

* https://github.com/expo/expo/search?l=JSON&q=react-native-web

It also remove the deprecated and unnecessary Next.js types package and ensures that Next `^9` is used in the workspace. To do that, I had to bump explicit `next` dev dependency in `dev-tools`, otherwise dependency was always resolved to `^8.0.0` which do not includes types and was a miss-match for most of `peerDependencies` listed in other packages.

Currently there is no good types for `NextConfig` - https://github.com/vercel/next.js/issues/10254.

Also as seen below Next.js don't care about that and uses `any`:

Screenshot 2021-04-09 at 12 55 07

CC: @EvanBacon 

# Test Plan

Command `yarn` && `yarn build` executed on workspace and in `@expo/next-adapter` package works correctly.

### DIFF
--- a/packages/next-adapter/src/withExpo.ts
+++ b/packages/next-adapter/src/withExpo.ts
@@ -1,9 +1,8 @@
 import { getBareExtensions } from '@expo/config/paths';
 import { withUnimodules } from '@expo/webpack-config/addons';
 import { AnyConfiguration } from '@expo/webpack-config/webpack/types';
-import { NextConfig } from 'next';
 
-export default function withExpo(nextConfig: NextConfig = {}): NextConfig {
+export default function withExpo(nextConfig: any = {}): any {
   return {
     ...nextConfig,
     pageExtensions: getBareExtensions(['web']),
--- a/packages/dev-tools/package.json
+++ b/packages/dev-tools/package.json
@@ -48,7 +48,7 @@
     "graphql-tag": "^2.9.1",
     "has-ansi": "^3.0.0",
     "http-proxy-middleware": "^0.18.0",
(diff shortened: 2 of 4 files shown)


packages/next-adapter/src/withExpo.ts
```diff
@@ -1,9 +1,8 @@
 import { getBareExtensions } from '@expo/config/paths';
 import { withUnimodules } from '@expo/webpack-config/addons';
 import { AnyConfiguration } from '@expo/webpack-config/webpack/types';
-import { NextConfig } from 'next';
 
-export default function withExpo(nextConfig: NextConfig = {}): NextConfig {
+export default function withExpo(nextConfig: any = {}): any {
```
EvanBacon, line 5: lol ok

EvanBacon (approved):
LGTM

packages/next-adapter/src/withExpo.ts
Simek: Can change it in next PR to `Record<string, any>` if you want, but probably we cannot go more strict than that, without writing the types by hand. 😉

PR #3378, opened 2021-04-10
analytics Id from username to userId 

kgc00:
# Why

Analytics was using `user.username` (account name) as an ID to track user activity with.  

Users can change their account name, and if they do, we would register them as a new user instead of tracking the change.  Analytics docs recommend using a database id https://segment.com/docs/connections/spec/identify/#user-id 

# How

update the assignment of the ID inside the `User` class to `user.userId` from `user.username`.

# Test Plan

fire some events locally and make sure they are getting the proper ID assigned.

### DIFF
--- a/packages/xdl/src/User.ts
+++ b/packages/xdl/src/User.ts
@@ -453,7 +453,7 @@ export class UserManagerInstance {
         });
       }
 
-      Analytics.setUserProperties(user.username, {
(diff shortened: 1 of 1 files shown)


kgc00:
no reason to do this for existing project and mess with things that are working fine.  I will make the change solely for the new, unified project in https://github.com/expo/expo-cli/pull/3372.

PR #3386, opened 2021-04-14
Handle async storage package rename in upgrade command

fson:
# Why

`@react-native-community/async-storage` has been renamed to `@react-native-async-storage/async-storage`. The old `@react-native-community/async-storage` package will not work with SDK 41. 

#3342 added a `expo doctor` check for the old package in dependencies and a codemod to migrate imports to use the new package. This PR makes the `expo upgrade` command to automatically update the dependency in `package.json`.

# How

For SDK 41 or newer, if `@react-native-community/async-storage` exists in dependencies, it will be replaced with the appropriate version of `@react-native-async-storage/async-storage`.

# Test Plan

Tested using `EXPO_BETA=1 expod upgrade`:
Screen Shot 2021-04-14 at 15 16 29
Result diff:
Screen Shot 2021-04-14 at 15 15 48

### DIFF
--- a/packages/expo-cli/src/commands/upgrade.ts
+++ b/packages/expo-cli/src/commands/upgrade.ts
@@ -85,7 +85,7 @@ export async function getUpdatedDependenciesAsync(
   workflow: ExpoWorkflow,
   targetSdkVersion: TargetSDKVersion | null,
   targetSdkVersionString: string
-): Promise<DependencyList> {
+): Promise<{ dependencies: DependencyList; removed: string[] }> {
   // Get the updated version for any bundled modules
   const { exp, pkg } = getConfig(projectRoot);
   const bundledNativeModules = (await JsonFile.readAsync(
@@ -103,6 +103,8 @@ export async function getUpdatedDependenciesAsync(
     targetSdkVersion,
   });
 
+  const removed: string[] = [];
+
   // Add expo-random as a dependency if using expo-auth-session and upgrading to
   // a version where it has been moved to a peer dependency.
   // We can remove this when we no longer support SDK versions < 40, or have a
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -701,6 +727,17 @@ export async function upgradeAsync(
     Log.addNewLineIfNone();
   }
 
+  if (removed.includes('@react-native-community/async-storage')) {
+    Log.addNewLineIfNone();
+    Log.log(
+      chalk.bold(
+        `🚨 @react-native-community/async-storage has been renamed to @react-native-async-storage/async-storage. Run \`npx expo-codemod sdk41-async-storage './**/*'\` to rename imports`
```
brentvatne, line 734:
```suggestion
        `🚨 @react-native-community/async-storage has been renamed to @react-native-async-storage/async-storage. The dependency has been updated for you, you will now need to either update the imports in your source code manually, or run \`npx expo-codemod sdk41-async-storage './**/*'\`.`
```

brentvatne (approved):

packages/expo-cli/src/commands/upgrade.ts
```diff
@@ -701,6 +727,17 @@ export async function upgradeAsync(
...
+      )
+    );
+    Log.log(chalk.grey.bold(untouchedDependencies.join(', ')));
```
brentvatne, line 737: this looks like a duplicate of above?

brentvatne, line 737:
```suggestion
```

brentvatne (changes requested):

PR #3389, opened 2021-04-14
[codemod] don't throw on node 14

cruzach:
# Why

`expo-codemod` should support node 14

# Test Plan

seems to work afaict- run `expo-codemod/bin/expo-codemod.js sdk41-async-storage ./` with node 14, result:
```diff
diff --git a/App.js b/App.js
index 56bbe11..cb3ecc6 100644
--- a/App.js
+++ b/App.js
@@ -11,7 +11,7 @@ import {
   NativeModules,
 } from "react-native";
 import * as Linking from "expo-linking";
-import AsyncStorage from "@react-native-community/async-storage";
+import AsyncStorage from '@react-native-async-storage/async-storage';
```

### DIFF
--- a/packages/expo-codemod/bin/expo-codemod.js
+++ b/packages/expo-codemod/bin/expo-codemod.js
@@ -2,7 +2,7 @@
 const semver = require('semver');
 const version = process.versions.node;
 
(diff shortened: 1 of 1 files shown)


brentvatne:
let's go with https://github.com/expo/expo-cli/pull/3389 so we don't have to update this list in the future as supported node versions change :)

PR #3397, opened 2021-04-16
[config-plugins] feat: enable splash screen plugin for any SDK greater than 38

bbarthec:
# Why

Launching `eas build -p ios` for SDK 41 results in the following warning: 
image

# How

`splash-screen` plugin was enabled only for SDK 39 and 40 (and UNVERSIONED), but not for SDK 41.
I've changed the enabling rules for this plugin to be enabled for any SDK greater than 38 (including UNVERSIONED).

# Test Plan

Locally the problem was reproducible via `expo-cli prebuild`.
Launching this with command with this change results with no warning message anymore.

### DIFF
--- a/packages/config-plugins/src/android/SplashScreen.ts
+++ b/packages/config-plugins/src/android/SplashScreen.ts
@@ -44,7 +44,8 @@ export function getSplashScreenConfig(config: ExpoConfig): AndroidSplashScreenCo
 }
 
 export async function setSplashScreenAsync(config: ExpoConfig, projectRoot: string) {
-  const majorVersionString = config.sdkVersion?.split('.').shift();
+  const majorVersionString =
+    config.sdkVersion === 'UNVERSIONED' ? null : config.sdkVersion?.split('.').shift();
   const splashScreenIsSupported =
     (majorVersionString && Number(majorVersionString) >= 39) ||
     /** UNVERSIONED CASE */ !majorVersionString;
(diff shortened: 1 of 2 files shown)


brentvatne (approved):

brentvatne:
this solution will work well for now, thanks @bbarthec! it seems that the source of this problem is that developers can use the `splash` key in `app.json` without installing expo-splash-screen, and so we potentially don't have the config plugin available, so we have to fall back to this "unversioned" plugin. we'll need to think about how to resolve this more sustainably soon

packages/config-plugins/src/android/SplashScreen.ts
```diff
@@ -44,8 +44,11 @@ export function getSplashScreenConfig(config: ExpoConfig): AndroidSplashScreenCo
 }
 
 export async function setSplashScreenAsync(config: ExpoConfig, projectRoot: string) {
+  const majorVersionString = config.sdkVersion?.split('.').shift();
   const splashScreenIsSupported =
-    config.sdkVersion === '39.0.0' || config.sdkVersion === '40.0.0' || !config.sdkVersion;
+    (majorVersionString && Number(majorVersionString) >= 39) ||
```
EvanBacon, line 49: Would this error out on UNVERSIONED? (UNVERSIONED should always be used afaict).

packages/config-plugins/src/ios/SplashScreen.ts
```diff
@@ -38,8 +38,11 @@ export function getSplashScreen(config: ExpoConfig): IosSplashScreenConfig | und
 }
 
 export async function setSplashScreenAsync(config: ExpoConfig, projectRoot: string) {
+  const majorVersionString = config.sdkVersion?.split('.').shift();
   const splashScreenIsSupported =
-    config.sdkVersion === '39.0.0' || config.sdkVersion === '40.0.0' || !config.sdkVersion;
+    (majorVersionString && Number(majorVersionString) >= 39) ||
+    /** UNVERSIONED CASE */ !majorVersionString;
+
```
EvanBacon, line 45: Should also improve the warning message so users are aware that the plugin is disabled on the current expo version, and maybe also inform them to upgrade.

packages/config-plugins/src/android/SplashScreen.ts
brentvatne: yeah that's a good point. `UNVERSIONED".split('.').shift()` -> `"UNVERSIONED"` - so `!majorVersionString` wouldn't handle this case

brentvatne (approved):

PR #3398, opened 2021-04-16
[expo-cli][xdl] return correct project page url after expo publish

axeldelafosse:
# Why

We are currently returning the website project page URL without the correct path, i.e. `@username/projectname` instead of `@username/projects/projectname` and it can create a "clawback" if the user's project is named `snacks` or `projects`.

More context here: https://github.com/expo/universe/pull/7331#issuecomment-820683580

# How

Created correct project page URL

# Test Plan

Run `expo publish` and check if the URL is correct

### DIFF
--- a/packages/xdl/src/project/publishAsync.ts
+++ b/packages/xdl/src/project/publishAsync.ts
@@ -210,6 +210,10 @@ export interface PublishedProjectResult {
    */
   url: string;
   /**
+   * Project page URL
+   */
+  projectPageUrl: string;
+  /**
    * TODO: What is this?
    */
   ids: string[];
@@ -390,12 +394,22 @@ export async function publishAsync(
     });
   }
 
+  // Create project manifest URL
+  const url =
+    options.releaseChannel && options.releaseChannel !== 'default'
+      ? `${response.url}?release-channel=${options.releaseChannel}`
+      : response.url;
+
+  // Create project page URL
+  const websiteUrl = url.replace('exp.host', 'expo.io');
+  const splitUrl = websiteUrl.split('/');
+  splitUrl.splice(4, 0, 'projects');
+  const projectPageUrl = splitUrl.join('/');
+
   return {
     ...response,
-    url:
-      options.releaseChannel && options.releaseChannel !== 'default'
-        ? `${response.url}?release-channel=${options.releaseChannel}`
-        : response.url,
+    url,
+    projectPageUrl,
   };
 }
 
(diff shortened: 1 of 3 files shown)


packages/xdl/src/project/publishAsync.ts
```diff
@@ -390,12 +394,22 @@ export async function publishAsync(
...
+  const url =
+    options.releaseChannel && options.releaseChannel !== 'default'
+      ? `${response.url}?release-channel=${options.releaseChannel}`
+      : response.url;
+
+  // Create project page URL
+  const websiteUrl = url.replace('exp.host', 'expo.io');
+  const splitUrl = websiteUrl.split('/');
+  splitUrl.splice(4, 0, 'projects');
+  const projectPageUrl = splitUrl.join('/');
```
brentvatne, line 407: this feels a little brittle. we get the `url` from our `publish/new` endpoint: https://github.com/expo/expo-cli/blob/4f40c276ffe1c52f5bc13b914818ddc36646e460/packages/xdl/src/project/publishAsync.ts#L237

i imagine we should be changing the url on that endpoint rather than transforming it in expo-cli code

axeldelafosse: Yeah I agree

axeldelafosse: OK! This is merged https://github.com/expo/universe/pull/7395

packages/xdl/src/project/publishAsync.ts
```diff
@@ -390,12 +394,22 @@ export async function publishAsync(
...
+  // Create project manifest URL
+  const url =
+    options.releaseChannel && options.releaseChannel !== 'default'
+      ? `${response.url}?release-channel=${options.releaseChannel}`
+      : response.url;
+
+  // Create project page URL
+  const projectPageUrl =
+    options.releaseChannel && options.releaseChannel !== 'default'
+      ? `${response.projectPageUrl}?release-channel=${options.releaseChannel}`
```
brentvatne, line 406: we probably want to handle the case here `projectPageUrl` is undefined and fall back to `url` - just in case - otherwise we need to be sure to time this release around deploying the update to www and, if something happens and it gets rolled back, then we would have a weird output here

axeldelafosse: > eg: maybe we set projectPageUrl to null there is some unexpected value (like no projectPageUrl in response) and then, in the code that prints this value, we don't display a project page url in that case.

Sounds good- I added this check

brentvatne (approved):

PR #3404, opened 2021-04-21
[config-plugins] Make user interface style default to light if not specified

brentvatne:
# Why

This is the default behavior in Expo Go and standalone apps built with expo build.

# How

The most reasonable place to change this seemed to be in `getUserInterfaceStyle`, but really anywhere would do.

# Test Plan

Run tests.


EvanBacon (approved):
LGTM, maybe update the schema to reflect the default value.

brentvatne:
seems to already:

```
"userInterfaceStyle": {
  "description": "Configuration to force the app to always use the light or dark user-interface appearance, such as \"dark mode\", or make it automatically adapt to the system preferences. If not provided, defaults to `light`.",
  "type": "string",
  "fallback": "light",
  "enum": ["light", "dark", "automatic"]
},
```

PR #3407, opened 2021-04-21
Fix running dev client from terminal UI using wrong bundleIdentifier

fson:
# Why

Steps to repro:
- Change bundleIdentifier in app.json or Xcode project but not both
- Run a server with expo start --dev-client
- Press i to launch in Simulator.app
- From "Which bundle identifier should we use?" choose the one from Xcode project
- The CLI tries to use the other one
 
Screen Shot 2021-04-21 at 13 59 40

# How

The if-else condition in `configureBundleIdentifierAsync` was flipped, when you chose "from Xcode project" it actually updated the *Xcode project to use the bundle identifier from `app.json`*. This PR corrects this error.

# Test Plan

Verified the correct bundle identifier is used after this fix:
Screen Shot 2021-04-21 at 14 24 16

### DIFF
--- a/packages/xdl/src/BundleIdentifier.ts
+++ b/packages/xdl/src/BundleIdentifier.ts
@@ -71,7 +71,7 @@ However, if you choose the one defined in the Xcode project you'll have to updat
           },
         ],
       });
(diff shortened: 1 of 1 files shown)


EvanBacon (approved):
seems fine

PR #3418, opened 2021-04-22
[workspace] deduplicate yarn lock in `postinstall`

Simek:
# Why

* https://github.com/atlassian/yarn-deduplicate#duplicated-packages

TL/DR: this script ensures that only the newest packages which matches semver are used in the workspace.

I have committed this to several repos and it has been used in (some major examples): 
* React Native core repo (no `postinstall` hook), 
* React Native website,
* React Native Directory,
* Docusuaurs (for all packages and websites, in repo for several months, no issues).

# How

This PR adds `yarn-deduplicate` to the workspace dependencies and hooks it up into `postinstall` check, before the Expo dependencies validation, to ensure that those checks are run on final lock version.

# Test Plan

`yarn build` on the workspace level was successful on `localhost`, let's try CI now.

CC: @brentvatne @EvanBacon

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "format": "prettier --write .",
     "build": "lerna run prepare --stream",
-    "postinstall": "expo-yarn-workspaces check-workspace-dependencies",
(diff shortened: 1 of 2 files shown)


EvanBacon (approved):
seems fine

PR #3422, opened 2021-04-23
[configure-splash-screen]: remove core-js dependency

### LINKED ISSUE #3185: Remove core-js from @expo/configure-splash-screen
Large package to use for `matchAll`, also none of our packages should be polyfilling the node environment.

hehex9:
# Why

Closes #3185 

# How

Since node 10 will be dropped in expo-cli soon (and already dropped by react-native), i think it's safe to upgrade the `engines` to `>=12` and just removing the `matchAll` polyfill.

# Test Plan

tests passed on local


EvanBacon (approved):
Awesome! 😍

PR #3423, opened 2021-04-23
[schemer] bump AJV, bump other deps

### LINKED ISSUE #3137: [schemer] Update avj
Upgrade to the latest and resolve the peer dependency warning:

```
npm WARN ajv-keywords@3.5.2 requires a peer of ajv@^6.9.1 but none is installed. You must install peer dependencies yourself.
```

Simek:
# Why

Hopefully fixes #3137

# How

There was a lot of digging thought AJV releases (https://github.com/ajv-validator/ajv/releases), I have earlier some experience with AJV in React Native Directory and few other, smaller projects, but I'm still not sure if migration to the new version is fully correct.

Without the `strict: false` option `meta` was reported as invalid key (https://github.com/ajv-validator/ajv/releases/tag/v7.0.0). I think that I do not understand this change fully, maybe an additional schema should be passed by [`ajv. addMetaSchema`](https://ajv.js.org/api.html#ajv-addmetaschema-schema-object-object-key-string-ajv) to allow custom meta? 

Other changes description:
* `dataPath` renamed to `instancePath` (v8 breaking change - https://github.com/ajv-validator/ajv/releases/tag/v8.0.0)
* now by default internal AJV RegExps are using [unicode flag ("u")](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode), and it was a reason for test failures, hopefully there is an option to opt-out
* earlier build-in `formats` have been moved to the separate package (https://github.com/ajv-validator/ajv/releases/tag/v7.0.0), this probably might not be required, unless the `empty` value specified in old types is never passed (`full`) should be the default format, from what I have understand.

Also the TS support in AJV have been enhanced, there is no longer need to type `Options` in Schemer code.

# Test Plan

Successful `yarn build`, `yarn test` and `yarn test-integration` on `localhost`. Now it's time for CI.

CC: @EvanBacon


EvanBacon (approved):
LGTM, maybe try running `expo publish` in a project with a complex config to ensure the basic functionality still works e2e

EvanBacon (approved):

PR #3436, opened 2021-04-27
Added expo-dev-client to auto plugins

EvanBacon:
# Why

- Auto configure `expo-dev-client` on prebuild.

# Test Plan

- Can't test since the package doesn't exist yet.


packages/expo-cli/src/commands/eject/configureProjectAsync.ts
```diff
@@ -31,28 +31,19 @@ export const expoManagedPlugins = [
...
 // Plugins that need to be automatically applied, but also get applied by expo-cli if the versioned plugin isn't available.
 // These are split up because the user doesn't need to be prompted to setup these packages.
 const expoManagedVersionedPlugins = [
-  // 'expo-splash-screen',
-  // 'expo-facebook',
-  // 'expo-branch',
-  // 'expo-updates',
-  // 'expo-ads-admob',
-  'expo-apple-authentication',
-  'expo-document-picker',
```
fson, line 50: Is removing these a breaking change? (Do we need to add it in the changelog?)

EvanBacon: nope, they were moved to config-plugins a while ago, just forgot to remove them here.

brentvatne (approved):

PR #3437, opened 2021-04-27
[install] add config plugins automatically

EvanBacon:
# Why

1. Make setup guides easier
2. Help migrate people away from auto config plugins. Auto config plugins are bad form and shouldn't really exist since they add unexpected functionality to the project, we added them for easier migration from `expo build` to `eas build`. This will enable people to run `expo install expo-camera` and have `expo-camera` be added to their plugins array.
3. Improve migration. Guides can say "run expo install ... with all of your existing packages to add any existing config plugins". This isn't perfect since a package might need a plugin but not have one setup.

# How

When a user installs packages with `expo install`, the command will check if the newly installed packages have any "valid" config plugins included. 

For sanity purposes we'll only enable this feature with packages that use the app.plugin.js entry point, this will prevent packages like lodash from being confused for config plugins and corrupting the prebuild. 

If the Expo config doesn't exist, the user will be warned with what to add. Similarly if the config is dynamic then we'll warn them and bail out. Otherwise the plugins will be added and a log will be printed showing the results. 

Currently there is no mechanism for detecting if a plugin has required props, so the build could still fail if a plugin is added without those required secondary props.

# Test Plan

- `EXPO_DEBUG=1 expod install lodash` should not add lodash to plugins.
- `EXPO_DEBUG=1 expod install expo-notifications expo-brightness` should add expo-notifications and expo-brightness to plugins. Running this command again shouldn't modify the app.json. This means the plugins won't be overwritten.
- `EXPO_DEBUG=1 expod install expo-sms` should not add plugins.
- In a project with `app.config.js` - `EXPO_DEBUG=1 expod install expo-notifications` should log what to add.
- In a project without an Expo config - `EXPO_DEBUG=1 expod install expo-notifications` should log what to add and that the config needs to be created.

### DIFF
--- a/packages/expo-cli/src/commands/eject/ConfigValidation.ts
+++ b/packages/expo-cli/src/commands/eject/ConfigValidation.ts
@@ -253,7 +253,7 @@ export async function getOrPromptForPackage(projectRoot: string): Promise<string
   return packageName;
 }
 
-async function attemptModification(
+export async function attemptModification(
   projectRoot: string,
   edits: Partial<ExpoConfig>,
   exactEdits: Partial<ExpoConfig>
@@ -262,7 +262,7 @@ async function attemptModification(
     skipSDKVersionRequirement: true,
   });
   if (modification.type === 'success') {
-    Log.newLine();
+    Log.addNewLineIfNone();
   } else {
     warnAboutConfigAndThrow(modification.type, modification.message!, exactEdits);
   }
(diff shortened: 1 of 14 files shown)


packages/expo-cli/src/commands/eject/ConfigValidation.ts
```diff
@@ -271,12 +271,36 @@ async function attemptModification(
...
     Log.chalk.yellow(
-      'No Expo config was found. Please create an Expo config (`app.config.js` or `app.json`) in your project root.'
+      `No Expo config was found. Please create an Expo config (${Log.chalk.bold`app.json`} or ${Log
+        .chalk.bold`app.config.js`}) in your project root.`
     )
   );
 }
 
-function warnAboutConfigAndThrow(type: string, message: string, edits: Partial<ExpoConfig>) {
+export async function attemptAddingPluginsAsync(
```
ajsmth, line 280: I'm sure you have a good reason but I was confused about why this function is exported from the eject directory but used in the install command

ajsmth (commented):
> For sanity purposes we'll only enable this feature with packages that use the app.config.js entry point, this will prevent packages like lodash from being confused for config plugins and corrupting the prebuild.

I was a little bit confused by this - it looks like there is support in place for `app.json` and `app.config.js` - I dug through a little bit and it looks like we're loading plugins via `app.plugin.js` - is that what this refers to? Sorry I'm not up to speed on this yet!

Otherwise I just have the one nit - everything else worked well for me. 

I also ended up testing installing plugin modules consecutively to ensure it didn't overwrite previous plugins e.g `EXPO_DEBUG=1 expod install expo-notifications` `EXPO_DEBUG=1 expod install expo-brightness` should have both plugins in app.json

packages/expo-cli/src/commands/eject/ConfigValidation.ts
EvanBacon: will refactor

EvanBacon:
Ya I meant `app.plugin.js`, sorry for the mixup.

ajsmth (approved):

PR #3465, opened 2021-05-06
[config-plugins] ignore case sensitivity when reading scheme

wkozyra95:
# Why

reading target does not handle a few cases

# How

- handle the case when reading scheme returns undefined
- add better error message when scheme file does not exist
- ignore capitalization when looking for a scheme file.

# Test Plan

run `eas build -p ios` with (correct scheme name, scheme name with different capitalization, missing scheme file)


dsokal (approved):
Fix the linter warning.

PR #3472, opened 2021-05-10
[config-plugins] support PKCS keystores for android builds

wkozyra95:
# Why

PKCS do not have keyPassword, but gradle requires that field. It seems to work when kesytorePassword is passed as both values in that case.

# How

if keyPassword is not specified pass kesytorePassword

# Test Plan

run build locally with modified eas-build.gradle


quinlanj (approved):
thanks!

PR #3482, opened 2021-05-12
[config] fix return typing of modifyConfigAsync

jkhales:
# Why

`modifyConfigAsync` never returns an `ExpoConfig` object.

# How

delete type.

# Test Plan

yarn tsc


fson (approved):
Looks good to me. The `config` property of the return value is also not used anywhere in this repo, so this should not cause any issues.

Issue #3478, opened 2021-05-11
expo export - bundles not created


rbro:
When running expo export in a certain way, the bundles directory does not get created.

I am running expo-cli 4.4.6.

If I run the below:
```
expo init my-project --template blank
expo export --public-url https://localhost --output-dir test my-project
```

the export is successful, but the test directory doesn't have a subdirectory called bundles.

I get these warnings during the export
```
app.json specifies null: /tmp/expo/test/bundles/ios-946081275c3439ce9b3a4f4b53c9df98.js, but that directory does not exist.
app.json specifies null: /tmp/expo/test/bundles/android-30428fb2857685f5cc00947d717c746e.js, but that directory does not exist.
```

If I run ```mkdir test/bundles``` and then re-run the export, it creates the bundles correctly.

Thank you for your help.

jkhales:
Hi, can't seem to repro this.
<img width="831" alt="Screen Shot 2021-05-12 at 6 28 33 AM" src="https://user-images.githubusercontent.com/1220444/117982856-4800c200-b2eb-11eb-80a9-0cf8296d3984.png">

what OS are you on?

Also I believe there is a typo, you have a trailing `my-project` on your second command.

rbro:
I am running expo export from outside the my-project directory, so the trailing ```my-project``` is the path I want to export.  In your example, can you try running "cd .." to go to the parent directory and then running:

```expo export --public-url https://localhost --output-dir test my-project```

jkhales:
@rbro thanks for reporting this and providing a clear example 🙏  The fix is merged https://github.com/expo/expo-cli/pull/3485 and will be in the next `expo-cli` release!

rbro:
@jkhales Thanks so much for the quick fix.

PR #3485, opened 2021-05-13
[expo-cli] fix expo export so it finds the correct projectDir

jkhales:
# Why
`expo export` doesn't work when run outside of a project directory: https://github.com/expo/expo-cli/issues/3478

# How

Set the `absoluteOutputDir` by resolving to the `projectDir` instead of `process.cwd()`

Went through and enforced usage of `*PathToWrite` by both the section where the directories are created as well as the section where the files are written.

# Test Plan

Repro'd the bug in #3478, confirmed that this PR fixes it.

Tested that `expo export` works both inside and outside a project dir.


EvanBacon (approved):
looks good

PR #3488, opened 2021-05-14
[config-plugins] Add noop Swift file to fix builds with Swift deps

brentvatne:
# Why

Installing `@react-native-menu/menu`, `react-native-blurhash` and `stripe-react-native` Just Work.

Q: Does this increase base binary size by too much?
A: No, according to TestFlight estimates:

Screen Shot 2021-05-13 at 10 47 41 PM

# How

Added an empty Swift file included in the target.

# Test Plan

```
expo init hello # some managed project
cd hello
yarn add @react-native-menu/menu
expod prebuild
expo run:ios
```

# Follow up

Possibly remove [this plugin from expo-facebook](https://github.com/expo/expo/blob/master/packages/expo-facebook/plugin/src/withNoopSwiftFile.ts)? Note that these plugins play together nicely so it's not strictly necessary.

### DIFF
--- a/packages/config-plugins/src/ios/Swift.ts
+++ b/packages/config-plugins/src/ios/Swift.ts
@@ -4,6 +4,7 @@ import path from 'path';
 import { ConfigPlugin, XcodeProject } from '../Plugin.types';
 import { withXcodeProject } from '../plugins/ios-plugins';
 import { getAppDelegate, getSourceRoot } from './Paths';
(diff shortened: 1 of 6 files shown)


EvanBacon (changes requested):
Just do this https://github.com/expo/expo/blob/master/packages/expo-facebook/plugin/src/withNoopSwiftFile.ts

EvanBacon (approved):
LGTM

PR #3490, opened 2021-05-14
[dev-server] Support metro-inspector-proxy for hermes debugging

Kudo:
# Why

To support Hermes inspector, metro-inspector-proxy acts as the websocket proxy server role to pass _Chrome DevTools Protocol_ commands to Hermes on device.
For Chrome DevTools to discover devices, endpoints like `/json/version` or `/json/list` should be reachable.
However, for Expo + Metro case, these endpoints will return the manifest.

The root cause is that metro-inspector-proxy middleware is added [after server started](https://github.com/facebook/metro/blob/master/packages/metro/src/index.js#L241-L256).
Other middlewares, like custom extensive middleware or Expo manifest middleware [are added before that](https://github.com/facebook/metro/blob/master/packages/metro/src/index.js#L241-L256).
In theory, the manifest middleware should be put at last for unmatched routes to return the manifest.
Since metro-inspector-proxy is handled after the manifest middleware, `/json/version` and `/json/list` returns the manifest but not device inspector information.

# How

The reason for metro-inspector-proxy to do that because [it needs the httpServer to attach the websocket handler](https://github.com/facebook/metro/blob/master/packages/metro/src/index.js#L248).
Before there are better solution from metro,
the way to add yet another metro-inspector-proxy middleware before manifest middleware and make it work.
There will be two metro-inspector-proxy middlewares in fact, but should not take too much resources.

# Test Plan

1. `expo start --dev-client` in bare workflow apps
2. Expect the request to `http://localhost:8081/json/version` should return
```json
{
    "Browser": "Mobile JavaScript",
    "Protocol-Version": "1.1"
}
```
3. Expect the request to `http://localhost:8081/json/list` should return just an empty array `[]` (assuming no debuggable hermes on devices)

### DIFF
--- a/packages/dev-server/src/MetroDevServer.ts
+++ b/packages/dev-server/src/MetroDevServer.ts
@@ -4,6 +4,7 @@ import { createDevServerMiddleware } from '@react-native-community/cli-server-ap
 import bodyParser from 'body-parser';
 import http from 'http';
 import type Metro from 'metro';
+import type { InspectorProxy as MetroInspectorProxy } from 'metro-inspector-proxy';
 import resolveFrom from 'resolve-from';
 
 import LogReporter from './LogReporter';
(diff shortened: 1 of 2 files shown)


EvanBacon:
Looks fine. btw, the manifest middleware should actually be constrained to:
```js
app.get('/', manifestHandler);
app.get('/manifest', manifestHandler);
app.get('/index.exp', manifestHandler);
```

I think ideally we'd just have `/`, (and `manifest.json` maybe) since the manifest is our equivalent to the `index.html` in a standard web app. In a web app you'd expect `/` and `/index.html` to both work so we'd want to emulate that but for json.

Kudo:
#3491 from @EvanBacon saves me for doing this workaround. let me close this pr.

PR #3494, opened 2021-05-16
Add `originalFullName` to the public config

wschurman:
# Why

https://github.com/expo/expo-cli/pull/3376 added `currentFullName` for use as a stable identifier, but it is not stable through project transfers. `originalFullName` is, so it should be used instead in libraries like auth-session and notifications.

# How

1. The xdl schema was defined in https://github.com/expo/universe/pull/7590
2. This PR is the equivalent of https://github.com/expo/expo-cli/pull/3376 but for `originalFullName`. `currentFullName` is kept so that the CLI continues to function until we change the auth-session and notifications libraries to use `originalFullName`.
3. Next up, I'll change the auth-session and notifications libraries to use `originalFullName` similar to expo/expo#12465 and expo/expo#12464.

# Test Plan

Run same unit tests that were added in https://github.com/expo/expo-cli/pull/3376.

In a sample project run:
`~/expo/expo-cli/node_modules/.bin/expo config --type public`, see both `currentFullName` and `originalFullName` in output.

### DIFF
--- a/packages/config/src/Config.ts
+++ b/packages/config/src/Config.ts
@@ -22,7 +22,7 @@ import {
 import { ConfigError } from './Errors';
 import { getExpoSDKVersion } from './Project';
 import { getDynamicConfig, getStaticConfig } from './getConfig';
-import { getCurrentFullName } from './getCurrentFullName';
+import { getFullName } from './getFullName';
 import { withConfigPlugins } from './plugins/withConfigPlugins';
 import { withInternal } from './plugins/withInternal';
 import { getRootPackageJsonPath } from './resolvePackageJson';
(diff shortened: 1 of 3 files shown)


brentvatne (approved):

packages/config/src/Config.ts
```diff
@@ -143,9 +143,11 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}):
         delete configWithDefaultValues.exp.android.config;
       }
 
-      // This value will be overwritten when the manifest is being served from the host (i.e. not completely accurate).
+      // These value will be overwritten when the manifest is being served from the host (i.e. not completely accurate).
```
ide, line 146: These value**s**

PR #3495, opened 2021-05-17
[dev-server] Add /inspector endpoint to open electron for hermes inspector

Kudo:
# Why

For dev-menu to launch Hermes inspector.

# How

Through the metro-inspector-proxy (http://localhost:8081/json/list), we could get the url for devtools to open.
In this PR, we will launch Electron to open the inspector url because Chrome/Edge cannot open devtools:// scheme directly from command line.

# Test Plan

[ ] Add unit test
[ ] Use with-dev-client example to test, but to clarify why Hermes crashed right after inspecting


packages/dev-server/package.json
```diff
@@ -39,6 +39,7 @@
     "@expo/metro-config": "0.1.67",
     "@react-native-community/cli-server-api": "4.9.0",
     "body-parser": "1.19.0",
+    "electron": "^12.0.7",
```
EvanBacon, line 42: electron is way to big for us to add to expo-cli

brentvatne: can we pop open a web browser window instead?

brentvatne: if we need to use electron here we can have users install it separately and then launch into it. we try to avoid adding large dependencies (especially if they have native modules) because they can slow down installation of expo-cli and in the worst case even lead to install failures

Kudo: makes sense.
how about to have electron and this middleware as separated package, e.g. @expo/dev-server-plugin-inspector?
since expo-cli does not have plugin support yet, may to hard code for this during import.

brentvatne: that's a neat idea - i'm in support of that

brentvatne: something else to consider that @EvanBacon pointed out - maybe using chrome-devtools-as-a-webpage could work? i believe that's what was intended from the code removed in this commit: https://github.com/react-native-community/cli/commit/dd048c6bf0430803637500d2f163dd65299c628e

notice how you can serve devtools from specific revisions at this url: https://chrome-devtools-frontend.appspot.com/serve_rev/@180642/devtools.html

Kudo: originally i thought `chrome-devtools-frontend.appspot.com` is just a frontend and needs to served through devtool:// scheme. after deep dive, it seems to be feasible.

e.g.
https://chrome-devtools-frontend.appspot.com/serve_rev/@e3cd97fc771b893b7fd1879196d1215b622c2bed/inspector.html?experiments=true&v8only=true&ws=localhost%3A8081%2Finspector%2Fdebug%3Fdevice%3D0%26page%3D1
(the `ws` parameter may be different from environment)

but just one limitation:
to get the sourcemap in this case, the request to http://localhost:8081/index.map is now a cross-origin request.
[metro](https://github.com/facebook/metro/blob/master/packages/metro/src/Server.js#L461) and [cli](https://github.com/react-native-community/cli/blob/master/packages/cli-server-api/src/securityHeadersMiddleware.ts#L17) does not support such case.
we should allow https://chrome-devtools-frontend.appspot.com in `Origin` request header and add it to `Access-Control-Allow-Origin` response header.
We may need to further customize this behavior or worth to contribute pr back to upstream.

Kudo: side note for the serve_rev in chrome-devtools-frontend.appspot.com
the revision actually bind to the commit which updating the `chrome/VERSION` file.

e.g. to get the version corresponding to chrome 90.0.4430.212
1. go to https://chromium.googlesource.com/chromium/src.git/+log/refs/tags/90.0.4430.212/chrome/VERSION
2. click the first log item, e.g. _e3cd97f Incrementing VERSION to 90.0.4430.212 by chrome-release-bot · 12 days ago_
3. there will be full commit sha value and that's it: `e3cd97fc771b893b7fd1879196d1215b622c2bed`
so the prefix will be
https://chrome-devtools-frontend.appspot.com/serve_rev/@e3cd97fc771b893b7fd1879196d1215b622c2bed/inspector.html

Kudo: tried with some poc:
https://github.com/expo/expo-cli/pull/3495/commits/c5e5d6a0d0580e9fa0a99d24a116c48a034eef63
https://github.com/expo/expo-cli/pull/3495/commits/c6f0844c3dc9cdba98853c33c751b22fc54a0af5

basically we have two proposals so far:
1. open inspector in chrome (with hook for metro/cli), so this mechanism could be builtin into expo-cli and expo-dev-menu
2. open inspector in electron and as an extra module need to manually install.

personally prefer the first one, but please let me know how you think.

Kudo: great. will follow up in this design and move inspector back to expo-cli.

Kudo: one more update today.
with dev-client, the metro may serve from LAN instead of localhost.
the request to fetch the sourcemap may from, e.g. https://chrome-devtools-frontend... to http://192.168.123.123.. and be blocked by newer chrome for insecure content (https to http).
user could manually allow this in the chrome site setting but it's really subtle.

as an alternative but really tricky, poc in https://github.com/expo/expo-cli/pull/3495/commits/def18b9a968954c30f31f46b16c462b550fdff69.
which trying to add `--allow-running-insecure-content` launch argument.
considering there may be existing chrome and making the argument no effect,
also add the `--user-data-dir=${tempProfileDir}` to force chrome create a new browser process.
although it's getting a little tricky, but works without electron.
I will wait the two prs for `open` package to land and finish the change.

btw, there was also a proposal that i give up finally:
to serve devtools frontend on dev-server, but the total size of the devtools frontend asset take up to 75MB, which also largely increase expo-cli download size.

Kudo:
@EvanBacon thought the pr is finalized without further knowing changes. please help to re-review again. thanks.

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -171,3 +187,73 @@ function importMetroServerFromProject(projectRoot: string): typeof Metro.Server
...
+  if (
+    typeof req.headers.origin === 'string' &&
+    !req.headers.origin.match(/^https?:\/\/localhost:/) &&
+    !req.headers.origin.match(/^https:\/\/chrome-devtools-frontend\.appspot\.com/)
+  ) {
+    next(
+      new Error(
+        'Unauthorized request from ' +
+          req.headers.origin +
+          '. This may happen because of a conflicting browser extension. Please try to disable it and try again.'
```
EvanBacon, line 219: Maybe mention using incognito mode instead?

packages/dev-server/src/middleware/JSInspectorMiddleware.ts
```diff
@@ -0,0 +1,127 @@
+import type { NextHandleFunction } from 'connect';
+import fs from 'fs-extra';
+import type { IncomingMessage, ServerResponse } from 'http';
+import fetch from 'node-fetch';
+import open from 'open';
+import path from 'path';
+import { TLSSocket } from 'tls';
+import { URL } from 'url';
+
+export default function JSInspectorMiddleware(): NextHandleFunction {
```
EvanBacon, line 10: I think this casing isn't correct, the other middleware is `clientLogsMiddleware` but maybe it should be `getClientLogsMiddleware`. Regardless, capitalize in JS often implies using the `new` keyword to instantiate it.

packages/dev-server/src/middleware/JSInspectorMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+      const url = `${urlBase}?experiments=true&v8only=true&ws=${encodeURIComponent(ws)}`;
+      launchChromiumAsync(url);
+      res.end();
+    } else {
+      res.writeHead(405);
+    }
+  };
+}
+
+interface MetroInspectorProxyApp {
```
EvanBacon, line 53: xit: move to top

packages/dev-server/src/middleware/JSInspectorMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+      app: {
+        name: open.apps.edge,
+        arguments: launchArgs,
+      },
+      newInstance: true,
+      wait: true,
+    });
+  }
+
+  await fs.remove(tempProfileDir);
```
EvanBacon, line 126: wrap in `try/finally` to ensure fs.remove is always used no matter what.

packages/dev-server/src/middleware/__tests__/JSInspectorMiddleware-test.ts
```diff
@@ -0,0 +1,167 @@
...
+    const entities = JSON.parse(RESPONSE_FIXTURE) as { [key: string]: string }[];
+    const entity = entities.find(object => object.description === appId);
+
+    const req = createRequest(`http://localhost:8081/inspector?applicationId=${appId}`);
+    const res = createMockedResponse();
+    const next = jest.fn();
+    const mockFetch = fetch as jest.MockedFunction<typeof fetch>;
+    mockFetch.mockReturnValue(Promise.resolve(new Response(RESPONSE_FIXTURE)));
+
+    const middlewareAsync = JSInspectorMiddleware();
```
EvanBacon, line 29: ya this looks incorrect without the new keyword, so changing the method to lowercase would make the most sense imo.

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -171,3 +187,73 @@ function importMetroServerFromProject(projectRoot: string): typeof Metro.Server
...
   return require(resolvedPath);
 }
+
+function replaceMiddlewareWith(
+  app: ConnectServer,
+  sourceMiddleware: HandleFunction,
+  targetMiddleware: HandleFunction
+) {
+  const index = app.stack.findIndex(middleware => middleware.handle === sourceMiddleware);
+  if (index >= 0) {
```
EvanBacon, line 197: Maybe use `find` since the mutation is on an object.
```js
  const somn = app.stack.find(middleware =&gt; middleware.handle === sourceMiddleware);<br class="Apple-interchange-newline">
  if (somn) {
    somn.handle = ...
  }
```

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -50,8 +57,17 @@ export async function runMetroDevServerAsync(
     port: metroConfig.server.port,
     watchFolders: metroConfig.watchFolders,
   });
+
+  replaceMiddlewareWith(
```
EvanBacon, line 61: Add a comment here explaining why

packages/dev-server/src/middleware/JSInspectorMiddleware.ts
```diff
@@ -0,0 +1,127 @@
...
+async function queryInspectorTargetAsync(
+  origin: string,
+  appId: string
+): Promise<MetroInspectorProxyApp | null> {
+  const resp = await fetch(`${origin}/json/list`);
+  const apps: MetroInspectorProxyApp[] = await resp.json();
+
+  let target: MetroInspectorProxyApp | null = null;
+  for (const app of apps) {
+    if (app.description === appId && app.vm !== "don't use") {
```
EvanBacon, line 73: haha, why "don't use"?

EvanBacon (changes requested):
Couple of simple things, nice work.

packages/dev-server/src/middleware/JSInspectorMiddleware.ts
Kudo: This is from [metro-inspector-proxy](https://github.com/facebook/metro/blob/3aca11697323dacef70bb914399e5e064fc74870/packages/metro-inspector-proxy/src/Device.js#L135-L143).
seems to support inspecting for live reload or fast refresh, but still in experiment.
i am not pretty sure how stable it is, so to exclude it in the mean time.

Kudo:
Sorry to attach two commits (https://github.com/expo/expo-cli/pull/3495/commits/7cc99729888f9ed3600b6bf24331ac7eb115d46a, e616f946936c171e8cc37e8f18b9cfc9f421cea6) missing in previous fs-extra tweak.

CI seems broken even on master branch but cannot reproduce on my local. maybe the cache issue !?

Kudo:
Some resolved review comment cannot shown inline, because i change the file from `JSInspectorMiddleware.ts` to `createJsInspectorMiddleware.ts` and GH simply show Outdated.
maybe revisit the whole thing is necessary 😅

Kudo:
rebase master to fix CI broken. there are no further changes so far.

EvanBacon (approved):
Great work!

PR #3519, opened 2021-05-26
[xdl][cli] Adds general expo cli events to unified analytics

kgc00:
# Why

We'd like to better understand how folks are using the CLI to improve our offerings. This will add more stable analytics to the cli.

# How

Pulls the segment identify call for the `UnifiedClient` up to the entry point of the cli.  The identify call reads cached user data (only present when user has previously logged in via 'Username-Password-Authentication' (not token based).  If the user is logged in with valid credentials we identify them.  Otherwise do not identify them.  

JFYI 
- the entry point cli track call will not fire for folks authenticated with tokens or who are not logged in (although we have an identify call at the login site further in the cli)
- any analytics track calls will not fire if the user has not been identified 

Also, after identifying the user we track an 'action' event with the sub command that was run (start, init, build, etc).

# Test Plan

tested manually test for:
- logged out users do not get tracked
- logged in users do get tracked

added automated tests to the User Singleton to confirm:
- reading cached user data works only when the user was previously logged in via 'Username-Password-Authentication'

added automated tests for the top level 'action' track to confirm it tracks the correct event name
added automated tests for the analytics bootstrapper to confirm only identifies folks when they are logged in


EvanBacon (approved):

packages/expo-cli/src/exp.ts
```diff
@@ -849,10 +847,33 @@ async function writePathAsync() {
...
+  if (!userData?.userId) return;
+
+  UnifiedAnalytics.identifyUser(userData.userId, {
+    userId: userData.userId,
+    currentConnection: userData?.currentConnection,
+    username: userData?.username,
+    userType: '', // not available without hitting api
+  });
+
+  const command = process.argv[2];
```
kgc00, line 865: just want to confirm that we can assume argv[2] will always exist at this point in the app? i see `writePathAsync()` accesses it without a try/catch or if statement so I assume it's safe to do this here.

packages/expo-cli/src/__tests__/exp-test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+type CommandMockData = {
+  name: string;
+  _name: string;
+  description: string;
+  alias?: string;
+  _alias?: string;
+};
+
+function generateCommandJSON(): CommandMockData[] {
```
kgc00, line 68: copied from introspect but slightly adjusted to contain `_name` and `_alias` and not contain `options` or `group`. figured the use-case was different enough to warrant a new function rather than adjusting the one present in introspect.

packages/expo-cli/src/exp.ts
```diff
@@ -708,13 +709,42 @@ Command.prototype.asyncActionProjectDir = function (
...
+  if (!subCommand) return; // only track valid expo commands
+
+  UnifiedAnalytics.logEvent('action', {
+    action: `expo ${subCommand}`,
+    source: 'expo cli',
+    source_version: UnifiedAnalytics.version,
+  });
+}
+
+async function runAsync(programName: string) {
```
kgc00, line 743: this PR makes `runAsync` an async function to allow for reading cached user creds in the entrypoint of the cli

packages/xdl/src/User.ts
```diff
@@ -97,6 +97,10 @@ export class UserManagerInstance {
     this._getSessionLock = new Semaphore();
   }
 
+  constructor() {
+    this.getCachedUserDataAsync = this.getCachedUserDataAsync.bind(this);
```
brentvatne, line 101: i prefer defining the function with fat arrow binding instead, less likely for this to break in the future and more self explanatory

packages/xdl/src/User.ts
```diff
@@ -213,7 +217,7 @@ export class UserManagerInstance {
   }
 
   /**
-   * Quickly grab cached user data to bootstrap non-critical services (analytics)
+   * Returns cached user data without hitting our backend. Only works for 'Username-Password-Authentication' flow. Does not work with 'Access-Token-Authentication' flow.
    */
   async getCachedUserDataAsync(): Promise<UserData | null> {
```
brentvatne, line 222:
```suggestion
  getCachedUserDataAsync = async (): Promise<UserData | null> => {
```

brentvatne (approved):

PR #3521, opened 2021-05-26
Add analytics to dev client commands

fson:
# Why

In order to improve the experience of using dev client, we're recording a few key events from the CLI.

# How

Events:
- `expo start --dev-client`: event `dev client start command`
- `expo run:ios`: event `dev client run command` with property `platform: "ios"`
- `expo run:android`: event `dev client run command` with property `platform: "android"`

Properties included in all of these events:
- `account_name`: the `owner` of the project, if defined. Otherwise current username.
- `dev_client_version`: semver version of the currently installed `expo-dev-client` package (if installed)
- `is_managed`: `true` if the project has `managed` as it's default target, i.e. iOS and Android native projects don't exist.
- `status`: `started` when the command is starting, `ready` once the dev server is running (`expo start`) or app is running (`expo run`), and `finished` when the command exits
- `uptime_ms`: time (in milliseconds) since the CLI process started. In other words, time elapsed since the user ran the command.

New user trait:
- `primaryAccountId`

# Test Plan

Added some temporary logging to the `logEvent` method in `Analytics.ts`, ran `start --dev-client`, `run:ios` and `run:android` commands and observed that the events were sent.

### DIFF
--- a/packages/expo-cli/src/analytics/getDevClientProperties.ts
+++ b/packages/expo-cli/src/analytics/getDevClientProperties.ts
@@ -1,0 +1,27 @@
+import { ExpoConfig, getDefaultTarget } from '@expo/config';
+import JsonFile, { JSONValue } from '@expo/json-file';
+import memoize from 'lodash/memoize';
+import resolveFrom from 'resolve-from';
+
+const getDevClientVersion = memoize((projectRoot: string): JSONValue | undefined => {
+  try {
+    const devClientPackage = resolveFrom.silent(projectRoot, 'expo-dev-client/package.json');
+    if (devClientPackage) {
+      return JsonFile.read(devClientPackage).version;
+    }
+  } catch {}
+  return undefined;
+});
+
+const isManaged = memoize((projectRoot: string): boolean => {
+  return getDefaultTarget(projectRoot) === 'managed';
+});
+
+export default function getDevClientProperties(projectRoot: string, exp: ExpoConfig) {
+  return {
+    accountName: exp.currentFullName,
+    devClientVersion: getDevClientVersion(projectRoot),
+    isManaged: isManaged(projectRoot),
+    uptimeMs: process.uptime() * 1000,
+  };
+}
--- a/packages/expo-cli/src/analytics/getDevClientProperties.ts
+++ b/packages/expo-cli/src/analytics/getDevClientProperties.ts
@@ -1,7 +1,11 @@
-import { ExpoConfig, getDefaultTarget } from '@expo/config';
+import { ExpoConfig, getAccountUsername, getDefaultTarget } from '@expo/config';
 import JsonFile, { JSONValue } from '@expo/json-file';
 import memoize from 'lodash/memoize';
 import resolveFrom from 'resolve-from';
+
+const getAccountName = memoize((exp: Pick<ExpoConfig, 'owner'>) => {
+  return getAccountUsername(exp);
+});
 
 const getDevClientVersion = memoize((projectRoot: string): JSONValue | undefined => {
   try {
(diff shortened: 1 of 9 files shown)


packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -146,6 +153,38 @@ export async function runAndroidActionAsync(projectRoot: string, options: Option
...
+      ...getDevClientProperties(projectRoot, exp),
+    });
+  });
+  installCustomExitHook(() => {
+    UnifiedAnalytics.logEvent('dev client run command', {
+      status: 'finished',
+      platform: 'android',
+      ...getDevClientProperties(projectRoot, exp),
+    });
+    UnifiedAnalytics.flush();
```
kgc00, line 184: does this [flush](https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/exp.ts#L354) not catch this scenario?

packages/expo-cli/src/commands/run/ios/runIos.ts
```diff
@@ -83,6 +90,38 @@ export async function runIosActionAsync(projectRoot: string, options: Options) {
   }
 }
 
+function track(projectRoot: string, exp: ExpoConfig) {
```
kgc00, line 93: seems like the only dif from the android `track` command is the platform? If we don't expect custom handling in each case I would lean towards some shared function, but I think this is fine as well 👍

packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -146,6 +153,38 @@ export async function runAndroidActionAsync(projectRoot: string, options: Option
...
 
+function track(projectRoot: string, exp: ExpoConfig) {
+  UnifiedAnalytics.logEvent('dev client run command', {
+    status: 'started',
+    platform: 'android',
+    ...getDevClientProperties(projectRoot, exp),
+  });
+  StatusEventEmitter.once('bundleBuildFinish', () => {
+    // Send the 'bundle ready' event once the JS has been built.
+    UnifiedAnalytics.logEvent('dev client run command', {
```
kgc00, line 164: I know the rest of the cli uses hardcoded strings but it might be nice to transition to some enum of event types like in other codebases 🙂

packages/expo-cli/src/analytics/getDevClientProperties.ts
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    const devClientPackage = resolveFrom.silent(projectRoot, 'expo-dev-client/package.json');
+    if (devClientPackage) {
+      return JsonFile.read(devClientPackage).version;
+    }
+  } catch {}
+  return undefined;
+});
+
+const getProjectType = memoize((projectRoot: string): 'managed' | 'generic' => {
```
kgc00, line 20: nice

kgc00:
looks good

packages/expo-cli/src/commands/run/android/runAndroid.ts
fson: I think it won't handle the case where an error is thrown. Perhaps it should be moved to a `finally` statement.

packages/expo-cli/src/commands/run/android/runAndroid.ts
fson: I think I actually prefer this way. I was trying to use search to find where a certain event is triggered in EAS CLI codebase and it was made harder by the constants and additional abstractions.

More thorough TypeScript type for `logEvent` might be useful, though, and would not lose the searchability.

packages/expo-cli/src/commands/run/android/runAndroid.ts
kgc00: thats a good point. maybe something for a subsequent pr then

kgc00 (approved):

packages/expo-cli/src/commands/run/android/runAndroid.ts
fson: Tried to use `finally`, however it turns out it won't work for the error case like that. There's a `process.exit(1)` in the `catch` block which means the `finally` block won't execute. We would have to add `flush()` to the `catch` block instead.

However, there's another reason the default [flush](https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/exp.ts#L354) code won't help with these events. For commands like `expo start` or `expo run` that start the development server, that flush will happen as soon as the command action code has executed – i.e. as soon as the server has started. But I need this flush to happen when the server is shutting down and after logging the event.

If we add other events to trigger during shutdown in the future, we could see if it makes sense to add this to the default exit hook or something similar (however since the order of exit event handlers is not guaranteed, we'd have to make some changes for this to work). But the ones I added are the only shutdown events tracked at the moment, as far as I know, so I'm going to keep the code as it is for now.

PR #3522, opened 2021-05-27
[expo-cli][dev-server] Export hbc as bundle output if js runtime is hermes

Kudo:
# Why

If an app is running with hermes runtime, `expo export` for self hosting update should also generate hermes bytecode instead of plain javascript.

# How

Assuming project should setup with pre-defined shape which set the runtime in gradle.properties like expo/expo#13052.
If exporting project is running with hermes, right after calling metro to build js bundles, we should further call `hermesc` to compile hbc.
The sourcemap also needs to re-compose by metro-source-map.
The overall flow is reference from [react-native's react.gradle](https://github.com/facebook/react-native/blob/0.63-stable/react.gradle).

## To clarify
The HermesBundler is placed in @expo/dev-server but it's not necessary to. It is just that we need metro-source-map to compose sourcemap and @expo/dev-server already has the logic to import metro from project.

# Test Plan

### Unit Test

```
HermesBundler-test.ts
  parseGradleProperties
    ✅  should return array of key-value tuple (2 ms)
    ✅  should keep "=" in value if there are multiple "="
    ✅  should exclude comment (1 ms)
  getHermesBytecodeBundleVersionAsync
    ✅  should return hermes bytecode version 74 for plain.74.hbc (13 ms)
    ✅  should throw exception for plain javascript file (5 ms)
    ✅  should throw exception for nonexistent file
  isHermesBytecodeBundleAsync
    ✅  should return true for hermes bytecode bundle file
    ✅  should return false for plain javascript file
    ✅  should throw exception for nonexistent file
```

### E2E Test

```
export-test.ts
    ✅  should export hbc bundle if running with hermes (31292 ms)
```


packages/expo-cli/e2e/__tests__/export-test.ts
```diff
@@ -52,6 +55,66 @@ it('exports the project for a self-hosted production deployment', async () => {
...
+  api.cache(true);
+  return {
+    presets: ['babel-preset-expo'],
+  };
+};
+`
+    );
+
+    const dotExpoHomeDirectory = path.join(projectRoot, '../.expo');
+    await runAsync(
```
jkhales, line 90: Would you mind adding a test for:
` ['expo', 'export', '--output-dir', inputDir, '--experimental-bundle']`
so we can ensure it works with eas publish? 

https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/project/publish.ts#L157

packages/dev-server/src/__tests__/fixtures/plain.js
```diff
@@ -0,0 +1 @@
+console.log('Hello World');
```
jkhales, line 1: is this intentional?

packages/expo-cli/e2e/__tests__/export-test.ts
Kudo: would you mind if i testing this locally?
b/c the --experimental-bundle output format might change in near future, for those we talked about the asset mimetype or keeping original js bundle.

yet another reason, the e2e test takes longer time in CI.

btw
from my previous test, the command works fine and generated bundle is hbc.

packages/dev-server/src/HermesBundler.ts
```diff
@@ -0,0 +1,154 @@
...
+    case 'linux':
+      return 'linux64-bin';
+    case 'win32':
+      return 'win64-bin';
+    default:
+      throw new Error(`Unsupported host platform for Hermes compiler: ${os.platform()}`);
+  }
+}
+
+export function parseGradleProperties(content: string): [string, string][] {
```
EvanBacon, line 105: Maybe export a k/v pair (object).

packages/dev-server/src/HermesBundler.ts
```diff
@@ -0,0 +1,154 @@
...
+      'Missing module "metro-source-map/src/composeSourceMaps" in the project. ' +
+        'This usually means React Native is not installed. ' +
+        'Please verify that dependencies in package.json include "react-native" ' +
+        'and run `yarn` or `npm install`.'
+    );
+  }
+  return require(resolvedPath);
+}
+
+const HERMES_MAGIC_HEADER = 'c61fbc03c103191f';
```
EvanBacon, line 133: docs?

packages/dev-server/src/HermesBundler.ts
```diff
@@ -0,0 +1,154 @@
...
+import path from 'path';
+import process from 'process';
+import resolveFrom from 'resolve-from';
+
+export async function shouldBuildHermesBundleAsync(
+  projectRoot: string,
+  platform: Platform
+): Promise<boolean> {
+  if (platform === 'android') {
+    const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
```
jkhales, line 15: Is this file path stable even with kotlin? I noticed a bunch of checks regarding this stuff here: https://github.com/expo/expo-cli/blob/master/packages/config-plugins/src/android/Paths.ts

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -138,7 +141,29 @@ export async function bundleAsync(
...
+    return await Promise.all(
+      bundles.map(async (bundle: BundleOptions) => {
+        const bundleOutput = await buildAsync(bundle);
+        const shouldBuildHermesBundle = await shouldBuildHermesBundleAsync(
+          projectRoot,
+          bundle.platform
+        );
+
+        if (shouldBuildHermesBundle) {
+          options.logger.info({ tag: 'expo' }, `💿 Building hermes bundle for ${bundle.platform}`);
```
EvanBacon, line 153: I wonder if we should use a bit more general terms here like "bytecode bundle" or "hermes bytecode bundle", calling it a "hermes bundle" reminds me of when RN called the Metro bundler the "React Native packager". If in the future we want to support other engines, keeping it agnostic from the start could be a good thing. CC @brentvatne @ide

packages/dev-server/src/HermesBundler.ts
```diff
@@ -0,0 +1,154 @@
...
+  const tempHbcFile = path.join(tempDir, 'index.hbc');
+  const hermesCommand = getHermesCommand(projectRoot);
+  const args = ['-emit-binary', '-out', tempHbcFile, tempBundleFile, '-output-source-map'];
+  if (optimize) {
+    args.push('-O');
+  }
+  await spawnAsync(hermesCommand, args);
+  const hbc = await fs.readFile(tempHbcFile);
+
+  const sourcemap = await createHermesSourcemapAsync(projectRoot, map, `${tempHbcFile}.map`);
```
jkhales, line 57:
```suggestion
  const [hbc, sourcmap] = await Promise.all([fs.readFile(tempHbcFile), createHermesSourcemapAsync(projectRoot, map, `${tempHbcFile}.map`)]);
```
?

packages/dev-server/src/HermesBundler.ts
jkhales, line 105: is this the same thing as https://github.com/expo/expo-cli/blob/86510a293529f0e862b64cf94b1eba0645bb4bce/packages/config-plugins/src/android/Properties.ts#L15 ?

jkhales (approved):
✨

packages/expo-cli/e2e/__tests__/export-test.ts
jkhales: yes local is fine. 
Agreed it will change.

Just wanted to be sure you were aware of this path as eas cli relies on it.

packages/dev-server/src/HermesBundler.ts
```diff
@@ -0,0 +1,154 @@
...
+): Promise<boolean> {
+  if (platform === 'android') {
+    const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
+    if (!fs.existsSync(gradlePropertiesPath)) {
+      // TODO(kudo): Clarify default behavior in managed workflow for gradle.properties not exist
+      return false;
+    }
+    const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
+    for (const [key, value] of properties) {
+      if (key === 'JS_RUNTIME' && value === 'hermes') {
```
EvanBacon, line 22: This seems interesting. In gradle properties, the format is often like `foo.bar=true`, but we've opted for this env-variable type format `FOO_BAR=true`. If this logic is to stay here, we'll need to document that we require this value to be defined in order for hermes to work, we'd probably also need to spec which format ppl should use in their gradle.properties files.

packages/dev-server/src/HermesBundler.ts
Kudo: thinking so. it looks like that even from kotlin project, the file is still gradle.properties.

packages/dev-server/src/HermesBundler.ts
Kudo: that is really interesting. we might [doing wrong for years](https://github.com/facebook/react-native/blob/master/ReactAndroid/gradle.properties).

also having a side question for managed workflow.
assuming gradle.properties is not existed in managed workflow project, where do we specify the js runtime?
perhaps we should moving this setting directly into app.json?

packages/dev-server/src/HermesBundler.ts
Kudo: exactly. but dev-server has no dependency for config-plugins, that's why rewriting a code here.

packages/dev-server/src/MetroDevServer.ts
Kudo: prefer "hermes bytecode bundle" if the name is not too long even in other places.

btw, v8 use snapshot or code caching (w/ slightly different concept)

packages/expo-cli/e2e/__tests__/export-test.ts
Kudo: thanks. will make sure --experimental-bundle not broken.

Kudo (commented):
thanks for your great review. the information is really helpful to me.
will update code accordingly later after a sleep.

packages/dev-server/src/MetroDevServer.ts
ide: Perhaps "Hermes bytecode for the bundle" is the most accurate description.

However, it is important that the Expo runtime does not have the concept of the bundle. Expo is like the web, not React Native. The web does not have a concept of a bundle. You can create bundles with tools like Webpack and the browser just treats it like a JS file.

Similarly, Expo apps don't have bundles. They have a collection of assets and one "launch asset", which is the first asset we run to launch an update. The launch asset currently needs to be JS but it could also be HBC or another format in the future. We just need to support running those new formats. Additionally, the launch asset does not need to be a bundle. It could be an unbundled JS file. Expo does not care because Expo is like the web.

Kudo: "Hermes bytecode for the bundle"  sounds clear in this context.
Let me update the code to something like "Building Hermes bytecode for the bundle - platform[android]".
If there are any comments, could update further.

packages/dev-server/src/HermesBundler.ts
brentvatne: i think we probably want to add a property to app.json and app.config.js, yeah. so, possibly something like:

```json
{
  "expo" {
     "slug": "abcd",
     "engine": "hermes"
  }
}
```

one consideration here is that if we don't end up supporting hermes in expo go then this field may be a bit confusing.

Kudo:
thanks for all the great review.
please let me merge this pr first and will address the app.json in another pr.

PR #3523, opened 2021-05-27
Allow `expo start --dev-client` in managed projects

fson:
# Why

It should be possible to use dev client with a managed project.

# How

- Removed the "can only be used in bare workflow apps" warning
- Added a check to look up the schema in `app.json`/`app.config.*s` when there are no native project (managed project)
- ~Added a prompt for schema when it's not configured.~
- ~Prompt has a good default.~

Screen Shot 2021-06-08 at 15 17 02

# Test Plan

Created a managed project and verified that `expo start --dev-client` started the dev server. Tested with:
- `app.json` including a schema -> started without warnings
- no `scheme` -> error

Also set the `scheme` property in `app.json` to the URI scheme of an app that I had previously built and installed on my Android device. Running `expod start --dev-client` and pressing `a` opened the app.

### DIFF
--- a/packages/expo-cli/src/schemes.ts
+++ b/packages/expo-cli/src/schemes.ts
@@ -1,6 +1,10 @@
+import { getConfig } from '@expo/config';
 import { AndroidConfig, IOSConfig } from '@expo/config-plugins';
+import JsonFile from '@expo/json-file';
 import plist from '@expo/plist';
-import * as fs from 'fs-extra';
+import fs from 'fs';
+import path from 'path';
+import slugify from 'slugify';
 
 import { AbortCommandError } from './CommandError';
 import {
@@ -8,6 +12,7 @@ import {
   hasRequiredIOSFilesAsync,
 } from './commands/eject/clearNativeFolder';
 import Log from './log';
+import prompt from './prompts';
 
 export async function getSchemesForIosAsync(projectRoot: string) {
   try {
@@ -48,6 +53,11 @@ export async function getDevClientSchemeAsync(projectRoot: string): Promise<stri
     getSchemesForAndroidAsync(projectRoot),
   ]);
 
+  // Allow managed projects
+  if (!hasIos && !hasAndroid) {
+    return getManagedDevClientSchemeAsync(projectRoot);
+  }
+
   let matching: string;
   // Allow for only one native project to exist.
   if (!hasIos) {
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/schemes.ts
```diff
@@ -81,6 +91,55 @@ export async function getDevClientSchemeAsync(projectRoot: string): Promise<stri
...
+      /^[a-z0-9-]+$/.test(value) || 'Only lowercase characters/numbers (a-z, 0-9) or hyphen (-)',
+    message: `Please choose a URI scheme for your app`,
+    initial: slugify(exp.slug, {
+      // Remove non-allowed characters and additionally hyphens, which are allowed but not idiomatic.
+      remove: /[^a-z0-9]/,
+      // Convert to lower case.
+      lower: true,
+    }),
+  });
+  if (staticConfigPath) {
```
EvanBacon, line 128: Maybe a bit more like this https://github.com/expo/expo-cli/blob/8b73a072907df0c1ce9dc23688f1f8980c72cd46/packages/expo-cli/src/commands/utils/modifyConfigAsync.ts#L30-L51

packages/expo-cli/src/schemes.ts
```diff
@@ -81,6 +91,55 @@ export async function getDevClientSchemeAsync(projectRoot: string): Promise<stri
...
+    // if app.json exists we can update it
+    const config = JsonFile.read(staticConfigPath, { json5: true });
+    config.expo = { ...(config.expo as object), scheme };
+    await JsonFile.writeAsync(staticConfigPath, config);
+    Log.log(`Scheme '${scheme}' updated in ${path.basename(staticConfigPath)}`);
+  } else {
```
EvanBacon, line 134: this would just ignore the dynamic config and create a potentially unused file

fson: Existing dynamic config is handled by an earlier if – see https://github.com/expo/expo-cli/pull/3523/files#diff-ef67bc980d05133cc7b87b450cf4a04e1546cae27683bc12a749a925bb79cf09R105

If we land in this else branch it means there's no config (dynamic or static) at all so we just create an app.json.

brentvatne:
if in a managed app we run `expo start --dev-client`, then prompt the user to add a scheme and update it and launch the server, we will start the dev client server with a scheme that doesn't exist in an dev client build that was created prior to running that command 😮 

maybe we can modify the config plugin to add a deterministic scheme derived from the project slug, and use that for `expo start --dev-client`? we could alternatively depend on the bundle id / package name being added to scheme -- run `expo config --type introspect` and notice that the bundled id / package names are added to schemes.

fson:
That's a good point from @brentvatne. I have now removed the prompt for scheme and instead added autogeneration of schemes in the `expo-dev-client` config plugin here: https://github.com/expo/expo/pull/13147

I'll add support for the autogenerated schemes to the CLI in a separate pull request.

PR #3525, opened 2021-05-28
[expo-cli] upgrade @expo/prebuild-config to fix tsc

jkhales:
# Why

yarn tsc is broken

# How

ran `yarn add @expo/prebuild-config`

# Test Plan

tsc passes.

### DIFF
--- a/packages/expo-cli/package.json
+++ b/packages/expo-cli/package.json
@@ -77,7 +77,7 @@
     "@expo/osascript": "2.0.28",
     "@expo/package-manager": "0.0.43",
     "@expo/plist": "0.0.13",
-    "@expo/prebuild-config": "1.0.1",
+    "@expo/prebuild-config": "^1.0.1",
     "@expo/results": "^1.0.0",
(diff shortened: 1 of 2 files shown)


brentvatne:
looks like we probably need to republish to bump the version of `@expo/config-types` used in `@expo/prebuild-config` - we should do that instead of this change. it's not clear to me why changing to ^1.0.1 matters here because 1.0.1 is the latest published version and is also the version of the package within the expo-cli repo

jkhales:
> looks like we probably need to republish to bump the version of `@expo/config-types` used in `@expo/prebuild-config` - we should do that instead of this change. it's not clear to me why changing to ^1.0.1 matters here because 1.0.1 is the latest published version and is also the version of the package within the expo-cli repo

yea, this is a good point, what is going on... 😅

jkhales:
OK, went through and tracked this down a bit more, seems like there was just a chain of dirty node_modules that I had to rebuild in order to get up to date with: https://github.com/expo/expo-cli/pull/3514/files

PR #3530, opened 2021-05-31
[config-plugins] Support android.jsEngine in app.json

Kudo:
# Why

to support hermes for managed workflow
closes [ENG-1222](https://linear.app/expo/issue/ENG-1222)

# How

add config-plugins and prebuild-configs to read `android.jsEngine` from app.json and update `expo.jsEngine` in `gradle.properties`

# Test Plan

### Unit Test

```
packages/config-plugins/src/android/__tests__/JsEngine-test.ts
  jsEngine
    ✅  returns default engine if no engine is provided (2 ms)
    ✅  return the engine if provided
    ✅  set the property if no property is existed (2 ms)
    ✅  overwrite the property if an old property is existed
```

### DIFF
--- a/packages/config-plugins/src/android/Engine.ts
+++ b/packages/config-plugins/src/android/Engine.ts
@@ -4,7 +4,7 @@ import type { ConfigPlugin } from '../Plugin.types';
 import { withGradleProperties } from '../plugins/android-plugins';
 import type { PropertiesItem } from './Properties';
 
-export const GRADLE_PROP_KEY = 'JS_RUNTIME';
+export const GRADLE_PROP_KEY = 'expo.jsEngine';
 export const DEFAULT_ENGINE = 'hermes';
 
 export const withEngine: ConfigPlugin = config => {
--- a/packages/config-plugins/src/android/Engine.ts
+++ b/packages/config-plugins/src/android/Engine.ts
@@ -4,7 +4,7 @@ import type { ConfigPlugin } from '../Plugin.types';
 import { withGradleProperties } from '../plugins/android-plugins';
 import type { PropertiesItem } from './Properties';
 
-export const GRADLE_PROP_KEY = 'expo.jsEngine';
+export const ENGINE_PROP_KEY = 'expo.jsEngine';
 export const DEFAULT_ENGINE = 'jsc';
 
 export const withEngine: ConfigPlugin = config => {
@@ -23,11 +23,11 @@ export function setEngine(
   gradleProperties: PropertiesItem[]
 ): PropertiesItem[] {
   const oldPropIndex = gradleProperties.findIndex(
-    prop => prop.type === 'property' && prop.key === GRADLE_PROP_KEY
+    prop => prop.type === 'property' && prop.key === ENGINE_PROP_KEY
   );
   const newProp: PropertiesItem = {
     type: 'property',
-    key: GRADLE_PROP_KEY,
+    key: ENGINE_PROP_KEY,
     value: config.android?.engine ?? DEFAULT_ENGINE,
   };
 
--- a/packages/config-plugins/src/android/Engine.ts
+++ b/packages/config-plugins/src/android/Engine.ts
@@ -5,7 +5,7 @@ import { withGradleProperties } from '../plugins/android-plugins';
 import type { PropertiesItem } from './Properties';
 
 export const GRADLE_PROP_KEY = 'expo.jsEngine';
-export const DEFAULT_ENGINE = 'hermes';
+export const DEFAULT_ENGINE = 'jsc';
 
 export const withEngine: ConfigPlugin = config => {
   return withGradleProperties(config, config => {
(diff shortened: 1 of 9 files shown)


Kudo:
would like to further discussion some questions here:

1. `android.engine` in app.json
considering web does not support engine replacement and developers might prefer using javascriptcore for ios but hermes for android. that's why i am proposing to add the new field inside android section.
the pr for updating xdl schema will create after this design is confirmed.
for expo go or classic `expo build`, this field is not supported. is it okay?

2. `JS_RUNTIME` in gradle.properties
as @EvanBacon mentioned, the formal format of gradle.properties is `foo.bar=true`.
should we change to `expo.jsEngine=hermes` ?

3. more about gradle.properties
for user once ejected before sdk42, their `enableHermes` in `android/app/build.gradle` does not reference to gradle properties. just to confirm that we will not do anything for this because it's already ejected, right?

packages/config-types/src/ExpoConfig.ts
```diff
@@ -689,6 +689,10 @@ export interface Android {
    * Determines how the software keyboard will impact the layout of your application. This maps to the `android:windowSoftInputMode` property. Defaults to `resize`. Valid values: `resize`, `pan`.
    */
   softwareKeyboardLayoutMode?: 'resize' | 'pan';
+  /**
+   * Specifies the JavaScript engine. Defaults to `hermes`. Valid values: `hermes`, `javascriptcore`.
+   */
+  engine?: 'hermes' | 'javascriptcore';
```
brentvatne, line 695:
```suggestion
  engine?: 'hermes' | 'jsc';
```

i have a slight preference for jsc, given that it's the canonical abbreviated form of JavaScriptCore

packages/config-types/src/ExpoConfig.ts
```diff
@@ -689,6 +689,10 @@ export interface Android {
    * Determines how the software keyboard will impact the layout of your application. This maps to the `android:windowSoftInputMode` property. Defaults to `resize`. Valid values: `resize`, `pan`.
    */
   softwareKeyboardLayoutMode?: 'resize' | 'pan';
+  /**
+   * Specifies the JavaScript engine. Defaults to `hermes`. Valid values: `hermes`, `javascriptcore`.
```
brentvatne, line 693:
```suggestion
   * Specifies the JavaScript engine. Defaults to `hermes`. Valid values: `hermes`, `javascriptcore`.
```

should default to jsc for now

brentvatne:
> 1. `android.engine` in app.json
>    considering web does not support engine replacement and developers might prefer using javascriptcore for ios but hermes for android. that's why i am proposing to add the new field inside android section.
>    the pr for updating xdl schema will create after this design is confirmed.

yeah i think this makes sense. maybe the best way to start with this feature is to have `engine` only on `android` for now. when we add support for ios also, we can consider adding a top level `engine` key that will apply to both platforms unless overridden by platform specific key (this is how `splash` works, for example).

>    for expo go or classic `expo build`, this field is not supported. is it okay?

yup. we'll have to be sure to make it clear in any of our documentation around this. if documentation doesn't end up being enough, we can add warnings when loading an app that uses hermes in expo go in development.

> 2. `JS_RUNTIME` in gradle.properties
>    as @EvanBacon mentioned, the formal format of gradle.properties is `foo.bar=true`.
>    should we change to `expo.jsEngine=hermes` ?

i like this idea!

> 3. more about gradle.properties
>    for user once ejected before sdk42, their `enableHermes` in `android/app/build.gradle` does not reference to gradle properties. just to confirm that we will not do anything for this because it's already ejected, right?

right. they can upgrade and regenerate their project with `expo prebuild --clean` if they want to, but otherwise they'll need to update the field manually

Kudo:
> > 2. `JS_RUNTIME` in gradle.properties
> >    as @EvanBacon mentioned, the formal format of gradle.properties is `foo.bar=true`.
> >    should we change to `expo.jsEngine=hermes` ?
> 
> i like this idea!
>
`expo.jsEngine` PR is here: https://github.com/expo/expo/pull/13120
xdl-schemas update PR is at https://github.com/expo/universe/pull/7656

There are further action items for this PR, please help to review if anything need to be updated.

packages/config-plugins/src/android/Engine.ts
```diff
@@ -0,0 +1,41 @@
+import type { ExpoConfig } from '@expo/config-types';
+
+import type { ConfigPlugin } from '../Plugin.types';
+import { withGradleProperties } from '../plugins/android-plugins';
+import type { PropertiesItem } from './Properties';
+
+export const GRADLE_PROP_KEY = 'expo.jsEngine';
```
EvanBacon, line 7:
```suggestion
export const ENGINE_PROP_KEY = 'expo.jsEngine';
```
or somn like that

EvanBacon (approved):
Nice work

Kudo:
thanks for the review. sorry also to update a missing javascriptcore -> jsc commit.
will wait for xdl-schemas pr landed and merge this.

PR #3532, opened 2021-06-01
[xdl] Prepend manifest handler

brentvatne:
# Why

https://github.com/expo/expo/issues/13114 reported that serving an app in Expo Go doesn't work when there is an index.html in project root. I looked into it and it turns out that this is because static file serving from the React Native CLI server runs before our manifest handler. So, this is a little hack to move our manifest handler middleware to the front of the middleware stack.

# How

I'm not sure `stack` is meant to be part of the public API but looking at [the source code for `connect`](https://github.com/senchalabs/connect/blob/master/index.js) this seems fine to use.

# Test Plan

Follow steps from repro for https://github.com/expo/expo/issues/13114 then use Expo CLI with this patch.


Kudo (approved):
looks like there are no other ways to prepend a middleware but just to manipulate internal stack.
works for me for the reason.

EvanBacon (approved):
haha, that sucks. Maybe a src helps to avoid this in the future.

PR #3536, opened 2021-06-01
[expo-cli] Enhance Hermes support for expo export and publish

Kudo:
# Why

this pr addresses some missing parts of hermes support.
closes [ENG-1223](https://linear.app/expo/issue/ENG-1223)

# How

1. add hermes bytecode bundle to `printBundleSizes`.
2. add hermes bytecode bundle support for `expo publish`.
3. support managed workflow by reading the `android.jsEngine` config in app.json.
4. `android.jsEngine` in app.json is now the single source of truth to determine  if we need to generate hermes bytecode bundle.
5. introduce `maybeInconsistentEngineAsync()` to show an error prompt if there are inconsistencies between app config and native project.
Screen Shot 2021-06-08 at 4 32 59 PM

# Test Plan

1. expo export + managed workflow app with `android.jsEngine=hermes` in app.json.
    make sure exported assets has hermes bytecode bundle.

2. expo publish + managed workflow app with `android.jsEngine=hermes` in app.json.
    download the asset and make sure it is a hermes bytecode bundle.

### Unit Test

```
packages/dev-server/src/__tests__/HermesBundler-test.ts
  maybeInconsistentEngineAsync - android
    ✅  should return false for managed project
    ✅  should return false if "enableHermes: true" in app/build.gradle and "jsEngine: 'hermes'" in app.json (1 ms)
    ✅  should return true if "enableHermes: false" in app/build.gradle and "jsEngine: 'hermes'" in app.json
    ✅  should return false if "expo.jsEngine=hermes" in gradle.properties and "jsEngine: 'hermes'" in app.json (2 ms)
    ✅  should return false for default sdk41 bare project
    ✅  should return true for default sdk41 bare project but user enabled hermes manually
    ✅  should handle the inconsistency between app/build.gradle and gradle.properties
```

### DIFF
--- a/packages/dev-server/src/HermesBundler.ts
+++ b/packages/dev-server/src/HermesBundler.ts
@@ -15,14 +15,20 @@ export async function shouldBuildHermesBundleAsync(
 ): Promise<boolean> {
   if (target === 'managed') {
     const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
-    return platform === 'android' && exp.android?.jsEngine === 'hermes';
+    switch (platform) {
+      case 'android':
+        return exp.android?.jsEngine === 'hermes';
+      default:
+        return false;
+    }
   }
 
-  if (platform === 'android') {
-    const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
-    if (fs.existsSync(gradlePropertiesPath)) {
-      const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
-      return properties['expo.jsEngine'] === 'hermes';
+  if (target === 'bare') {
+    switch (platform) {
+      case 'android':
+        return isHermesEnabledForBareAndroidAsync(projectRoot);
+      default:
+        return false;
     }
   }
 
@@ -120,6 +126,15 @@ export function parseGradleProperties(content: string): Record<string, string> {
     result[key] = value;
   }
   return result;
+}
+
+async function isHermesEnabledForBareAndroidAsync(projectRoot: string): Promise<boolean> {
+  const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
+  if (fs.existsSync(gradlePropertiesPath)) {
+    const props = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
+    return props['export.jsEngine'] === 'hermes';
+  }
+  return false;
 }
 
 function importMetroSourceMapFromProject(projectRoot: string): typeof composeSourceMaps {
--- a/packages/dev-server/src/HermesBundler.ts
+++ b/packages/dev-server/src/HermesBundler.ts
@@ -8,31 +8,14 @@ import path from 'path';
 import process from 'process';
 import resolveFrom from 'resolve-from';
 
-export async function shouldBuildHermesBundleAsync(
-  projectRoot: string,
-  target: ProjectTarget,
-  platform: Platform
-): Promise<boolean> {
-  if (target === 'managed') {
-    const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
-    switch (platform) {
-      case 'android':
-        return exp.android?.jsEngine === 'hermes';
-      default:
-        return false;
-    }
+export function shouldBuildHermesBundle(projectRoot: string, platform: Platform): boolean {
+  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  switch (platform) {
+    case 'android':
+      return exp.android?.jsEngine === 'hermes';
+    default:
+      return false;
   }
-
-  if (target === 'bare') {
-    switch (platform) {
-      case 'android':
-        return isHermesEnabledForBareAndroidAsync(projectRoot);
-      default:
-        return false;
-    }
-  }
-
-  return false;
 }
 
 interface HermesBundleOutput {
--- a/packages/dev-server/src/HermesBundler.ts
+++ b/packages/dev-server/src/HermesBundler.ts
@@ -1,4 +1,5 @@
-import type { Platform } from '@expo/config';
+import type { Platform, ProjectTarget } from '@expo/config';
+import { getConfig } from '@expo/config';
 import spawnAsync from '@expo/spawn-async';
 import fs from 'fs-extra';
 import type { composeSourceMaps } from 'metro-source-map';
@@ -9,17 +10,21 @@ import resolveFrom from 'resolve-from';
 
 export async function shouldBuildHermesBundleAsync(
   projectRoot: string,
+  target: ProjectTarget,
   platform: Platform
 ): Promise<boolean> {
+  if (target === 'managed') {
+    const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+    return platform === 'android' && exp.android?.jsEngine === 'hermes';
+  }
+
   if (platform === 'android') {
     const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
-    if (!fs.existsSync(gradlePropertiesPath)) {
-      // TODO(kudo): Clarify default behavior in managed workflow for gradle.properties not exist
-      return false;
-    }
-    const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
-    if (properties['expo.jsEngine'] === 'hermes') {
-      return true;
+    if (fs.existsSync(gradlePropertiesPath)) {
+      const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
+      if (properties['expo.jsEngine'] === 'hermes') {
+        return true;
+      }
     }
   }
   return false;
--- a/packages/dev-server/src/HermesBundler.ts
+++ b/packages/dev-server/src/HermesBundler.ts
@@ -22,11 +22,10 @@ export async function shouldBuildHermesBundleAsync(
     const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
     if (fs.existsSync(gradlePropertiesPath)) {
       const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
-      if (properties['expo.jsEngine'] === 'hermes') {
-        return true;
-      }
+      return properties['expo.jsEngine'] === 'hermes';
     }
   }
+
   return false;
 }
 
--- a/packages/dev-server/src/HermesBundler.ts
+++ b/packages/dev-server/src/HermesBundler.ts
@@ -1,5 +1,4 @@
-import type { Platform, ProjectTarget } from '@expo/config';
-import { getConfig } from '@expo/config';
+import type { ExpoConfig, Platform } from '@expo/config';
 import spawnAsync from '@expo/spawn-async';
 import fs from 'fs-extra';
 import type { composeSourceMaps } from 'metro-source-map';
@@ -8,11 +7,10 @@ import path from 'path';
 import process from 'process';
 import resolveFrom from 'resolve-from';
 
-export function shouldBuildHermesBundle(projectRoot: string, platform: Platform): boolean {
-  const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+export function isEnableHermesManaged(expoConfig: ExpoConfig, platform: Platform): boolean {
   switch (platform) {
     case 'android':
-      return exp.android?.jsEngine === 'hermes';
+      return expoConfig.android?.jsEngine === 'hermes';
     default:
       return false;
   }
@@ -111,12 +109,41 @@ export function parseGradleProperties(content: string): Record<string, string> {
   return result;
 }
 
-async function isHermesEnabledForBareAndroidAsync(projectRoot: string): Promise<boolean> {
+export async function maybeInconsistentEngineAsync(
+  projectRoot: string,
+  platform: string,
+  isHermesManaged: boolean
+): Promise<boolean> {
+  switch (platform) {
+    case 'android':
+      return maybeInconsistentEngineAndroidAsync(projectRoot, isHermesManaged);
+    default:
+      return false;
+  }
+}
+
+async function maybeInconsistentEngineAndroidAsync(
+  projectRoot: string,
+  isHermesManaged: boolean
+): Promise<boolean> {
+  // Trying best to check android native project if by chance to be consistent between app config
+
+  // Check android/app/build.gradle for "enableHermes: true"
+  const appBuildGradlePath = path.join(projectRoot, 'android', 'app', 'build.gradle');
+  if (fs.existsSync(appBuildGradlePath)) {
+    const content = await fs.readFile(appBuildGradlePath, 'utf8');
+    const isHermesBare = content.search(/^\s*enableHermes:\s*true,?\s+/m) >= 0;
+    return isHermesManaged !== isHermesBare;
+  }
+
+  // Check gradle.properties from prebuild template
   const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
   if (fs.existsSync(gradlePropertiesPath)) {
     const props = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
-    return props['export.jsEngine'] === 'hermes';
+    const isHermesBare = props['expo.jsEngine'] === 'hermes';
+    return isHermesManaged !== isHermesBare;
   }
+
   return false;
 }
 
--- a/packages/dev-server/src/MetroDevServer.ts
+++ b/packages/dev-server/src/MetroDevServer.ts
@@ -1,4 +1,5 @@
 import Log from '@expo/bunyan';
+import type { ProjectTarget } from '@expo/config';
 import * as ExpoMetroConfig from '@expo/metro-config';
 import {
   createDevServerMiddleware,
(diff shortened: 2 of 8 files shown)


packages/xdl/src/project/createBundlesAsync.ts
```diff
@@ -71,18 +113,15 @@ export async function createBundlesAsync(
...
-  }
+  const config = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const isLegacy = isLegacyImportsEnabled(config.exp);
   const platforms: Platform[] = ['android', 'ios'];
   const [android, ios] = await bundleAsync(
     projectRoot,
+    publishOptions.target ?? getDefaultTarget(projectRoot),
     {
-      target: publishOptions.target,
+      // If not legacy, delete the target option to prevent warnings from being thrown.
```
jkhales, line 123:
```suggestion
      // If not legacy, ignore the target option to prevent warnings from being thrown.
```

packages/xdl/src/project/createBundlesAsync.ts
```diff
@@ -71,18 +113,15 @@ export async function createBundlesAsync(
...
-  // If not legacy, delete the target option to prevent warnings from being thrown.
-  if (!isLegacy) {
-    delete publishOptions.target;
-  }
+  const config = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const isLegacy = isLegacyImportsEnabled(config.exp);
   const platforms: Platform[] = ['android', 'ios'];
   const [android, ios] = await bundleAsync(
     projectRoot,
+    publishOptions.target ?? getDefaultTarget(projectRoot),
```
jkhales, line 121:
```suggestion
```
typo?

packages/xdl/src/project/publishAsync.ts
```diff
@@ -93,17 +93,17 @@ export async function publishAsync(
     useDevServer: Env.shouldUseDevServer(exp),
   });
 
-  printBundleSizes(bundles);
+  printBundleSizes(bundles, true);
```
jkhales, line 96:
```suggestion
  printBundleSizes(bundles,  /* isHermesPreferable* / true);
```

packages/xdl/src/project/publishAsync.ts
```diff
@@ -93,17 +93,17 @@ export async function publishAsync(
...
 
   await ProjectAssets.publishAssetsAsync({ projectRoot, exp, bundles });
 
-  const androidBundle = bundles.android.code;
-  const iosBundle = bundles.ios.code;
+  const androidBundle = bundles.android.hermesBytecodeBundle || bundles.android.code;
+  const iosBundle = bundles.ios.hermesBytecodeBundle || bundles.ios.code;
```
jkhales, line 101:
```suggestion
  const androidBundle = bundles.android.hermesBytecodeBundle ?? bundles.android.code;
  const iosBundle = bundles.ios.hermesBytecodeBundle ?? bundles.ios.code;
```

packages/xdl/src/project/publishAsync.ts
```diff
@@ -93,17 +93,17 @@ export async function publishAsync(
...
 
   const hasHooks = validPostPublishHooks.length > 0;
 
-  const androidSourceMap = hasHooks ? bundles.android.map : null;
-  const iosSourceMap = hasHooks ? bundles.ios.map : null;
+  const androidSourceMap = hasHooks ? bundles.android.hermesSourcemap || bundles.android.map : null;
+  const iosSourceMap = hasHooks ? bundles.ios.hermesSourcemap || bundles.ios.map : null;
```
jkhales, line 106:
```suggestion
  const androidSourceMap = hasHooks ? bundles.android.hermesSourcemap ?? bundles.android.map : null;
  const iosSourceMap = hasHooks ? bundles.ios.hermesSourcemap ?? bundles.ios.map : null;
```

packages/xdl/src/project/createBundlesAsync.ts
```diff
@@ -24,17 +24,59 @@ type PackagerOptions = {
...
+// If `isHermesPreferable` is true, will try use hermes bytecode bundle as primary output.
+// This if for classic `expo export` or `expo publish` flows.
+export function printBundleSizes(
+  bundles: { android: BundleOutput; ios: BundleOutput },
+  isHermesPreferable: boolean
+) {
+  const files: [string, string | Uint8Array][] = [];
+
+  if (isHermesPreferable) {
+    if (bundles.ios.hermesBytecodeBundle) {
```
jkhales, line 36:
```suggestion
  if (isHermesPreferable && bundles.ios.hermesBytecodeBundle) {
```
If we do this for each of the if/else statements, can we get rid of half of these?

packages/xdl/src/project/createBundlesAsync.ts
```diff
@@ -24,17 +24,59 @@ type PackagerOptions = {
...
+    if (bundles.android.hermesSourcemap) {
+      files.push([chalk.dim('index.android.js.map (Hermes)'), bundles.android.hermesSourcemap]);
+    } else if (bundles.android.map) {
+      files.push([chalk.dim('index.android.js.map'), bundles.android.map]);
+    }
+  } else {
+    files.push(['index.ios.js', bundles.ios.code], ['index.android.js', bundles.android.code]);
+    if (bundles.ios.hermesBytecodeBundle) {
+      files.push(['index.ios.js (Hermes)', bundles.ios.hermesBytecodeBundle]);
+    }
```
jkhales, line 62: Why are we pushing a `hermes` bundle here if `isHermesPreferable === false`?

packages/expo-cli/src/commands/export/exportAppAsync.ts
```diff
@@ -97,7 +97,7 @@ export async function exportAppAsync(
     useDevServer: Env.shouldUseDevServer(exp),
   });
 
-  printBundleSizes(bundles);
+  printBundleSizes(bundles, true);
```
jkhales, line 100:
```suggestion
  printBundleSizes(bundles, /* isHermesPreferable */ true);
```

jkhales (approved):
✨

packages/dev-server/src/HermesBundler.ts
```diff
@@ -9,17 +10,21 @@ import resolveFrom from 'resolve-from';
 
 export async function shouldBuildHermesBundleAsync(
   projectRoot: string,
+  target: ProjectTarget,
   platform: Platform
 ): Promise<boolean> {
+  if (target === 'managed') {
+    const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+    return platform === 'android' && exp.android?.jsEngine === 'hermes';
+  }
```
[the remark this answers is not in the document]
Kudo: Once tried but later feel the original code might be clear after adding ios support.
Sorry for editing your suggestion code b/c trying to commit the suggestion before.

```ts
  if (target === 'managed') {
    switch (platform) {
      case 'android':
        return exp.android?.jsEngine === 'hermes';
      case 'ios':
        return exp.ios?.jsEngine === 'hermes';
      default:
        return false;
    }
  }

  // bare workflow

  if (platform === 'ios') {
    // Read settings somewhere or maybe directly from Podfile
    if (fs.existsSync(...)) {
      const settings = parseIosSettings(...); 
      return settings['jsEngine'] === 'hermes';
    }
  }

  if (platform === 'android') {
    const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
    if (fs.existsSync(gradlePropertiesPath)) {
      const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
      return properties['expo.jsEngine'] === 'hermes';
    }
  }
  return false;
```

v.s.

```ts
  if (platform === 'web') {
    return false;
  }

  if (platform === 'ios') {
    if (target === 'managed') {
      return exp.ios?.jsEngine === 'hermes';
    }

    // Read settings somewhere or maybe directly from Podfile
    if (fs.existsSync(...)) {
      const settings = parseIosSettings(...); 
      return settings['jsEngine'] === 'hermes';
    }
    return false;
  }

  // platform === 'android'
  if (target === 'managed') {
    return exp.android?.jsEngine === 'hermes';
  }

  const gradlePropertiesPath = path.join(projectRoot, 'android', 'gradle.properties');
  if (fs.existsSync(gradlePropertiesPath)) {
    const properties = parseGradleProperties(await fs.readFile(gradlePropertiesPath, 'utf8'));
    return properties['expo.jsEngine'] === 'hermes';
  }

  return false;
```

packages/xdl/src/project/createBundlesAsync.ts
Kudo: if the else block is needed, this style may not benefit a lot ?
let me answer why we need the else block below.

packages/xdl/src/project/createBundlesAsync.ts
Kudo: this is for eas update. saying that we will upload both the plaintext js bundle and hermes bytecode bundle to our servers.
is the design making sense?

personally, i am not sure if we need to do this in the meantime.
some benefit foreseeing is for dev-clients to load plaintext javascript bundle if hermes bytecode version is incompatible.
cc @brentvatne @ide 

if we don't need design, the code here could be simpler.

Kudo (commented):
thanks for the great review. learn to write clearer code.
some uncertain comments replied inline. please take a look.
thanks.

packages/xdl/src/project/createBundlesAsync.ts
jkhales: Sounds good. Agreed that this comment only applies if we can get rid of the else block.

packages/dev-server/src/HermesBundler.ts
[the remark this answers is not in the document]
jkhales: both of those alternatives seem nice to me. The one suggestion I would have would be to change the comments, `// bare workflow`, `// platform === 'android'` into `if` or `switch` clauses. 

This would keep the code symmetric between bare/managed and ios/android and also demarcate the relevant code block.

packages/xdl/src/project/createBundlesAsync.ts
jkhales: I believe the last time we (@brentvatne @ide @esamelson) discussed this, we were going to avoid fallback to js due to the complexity and leave it to the client to bump their runtime version correctly. Would be curious to know if I misunderstood this though.

packages/dev-server/src/HermesBundler.ts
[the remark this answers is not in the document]
Kudo: make sense. let me update the code accordingly.

packages/xdl/src/project/createBundlesAsync.ts
Kudo: just revisited our discussion and yes you are right.
sorry for getting lost from the beginning.

packages/xdl/src/project/createBundlesAsync.ts
Kudo: let me remove `isHermesPreferable` entirely as we don't need both hermes bytecode and plaintext js bundle.

packages/xdl/src/project/createBundlesAsync.ts
Kudo: not really a typo.
this target here is for HermesBundler to determine workflows.
the target below is for [metro-config](https://github.com/expo/expo-cli/blob/a162e4d4166736cbb824b05ea8175bd58a971433/packages/metro-config/src/ExpoMetroConfig.ts#L72-L82) which seems to be removed in the future.

Kudo:
@jkhales just updated the code. hopefully it's getting clear. please take a look if you get a chance. thanks for your review.

packages/xdl/src/project/createBundlesAsync.ts
jkhales: ah ok i see.

jkhales (approved):
looks great 😎

brentvatne:
@Kudo - i think it is preferable to only include one type of bundled, js or hbc. if a developer has indicated that their app uses hermes, then we should only try to run their app on hermes, so we don't need to upload both js and hbc.

packages/xdl/src/project/createBundlesAsync.ts
```diff
@@ -71,18 +84,15 @@ export async function createBundlesAsync(
...
-  // If not legacy, delete the target option to prevent warnings from being thrown.
-  if (!isLegacy) {
-    delete publishOptions.target;
-  }
+  const config = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const isLegacy = isLegacyImportsEnabled(config.exp);
   const platforms: Platform[] = ['android', 'ios'];
   const [android, ios] = await bundleAsync(
     projectRoot,
+    publishOptions.target ?? getDefaultTarget(projectRoot),
```
brentvatne, line 92: ideally i'd like to avoid adding any new uses for the target flag. what we have been doing in similar cases to this instead is we:

- check for the existence of the ios/android project
- use that to read the information we need if it's available
- otherwise we read from expo app config.

packages/dev-server/src/HermesBundler.ts
```diff
@@ -9,19 +10,28 @@ import resolveFrom from 'resolve-from';
 
 export async function shouldBuildHermesBundleAsync(
   projectRoot: string,
+  target: ProjectTarget,
```
brentvatne, line 13: let's skip switching behavior based on target and instead try to read from native project before app config, as described in comment below

brentvatne (changes requested):

packages/xdl/src/project/createBundlesAsync.ts
Kudo: based on this, `expo publish --target` and `expo export --target` will not be supported, at least for hermes support.
is this right?

brentvatne: correct!

brentvatne: if anything is unclear (i imagine it is) then let's chat about it tomorrow :)

Kudo: thanks for the briefly explanation. 
that would be fine if we don't support the `--target` which caused confusion and make the code clearer.

just thinking about app config as single source of truth for a long time.
as mentioned, the assumption for user to setup native project properly will be a problem.
an idea is that we could show a warning during export/publish if finding inconsistencies between native project and app config.
e.g. app.json setups as hermes but gradle.properties setup for jsc, we could show a warning prompt.

let's talk about this later.

Kudo: introduced the `maybeInconsistentEngineAsync()` to show the error prompt.
also updated the PR description with screenshot.
please help to review again. thanks.

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -162,12 +168,24 @@ export async function bundleAsync(
...
+
+        const maybeInconsistentEngine = await maybeInconsistentEngineAsync(
           projectRoot,
-          bundle.platform
+          bundle.platform,
+          isHermesManaged
         );
+        if (maybeInconsistentEngine) {
+          throw new Error(
+            `Inconsistent Hermes settings between app config and ${bundle.platform} native project.\n` +
```
brentvatne, line 180: let's use proper casing for this - `android` -> `Android`, `ios` -> `iOS`

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -162,12 +168,24 @@ export async function bundleAsync(
...
+              `Enabling Hermes in app.json or app.config.js: ${isHermesManaged.toString()}\n` +
```
brentvatne, line 181: let's use [getConfigFilePaths](https://github.com/expo/expo-cli/blob/a02c8997980d59007478d2c29a792276962147c3/packages/config/src/Config.ts#L257-L272) to output the config file used. if dynamic config path is available, we specify that, otherwise we use the static path

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -162,12 +168,24 @@ export async function bundleAsync(
...
+              `Please check the following native project settings if there are any inconsistencies:\n` +
+              '  - android/gradle.properties\n' +
+              '  - android/app/build.gradle\n'
+          );
```
brentvatne, line 185:
```suggestion
            `JavaScript engine configuration is inconsistent between ${configFileName) and ${bundle.platform} native project.\n` +
              `In ${configFileName}: Hermes is ${isHermesManaged ? 'enabled' : 'not enabled'}\n` +
              `In ${bundle.platform} native project: Hermes is ${isHermesManaged ? 'not enabled' : 'enabled'}\n`
              `Please check the following files for inconsistencies:\n` +
              '  - android/gradle.properties\n' +
              '  - android/app/build.gradle\n'
          );
```

what do you think about this? i think it reads a bit more clearly. `configFileName` refers to the other comment abou `getConfigFilePaths`

brentvatne (approved):

brentvatne (approved):
👌

Kudo:
let me merge this first. will create another pr if a document link to expo.fyi is ready.

zhigang1992:
hey guys, if I add android.jsEngine, then it throws

image

without it, it throws

image

my `expo-cli` is on version `4.7.1`

Kudo:
@zhigang1992 this feature is supported for SDK 42 or higher.
and [SDK 42 is just in beta now](https://blog.expo.io/expo-sdk-42-beta-is-now-available-e505770389e7).
if you feel comfortable to try SDK 42, please try `EXPO_BETA=1 expo upgrade` to upgrade your app.

zhigang1992:
@Kudo but it's give me error when I upgrade the `expo-cli`. I have not upgraded to SDK 42. and I'm using bare flow.

Should the error in second screenshot protect against old version of SDKs?

brentvatne:
@zhigang1992 - you can use an older version of expo-cli for now, we'll add version validation to this!

PR #3547, opened 2021-06-04
[expo-cli][config][xdl] support runtimeVersion in classic updates

jkhales:
# Why

Support `runtimeVersion` in classic updates. 
Companion PR to https://github.com/expo/universe/pull/7663.

If the user wishes to publish with a `runtimeVersion`, they can add it to the app.json and the `sdkVersion` will be ignored.

Screen Shot 2021-06-04 at 12 58 56 PM

If the user wishes to go back to `sdkVersion`, they just need to delete the `runtimeVersion` key from the app.json.

# How

 - Made @expo/config not define `sdkversion` <-> `runtimeVersion` is not defined in the config.
   - `@expo/config.getConfig` is called here: https://github.com/expo/expo-cli/blob/master/packages/xdl/src/project/getPublishExpConfigAsync.ts#L33 when we publish an update.
 - In expo-cli, updated types and added extra column where necessary.

# Test Plan

Tested in demo repo against https://github.com/expo/universe/pull/7663 running on docker.

- [x] expo publish
- [x] expo publish:details
With only sdkVersion:
  
Screen Shot 2021-06-04 at 12 51 28 PM
With only runtimeVersion:
    
    
Screen Shot 2021-06-04 at 12 51 58 PM
With both:
  
Screen Shot 2021-06-04 at 12 52 26 PM

- [x] expo publish:set

Screen Shot 2021-06-04 at 12 52 55 PM

- [x] expo publish:details

Screen Shot 2021-06-04 at 12 53 49 PM

### DIFF
--- a/packages/xdl/src/project/getPublishExpConfigAsync.ts
+++ b/packages/xdl/src/project/getPublishExpConfigAsync.ts
@@ -30,11 +30,20 @@ export async function getPublishExpConfigAsync(
   options.releaseChannel = options.releaseChannel || 'default';
 
   // Verify that exp/app.json and package.json exist
-  const { exp: privateExp } = getConfig(projectRoot);
-  const { hooks } = privateExp;
-  const { exp, pkg } = getConfig(projectRoot, { isPublicConfig: true });
-
+  const {
+    exp: { hooks, runtimeVersion },
+  } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+  const { exp, pkg } = getConfig(projectRoot, {
+    isPublicConfig: true,
+    // enforce sdk validation if user is not using runtimeVersion
+    skipSDKVersionRequirement: !!runtimeVersion,
+  });
   const { sdkVersion } = exp;
+  if (!sdkVersion) {
+    throw new Error(
+      'Config is missing an SDK version. See https://docs.expo.io/bare/installing-updates/'
+    );
+  }
 
   // Only allow projects to be published with UNVERSIONED if a correct token is set in env
   if (sdkVersion === 'UNVERSIONED' && !Env.maySkipManifestValidation()) {
--- a/packages/xdl/src/project/getPublishExpConfigAsync.ts
+++ b/packages/xdl/src/project/getPublishExpConfigAsync.ts
@@ -39,12 +39,6 @@ export async function getPublishExpConfigAsync(
     skipSDKVersionRequirement: !!runtimeVersion,
   });
   const { sdkVersion } = exp;
-  if (!sdkVersion) {
-    throw new Error(
-      'Config is missing an SDK version. See https://docs.expo.io/bare/installing-updates/'
-    );
-  }
-
   // Only allow projects to be published with UNVERSIONED if a correct token is set in env
   if (sdkVersion === 'UNVERSIONED' && !Env.maySkipManifestValidation()) {
     throw new XDLError('INVALID_OPTIONS', 'Cannot publish with sdkVersion UNVERSIONED.');
--- a/packages/expo-cli/src/commands/publish/getUsageAsync.ts
+++ b/packages/expo-cli/src/commands/publish/getUsageAsync.ts
@@ -52,7 +52,7 @@ async function _getUsageAsync(projectRoot: string): Promise<string> {
   const timeDifferenceString = _getTimeDifferenceString(new Date(), new Date(publishedTime));
 
   return (
-    `--release-channel and --sdk-version arguments are required. \n` +
+    `--release-channel and either --sdk-version or --runtime-version arguments are required. \n` +
     `For example, to roll back the revision [${revisionId}] on release channel [${channel}] (published ${timeDifferenceString}), \n` +
     `run: expo publish:rollback --release-channel ${channel} --sdk-version ${sdkVersion}`
   );
@@ -84,7 +84,7 @@ function _getTimeDifferenceString(t0: Date, t1: Date): string {
 
 function _getGenericUsage(): string {
   return (
-    `--release-channel and --sdk-version arguments are required. \n` +
+    `--release-channel and either --sdk-version or --runtime-version arguments are required. \n` +
     `For example, to roll back the latest publishes on the default channel for sdk 37.0.0, \n` +
     `run: expo publish:rollback --release-channel defaul --sdk-version 37.0.0 \n` +
     `To rollback a specific platform, use the --platform flag.`
--- a/packages/expo-cli/src/commands/publish/publishHistoryAsync.ts
+++ b/packages/expo-cli/src/commands/publish/publishHistoryAsync.ts
@@ -5,6 +5,7 @@ import { getPublishHistoryAsync, HistoryOptions, Publication } from '../utils/Pu
 import * as table from '../utils/cli-table';
 
 const HORIZ_CELL_WIDTH_SMALL = 15;
+const HORIZ_CELL_WIDTH_MEDIUM = 20;
 const HORIZ_CELL_WIDTH_BIG = 40;
 
 export async function actionAsync(projectRoot: string, options: HistoryOptions) {
@@ -26,11 +27,19 @@ export async function actionAsync(projectRoot: string, options: HistoryOptions) 
     );
     Log.log(generalTableString);
 
+    const hasRuntimeVersion = result.queryResult.some(
+      (publication: Publication) => !!publication.runtimeVersion
+    );
+    const hasSdkVersion = result.queryResult.some(
+      (publication: Publication) => !!publication.sdkVersion
+    );
+
     // Print info specific to each publication
     const headers = [
       'publishedTime',
       'appVersion',
-      'sdkVersion',
+      ...(hasSdkVersion ? ['sdkVersion'] : []),
+      ...(hasRuntimeVersion ? ['runtimeVersion'] : []),
       'platform',
       'channel',
       'publicationId',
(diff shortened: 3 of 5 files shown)


packages/config/src/Config.ts
```diff
@@ -473,10 +473,12 @@ function ensureConfigHasDefaultValues({
...
-    sdkVersion = getExpoSDKVersion(projectRoot, expWithDefaults);
-  } catch (error) {
-    if (!skipSDKVersionRequirement) throw error;
+  if (!exp.runtimeVersion) {
+    try {
+      sdkVersion = getExpoSDKVersion(projectRoot, expWithDefaults);
+    } catch (error) {
+      if (!skipSDKVersionRequirement) throw error;
+    }
   }
```
EvanBacon, line 482: We should always perform the SDK version check, if you don't want it then pass `skipSDKVersionRequirement` to `getConfig`.

packages/expo-cli/src/commands/export/mergeAppDistributions.ts
```diff
@@ -7,7 +7,7 @@ import { Project, XDLError } from 'xdl';
 
 import Log from '../../log';
 
-type SelfHostedIndex = ExpoAppManifest & {
+type SelfHostedIndex = ExpoAppManifestWithSdk & {
```
jkhales, line 10: Basing the sdk requirement off of this function: https://github.com/expo/expo-cli/pull/3547/files#diff-1a0c6311e2ebf725a99acb8988c676a4f260ca080d36bf4eee0dc8b617a0125fR15

packages/xdl/src/ProjectAssets.ts
```diff
@@ -334,18 +336,20 @@ async function collectAssets(
...
-      const absolutePath = path.resolve(projectRoot, assetPath);
-      const contents = await fs.readFile(absolutePath);
-      const hash = md5hex(contents);
-      manifestAssets.push({ files: [absolutePath], fileHashes: [hash], hash });
-      return urljoin(hostedAssetPrefix, hash);
-    },
-    strict: true,
-  });
+  if (exp.sdkVersion) {
+    await resolveManifestAssets({
```
jkhales, line 340: If you dig through `resolverManifestAssets` it ends with a call to https://github.com/expo/expo-cli/pull/3547/files#diff-96a934c1e14024d198b87db4ce862a393b9829d562f3d7f0e1c2048eb4153a62R63

which requires an `sdkVersion`

jkhales (commented):
I moved everthing back outside the `getConfig` and am using `skipSDKVersionRequirement`

This definitely feels a lot safer as only getPublishExpConfigAsync will potentially serve a config missing an `sdkVersion` now.

packages/xdl/src/start/ManifestHandler.ts
```diff
@@ -231,14 +231,16 @@ export async function getManifestResponseAsync({
...
   manifest.hostUri = await UrlUtils.constructHostUriAsync(projectRoot, hostname);
-  // Resolve all assets and set them on the manifest as URLs
-  await ProjectAssets.resolveManifestAssets({
-    projectRoot,
-    manifest: manifest as ExpoAppManifest,
-    async resolver(path) {
-      return manifest.bundleUrl!.match(/^https?:\/\/.*?\//)![0] + 'assets/' + path;
-    },
-  });
+  if (manifest.sdkVersion) {
```
EvanBacon, line 234: What happens in a local development project when the sdkVersion isn't used? are assets not available?

packages/xdl/src/project/getPublishExpConfigAsync.ts
```diff
@@ -30,22 +31,27 @@ export async function getPublishExpConfigAsync(
...
+  const { exp, pkg } = getConfig(projectRoot, {
+    isPublicConfig: true,
+    skipSDKVersionRequirement: !!runtimeVersion,
+  });
 
-  const { sdkVersion } = exp;
+  // we can't publish with both sdkVersion and runtimeVerson specified, so ensure we pick one or the other.
+  const xorExp = runtimeVersion
+    ? ({ ...exp, sdkVersion: undefined } as ExpoAppManifestWithRuntimeVersion)
+    : (exp as ExpoAppManifestWithSdk);
```
EvanBacon, line 45: Maybe it'd be worth writing a simple unit test on `getPublishExpConfigAsync` which provides two different types of configs (runtimeVersion, sdkVersion) and gets the expected results.

packages/xdl/src/start/ManifestHandler.ts
jkhales: I'm reworking this to try and include `sdkVersion` whenever possible. Namely I am doing (`sdk` OR `rtv`) instead of (`sdk` XOR `rtv`). A lot of context in #updates channel if you are curious.

What are the situations that sdkVersion would not be available? Are there any now that this PR is doing OR instead of XOR?

jkhales:
So, curious right now if there is a reasonable flow where we publish with *only* an RTV?

The two flows that makes sense to me are `SDK` only and `SDK + RTV`.

packages/expo-cli/src/commands/publish-modify.ts
```diff
@@ -61,14 +61,22 @@ export default function (program: Command) {
     .helpGroup('publish')
     .option('--channel-id <channel-id>', 'This flag is deprecated.')
     .option('-c, --release-channel <channel-name>', 'The channel to rollback from. (Required)')
-    .option('-s, --sdk-version <version>', 'The sdk version to rollback. (Required)')
+    .option(
+      '-s, --sdk-version <version>',
+      'The sdk version to rollback. (Required if runtime version is not specified)'
```
brentvatne, line 66:
```suggestion
      'The SDK version to rollback. (Required if runtime version is not specified)'
```

(not introduced in this PR but we tend to capitalize SDK)

packages/expo-cli/src/commands/publish-modify.ts
```diff
@@ -61,14 +61,22 @@ export default function (program: Command) {
...
+    )
+    .option(
+      '-rtv, --runtime-version <version>',
+      'The runtime version to rollback. (Required if sdk version is not specified)'
```
brentvatne, line 70:
```suggestion
      'The runtime version to rollback. (Required if SDK version is not specified)'
```

packages/expo-cli/src/commands/publish-modify.ts
```diff
@@ -139,7 +147,7 @@ async function _getUsageAsync(projectRoot: string): Promise<string> {
   const timeDifferenceString = _getTimeDifferenceString(new Date(), new Date(publishedTime));
 
   return (
-    `--release-channel and --sdk-version arguments are required. \n` +
+    `--release-channel and --sdk-version or --runtime-version arguments are required. \n` +
```
brentvatne, line 150:
```suggestion
    `--release-channel and either --sdk-version or --runtime-version arguments are required. \n` +
```

packages/expo-cli/src/commands/publish-modify.ts
```diff
@@ -171,7 +179,7 @@ function _getTimeDifferenceString(t0: Date, t1: Date): string {
 
 function _getGenericUsage(): string {
   return (
-    `--release-channel and --sdk-version arguments are required. \n` +
+    `--release-channel and --sdk-version or --runtime-version arguments are required. \n` +
```
brentvatne, line 182:
```suggestion
    `--release-channel and either --sdk-version or --runtime-version arguments are required. \n` +
```

packages/xdl/src/project/Doctor.ts
```diff
@@ -176,6 +176,11 @@ async function _validateExpJsonAsync(
   ProjectUtils.clearNotification(projectRoot, 'doctor-problem-checking-watchman-version');
 
   const sdkVersion = exp.sdkVersion;
+  if (!sdkVersion) {
+    throw new Error(
+      'Config is missing an SDK version. See https://docs.expo.io/bare/installing-updates/'
```
brentvatne, line 181: i'm not sure i understand this one -- why are we linking to the bare workflow instructions for installing expo-updates if sdk version is missing? is it still not possible to use expo-updates without setting an sdk version?

jkhales: I was assuming that a missing SDK version was evidence that they were trying to do something with a runtime version in bare.

This part of the PR was just me trying to improve validation and not necessary to the rest. So I think I will remove it as I don't have enough context about the issue.

packages/xdl/src/start/ManifestHandler.ts
```diff
@@ -231,6 +231,11 @@ export async function getManifestResponseAsync({
   manifest.debuggerHost = await UrlUtils.constructDebuggerHostAsync(projectRoot, hostname);
   manifest.logUrl = await UrlUtils.constructLogUrlAsync(projectRoot, hostname);
   manifest.hostUri = await UrlUtils.constructHostUriAsync(projectRoot, hostname);
+  if (!manifest.sdkVersion) {
+    throw new Error(
+      'Config is missing an SDK version. See https://docs.expo.io/bare/installing-updates/'
+    );
+  }
```
jkhales, line 238: removing this check as well.

packages/xdl/src/project/getPublishExpConfigAsync.ts
```diff
@@ -30,16 +30,26 @@ export async function getPublishExpConfigAsync(
...
+    isPublicConfig: true,
+    // enforce sdk validation if user is not using runtimeVersion
+    skipSDKVersionRequirement: !!runtimeVersion,
+  });
   const { sdkVersion } = exp;
+  if (!sdkVersion) {
+    throw new Error(
+      'Config is missing an SDK version. See https://docs.expo.io/bare/installing-updates/'
+    );
+  }
```
jkhales, line 46: This is included as although we no longer enforce SDK validation when using runtimeVersion, we still need the SDK version to be defined in order to prevent various expo libraries that require the SDK version to be defined from crashing.

brentvatne: i don't think any libraries require sdk version to be defined - what are some examples you had in mind?

jkhales: OK removed after slack convo.

brentvatne (approved):

PR #3549, opened 2021-06-05
Use expo run commands in native projects

EvanBacon:
# Why

- https://blog.expo.io/introducing-expo-run-commands-835ae8da4813
- https://github.com/expo/create-react-native-app/pull/885
- https://github.com/expo/expo/pull/13174


brentvatne:
let's merge this when you can do this successfully:

````
expo init project # pick bare template
cd project
expo run:ios
```

right now it will not work because no shared scheme. also, it will show a qr code with dev client prefix and have a bunch of dev client messaging, and we don't need to show any of this if not using dev client.

PR #3562, opened 2021-06-09
Remove bare typescript template

EvanBacon:
# Why

This template is out of sync with the other bare template, things like this https://github.com/expo/expo/pull/13216 will potentially cause confusion as the templates diverge.

# How

Remove `expo-template-bare-typescript` from init


brentvatne (approved):

hadnet:
So when will it come back?

hadnet:
Now when I choose minimal I got an error when starting using `yarn ios`. It works only with `expo start` command.

brentvatne:
@hadnet - we decided not to bother maintaining a separate template because it's so easy to convert the project to typescript. just rename App.js to App.tsx and run `expo start` and we do it for you

PR #3563, opened 2021-06-09
Patch android status bar to match the color of the splash screen

EvanBacon:
# Why

Prevent the status bar from flashing to the background color by setting the androidStatusBar.backgroundColor to the splash screen background color. This will cause the status bar to be an unexpected color in the app unless `<StatusBar style="auto"/>` is used (which seems reasonable enough).

This PR also drops support for `translucent` as the backgroundColor as this breaks things

### DIFF
--- a/packages/config-plugins/src/android/StatusBar.ts
+++ b/packages/config-plugins/src/android/StatusBar.ts
@@ -1,5 +1,5 @@
 import { ExpoConfig } from '@expo/config-types';
-import { assert } from 'console';
+import assert from 'assert';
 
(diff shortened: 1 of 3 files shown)


packages/config-plugins/src/android/StatusBar.ts
```diff
@@ -1,4 +1,5 @@
 import { ExpoConfig } from '@expo/config-types';
+import { assert } from 'console';
```
EvanBacon, line 2:
```suggestion
import assert from 'assert';
```

PR #3589, opened 2021-06-19
WIP: feat(customize:web): adds support for non-interactive automatic mode

jduarter:
# Why

- adds support for non-interactive automatic mode only for the "customize:web" command.
- I need this to be able to run the command automatically in a CI/CD pipeline 

# How

- I basically have added a new `--yes` parameter to the `customize:web` command of `expo-cli`.
- In this operation mode, no prompt is shown, and all files are automatically selected for ejection.

# Test Plan

### DIFF
--- a/packages/expo-cli/src/commands/customize.ts
+++ b/packages/expo-cli/src/commands/customize.ts
@@ -9,7 +9,7 @@ import prompts from 'prompts';
 
 import Log from '../log';
 
-type Options = { force: boolean };
+type Options = { force: boolean; yes: boolean };
 
 async function maybeWarnToCommitAsync(projectRoot: string) {
   let workingTreeStatus = 'unknown';
@@ -82,7 +82,10 @@ async function generateFilesAsync({
   await Promise.all(promises);
 }
 
-export async function action(projectRoot: string = './', options: Options = { force: false }) {
+export async function action(
+  projectRoot: string = './',
+  options: Options = { force: false, yes: false }
+) {
   // Get the static path (defaults to 'web/')
   // Doesn't matter if expo is installed or which mode is used.
   const { exp } = getConfig(projectRoot, {
@@ -110,9 +113,11 @@ export async function action(projectRoot: string = './', options: Options = { fo
       title: file,
       value: file,
       // @ts-ignore: broken types
-      disabled: !options.force && exists,
+      disabled: !options.force && !options.yes && exists,
       description:
-        options.force && exists ? chalk.red('This will overwrite the existing file') : '',
+        (options.force || options.yes) && exists
+          ? chalk.red('This will overwrite the existing file')
+          : '',
     });
   }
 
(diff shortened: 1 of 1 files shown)


jduarter:
Captura de pantalla 2021-06-19 a las 13 08 01

EvanBacon (changes requested):
1. We use `--non-interactive` already
2. I don't like the idea of creating everything by default, how would someone run the command and create a single thing? or nothing? We wouldn't be able to add any new _optional_ files in the future because that would break the "create everything" behavior.

jduarter:
> * We use `--non-interactive` already
> * I don't like the idea of creating everything by default, how would someone run the command and create a single thing? or nothing? We wouldn't be able to add any new _optional_ files in the future because that would break the "create everything" behavior.

Thanks for your comments, @EvanBacon!

I guess the existing `--non-interactive` modifier is not possible to be used in the case of `customize:web` as it will assume no files should be exported by default, right? (all options of files to be generated are disabled by default).

For the second comment, I can add a new parameter accepting a filename (or even better, a list of files) to be generated. The resulting functionality would then generate (as an example) the `babel.config.js` and `webpack.config.js` files if we invoke the command with the arguments `customize:web --non-interactive --file=babel.config.js --file=webpack.config.js` (only if such files doesn't exist yet).

If you agree, I can change the `--yes` to use `--non-interactive` and add the `--file` option to it. Also, I will be adding the ``--non-interactive`` command to the help so it's documented.

Additionally, I think it can be very helpful to add something like `--json` to make Expo output in JSON so it can be easily interpreted by external tools.

Thanks again for your attention, it's a pleasure to be able to contribute to this project.

EvanBacon:
I’m thinking maybe we should use a comma delimited list (which we do in other places), or a glob pattern (which we don’t do anywhere afaik). The benefit of the glob pattern is that it’s a bit more flexible, and file paths could be annoying to type out exactly.

Something like `--select=*`, `--select web/index.html,webpack.config.js`, the syntax is inspired by [debug](https://www.npmjs.com/package/debug)

jduarter:
I also like the idea of using globs!

I will do the following:

- [ ]  check if there is any glob mechanism already inside the package.
- [ ]  (implement the most suitable glob-lib otherwise, and create a function that can be used in the entire package for several purposes)
- [ ]  remove the initial proposed `--yes` and develop the `--select` option.
- [ ] add both the `--select` and `--non-interactive` parameters to the command help & other documentation describing the functionality of `customize:web`.

I will mark this one as `WIP`, I will keep you posted.

Greetings!

EvanBacon:
This appears to be resolved in the latest, versioned Expo CLI (`npx expo customize metro.config.js`). If you encounter this issue in the future, please open an issue on [expo/expo](https://github.com/expo/expo/issues/new?assignees=&labels=needs+review%2CCLI&template=bug_report_cli.yml).

jduarter:
Awesome, thx, Evan. I will try it when I have some time.

PR #3600, opened 2021-06-23
Skip unprovided mods and warn

EvanBacon:
# Why

As new versions of config plugins come out there may be mismatched providers, this PR will prevent them from breaking and instead warn that a mod will be skipped in expo prebuild, eject, and config.

# Test Plan

In a project:

`plugin.js`
```js
module.exports = (config) => {
  config.mods = {};
  config.mods.ios = {};
  config.mods.ios.foobar = () => null;
  return config;
};
```

`app.json`

```json
{ "expo": { "plugins": ["./plugin"] } }
```

### DIFF
--- a/packages/config-plugins/src/plugins/mod-compiler.ts
+++ b/packages/config-plugins/src/plugins/mod-compiler.ts
@@ -1,4 +1,3 @@
-import chalk from 'chalk';
 import path from 'path';
 
 import { ExportedConfig, Mod, ModConfig, ModPlatform } from '../Plugin.types';
--- a/packages/config-plugins/src/plugins/mod-compiler.ts
+++ b/packages/config-plugins/src/plugins/mod-compiler.ts
@@ -1,8 +1,10 @@
+import chalk from 'chalk';
 import path from 'path';
 
 import { ExportedConfig, Mod, ModConfig, ModPlatform } from '../Plugin.types';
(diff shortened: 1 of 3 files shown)


packages/config-plugins/src/plugins/mod-compiler.ts
```diff
@@ -164,9 +165,11 @@ export async function evalModsAsync(
           if (assertMissingModProviders !== false) {
             throw new PluginError(errorMessage, 'MISSING_PROVIDER');
           } else {
-            if (config._internal?.isDebug) {
-              console.warn(errorMessage);
-            }
+            Warnings.addWarningForPlatform(
+              platformName as ModPlatform,
+              `${platformName}.${modName}`,
+              `Skipping: Initial base modifier for "${platformName}.${modName}" is not a provider and therefore will not provide modResults to child mods`
```
brentvatne, line 171:
```suggestion
              `Skipping: Initial base modifier for "${platformName}.${modName}" is not a provider and therefore will not provide modResults to child mods. This may be due to an outdated version of Expo CLI.`
```

brentvatne (approved):

packages/config-plugins/src/plugins/mod-compiler.ts
EvanBacon: Could be problematic if you aren't using expo-cli

PR #3602, opened 2021-06-24
[dev-server] Fix incorrect inspector request from android emulator loopback

Kudo:
# Why

For dev-client running on android emulator, user may try to load bundles from http://10.0.2.2:8081/.
This is the [host loopback address](https://developer.android.com/studio/run/emulator-networking)
To query available inspector, we tried to send `/json/list` request to metro server but using HOST address in request headers.
In this case, HOST is 10.0.2.2:8081. For metro server, sending requests to 10.0.2.2:8081 is usually unreachable.

# How

Should use metro's local server address instead of request header HOST address.

# Test Plan

1. Use dev-client to load http://10.0.2.2:8081 bundle on android emulator.
2. Open dev-menu and see if there are any timeout errros.


lukmccall (approved):
LGTM 🚀 Thanks for taking care of it 🥇

PR #3606, opened 2021-06-24
feat: serve new expo-updates format manifests

wschurman:
# Why

We'd like to move `expo start` to use the [new format manifests](https://docs.expo.io/technical-specs/expo-updates-0/). This PR starts that process by serving the new format manifest at a separate URL on the local webserver.

Next up:
- figure out how to embed scopeKey in this manifest format (top level field on the manifest probably)
- figure out how to sign scopeKey for Expo Go (cert chain probably)
- figure out how this interacts with eas-cli for embedding things like extra.eas.projectId in the manifest and signing the scopeKey.
- create a way to run `expo start --new-manifest-format` (or something like that) that will serve the new manifest and use it for the development session and QR code by default

# `extra` field

This PR is where we're designing the format of the new manifest `extra` field. It's format has been discussed before (https://github.com/expo/universe/pull/7557, https://github.com/expo/eas-cli/pull/402) but we hadn't decided what the criteria for organization within the field should be. This PR formalizes it a bit with a definition of the sub fields:
- `eas` - contains Expo Application Services fields like `projectId`
- `expoClient` - contains fields necessary for the expo client. This is the non-development fields from the classic manifest, which is also really just the app.json (with assets "resolved").
- `expoGo` - contains fields that only are used within Expo Go for development like things about the packager, etc.

The idea is that app.json is fairly difficult to split into distinct logical units (library configs) in a maintainable way. We could require metadata in the version schema indicating what library the field belongs to, but doing that categorization this early in the process seemed like a non-starter. Instead, we put the entire app.json in it's own extra field for now, and then move specific app.json fields into their own library keys once we have a good way to separate them out. One future idea is to have a `libraries` key in app.json that is essentially a merge field for the `extra` field.

This approach was decided upon because it is easy to transition to from classic manifests in the client, clear how to transfer away from for separating out library-specific config into their own keys, and easy to maintain here in the CLI for now.

# How

Serve the new manifest format at `/update-manifest-experimental`.

# Test Plan

- Run the new test.
- `EXPO_DEBUG=true ~/expo/expo-cli/node_modules/.bin/expo start` in a project, navigate to http://192.168.86.113:19000/update-manifest-experimental?platform=ios, see new format manifest
- Load http://192.168.86.113:19000/update-manifest-experimental?platform=ios in the android simulator in combination with https://github.com/expo/expo/pull/13398

### DIFF
--- a/packages/config/src/Config.types.ts
+++ b/packages/config/src/Config.types.ts
@@ -59,32 +59,56 @@ export type HookArguments = {
   log: (msg: any) => void;
 };
 
-export type ExpoAppManifest = ExpoConfig & {
-  sdkVersion: string;
-  bundledAssets?: string[];
-  isKernel?: boolean;
-  kernel?: { androidManifestPath?: string; iosManifestPath?: string };
-  assetUrlOverride?: string;
-  publishedTime?: string;
-  commitTime?: string;
-  releaseId?: string;
-  revisionId?: string;
-  mainModuleName?: string;
+export type ExpoGoConfig = {
+  mainModuleName: string;
   // A string that flipper checks to determine if Metro bundler is running
   // by adding it to the manifest, we can trick Flipper into working properly.
   // https://github.com/facebook/flipper/blob/9ca8bee208b7bfe2b8c0dab8eb4b79688a0c84bc/desktop/app/src/dispatcher/metroDevice.tsx#L37
-  __flipperHack?: 'React Native packager is running';
-  env?: Record<string, any>;
-  bundleUrl?: string;
-  debuggerHost?: string;
-  logUrl?: string;
-  hostUri?: string;
-  id?: string;
-  developer?: {
+  __flipperHack: 'React Native packager is running';
+  debuggerHost: string;
+  logUrl: string;
+  developer: {
     tool: string | null;
     projectRoot?: string;
   };
+  packagerOpts: {
+    [key: string]: any;
+  };
 };
+
+export type ExpoAppManifest = ExpoConfig &
+  Partial<ExpoGoConfig> & {
+    sdkVersion: string;
+    bundledAssets?: string[];
+    isKernel?: boolean;
+    kernel?: { androidManifestPath?: string; iosManifestPath?: string };
+    assetUrlOverride?: string;
+    publishedTime?: string;
+    commitTime?: string;
+    releaseId?: string;
+    revisionId?: string;
+    env?: Record<string, any>;
+    bundleUrl?: string;
+    hostUri?: string;
+    id?: string;
+  };
+
+export interface ExpoUpdatesManifestAsset {
+  url: string;
+  key: string;
+  contentType: string;
+  hash?: string;
+}
+
+export interface ExpoUpdatesManifest {
+  id: string;
+  createdAt: string;
+  runtimeVersion: string;
+  launchAsset: ExpoUpdatesManifestAsset;
+  assets: ExpoUpdatesManifestAsset[];
+  metadata: { [key: string]: string };
+  extra: { [key: string]: any };
+}
 
 export type Hook = {
   file: string;
--- a/packages/xdl/src/ProjectAssets.ts
+++ b/packages/xdl/src/ProjectAssets.ts
@@ -1,6 +1,7 @@
 import { ExpoAppManifest, ExpoConfig } from '@expo/config';
 import { BundleAssetWithFileHashes, BundleOutput } from '@expo/dev-server';
 import assert from 'assert';
+import crypto from 'crypto';
 import FormData from 'form-data';
 import fs from 'fs-extra';
 import chunk from 'lodash/chunk';
@@ -8,6 +9,7 @@ import get from 'lodash/get';
 import set from 'lodash/set';
 import uniqBy from 'lodash/uniqBy';
 import md5hex from 'md5hex';
+import mime from 'mime';
 import minimatch from 'minimatch';
 import path from 'path';
 import urljoin from 'url-join';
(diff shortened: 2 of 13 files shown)


jkhales (commented):
The `extra` field is supposed to be for optional third party configuration. 

Adding `expoClientConfig` makes me feel like we are doing something wrong as the expo client is supposed to be part of the official protocol.

What are we putting in here that can't be stored in the `eas` field?

ide:
I haven't read through any of the code but have one preliminary comment on the field names: everything in the manifest is configuration. Instead of `expoClientConfig` and `expoGoConfig`, we could name them `expoClient` and `expoGo` and it'd make as much sense or even be clearer by not implying that fields without "config" aren't about configuration ("extra.notifications" would still contain notification configuration data, for example).

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -0,0 +1,148 @@
...
+    id: uuid.v4(),
+    createdAt: new Date().toISOString(),
+    runtimeVersion,
+    launchAsset: {
+      key: mainModuleName,
+      contentType: 'application/javascript',
+      url: bundleUrl,
+    },
+    assets,
+    metadata: {}, // required for the client to detect that this is an expo-updates manifest
```
esamelson, line 97: we should change how we detect this in expo-updates; afaik this field is no longer used for anything else

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -0,0 +1,148 @@
...
+    projectRoot,
+    platform,
+    projectSettings,
+    bundleUrlPackagerOpts,
+    mainModuleName,
+    hostname,
+  });
+
+  // resolve all assets and set them on the manifest as { rawUrl?: string, assetKey?: string } objects
+  const assets = await ProjectAssets.resolveAndCollectExpoUpdatesManifestAssets(
```
esamelson, line 81: it might be helpful to add a comment somewhere clarifying that "assets" here refers to "manifest assets" (e.g. `notification.icon`. `splash.image`, etc) rather than assets `require`d in JS. (at least, that's my understanding of what's happening here).

up to now, only the latter have been added to the `assets` field in the new manifest. but i think i understand the reasoning behind this, so that these "manifest assets" all get downloaded onto the device ahead of time by expo-updates, correct?

esamelson (commented):
I don't consider myself qualified enough on the inner workings of this expo-cli code to give a thorough review, but left a couple comments and questions. Also want to echo James's comment about naming -- the word "Config" in `expoGoConfig` and `expoClientConfig` seems redundant.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: This is doing two things:
- Collect each local asset from `notification.icon`, `splash.image`, etc and place in manifest assets. These are the same assets that are collected during an EAS branch publish.
- Put a new sibling field for each field containing a URL and reference to the asset. For example, for `notification.icon = /path/to/local/image.png` this will add a new field: `notification.iconAsset = { assetKey: <key in assets>, rawUrl: <URL to fetch asset from> }`.

I'll add a comment.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: https://linear.app/expo/issue/ENG-1486/change-how-we-detect-eas-manifests-in-client-code

packages/config/src/Config.types.ts
```diff
@@ -59,29 +59,53 @@ export type HookArguments = {
...
+  hash?: string;
+}
+
+export interface ExpoUpdatesManifest {
+  id: string;
+  createdAt: string;
+  runtimeVersion: string;
+  launchAsset: ExpoUpdatesManifestAsset;
+  assets: ExpoUpdatesManifestAsset[];
+  metadata: object;
```
ide, line 105: Is this intentionally not `{ [key: string]: any }`?

packages/config/src/Config.types.ts
```diff
@@ -59,29 +59,53 @@ export type HookArguments = {
...
+export interface ExpoUpdatesManifestAsset {
+  url: string;
+  key: string;
+  contentType: string;
+  hash?: string;
+}
+
+export interface ExpoUpdatesManifest {
+  id: string;
+  createdAt: string;
```
ide, line 101: Do we want this to be a parsed Date object or do we want the timestamp string verbatim?

packages/xdl/src/ProjectAssets.ts
```diff
@@ -349,3 +395,29 @@ async function collectAssets(
...
+): Promise<{ url: string; hash: string; key: string; contentType: string }[]> {
+  const manifestAssets: { url: string; hash: string; key: string; contentType: string }[] = [];
+  await resolveExpoUpdatesManifestAssets({
+    projectRoot,
+    manifest: exp,
+    async assetKeyResolver(assetPath) {
+      const absolutePath = path.resolve(projectRoot, assetPath);
+      const contents = await fs.readFile(absolutePath);
+      const fileType = await FileType.fromBuffer(contents);
+      const hash = md5hex(contents);
```
ide, line 412: I believe this is coupled to something (expo-asset?). Would be good to have a comment here to highlight that coupling since it's non-obvious why we use MD5 here.

packages/xdl/src/ProjectAssets.ts
```diff
@@ -349,3 +395,29 @@ async function collectAssets(
...
+      manifestAssets.push({
+        url: urlResolver(assetPath),
+        hash,
+        key: assetPath,
+        contentType: fileType?.mime ?? '',
```
ide, line 417: Does EAS not accept null or omitted content types?

packages/xdl/src/ProjectAssets.ts
```diff
@@ -60,20 +61,65 @@ export async function resolveGoogleServicesFile(projectRoot: string, manifest: E
...
+  projectRoot: string;
+  manifest: ExpoConfig;
+  assetKeyResolver: (assetPath: string) => Promise<string>;
+}): Promise<void> {
+  const assetSchemas = await getAssetFieldPathsForManifestAsync(manifest);
+  // Get the URLs
+  const assetInfos = await Promise.all(
+    assetSchemas.map(async manifestField => {
+      const pathOrURL = get(manifest, manifestField);
+      if (pathOrURL.match(/^https?:\/\/(.*)$/)) {
```
ide, line 84: We don't need support for capture groups:
```suggestion
      if (/^https?:\/\//.test(pathOrURL)) {
```

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+} from '../internal';
+import {
+  getBundleUrlAsync,
+  getExpoGoConfig,
+  getPackagerOptionsAsync,
+  stripPort,
+} from './ManifestHandler';
+
+function getPlatformFromRequest(req: express.Request | http.IncomingMessage): string {
+  const url = req.url ? parse(req.url, true) : null;
```
ide, line 24:
```suggestion
  const url = req.url ? parse(req.url, /* parseQueryString */ true) : null;
```

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+}: {
+  projectRoot: string;
+  platform: string;
+  host?: string;
+}): Promise<{
+  body: ExpoUpdatesManifest;
+  headers: Map<string, number | string | readonly string[]>;
+}> {
+  const headers = new Map<string, any>();
+  // set required headers for expo-updates manifest specification
```
ide, line 45: technically Expo Updates is the protocol and expo-updates is the client library implementation
```suggestion
  // set required headers for Expo Updates manifest specification
```

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+
+  const expoUpdatesManifest = {
+    id: uuid.v4(),
+    createdAt: new Date().toISOString(),
+    runtimeVersion,
+    launchAsset: {
+      key: mainModuleName,
+      contentType: 'application/javascript',
+      url: bundleUrl,
```
ide, line 98: A few assumptions here may change with Hermes: the content type won't be JS and the URL will point to an HBC file instead of a JS bundle.

ide (changes requested):
Looks mostly good, a few changes.

I'd like to think through the name "expoClient" a bit more. It might be right but a name like "classicManifest" might also make sense.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: Yep, should be easy to change at that time.

wschurman:
> I'd like to think through the name "expoClient" a bit more. It might be right but a name like "classicManifest" might also make sense.

I don't really have a strong opinion on this. `expoClient` feels correct to me since it is configuration for the client (which is a bunch of libraries bundled together), and when we move configuration out of expoClient it will go in library specific fields.

packages/config/src/Config.types.ts
ide: It needs to be a string-valued dictionary. The metadata field is used for selecting which update to run when doing rollouts.

packages/xdl/src/ProjectAssets.ts
ide: In this case, EAS Update defaults to `application/octet-stream` (i.e. unknown bytes). I think it would be good for EAS Update and the development server to share this behavior so publishing doesn't change as many things.

ide (approved):

packages/xdl/src/ProjectAssets.ts
wschurman: Ah, I see that we just use the extension in EAS cli and fall back to `application/octet-stream`: https://github.com/expo/eas-cli/blob/d50bdedcec70e8e001730ed6b4946d12b0c639ca/packages/eas-cli/src/project/publish.ts

I'll use that strategy.

Issue #3625, opened 2021-07-02
uri-scheme tool crashes on 'empty' url type entry


kbrandwijk:
If you happen to have an empty url type in your iOS project (as in, clicked the + sign once on url types accidentally, but didn't enter anything) then the cli will crash ugly.

```text
npx: installed 1 in 1.207s

Aborting run
An unexpected error was encountered. Please report it as a bug:
TypeError: Cannot read property 'includes' of undefined
    at /Users/*****/.npm/_npx/48240/lib/node_modules/uri-scheme/build/index.js:1:418895
    at Array.some (<anonymous>)
    at Object.hasScheme (/Users/kim/.npm/_npx/48240/lib/node_modules/uri-scheme/build/index.js:1:418862)
    at Object.addAsync (/Users/kim/.npm/_npx/48240/lib/node_modules/uri-scheme/build/index.js:1:481738)
    at Object.addAsync (/Users/kim/.npm/_npx/48240/lib/node_modules/uri-scheme/build/index.js:1:486127)
    at async Command.<anonymous> (/Users/kim/.npm/_npx/48240/lib/node_modules/uri-scheme/build/index.js:1:477579)
```

kbrandwijk:
I have tracked this down to this line:

https://github.com/expo/expo-cli/blob/8f551d89859a42ce0dfcf357b14988f3c924e5ff/packages/config-plugins/src/ios/Scheme.ts#L91

I'll come up with a PR for this.

PR #3627, opened 2021-07-02
Handle empty entries when parsing URL types in iOS projects

kbrandwijk:
# Why

As described in #3625, a newly created URL type entry in XCode doesn't have any properties yet, so it's not safe to assume the CFBundleURLSchemes property exists and is an array, in fact it crashes the `uri-scheme` CLI tool.

# How

This fix adds a type check in the same coding style as the one right above it to ensure the existence of the array before trying to access it.

# Test Plan

As I am not extremely familiar with this huge monorepo, I am not aware of any automated test cases covering this functionality. However, the fix is small and self-explanatory, and aligns perfectly with the stacktrace thrown by the `uri-scheme` CLI tool.


EvanBacon (approved):
LGTM thanks!

Issue #151, opened 2017-05-09
Restart packager automatically (or prompt to restart) when relevant files change


brentvatne:
- [ ] babelrc
- [ ] perhaps some fields in package.json (https://github.com/expo/xde/issues/2)

PR #3638, opened 2021-07-07
Warn about babel config changing

EvanBacon:
# Why

Resolve https://github.com/expo/expo-cli/issues/151

# How

Observe file changes for babel config and log a warning to restart the server. 

# Test Plan

- `expo start` 
- modify the babel.config.js 
- See logs in terminal
- Repeat but with `expo web`


brentvatne:
nice 👌

PR #3651, opened 2021-07-10
[run][android] fix cannot connect metro in a pure managed app

Kudo:
# Why

`expo run:android` cannot connect metro server because reverse tunnel not setting up.

# How

Add adb reverse

# Test Plan

```sh
expo init testapp
# select the managed blank template

expo run:android
# app should start without metro connection failures.
```


EvanBacon (approved):
lgtm thanks

PR #3667, opened 2021-07-14
[run:ios] Added experimental pre-bundling functionality

EvanBacon:
# Why

Production builds with basic JS syntax errors are hard to provide proper error stack traces for, and they can be caught a lot quicker if we run the bundler before building the native runtime.

# How

Added `EXPO_PREBUNDLE=1` which enables bundling before in Release builds. This disables the bundling script in the Xcode project, and copies the results into the binary after it finishes building.

We don't currently have a complete replacement for RN cli bundle command so for now we just invoke the RN CLI with some parameters.

# Test Plan

- `EXPO_PREBUNDLE=1 expo run:ios --configuration Release` you should see the RN CLI output.
- The app should install and open on a device with no issues. Running in EXPO_DEBUG shows the output folders which you can inspect.
- Perform the command again after adding a missing import to the project, it should fail fast.


EvanBacon:
Stale, closing

PR #3668, opened 2021-07-15
[config-plugins] Fix addJavaImports broken for kotlin files

Kudo:
# Why

`addJavaImports` does not add imports after package line for kotlin files. kotlin does not end lines with semicolon.

# How

for the package regular expression, the semicolon is optional.

# Test Plan

Test splash-screen config-plugins with MainActvivity.kt


EvanBacon (approved):
lgtm, thanks

PR #3685, opened 2021-07-17
[expo-cli] Created `expo bundle` command

EvanBacon:
# Why

Give us control over the static bundling for offline support. 

- [x] Use babel-preset-expo by default.
- [x] Use expo/metro-config by default.
- [x] Respect expo.entryPoint, expo/AppEntry.js, and platform specific entry points.
- [x] Unify Expo Metro logging.
- [x] Move default bundle logic out of native scripts and into JS.
- Related ENG-1075 ENG-1155 ENG-1156

# How

Created `expo bundle` which can only bundle a single platform at a time (different to `expo export` or `expo publish`). I didn't add support for the old non-devserver option that we support in `expo start`, or targets (managed/bare).

# Test Plan

In `packages/expo-cli`:
```
yarn test:e2e --testPathPattern bundle-test
```

- [x] Locally `expo bundle --ios` and `expo bundle --android`
  - Compare to `npx react-native bundle --entry-file index.js --bundle-output ios-build/index.jsbundle --minify --assets-dest ./ios-build`
- [x] Test argument parsing.
- [x] Test MetroDevServer `bundleAsync` persisting contents.
- [x] Ensure the expected amount of assets are produced.
- [x] Ensure the source maps don't contain the user's absolute computer path.
- [x] Ensure multi-resolution images are copied as expected on iOS and Android.


packages/expo-cli/src/commands/bundleAsync.ts
```diff
@@ -0,0 +1,87 @@
+import { getConfig } from '@expo/config';
+import { bundleAsync } from '@expo/dev-server';
+import assert from 'assert';
+import fs from 'fs-extra';
+import path from 'path';
+import { printBundleSizes, ProjectUtils } from 'xdl';
+
+import { resolveEntryPoint } from '../../../xdl/build/internal';
```
EvanBacon, line 8: fix

brentvatne (commented):
the tests look good. one thing that would make them even better is if we were to run them against the latest expo package version on expo/expo as well, in order to catch regressions earlier on.

EvanBacon:
closing in favor of `npx expo export`

PR #3686, opened 2021-07-17
[prebuild-config] Also set CFBundleName

brentvatne:
# Why

Screen Shot 2021-07-17 at 9 13 33 AM

# How

Use existing `withName` plugin

# Test Plan

Snapshots looked good

### DIFF
--- a/packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
+++ b/packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
@@ -157,7 +157,7 @@ Object {
       "CFBundleExecutable": "$(EXECUTABLE_NAME)",
       "CFBundleIdentifier": "$(PRODUCT_BUNDLE_IDENTIFIER)",
       "CFBundleInfoDictionaryVersion": "6.0",
-      "CFBundleName": "$(PRODUCT_NAME)",
+      "CFBundleName": "my cool app",
       "CFBundlePackageType": "APPL",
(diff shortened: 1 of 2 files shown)


packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
```diff
@@ -157,7 +157,7 @@ Object {
       "CFBundleExecutable": "$(EXECUTABLE_NAME)",
       "CFBundleIdentifier": "$(PRODUCT_BUNDLE_IDENTIFIER)",
       "CFBundleInfoDictionaryVersion": "6.0",
-      "CFBundleName": "$(PRODUCT_NAME)",
```
EvanBacon, line 160: Maybe we should be changing `PRODUCT_NAME` instead.

EvanBacon: We use `$(PRODUCT_NAME)` in a lot of permission messages so we probably want to resolve that.

brentvatne: an aside - i think in most cases we shouldn't use `PRODUCT_NAME` in permission messages - there is only one example of doing this in the ios hig: https://developer.apple.com/design/human-interface-guidelines/ios/app-architecture/accessing-user-data/

Screen Shot 2021-07-28 at 1 27 40 PM

but i agree that changing `PRODUCT_NAME` instead of `CFBundleName` is probably preferable here

Issue #3630, opened 2021-07-02
yarn next dev fails after upgrading to sdk 42


wallstar:
After upgrading to sdk 42, I cannot develop using Nextjs.  When I run yarn next dev, I'm getting the following error:
 **"Error: Cannot find module 'jsonfile/utils'"**

When I switch back to sdk 41, it works normally.
Versions

- "next": "^10.0.7"
- "@expo/next-adapter": "~2.1.52",
- "expo": "^42.0.0",
- expo-cli: 4.7.2

axeldelafosse:
Hey @wallstar! Can you try upgrading to Next.js 11 and @expo/next-adapter 3.0.1 please?

wallstar:
@axeldelafosse thanks for getting back to me! I can give it a shot next week, but doesn't Next.js 11 require React 17 which isn't yet supported by expo?

axeldelafosse:
Yes it does but don't worry you can use React 17 with Expo

wallstar:
@axeldelafosse worked for me, thanks! I did have to disable webpack 5 in my next.config.js, but I'm not sure if that's related

axeldelafosse:
Cool! You should also be able to use Webpack 5. What errors are you getting? You can check here for more context: https://github.com/expo/expo-cli/issues/3579.

Feel free to close this issue when everything works for you!

wallstar:
Getting this error:

ValidationError: Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration[0].module.rules[7] should be one of these:
   ["..." | object { compiler?, dependency?, descriptionData?, enforce?, exclude?, generator?, include?, issuer?, issuerLayer?, layer?, loader?, mimetype?, oneOf?, options?, parser?, realResource?, resolve?, resource?, resourceFragment?, resourceQuery?, rules?, scheme?, sideEffects?, test?, type?, use? }, ...]

axeldelafosse:
Do you have a custom Webpack config? Can you share your `next.config.js` please?

wallstar:
`const { withExpo } = require('@expo/next-adapter');
const withFonts = require('next-fonts');
const withPlugins = require('next-compose-plugins');
const withImages = require('next-images');
const withBundleAnalyzer = require('@next/bundle-analyzer')({
  enabled: process.env.ANALYZE === 'true',
});
const withTM = require('next-transpile-modules')([
  'moti',
  'react-native-web',
]);

[...]

Issue #3579, opened 2021-06-16
Support NextJS 11 


marklawlor:
`@expo/next-adapter` only supports NextJS 9.x. which is now two major versions behind. The Next adapter is broken for two reasons

- Compile error 'Missing class properties transform' when using Next >=10.0.2 https://github.com/expo/expo-cli/issues/3226
- Please upgrade @expo/webpack-config's webpack dependency to 5.x https://github.com/expo/expo-cli/issues/2881

I'm creating this issue to discuss just updating just `@expo/next-adapter` (not Expo web) to the latest NextJS version with webpack 5 support.

Here is a working proof of concept https://github.com/marklawlor/expo-nextjs-webpack-5.

@EvanBacon Whats the best way to PR these patches without breaking Expo web?

In https://github.com/expo/expo-cli/issues/2881 you indicated that you wished to wait for Create React App to update, but `@expo/next-adapter` only uses the `withUnimodules` addon which requires minimal changes which could be added in a non-breaking manner behind flags.

## PRs

- https://github.com/expo/expo/pull/13327
- https://github.com/expo/expo-cli/pull/3586

### My expo app has a JSX transform error

NextJS 11 w/ React 17 uses the new JSX transform. You should enable `useTransformReactJsxExperimental` for all platforms.

### "React.createElement is not a function" 

The JSX transform does not require you to import React, so it may have been removed from a shared component. Consider switching to `useTransformReactJsxExperimental` for all platforms.

### I'm getting Postcss errors from NextJS

Run `npm list postcss`. NextJS has a peerDependancy on PostCSS 8 and you may have to install this peerDependancy yourself.

EvanBacon:
You can safely make changes without breaking expo web by simply modifying the webpack config in the next-adapter. Happy to review a PR.

marklawlor:
@EvanBacon Unfortunately the changes require updating `html-webpack-plugin` (a dependency of `@expo/webpack-config`) to the next major version which [drops support for webpack 4](https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md). So I'm not sure how to completely PR these changes without breaking Expo web.

The easiest option is PR what I can, and then people just need to use npm/yarn resolutions for `html-webpack-plugin` until `@expo/webpack-config` is updated. Or is there a way we could start a `webpack5` branch (similar to create-react-app) and start  the progression to webpack5?

Also, thanks for publishing the new version of `babel-preset-expo`

[...]

PR #3691, opened 2021-07-19
[next-adapter] update template to transpile `react-native-web`

axeldelafosse:
# Why

We added support for Next.js 11 and Webpack since `@expo/next-adapter@3.0.0`

# How

Updated the required `next.config.js` to transpile `react-native-web`

# Test Plan

Use the next-adapter generated template and make sure that it's working

### DIFF
--- a/packages/next-adapter/template/next.config.js
+++ b/packages/next-adapter/template/next.config.js
@@ -1,7 +1,9 @@
 // Learn more: https://github.com/expo/expo/blob/master/docs/pages/versions/unversioned/guides/using-nextjs.md#withexpo
 
 const { withExpo } = require('@expo/next-adapter');
+const withPlugins = require('next-compose-plugins');
+const withTM = require('next-transpile-modules')(['react-native-web']);
 
-module.exports = withExpo({
-  projectRoot: __dirname,
-});
+const nextConfig = {};
+
+module.exports = withPlugins([withTM, [withExpo, { projectRoot: __dirname }]], nextConfig);
--- a/packages/next-adapter/src/customize/manifest.ts
+++ b/packages/next-adapter/src/customize/manifest.ts
@@ -69,9 +69,12 @@ function getDependencies(
   const dependencies = ['react-native-web', 'next'].filter(
     dependency => !resolveFrom.silent(projectRoot, dependency)
   );
-  const devDependencies = ['@expo/next-adapter', 'babel-preset-expo'].filter(
-    dependency => !resolveFrom.silent(projectRoot, dependency)
-  );
+  const devDependencies = [
+    '@expo/next-adapter',
+    'babel-preset-expo',
+    'next-compose-plugins',
+    'next-transpile-modules',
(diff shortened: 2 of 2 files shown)


packages/next-adapter/template/next.config.js
```diff
@@ -1,7 +1,9 @@
 // Learn more: https://github.com/expo/expo/blob/master/docs/pages/versions/unversioned/guides/using-nextjs.md#withexpo
 
 const { withExpo } = require('@expo/next-adapter');
+const withPlugins = require('next-compose-plugins');
+const withTM = require('next-transpile-modules')(['react-native-web']);
```
EvanBacon, line 5: react-native-web is compiled afaict. https://github.com/necolas/react-native-web/blob/98dcefd9041c10cc95a2385202498b274d926a6b/packages/react-native-web/package.json#L8-L9

axeldelafosse: Yeah I know that it's compiled ahead of time and shouldn't be transpiled -- I have multiple projects where it's working correctly but it looks like the template doesn't compile if we don't transpile `react-native-web`. More context here: https://github.com/expo/expo-cli/issues/3579#issuecomment-882132690

This is weird and to be honest I think I might be missing something. Any idea why this is happening?

EvanBacon: Have you already tried modifying the RNW babel plugin to use cjs instead? https://github.com/vercel/next.js/blob/e65c56e7e6c16c3c971984ce7acd161bca6b9c2f/examples/with-react-native-web/babel.config.js#L3

axeldelafosse: It doesn't work :(

axeldelafosse: Following up on this, it looks like this error appears only when not transpiling any other modules: https://github.com/expo/expo-cli/issues/3630#issuecomment-888101730. @marklawlor can you confirm that this issue disappeared for you too?

marklawlor: @axeldelafosse I've noticed that all reports seem to be on a hello world template with no other dependencies. Once you add another dependency to `next-transpile-modules` it seems you can remove `react-native`.

I'm not sure why, and I haven't looked into it at all. When I updated `next-adapter` none of the logic on the Expo side was changed (just function signatures) - however `next-transpile-modules` was bumped like 3 major versions. The transpilation code inside `webpack-config` has comments that it was adapted from `next-transpile-modules` so I assume there is now some incompatibility edge case.

In terms of effort, I'm not sure if exploring this bug is worth it - especially if the workaround is a simple documentation change that doesn't seem to have any side effects. I personally don't have the current capacity to look into it and would rather put that effort into native ESM support.

EvanBacon:
Closing in favor of the guide now saying this https://github.com/expo/expo-cli/blob/main/packages/next-adapter/README.md#nextjs-configuration

PR #3693, opened 2021-07-19
[config] Replace babel with sucrase

EvanBacon:
# Why

[sucrase](https://github.com/alangpierce/sucrase) is [smaller](https://packagephobia.com/result?p=sucrase) (upwards of 6x smaller) and faster than babel (for our uses cases, it's over 2x faster). The downside to using sucrase is that you don't have access to the wide plugin system that babel has, but our app.config.js transpilation cannot be customized so this doesn't apply to us.

The `app.config.ts` in `expo/apps/native-component-list` is a good test because it's very comprehensive:

| Babel   | Sucrase |
|--------|---------|
| 600ms | 240ms  |

For error handling:

| Babel   | Sucrase |
|--------|---------|
| 320ms |    60ms  |

Our tests cover proper error formatting, I do import babel/code-frame to style the error messages so users don't notice a big change in DX.

# Test Plan

- All existing unit tests
- NCL runs
- class properties still work
- error handling still works


packages/config/src/__tests__/getConfig-test.ts
```diff
@@ -136,7 +138,7 @@ describe(getDynamicConfig, () => {
   it(`throws a useful error for dynamic configs with a syntax error`, () => {
     const paths = getConfigFilePaths('syntax-error');
     expect(() => getDynamicConfig(paths.dynamicConfigPath, mockConfigContext)).toThrowError(
-      'Unexpected token (3:4)'
+      /Error (?:.*) \(5:7\)/
```
ide, line 141: Don't need the capture group
```suggestion
      /Error .* \(5:7\)/
```

packages/config/src/evalConfig.ts
```diff
@@ -22,23 +21,43 @@ export function evalConfig(
...
+      error.codeFrame = codeFrame;
+      error.message += `\n${codeFrame}`;
+    }
+    throw error;
+  }
+}
+
+function extractLocationFromSyntaxError(
+  error: Error | any
+): { line?: number; column?: number } | null {
```
ide, line 49: Usually line is always provided - can this be nonoptional?

packages/config/src/evalConfig.ts
```diff
@@ -22,23 +21,43 @@ export function evalConfig(
...
-    configFile: false,
-    comments: false,
-    ignore: [/node_modules/],
-    filename: 'unknown',
-    presets: [getBabelPreset()],
-    // Retain lines for improved error reporting.
-    retainLines: true,
-  });
+    const result = requireString(code, configFile);
+    return resolveConfigExport(result, configFile, request);
```
ide, line 32: Move these out of the try-catch since presumably we don't want to apply code frame formatting to errors in these method calls. (Keep try-blocks as narrow as possible so that they catch errors only from the code we expect to throw an error.)

EvanBacon: `requireString` can throw syntax errors, other one could be fine to move out.

brentvatne (approved):

Issue #3663, opened 2021-07-14
Creating a new project does not respect `init.defaultBranch` setting of git


trajano:
It always uses `main` even if it was set to `master`

PR #3702, opened 2021-07-20
Remove custom branch setting

EvanBacon:
# Why

- resolve https://github.com/expo/expo-cli/issues/3663


trajano:
Your test looks a bit weird. in that it assumes the default is `master` but that can change at any time.  Instead your test should include setting the init.defaultbranch (somehow)

PR #3708, opened 2021-07-21
Fix hermesc.exe resolving on windows

Kudo:
# Why

fix https://github.com/expo/expo/issues/13626

# How

on windows, the resolving path should include `.exe` suffix.

# Test Plan

setup jsEngine as hermes
```sh
$ jq .expo.android.jsEngine app.json
hermes
```

expo export on macos and windows


EvanBacon (approved):
LGTM

PR #3710, opened 2021-07-21
[config-plugins] Fix multitasking ipad orientations

EvanBacon:
# Why

If you build an app with SDK 43, requireFullScreen will be false, after the build is successfully compiled, Transporter will fail with:

```
ERROR ITMS-90474: "Invalid Bundle. iPad Multitasking support requires these orientations: 'UIInterfaceOrientationPortrait,UIInterfaceOrientationPortraitUpsideDown,UIInterfaceOrientationLandscapeLeft,UIInterfaceOrientationLandscapeRight'. Found 'UIInterfaceOrientationPortrait,UIInterfaceOrientationPortraitUpsideDown' in bundle 'com.bacon.app'."
```

Xcode bootstraps a project with:

```js
{
  'UISupportedInterfaceOrientations~ipad': {
   'UIInterfaceOrientationPortrait',
    'UIInterfaceOrientationPortraitUpsideDown',
    'UIInterfaceOrientationLandscapeLeft',
    'UIInterfaceOrientationLandscapeRight',
  }
}
```

This PR proposes that we inject these values when requireFullScreen is false. There is currently no way to undo adding these after they've been added, but it seems that we should mostly just default to users having these. 

# Test Plan

- [x] Unit test
- [x] Build and submit an app using `ios.requireFullScreen: false` without any issues

### DIFF
--- a/packages/config-plugins/src/ios/RequiresFullScreen.ts
+++ b/packages/config-plugins/src/ios/RequiresFullScreen.ts
@@ -30,12 +30,50 @@ export function getRequiresFullScreen(config: Pick<ExpoConfig, 'ios' | 'sdkVersi
   }
 }
 
+const iPadInterfaceKey = 'UISupportedInterfaceOrientations~ipad';
+
 const requiredIPadInterface = [
   'UIInterfaceOrientationPortrait',
   'UIInterfaceOrientationPortraitUpsideDown',
   'UIInterfaceOrientationLandscapeLeft',
   'UIInterfaceOrientationLandscapeRight',
 ];
+
+function isStringArray(value: any): value is string[] {
+  return Array.isArray(value) && value.every(value => typeof value === 'string');
+}
+
+function hasMinimumOrientations(masks: string[]): boolean {
+  return requiredIPadInterface.every(mask => masks.includes(mask));
+}
+
+/**
+ * Require full screen being disabled requires all ipad interfaces to to be added,
+ * otherwise submissions to the iOS App Store will fail.
+ *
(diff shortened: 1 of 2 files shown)


packages/config-plugins/src/ios/RequiresFullScreen.ts
```diff
@@ -29,13 +30,44 @@ export function getRequiresFullScreen(config: Pick<ExpoConfig, 'ios' | 'sdkVersi
...
+  if (!requiresFullScreen) {
+    if (Array.isArray(infoPlist['UISupportedInterfaceOrientations~ipad'])) {
+      if (
+        !(infoPlist['UISupportedInterfaceOrientations~ipad'] as string[]).every(mask =>
+          requiredIPadInterface.includes(mask)
+        )
+      ) {
+        WarningAggregator.addWarningIOS(
+          'ios.requireFullScreen',
+          `iPad multitasking requires all UISupportedInterfaceOrientations~ipad to be defined. The Info.plist currently defines insufficient values that will be overwritten. Existing: ${infoPlist[
```
brentvatne:
```suggestion
          `iPad multitasking requires all UISupportedInterfaceOrientations~ipad to be defined. The Info.plist currently defines values that are incompatible with UIRequiresFullScreen and will be overwritten. Existing: ${infoPlist[
```

brentvatne (approved):

barthap (approved):
Shall we also update the https://expo.fyi/ipad-requires-fullscreen? It is displayed when `eas submit` encounters the `ITMS-90474` error.

brentvatne:
@barthap - it seems like it will be difficult for users to get into the `ITMS-90474` state after this config plugin lands and is deployed (would need to update expo-cli on eas build after this is merged and published).

if we can think of a way where this can go wrong, we should update the doc to handle that.

a good general strategy in this situation to help people resolve the issue is to explain why this error is happening precisely (if `UIRequiresFullScreen` is false, ... x, y, z are required. if `UIRequiresFullScreen` is true, .....), and help users learn what they can do to inspect their app config to determine why those constraints don't apply (eg: through `expo config` for managed users).

Issue #2274, opened 2020-06-16
Support XDG Base Directory specification


distefam:
## Description

Expo should place its configuration files according to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

## Expected Behavior

Follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and place the contents of the aforementioned `.expo` directory in `$HOME/config/expo`.

## Observed Behavior

Expo puts a hidden directory (`.expo`) in the user's home directory when run.

## Reproducible Demo

1. Install expo using `npm install -g expo-cli`
2. Run `expo start` from within project directory
3. Run `ls -lA $HOME | grep expo` and observe `.expo` directory

brentvatne:
Hi there!

This PR starts with a very strong assumption:

> Expo should place its configuration files according to the XDG Base Directory Specification.

Can you explain the reasoning further please? We're open to being convinced, but we would need to hear more.

distefam:
You're right, _should_ is strong language. It would be very nice in my opinion, however. Basically, on Linux systems installed programs used to litter one's home directory with a bunch of hidden files, "dotfiles". The XDG Base Directory specification linked above is an attempt to migrate those hidden files to a standard directory structure. 

This makes the output of `ls -la` much more useful when in one's home directory, makes dotfile syncing easier as they are now contained in a single place (`$HOME/.config`) and generally is much less cluttered.

The [Archlinux wiki maintains a list of programs](https://wiki.archlinux.org/index.php/XDG_Base_Directory#Supported) (incomplete but pretty substantial) that adhere to or at least have workarounds to accommodate the spec.

erayaydin:
Any news about this improvement? I think we can add this issue to Arch Linux Wiki for tracking @distefam . XDG Base Directory Specification is important for most of the GNU/Linux users.

codingjerk:
@brentvatne
Can I propose a code that will keep backward compatibility? There are a lot of advantages to segregate configs and persistent/temporary data from each other.

@distefam
The files in `.expo` directory are not configs, so it will be better to store them in `$XDG_CACHE_HOME` aka `.cache` or even in `$XDG_DATA_HOME` aka `.local/share`.

brentvatne:
@codingjerk - i'm open to a pr that is backwards compatible!

[...]

PR #3722, opened 2021-07-27
Initial XDG_CONFIG_HOME support, see #2274

jedahan:
# Why

# How

# Test Plan


EvanBacon:
1. Need to fill out the entire PR template.
2. This support isn't robust and certain features like eas-cli and expo-updates would break.
3. The [source of truth for the CLI has moved](https://github.com/expo/expo/search?q=homedir).

PR #3725, opened 2021-07-27
[config-plugins] Added window soft input mode

EvanBacon:
# Why

- Resolve https://forums.expo.dev/t/softwarekeyboardlayoutmode-pan-works-in-expo-go-and-expo-build-android-but-not-in-eas-build/54978

# Test Plan

Tests should cover all important cases

### DIFF
--- a/packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
+++ b/packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
@@ -467,6 +467,7 @@ Object {
                   "android:label": "@string/app_name",
                   "android:name": ".MainApplication",
                   "android:theme": "@style/AppTheme",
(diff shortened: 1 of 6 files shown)


packages/prebuild-config/src/plugins/__tests__/__snapshots__/withDefaultPlugins-test.ts.snap
```diff
@@ -1264,6 +1265,7 @@ Object {
                   "android:roundIcon": "@mipmap/ic_launcher_round",
                   "android:theme": "@style/AppTheme",
                   "android:usesCleartextTraffic": "true",
+                  "android:windowSoftInputMode": "adjustPan",
```
brentvatne, line 1268: i believe this belongs on the activity level -- MainActivity rather than MainApplication

EvanBacon: good catch

PR #3737, opened 2021-08-02
[expo cli] start project using application identifier

EvanBacon:
# Why

Not all projects have a shared deep linking scheme, which means expo-cli currently cannot open all projects. This PR aims to restructure the "open on device" functionality to support using an application identifier as a fallback.

# How

- Guess the application identifier for a dev client project.
- When the shared scheme cannot be found, disable the QR code in Terminal UI.

# Test Plan

- Open a project bootstrapped with `npx react-native init` by running `expo start --dev-client --ios --android`


packages/xdl/src/Android.ts
```diff
@@ -815,32 +836,76 @@ export async function openProjectAsync({
...
+
+    if (!applicationId) {
+      return {
+        success: false,
+        error:
+          'Cannot resolve application identifier or URI scheme to open the native Android app.\nBuild the native app with `expo run:android` or `eas build -p android`',
+      };
+    }
+
+    const mainActivity = 'MainActivity';
```
EvanBacon, line 883: https://github.com/expo/expo-cli/pull/3741

PR #3743, opened 2021-08-03
[webpack config] reduce font loading complexity

EvanBacon:
# Why

Don't attempt to load fonts from packages that don't exist. Use node module resolution to determine where a package is located.


EvanBacon:
We'll drop this with the webpack 5 upgrade https://github.com/expo/expo-cli/pull/3763

PR #3745, opened 2021-08-05
[xdl] Created CoreSimulator tools

EvanBacon:
# Why

`xcrun simctl` is very slow, and as we increase the amount of features we support, everything gets really sluggish. For example, reopening an iOS project in Expo Go can take around ~2.8s.

This PR introduces some alternative methods to simctl which uses the filesystem and generated plists to infer the state of the Simulator (when `EXPO_USE_CORE_SIM=1`). Resulting in up to 7x faster flows when pressing "i" to reopen your app. It also improves the startup times, upgrading, installing, uninstalling, and device listing times.

# Test Plan

- `export EXPO_USE_CORE_SIM=1` to enable the feature (hidden behind a flag for testing)
- `EXPO_PROFILE=1 expo start --ios` -- then press `i`
  - Close the simulator and press `i` to reopen the simulator.
  - Close the Simulator.app and press `i` to reopen and reboot the simulator.
  - Delete the Expo Go app, and press `i` to install and open the app.


brentvatne (approved):
let's land this and publish soon, then bump the version of expo-cli in expo/expo and set `EXPO_USE_CORE_SIM=1` in direnv so we use it internally

PR #3751, opened 2021-08-05
[xdl] pass appropriate headers when making getManifestCall

jkhales:
# Why

`expo publish` with a runtime version was failing for the project described here:

https://exponent-internal.slack.com/archives/C020VQQ8WAE/p1628129369013400

# How

Pass the correct runtime headers when making a `getManifestAsync` call.

I'm unfamiliar with how this part of classic publishing works, but it seems safe to assume that where we were requesting an sdk runtime manifest, we should now be getting a runtime OR sdk manifest.

`grep -i` for other instances of `expo-sdk-version` and `exponenet-sdk-version` and found none.

# Test Plan

Repro'd error
Confirmed publish works with rtv.
Confirmed publish still works without rtv.

### DIFF
--- a/packages/xdl/src/project/publishAsync.ts
+++ b/packages/xdl/src/project/publishAsync.ts
@@ -133,15 +133,21 @@ export async function publishAsync(
     exp.android?.publishManifestPath ||
     EmbeddedAssets.shouldEmbedAssetsForExpoUpdates(projectRoot, exp, pkg, target)
   ) {
+    const sdkOrRuntimeVersion = exp.runtimeVersion
+      ? {
+          'expo-runtime-version': exp.runtimeVersion,
+        }
+      : { 'expo-sdk-version': exp.sdkVersion };
+
     [androidManifest, iosManifest] = await Promise.all([
       ExponentTools.getManifestAsync(response.url, {
-        'Exponent-SDK-Version': exp.sdkVersion,
+        ...sdkOrRuntimeVersion,
         'Exponent-Platform': 'android',
         'Expo-Release-Channel': options.releaseChannel,
         Accept: 'application/expo+json,application/json',
       }),
       ExponentTools.getManifestAsync(response.url, {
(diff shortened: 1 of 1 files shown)


jkhales:
There is the additional question of the excessively noisy error message.

Curious what the thinking behind this line may have been: https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/ExponentTools.ts#L29

On one hand we could gate it behind `EXPO_DEBUG` on the other hand even when the operation is successful we get a pretty noisy output here: https://github.com/expo/expo-cli/blob/master/packages/xdl/src/detach/ExponentTools.ts#L33

So curious if this is intentional for this workflow?

esamelson (approved):
This fix seems right to me. Thanks for finding the issue here 🙏 

I think (though not 100% sure) https://github.com/expo/expo-cli/blob/722bae2f13273608af5cd3b5d19a8baa0c177cd0/packages/xdl/src/detach/ExponentTools.ts#L33 exists so that the build logs in turtle v1 will show the exact manifest/update embedded in the app. I don't think we'd want to remove it, unless it's different from what I'm thinking. We might be able to move it around or only log it if we're in a turtle v1 context, though.

My best guess about https://github.com/expo/expo-cli/blob/722bae2f13273608af5cd3b5d19a8baa0c177cd0/packages/xdl/src/detach/ExponentTools.ts#L29 is that I think it would generally indicate something has gone wrong on our end (which it has in this case) so the more information we have, the better. But definitely worth revisiting if you think we could improve the output and/or if people are running into this more frequently. @quinlanj might be able to provide more context, too.

quinlanj (approved):
This looks right to me! It's been a really long time since I last looked at classic updates (self hosting in particular), but this change shouldn't affect it. 

As for the logging output -- I don't think it was put there for any special reason, looks like run of the mill error/debug handling to me (sometimes www outages cause this)

PR #3752, opened 2021-08-05
[xdl][expo-cli] Introduces RudderStack

kgc00:
# Why

We are deprecating Segment usage on our platform in favor of RudderStack. This PR will introduce RudderStack as our analytics client.

# How

- adds the RudderStack client to the `XDL` `Analytics` file
- adds types for the RudderStack analytics client
- keeps Segment around for now to ensure event volumes match
- the RudderStack client is using a fork that has all logs go through Winston. This way we can silence the error logs.

# Test Plan

- I've run `XDL` commands locally and confirmed they reach the proper destinations. 
- This PR logs analytics events to both Segment and RudderStack. Once this is merged we'll need to confirm event volumes match. Afterwards, we'll follow up with a PR to remove the `segmentClient` from `Analaytics`.

### DIFF
--- a/packages/expo-cli/src/exp.ts
+++ b/packages/expo-cli/src/exp.ts
@@ -18,6 +18,7 @@ import {
   ApiV2,
   Binaries,
   Config,
+  Env,
   Logger,
   LogRecord,
   LogUpdater,
@@ -655,8 +656,27 @@ Command.prototype.asyncActionProjectDir = function (
 };
 
 export async function bootstrapAnalyticsAsync(): Promise<void> {
-  Analytics.initializeClient('vGu92cdmVaggGA26s3lBX6Y5fILm8SQ7', packageJSON.version);
-  UnifiedAnalytics.initializeClient('u4e9dmCiNpwIZTXuyZPOJE7KjCMowdx5', packageJSON.version);
+  if (Env.shouldDisableAnalytics()) {
+    return; // do not allow E2E to fire events
+  }
+
+  Analytics.initializeClient(
+    'vGu92cdmVaggGA26s3lBX6Y5fILm8SQ7',
+    {
+      apiKey: '1wHTzmVgmZvNjCalKL45chlc2VN',
+      dataPlaneUrl: 'https://cdp.expo.dev',
+    },
+    packageJSON.version
+  );
+
+  UnifiedAnalytics.initializeClient(
+    'u4e9dmCiNpwIZTXuyZPOJE7KjCMowdx5',
+    {
+      apiKey: '1wabJGd5IiuF9Q8SGlcI90v8WTs',
+      dataPlaneUrl: 'https://cdp.expo.dev',
+    },
+    packageJSON.version
+  );
 
   const userData = await profileMethod(
     UserManager.getCachedUserDataAsync,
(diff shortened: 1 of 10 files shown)


packages/xdl/src/Analytics.ts
```diff
@@ -1,19 +1,115 @@
...
+    options?: {
+      flushAt?: number; // queue size
+      flushInterval?: number; // ms
+      // the max number of elements that the SDK can hold in memory,
+      // this is different than the Redis list created when persistence is enabled
+      maxInternalQueueSize?: number;
+      logLevel?: string;
+    }
+  ): RudderStackClient;
+  createPersistenceQueue(
```
wkozyra95, line 82: add a comment to warn that this function is never called in client side code

wkozyra95 (approved):
LGTM

I recommend testing bellow cases before merging:
- add `0.0.0.0   cdp.expo.dev` to `/etc/hosts` and check if there is no user-facing errors(amplitude client threw error in eas cli in that case)
- start redis (e.g. in docker) with default port 6379 and make sure nothing is written there

packages/xdl/src/Analytics.ts
kgc00: i've moved the typing to a d.ts file in the library

kgc00:
> LGTM
> 
> I recommend testing bellow cases before merging:
> 
> * add `0.0.0.0   cdp.expo.dev` to `/etc/hosts` and check if there is no user-facing errors(amplitude client threw error in eas cli in that case)
> * start redis (e.g. in docker) with default port 6379 and make sure nothing is written there

cool, great tip about `/etc/hosts`! Yup, we're using the expo fork of rudder node so that we can silence any errors. it's working properly!

packages/expo-cli/src/exp.ts
```diff
@@ -655,8 +655,30 @@ Command.prototype.asyncActionProjectDir = function (
 };
 
 export async function bootstrapAnalyticsAsync(): Promise<void> {
-  Analytics.initializeClient('vGu92cdmVaggGA26s3lBX6Y5fILm8SQ7', packageJSON.version);
-  UnifiedAnalytics.initializeClient('u4e9dmCiNpwIZTXuyZPOJE7KjCMowdx5', packageJSON.version);
+  const isE2e = getenv.boolish('E2E', false) || getenv.boolish('CI', false); // do not allow E2E to fire events
```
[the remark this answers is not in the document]
kgc00: oh wow, this is much cleaner! 👍

packages/xdl/src/Env.ts
```diff
@@ -27,3 +27,7 @@ export function maySkipManifestValidation(): boolean {
 export function shouldUseDevServer(exp: ExpoConfig) {
   return Versions.gteSdkVersion(exp, '40.0.0') || getenv.boolish('EXPO_USE_DEV_SERVER', false);
 }
+
+export function isE2EOrCI() {
```
wkozyra95, line 31: maybe sth less specific would be cleaner, e.g shouldEnableAnalytics,shouldEnableTracking

kgc00: I do like keeping it generic so other places in the CLI can also use this function to disable/enable behaviour on CI, but I'm happy to make the change. we can always revert it later on if we need to use it somewhere else.

PR #3758, opened 2021-08-10
Added fallback support for complex characters in prebuild

EvanBacon:
# Why

- resolve https://github.com/expo/eas-cli/issues/458 
- When name sanitization strips all characters from a name, fallback to "app".
- I've also added a step to use slugify if all characters are stripped, this would attempt to add extra diversity to project names.
- Unified name sanitization across the repo.

# How

# Test Plan

- `expo prebuild` in an app named: `あいう`

### DIFF
--- a/packages/config-plugins/src/ios/utils/Xcodeproj.ts
+++ b/packages/config-plugins/src/ios/utils/Xcodeproj.ts
@@ -1,6 +1,7 @@
 import { ExpoConfig } from '@expo/config-types';
 import assert from 'assert';
 import path from 'path';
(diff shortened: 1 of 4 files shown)


brentvatne (approved):

buraks:
From the looks of it, something else is broken for Android builds on EAS with special characters in the app name.

See https://github.com/expo/eas-cli/issues/458#issuecomment-1044350619

Everything works fine on iOS, only Android. If I change the app name the build is successful on Android, only when I have nonEnglish characters does the build fails

Issue #3461, opened 2021-05-04
Executing expo export modifies AndroidManifest.xml


ex3ndr:
I have unfortunate behaviour that i wasn't able to disable. I am using `expo-updates` with my own CDN and due some strange bugs i decided to put updates for each runtime version to separate folders to avoid accidentally pushing update to wrong folder and keep CDN fully stateless.

Unfortunatelly, it doesn't work well since i am unable to configure updates endpoint via gradle script because running `expo export` overwites current link in manifest that breaks my configuration and there are no way to disable this behaviour.

Expected behaviour: Not updating manifest or being able to disable this.

Suggestion for fix: Use gradle templating and allow gradle to resolve update url instead of writing to manifest.

jkhales:
Hi! That is frustrating. To understand a bit more, is this new behaviour or is this a suggestion for improvement?

ex3ndr:
This is current behaviour on current expo/expo-cli libraries, it works fine until you start to hack around with updates :) I suggest at least be able to avoid modification of manifest and there are no such behaviour for ios, only on Android.

jkhales:
> This is current behaviour on current expo/expo-cli libraries, it works fine until you start to hack around with updates :) I suggest at least be able to avoid modification of manifest and there are no such behaviour for ios, only on Android.

OK great, I am going to look into fixing this for you.

brentvatne:
maybe this would be well suited for an env var, and i think if we make it possible to opt-out of this behavior in export we should do the same for publish

zlwu:
Any progress on this topic?

ex3ndr:
Yeah, I already hit this issue in three different companies :( Almost everyone need some native modules and therefore ejected and begin to manually track runtime version.

jkhales:
Haven't forgotten about this! Just a lot of thing in the "TODO" box!

jkhales:
This is live as of expo-cli@4.11

PR #3764, opened 2021-08-12
[xdl] only configure native files when appropriate 

jkhales:
# Why

`expo publish` and `expo export` have an overly aggressive side effect that modifies the `Expo.plist` and `AndroidManifest.xml`. c.f. https://github.com/expo/expo-cli/issues/3461

# How

The side effect is in `_maybeConfigureExpoUpdatesEmbeddedAssetsAsync` renamed to `_maybeRunModifiedExpoUpdatesPluginAsync`.

This area of the code has multiple things wrong with it. 
1. The code modification is triggered every time, however judging from the log message it should be only happen when the there is the placeholder url `YOUR-APP-URL-HERE`. (Note that `isLikelyFirstPublish` is set, but never used to gate anything other than the log message).
2. We only check the `Expo.plist` for the placeholder url, we should do both and support platform specific flows.
3. We use custom code instead of importing from `config-plugins`

# Test Plan

Tested in a demo repo and confirmed
1. It modifies the native files if the update url is the place holder url.
2. It does it platform specific
3. It does not modify the `Expo.plist`/`AndroidManifest.xml` once a non-place holder url has been set.

### DIFF
--- a/packages/xdl/src/EmbeddedAssets.ts
+++ b/packages/xdl/src/EmbeddedAssets.ts
@@ -311,8 +311,8 @@ async function _maybeRunModifiedExpoUpdatesPluginAsync(config: EmbeddedAssetsCon
         let currentlyConfiguredValue = currentlyConfiguredExpoPlist[key];
         const inferredValue = expoPlistForProject[key];
         if (key === RELEASE_CHANNEL && inferredValue) {
-          // A client with an undefined release channel  is mapped to
-          // 'default' in the server so avoid logging an unneccessary warning.
+          // A client with an undefined release channel is mapped to
+          // 'default' in the server, so avoid logging an unneccessary warning.
           currentlyConfiguredValue = currentlyConfiguredValue ?? 'default';
         }
         if (currentlyConfiguredValue !== inferredValue) {
@@ -411,8 +411,8 @@ async function _maybeRunModifiedExpoUpdatesPluginAsync(config: EmbeddedAssetsCon
           x => x['android:name'] === key
         )?.['android:value'];
         if (key === RELEASE_CHANNEL && inferredValue) {
-          // A client with an undefined release channel  is mapped to
-          // 'default' in the server so avoid logging an unneccessary warning.
+          // A client with an undefined release channel is mapped to
+          // 'default' in the server, so avoid logging an unneccessary warning.
           currentlyConfiguredValue = currentlyConfiguredValue ?? 'default';
         }
         if (inferredValue !== currentlyConfiguredValue) {


packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -253,35 +263,44 @@ async function _maybeConfigureExpoKitEmbeddedAssetsAsync(config: EmbeddedAssetsC
...
+  const expoPlistPath = path.join(supportingDirectory, 'Expo.plist');
+  if (fs.existsSync(expoPlistPath)) {
+    const expoPlistForProject = plist.parse(await readFileSync(expoPlistPath, 'utf8'));
+    const currentIOSUpdatesURL = expoPlistForProject[IOSConfig.Updates.Config.UPDATE_URL];
+    if (currentIOSUpdatesURL === PLACEHOLDER_URL) {
+      isLikelyFirstIOSPublish = true;
+
+      // The username is only used for defining a default updates URL.
+      // Since we overwrite the URL below the username is superfluous.
+      IOSConfig.Updates.setUpdatesConfig(exp, expoPlistForProject, /*expoUsername*/ null);
```
jkhales, line 289: This edits *more* properties than before, but they are all appropriate and their absence was arguably another defect. Namely:

1. `EXUpdatesCheckOnLaunch`
2. `EXUpdatesLaunchWaitMs`
3. `EXUpdatesEnabled`

Further it will set the `EXUpdatesRuntimeVersion` instead of the `EXUpdatesSDKVersion` when appropriate.

https://github.com/expo/expo-cli/blob/master/packages/config-plugins/src/ios/Updates.ts#L68

And analogously for Android

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -290,54 +309,62 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
...
+    let androidManifest = await readAndroidManifestAsync(androidManifestXmlPath);
+    const currentAndroidUpdateURL = getMainApplicationMetaDataValue(
+      androidManifest,
+      AndroidManifestKeyForUpdateURL
     );
+    if (currentAndroidUpdateURL === PLACEHOLDER_URL) {
+      isLikelyFirstAndroidPublish = true;
+      // The username is only used for defining a default updates URL.
+      // Since we overwrite the URL below the username is superfluous.
+      androidManifest = AndroidConfig.Updates.setUpdatesConfig(
```
jkhales, line 323: See https://github.com/expo/expo-cli/pull/3764/files#r687724121

packages/xdl/src/detach/IosPlist.ts
```diff
@@ -93,18 +93,4 @@ async function createBlankAsync(plistPath: string, plistName: string) {
   fs.removeSync(tmpConfigFile);
 }
 
-async function cleanBackupAsync(plistPath: string, plistName: string, restoreOriginal = true) {
```
jkhales, line 96: No longer called anywhere.

This entire file seems like it should be phased out eventually.

EvanBacon: most of this xdl stuff was supposed to be legacy turtle v1 things that we replaced with config plugins, but they're still referenced a bit in export/publish

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -290,54 +309,62 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
...
   }
 
-  if (isLikelyFirstPublish) {
+  if (isLikelyFirstIOSPublish || isLikelyFirstAndroidPublish) {
+    let platformSpecificMessage: string;
+
+    if (isLikelyFirstIOSPublish && !isLikelyFirstAndroidPublish) {
+      platformSpecificMessage =
+        '🚀 It looks like this your first iOS publish for this project! ' +
+        "We've automatically set some configuration values in Expo.plist. ";
```
EvanBacon, line 355:
```suggestion
        "Automatically setup Expo updates in the Expo.plist";
```

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -1,6 +1,14 @@
...
-import { IOSConfig } from '@expo/config-plugins';
-import fs from 'fs-extra';
+import { AndroidConfig, IOSConfig } from '@expo/config-plugins';
+import {
+  addMetaDataItemToMainApplication,
+  getMainApplicationMetaDataValue,
+  getMainApplicationOrThrow,
+  readAndroidManifestAsync,
+  writeAndroidManifestAsync,
+} from '@expo/config-plugins/build/android/Manifest';
```
EvanBacon, line 9: Use `AndroidConfig.Manifest.getMainApplicationOrThrow` etc.

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -1,6 +1,14 @@
...
+import plist from '@expo/plist';
+import fs, { readFileSync, writeFileSync } from 'fs-extra';
```
EvanBacon, line 11:
```suggestion
import fs from 'fs-extra';
```

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -290,54 +309,62 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
...
+      platformSpecificMessage =
+        '🚀 It looks like this your first publish for this project! ' +
+        "We've automatically set some configuration values in Expo.plist and AndroidManifest.xml. ";
+    }
+
     logger.global.warn(
-      '🚀 It looks like this your first publish for this project! ' +
-        "We've automatically set some configuration values in Expo.plist and AndroidManifest.xml. " +
+      platformSpecificMessage +
         "You'll need to make a new build with these changes before you can download the update " +
```
EvanBacon, line 368:
```suggestion
        "The native app must be rebuilt, and downloaded by your users before they can access the new JavaScript bundle " +
```

EvanBacon (approved):
Looks fine from what I can tell, just some print and import fixes.

esamelson (changes requested):
>     1. The code modification is triggered every time

The platform specific flows seem like a good change, but this original behavior was intentional. If the URL ever changes after it's set initially (e.g. the app's slug changes, the username or owner is different) or the developer publishes to a different release channel, then when they publish an update it will never be picked up by existing builds. This is why we set these values on each publish, not just the initial one.

The idea was that modifying the native build files on every publish would therefore pick up all of these changes and would at the very least be a signal to developers that something unintended has changed.

Another possible solution would be to compare the existing values to what we _would_ set them to and output a warning message in `expo publish`/`expo export` if they're different. But I don't think we want to remove this check entirely, as this PR currently does.

jkhales:
Pic of new warning messages:

Screen Shot 2021-08-19 at 8 32 33 AM

esamelson (changes requested):
Code looks good, just a couple suggestions re: wording, and could we also:
1. only show the warning for critical properties that matter for update compatibility (url, sdk/rtv, release channel) rather than all properties?
2. change the wording in the screenshot to communicate that these properties matter for update compatibility -- maybe something like
```
Your app.json (key) is set to (value), but (key) in Expo.plist is set to (value).
Make sure the value in app.json matches the builds you're intending to update.
```

packages/xdl/src/EmbeddedAssets.ts
```diff
@@ -290,55 +325,98 @@ async function _maybeConfigureExpoUpdatesEmbeddedAssetsAsync(config: EmbeddedAss
...
   }
 
-  if (isLikelyFirstPublish) {
+  if (isLikelyFirstIOSPublish || isLikelyFirstAndroidPublish) {
+    let platformSpecificMessage: string;
+
+    if (isLikelyFirstIOSPublish && !isLikelyFirstAndroidPublish) {
+      platformSpecificMessage =
+        '🚀 It looks like this your first iOS publish for this project! ' +
+        'Automatically setup Expo updates in the Expo.plist. ';
```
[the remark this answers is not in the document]
jkhales: Heh, this is a change suggested by @EvanBacon . Do you have any opposition to this proposed wording?

jkhales:
>Your app.json (key) is set to (value), but (key) in Expo.plist is set to (value).
Make sure the value in app.json matches the builds you're intending to update.

Changed the wording to follow the spirit of this suggestion. (Realized the messages should be a bit more personalized as well).

Screen Shot 2021-08-20 at 9 07 39 AM

brentvatne:
i'd suggest looking at the output of a command like `expo prebuild` or `eas build` and notice how we use `chalk` to highlight certain keywords. also, this message is quite long so it may be better to include a short summary of it and then link to a fyi page for more information.

jkhales:
Modernized logging following @brentvatne 's suggestion:

Screen Shot 2021-08-23 at 10 26 27 AM

Created a PR for an FYI page (linked to in the `learn more` log output):
https://github.com/expo/fyi/pull/40

packages/xdl/src/EmbeddedAssets.ts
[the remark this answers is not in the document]
jkhales: Going to revert to @esamelson suggestions.

PR #3771, opened 2021-08-14
expo.io -> expo.dev

brentvatne:
# Why

We have a new domain, we should use it everywhere.

# How

Search and replace but careful not to clobber `expo.ios.xyz` strings. Also skipped over schemas, because that needs to be handled in universe.

# Test Plan

👀

### DIFF
--- a/packages/dev-tools/components/ProjectManagerPublishingSection.js
+++ b/packages/dev-tools/components/ProjectManagerPublishingSection.js
@@ -283,7 +283,7 @@ export default class ProjectManagerPublishingSection extends React.Component {
         <p className={STYLES_PARAGRAPH}>
           Once you publish your project, you will be able to view it at&nbsp;
           <span className={STYLES_EMPHASIS}>
-            https://expo.io/@{this.props.user.username}/projects/{this.state.config.slug}
+            https://expo.dev/@{this.props.user.username}/projects/{this.state.config.slug}
           </span>
           .
         </p>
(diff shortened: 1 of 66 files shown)


packages/dev-tools/server/graphql/GraphQLSchema.ts
```diff
@@ -287,7 +287,7 @@ const typeDefs = graphql`
     openSimulator(platform: Platform!): OpenSimulatorResult
     # Starts WebPack server
     openWeb: OpenWebResult
-    # Publishes the current project to expo.io
+    # Publishes the current project to expo.dev
```
ide, line 290:
```suggestion
    # Publishes the current project to expo.dev (classic updates)
```

packages/dev-tools/components/ProjectManagerPublishingSection.js
```diff
@@ -283,7 +283,7 @@ export default class ProjectManagerPublishingSection extends React.Component {
         <p className={STYLES_PARAGRAPH}>
           Once you publish your project, you will be able to view it at&nbsp;
           <span className={STYLES_EMPHASIS}>
-            https://expo.io/@{this.props.user.username}/projects/{this.state.config.slug}
+            https://expo.dev/@{this.props.user.username}/projects/{this.state.config.slug}
```
ide, line 286: Unrelated but we really need to be using `encodeURIComponent`. Here the worst an attack can do is create a broken URL but this kind of programming could open vulnerabilities in other contexts.

ide (approved):

FiberJW (approved):

PR #3773, opened 2021-08-16
[config-plugins] add vendor paths used by Bundlers to ignoredPaths

javascripter:
# Why
`expo run:ios` may fail on projects that use Bundler to manage CocoaPods and Fastlane versions.

With `Gemfile` at ios project root:
```
source 'https://rubygems.org'

gem 'cocoapods'
gem 'cocoapods-keys'

gem 'fui', '~> 0.3.0'
gem 'xcpretty'
gem 'second_curtain', '~> 0.2.3'
gem 'fastlane'
```
When you run `cd ios && bundle exec pod install`, Bundler saves files to `vendor/bundle` paths such as `ios/vendor/bundle` and `android/vendor/bundle`, which may have unrelated xcodeproj files.

For example, after `bundle exec pod install`, my project end up having files like below:

```
  'XXXX/android/vendor/bundle/ruby/2.5.0/gems/fastlane-2.157.2/fastlane/swift/FastlaneSwiftRunner/FastlaneSwiftRunner.xcodeproj',
  'XXXX/ios/ExampleApp.xcodeproj',
  'XXXX/ios/Pods/Pods.xcodeproj',
  'XXXX/ios/vendor/bundle/ruby/2.5.0/gems/fastlane-2.157.2/fastlane/swift/FastlaneSwiftRunner/FastlaneSwiftRunner.xcodeproj',
```
These cause `expo run:ios` to fail because it picks up the first xcodeproj found.

# How

I added `vendor` to ignoredPaths.

# Test Plan
`expo run:ios` successfully builds on projects that uses Bundler after this change.

### DIFF
--- a/packages/config-plugins/src/ios/Paths.ts
+++ b/packages/config-plugins/src/ios/Paths.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 import { UnexpectedError } from '../utils/errors';
 import * as WarningAggregator from '../utils/warnings';
 
-const ignoredPaths = ['**/@(Carthage|Pods|node_modules)/**'];
+const ignoredPaths = ['**/@(Carthage|Pods|vendor|node_modules)/**'];
 
 interface ProjectFile<L extends string = string> {
   path: string;
--- a/packages/expo-cli/src/commands/run/ios/resolveOptionsAsync.ts
+++ b/packages/expo-cli/src/commands/run/ios/resolveOptionsAsync.ts
@@ -26,7 +26,7 @@ export type ProjectInfo = {
   name: string;
 };
 
-const ignoredPaths = ['**/@(Carthage|Pods|node_modules)/**'];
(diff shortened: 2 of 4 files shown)


packages/config-plugins/src/ios/Paths.ts
```diff
@@ -5,7 +5,7 @@ import * as path from 'path';
 import { UnexpectedError } from '../utils/errors';
 import * as WarningAggregator from '../utils/warnings';
 
-const ignoredPaths = ['**/@(Carthage|Pods|node_modules)/**'];
+const ignoredPaths = ['**/@(Carthage|Pods|vendor|node_modules)/**'];
```
EvanBacon, line 8: We use this pattern in a few places, can you check to see if they need updated too.

javascripter: Thanks. I confirmed the patterns needed to be updated in other places as well (`eas build:configure` failed for the same reason as well). I updated other files.

EvanBacon (approved):
lgtm

PR #3782, opened 2021-08-18
fix: only show supported platforms in dev tools and terminal UIs

cruzach:
# Why

Got a request to hide the option to open in web, since if you're using `@expo/next-adapter` you need to use `yarn next dev` instead. Seems like we should hide any options that is excluded from `platforms` 

# How

put all the 'open in X platform' behind a flag that depends on the `platforms` key of your app config

# Test Plan

`expod start`

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.ts
+++ b/packages/expo-cli/src/commands/start/TerminalUI.ts
@@ -35,6 +35,7 @@ type StartOptions = {
   nonPersistent?: boolean;
   maxWorkers?: number;
   webOnly?: boolean;
+  platforms?: ('android' | 'ios' | 'web')[];
 };
 
 const printHelp = (): void => {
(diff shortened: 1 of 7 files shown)


packages/expo-cli/src/commands/start/TerminalUI.ts
```diff
@@ -56,23 +57,24 @@ const printUsageAsync = async (
...
-    ['a', `open Android`],
-    ['shift+a', `select a device or emulator`],
-    isMac && ['i', `open iOS simulator`],
-    isMac && ['shift+i', `select a simulator`],
-    ['w', `open web`],
+    platforms.includes('android') && ['a', `open Android`],
+    platforms.includes('android') && ['shift+a', `select a device or emulator`],
+    platforms.includes('ios') && isMac && ['i', `open iOS simulator`],
+    platforms.includes('ios') && isMac && ['shift+i', `select a simulator`],
+    platforms.includes('web') && ['w', `open web`],
```
EvanBacon, line 77: This would still be shown in projects that use Next.js

cruzach: not if you set `"platforms": ["ios", "android"]` in app config

cruzach:
Closing because we don't rely on the `platforms` key of the app config for anything, and so we don't want to _start_ relying on it for this

Issue #2809, opened 2020-10-19
Launch app in multiple emulators simultaneously


kevgrig:
## Description

I was very excited by the new feature to target multiple emulators in [SDK 39](https://blog.expo.io/expo-sdk-39-is-now-available-4c10aa825e3f) as this will greatly improve my development experience:

> You can now pick which device or simulator to run your app on.

However, when I use `Shift+a`, although I see two emulators, selecting either one and typing `a` only opens the app on one of the emulators.

## Expected Behavior

Switch emulators with `Shift+a` and launch the app simultaneously in both emulators.

## Observed Behavior

App only ever opens in one emulator even though I try to switch emulators with `Shift+a`

## Environment

```
$ expo diagnostics

  Expo CLI 3.28.0 environment info:
    System:
      OS: Linux 5.8 Fedora 32 (Workstation Edition) 32 (Workstation Edition)
      Shell: 5.0.17 - /bin/bash
    Binaries:
      Node: 12.18.3 - /usr/bin/node
      Yarn: 1.21.1 - /usr/bin/yarn
      npm: 6.14.6 - /usr/bin/npm
    npmPackages:
      expo: ^39.0.0 => 39.0.3 
      react: 16.13.1 => 16.13.1 
      react-native: https://github.com/expo/react-native/archive/sdk-39.0.3.tar.gz => 0.63.2 
    npmGlobalPackages:
      expo-cli: 3.27.4
    Expo Workflow: managed
```

Phone/emulator/simulator platform, model and version: Pixel 2 (API 30) and Pixel 3a (API 30)

## Reproducible Demo

1. Start both emulators and start expo:
    ```
    $ pkill -9 -f adb
    $ ~/Android/Sdk/emulator/emulator -avd Pixel_2_API_30 &>/tmp/emulator1.log &
    $ ~/Android/Sdk/emulator/emulator -avd Pixel_3a_API_30 &>/tmp/emulator2.log &
    $ expo start -c
    [...]
    ```
1. Type `Shift+a` to list emulators:
    ```
    ? Select a device/emulator › 
    ❯    (emulator)
         (emulator)
    ```
1. Press `Enter` and then type `a` and the app launches in one emulator.
1. Type `Shift+a` again and select the second emulator:
    ```
    ? Select a device/emulator › 
        (emulator)
    ❯   (emulator)
    ```
1. Type `a` again but the app still launches in the same emulator as before.

kevgrig:
1. List all AVDs. For example:
[...]

PR #3793, opened 2021-08-23
[expo-cli] Open all attached emulators

EvanBacon:
# Why

- Resolve https://github.com/expo/expo-cli/issues/2809
- I built the feature but I'm not sure if it makes total sense, need to test it in the dev-client context a bit more before I can determine if the slowdown / degraded error handling is worth it.

# Test Plan

- Press `a` to open an Emulator, `A` to pick another emulator. Open on both by pressing `a` and not `A`


EvanBacon:
Stale, this feature may just work™ in the new versioned CLI.

PR #3796, opened 2021-08-24
[config-plugins] Add hermes support for ios Podfile.properties.json

Kudo:
# Why

add hermes support for ios.

# How

- xdl-schema related pr: https://github.com/expo/universe/pull/8063
- client related pr: https://github.com/expo/expo/pull/14119
- add `podfileProperties` mod and `withPodfileProperties` plugin
- add `withJsEnginePodfileProps` plugin and will read `config.ios.jsEngine`, then `config.jsEngine` and generate `Podfile.properties.json` accordingly.

# Test Plan

```
 PASS   @expo/config-plugins  src/ios/__tests__/JsEngine-test.ts
  jsEngine
    ✓ returns default engine if no engine is provided (2 ms)
    ✓ return the engine if provided
    ✓ return the engine where platform override config has higher priority (1 ms)
    ✓ set the property from shared `jsEngine` config
    ✓ set the property from platform `jsEngine` override config
    ✓ overwrite the property if an old property is existed (1 ms)
 PASS   @expo/prebuild-config  src/plugins/__tests__/withDefaultPlugins-test.ts
  built-in plugins
    ✓ create Podfile.properties.json file for backward compatible (196 ms)
```


packages/config-plugins/src/ios/__tests__/JsEngine-test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const config: Partial<ExpoConfig> = { jsEngine: 'hermes', ios: { jsEngine: 'jsc' } };
+    expect(getJsEngine(config)).toBe('jsc');
+  });
+
+  it('set the property from shared `jsEngine` config', () => {
+    const config: Partial<ExpoConfig> = { jsEngine: 'hermes' };
+    const podfileProperties = {};
+
+    expect(setJsEngine(config, podfileProperties)).toMatchObject({
+      [`${JS_ENGINE_PROP_KEY}`]: 'hermes',
```
EvanBacon, line 26:
```suggestion
      [JS_ENGINE_PROP_KEY]: 'hermes',
```
??

packages/config-types/src/ExpoConfig.ts
```diff
@@ -3,6 +3,11 @@
  * The standard Expo config object defined in `app.config.js` files.
  */
 
+/**
+ * Specifies the JavaScript engine for apps. Supported only on EAS Build. Defaults to `jsc`. Valid values: `hermes`, `jsc`.
+ */
+export type JsEngine = 'hermes' | 'jsc';
```
EvanBacon, line 9: This file is generated, first you need to add the types to expo/universe (example: https://github.com/expo/universe/pull/7876) then you can run:

```
yarn generate --path ../../../universe/server/www/xdl-schemas/UNVERSIONED-schema.json
```
[reference](https://github.com/expo/expo-cli/blob/master/packages/config-types/README.md#contributing)

EvanBacon (changes requested):
Looks good so far, should update the integration test in prebuild-config and add a mock file for `Podfile.properties.json`.

zhigang1992:
If we are adding Hermes engine, does it make sense to expose use_flipper! behind a flag as well.

Flipper is rather useful when debugging Hermes code.

PR #3803, opened 2021-08-27
[config-cli][xdl] add getRuntimeVersion method

jkhales:
# Why

A method to parse the increased complexity of the `runtimeVersion` field from [increased complexity](https://github.com/expo/universe/pull/8081).

# How

1. Added `getNativeBuildVersion` to handle new 'nativeVersion' policy.
2. 'sdkVersion' policy triggers `exposdk:${sdkVersion`}.
3. Allow platform specific overrides

# Test Plan

Added unit tests.


packages/config/src/Config.ts
```diff
@@ -488,8 +488,28 @@ function ensureConfigHasDefaultValues({
...
+        runtimeVersion: getRuntimeVersionNullable(expWithDefaults, 'android') ?? undefined,
+      } as any) //TODO-JJ remove this cast in SDK 43 https://linear.app/expo/issue/ENG-1869/remove-tempruntimeversion-in-expoconfig
+    : undefined;
+
   return {
-    exp: { ...expWithDefaults, sdkVersion, platforms },
+    exp: {
+      ...expWithDefaults,
+      sdkVersion,
+      runtimeVersion: getRuntimeVersionNullable(expWithDefaults) ?? undefined,
```
esamelson, line 508: Will this throw if a policy is specified at the root level?

Even if not, I'm sort of concerned about the idea of having a field here that sometimes evaluates to the correct value and is sometimes null -- it seems like a bad idea for any other code to rely on this value. Do you have a sense of what code might use this particular field?

packages/config/src/Project.ts
```diff
@@ -25,3 +29,73 @@ export function getExpoSDKVersion(
...
+  },
+  platform: 'android' | 'ios'
+): string {
+  if (!config.version) {
+    throw new ConfigError('Missing "version" field', 'INVALID_CONFIG');
+  }
+  switch (platform) {
+    case 'ios': {
+      if (!config.ios?.buildNumber) {
+        throw new ConfigError(
```
esamelson, line 46: I think buildNumber and versionCode might have default values if they're unset. It might be a nicer UX to just use the default values here, too -- as if we were reading them from the actual native build files (even though we're really not).

packages/config/src/__tests__/Project-test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const version = '1';
+    const buildNumber = '2';
+    expect(
+      getRuntimeVersionNullable(
+        { version, runtimeVersion: { policy: 'nativeBuildVersion' }, ios: { buildNumber } },
+        'ios'
+      )
+    ).toBe(`${version}(${buildNumber})`);
+  });
+  it('returns null if no runtime version is found', () => {
```
esamelson, line 62: If we do go with a sometimes-null root-level runtimeVersion field, I think we may want to differentiate between projects that don't have a runtimeVersion at all (and therefore should use SDK version) and projects that do but don't have a single root-level runtimeVersion (and you need to look in the platform objects).

esamelson (commented):
The implementation looks good -- this intermediate config step is kind of inherently funky though. Left a few comments.

packages/config/src/Config.ts
jkhales: Yea, good point. 

I was thinking about support for classic updates here. But thinking about it, maybe we can use the platform specific runtimes for them as well... 🤔

packages/config-plugins/src/android/Version.ts
```diff
@@ -34,6 +34,7 @@ export function setVersionName(config: Pick<ExpoConfig, 'version'>, buildGradle:
 }
 
 export function getVersionCode(config: Pick<ExpoConfig, 'android'>) {
+  // TODO-JJ import this from  @expo/config once new version is published
   return config.android?.versionCode ?? 1;
```
EvanBacon, line 38: config -> config-plugins -> config is a cyclic dep

jkhales: good point. Is there a reason why @expo/config-plugins is being treated as a parent? Seems a bit awkward. 🤔

jkhales: After discussing in slack with @EvanBacon, have come to understand that config-plugins access the expoConfig via `getConfig`. Thus if we set the default values there (which is part of this PR below), we can delete these "default ensuring" functions once this PR is published..

packages/config/src/Config.ts
```diff
@@ -488,8 +494,35 @@ function ensureConfigHasDefaultValues({
...
+  // TODO(JJ) remove top level runtimeVersion in favor of platform specific definitions
+  // While there is code that expects either a string or undefined here, pass only non-object runtime Version
+  const topLevelRuntimeVersion =
+    typeof expWithDefaults.runtimeVersion === 'string' ? expWithDefaults.runtimeVersion : undefined;
+
   return {
-    exp: { ...expWithDefaults, sdkVersion, platforms },
+    exp: {
+      ...expWithDefaults,
+      version: getVersion(expWithDefaults),
```
jkhales, line 519: @EvanBacon & @esamelson does it feel right to you guys to be falling back to the default `version`, `versionCode`, and `buildNumber` fields here?

Since we are formatting the `runtimeVersion` which is a function of these values, it seems correct to ensure that these values are explicit.

packages/config-plugins/src/android/Version.ts
```diff
@@ -34,7 +34,7 @@ export function setVersionName(config: Pick<ExpoConfig, 'version'>, buildGradle:
 }
 
 export function getVersionCode(config: Pick<ExpoConfig, 'android'>) {
-  return config.android?.versionCode ?? 1;
+  return config.android?.versionCode;
```
jkhales, line 37: Repeating [comment](https://github.com/expo/expo-cli/pull/3803#discussion_r700187819) here:

After discussing in slack with @EvanBacon, have come to understand that config-plugins access the expoConfig via getConfig. Thus if we set the default values there (which is part of this PR below), we can delete these "default ensuring" functions once this PR is published..

packages/config-plugins/src/utils/__tests__/Updates-test.ts
```diff
@@ -1,20 +1,93 @@
...
+  it(`returns correct legacy urls if 'updates.url' is not provided, but 'slug' and ('username'|'owner') are provided.`, () => {
+    expect(getUpdateUrl({ slug: 'my-app' }, 'user')).toBe('https://exp.host/@user/my-app');
+    expect(getUpdateUrl({ slug: 'my-app', owner: 'owner' }, 'user')).toBe(
+      'https://exp.host/@owner/my-app'
+    );
+    expect(getUpdateUrl({ slug: 'my-app', owner: 'owner' }, null)).toBe(
+      'https://exp.host/@owner/my-app'
+    );
+  });
 });
```
jkhales, line 24: Just wrapped in a describe.

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -19,3 +25,82 @@ export function getUpdateUrl(
...
+      );
+    }
+  }
+}
+
+/**
+ * Compute runtime version policies.
+ * @return an expoConfig with only string valued platform specific runtime versions.
+ */
+export const withRuntimeVersion: ConfigPlugin = config => {
```
jkhales, line 58: This will be used here: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/project/expoConfig.ts in a follow up PR.

Not yet clear how much classic updates will support these new runtime version policies. For now, classic updates expects a string in a top level runtime version, or uses the sdkVersion.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -20,13 +20,18 @@ import {
...
   const url = req.url ? parse(req.url, /* parseQueryString */ true) : null;
   const platform = url?.query.platform || req.headers['expo-platform'];
   if (!platform) {
     throw new Error('Must specify expo-platform header or query parameter');
   }
-  return String(platform);
+
+  const stringifiedPlatform = String(platform);
+  if (!['android', 'ios'].includes(stringifiedPlatform)) {
+    throw new Error(`platform must be "android" or "ios". Recieved: "${platform}"`);
```
jkhales, line 32: Web is not supported for now: https://exponent-internal.slack.com/archives/C013ZK4SA12/p1630699514008700

jkhales: (please correct me if I'm misunderstanding this...)

jkhales: talking with @wschurman in general this endpoint is in development so very low risk.

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -1,10 +1,16 @@
-import { ExpoConfig } from '@expo/config-types';
+import { Android, ExpoConfig, IOS } from '@expo/config-types';
+import { getRuntimeVersionForSDKVersion } from '@expo/sdk-runtime-versions';
```
jkhales, line 2: @esamelson did you have a reason for wanting to put `getRuntimeVersion` into a fully separate module like `getRuntimeVersionForSDKVersion` instead of just putting it in `config-plugins` like this PR is proposing?

packages/config-plugins/src/android/Updates.ts
```diff
@@ -35,10 +35,18 @@ export const withUpdates: ConfigPlugin<{ expoUsername: string | null }> = (
   });
 };
 
-export function getRuntimeVersion(
+/**
+ * runtime version maybe null in projects using classic updates. In that
+ * case we use SDK version
+ */
+export function getRuntimeVersionNullablle(
```
esamelson, line 42:
```suggestion
export function getRuntimeVersionNullable(
```

packages/config-plugins/src/utils/Updates.ts
esamelson: No, I meant "module" as in "ts file in a package" rather than "package," but I suppose I could have just written "method" instead. Sorry for the confusion! I think this is good, as long as this method is also accessible by the publish scripts (which it looks like it is / will be).

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -19,3 +25,82 @@ export function getUpdateUrl(
...
+    }
+  }
+}
+
+/**
+ * Compute runtime version policies.
+ * @return an expoConfig with only string valued platform specific runtime versions.
+ */
+export const withRuntimeVersion: ConfigPlugin = config => {
+  if (config.ios || config.runtimeVersion) {
```
esamelson, line 59: Will we end up with the correct result here if `config.ios` is truthy but both `config.runtimeVersion` and `config.ios.runtimeVersion` are undefined? Wondering if this should be checking `if (config.ios?.runtimeVersion || config.runtimeVersion)` instead.

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -19,3 +25,82 @@ export function getUpdateUrl(
...
+  const runtimeVersion = config[platform]?.runtimeVersion ?? config.runtimeVersion;
+  if (!runtimeVersion) {
+    throw new Error(
+      `There is neither a value or a policy set for the runtime version on "${platform}"`
+    );
+  }
+
+  if (typeof runtimeVersion === 'string') {
+    return runtimeVersion;
+  } else if (runtimeVersion['policy'] === 'nativeVersion') {
```
esamelson, line 92: any reason not to do
```suggestion
  } else if (runtimeVersion.policy === 'nativeVersion') {
```
here and below?

esamelson (commented):
a few small comments but overall this is looking great!

packages/config-plugins/src/utils/Updates.ts
jkhales: great catch, thank you!

esamelson (approved):

PR #3805, opened 2021-08-28
[expo-cli][xdl] Removes Segment

kgc00:
# Why

Volumes are look even between Segment and RudderStack. Removing Segment to finish the migration to RudderStack.

# How

Removed Segment from xdl and expo-cli. 

Slight refactor to the `Env.shouldDisableAnalytics` method  => `Env.shouldEnableAnalytics`. This way I could pull it out of `boostrapAnalyticsAsync` and remove the mock for it inside `exp.ts`. It's a bit simpler and cleaner, imo.

# Test Plan

Tested locally to confirm events fired outside of tests.

### DIFF
--- a/packages/expo-cli/src/__tests__/exp-test.ts
+++ b/packages/expo-cli/src/__tests__/exp-test.ts
@@ -11,12 +11,6 @@ Command.prototype.asyncAction = jest.fn().mockImplementation(_ => program);
 Command.prototype.allowOffline = jest.fn().mockImplementation(_ => program);
 Command.prototype.urlOpts = jest.fn().mockImplementation(_ => program);
 
-jest.mock('xdl', () => {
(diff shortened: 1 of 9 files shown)


packages/expo-cli/src/__tests__/exp-test.ts
```diff
@@ -11,12 +11,6 @@ Command.prototype.asyncAction = jest.fn().mockImplementation(_ => program);
 Command.prototype.allowOffline = jest.fn().mockImplementation(_ => program);
 Command.prototype.urlOpts = jest.fn().mockImplementation(_ => program);
 
-jest.mock('xdl', () => {
```
kgc00, line 14: no longer necessary since this check got pulled above `bootstrapAnalyticsAsync`

wkozyra95 (approved):

PR #3817, opened 2021-09-02
[config-plugins] Add plugin to restrict android permissions

EvanBacon:
# Why

fallback idea for ENG-1226

### DIFF
--- a/packages/config-plugins/src/android/Permissions.ts
+++ b/packages/config-plugins/src/android/Permissions.ts
@@ -1,4 +1,5 @@
 import { ExpoConfig } from '@expo/config-types';
+import assert from 'assert';
 
 import { ConfigPlugin } from '../Plugin.types';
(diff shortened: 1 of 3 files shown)


packages/config-plugins/src/android/Permissions.ts
```diff
@@ -22,6 +23,76 @@ export const withPermissions: ConfigPlugin<string[] | void> = (config, permissio
...
+ *
+ * @param manifestPermissions manifest `uses-permissions` array.
+ * @param permission `android:name` of the permission to restrict
+ * @returns
+ */
+function ensureBlockedPermission(
+  manifestPermissions: ManifestUsesPermission[],
+  permission: string
+) {
+  // Remove permission if it currently exists
```
brentvatne, line 86: :+1:

brentvatne (approved):

Issue #3820, opened 2021-09-02
Publish fails with expo-cli 4.11.0


nemesis:
### Summary

I'm trying to publish a JavaScript bundle to a specific release channel in the bare workflow, and the `expo publish` fails with an error on `expo-cli@4.11.0`, but works fine with `expo-cli@4.10.0`.

### Environment

```
Expo CLI 4.11.0 environment info:
    System:
      OS: macOS 11.4
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 15.14.0 - ~/.nvm/versions/node/v15.14.0/bin/node
      Yarn: 1.22.11 - /usr/local/bin/yarn
      npm: 7.7.6 - ~/.nvm/versions/node/v15.14.0/bin/npm
      Watchman: 2021.08.30.00 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.10.1 - ~/.rvm/gems/ruby-2.7.2/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    IDEs:
      Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~41.0.1 => 41.0.1
      react: 16.13.1 => 16.13.1
      react-dom: 16.13.1 => 16.13.1
      react-native: ~0.63.4 => 0.63.4
      react-native-web: ~0.13.12 => 0.13.18
    npmGlobalPackages:
      expo-cli: 4.11.0
    Expo Workflow: bare
```

### Please specify your device/emulator/simulator platform, model and version

not applicable

### Error output

```
› Expo SDK: 41.0.0
› Release channel: default
› Workflow: Bare

Building optimized bundles and generating sourcemaps...

Publish Assets:
- Asset target: bare


Expo Metro config:
- Bundler target: bare
- Legacy: false
- Extensions: ts, tsx, js, jsx, json
- React Native: project_dir/node_modules/react-native
- Babel config: project_dir/babel.config.js


Expo Metro config:
- Bundler target: bare
- Legacy: false
- Extensions: ts, tsx, js, jsx, json
- React Native: project_dir/node_modules/react-native

Starting Metro Bundler
iOS Bundling complete 3602ms

Bundle                     Size
┌ index.ios.js          3.31 MB
├ index.android.js      3.33 MB
├ index.ios.js.map       9.5 MB
[...]

PR #3825, opened 2021-09-03
[xdl] throw if AndroidManifest.xml is missing a MainApplication

jkhales:
# Why

https://github.com/expo/expo-cli/issues/3820

# How

Use method that throws with an informative error if MainApplication is missing. Also filters for only MainApplication as that is all that is needed.

# Test Plan

tested in demo repo with MainApplication both present and missing.


EvanBacon (approved):
lgtm

PR #3828, opened 2021-09-06
Init using `main` as the default git branch

sethbattin:
# Why

I tried expo this weekend, and it made a repo for me with the default branch as `master`.  This just meant I had never updated my local git settings to do otherwise (as the manual explains https://git-scm.com/docs/git-init#Documentation/git-init.txt--bltbranch-namegt).  Even though it was easy to fix, I thought we might as well make the default avoid the non-inclusive language compiled into git.

---
This patch is more of a proposal than a change request.  If the spirit seems good to the maintainers, I can proceed with the remainder of the contributor guildelines, add tests, yada yada.

# How

I modeled my change on GitHub's new repo instructions.  It assumes that if someone had never updated their global settings, then they probably wouldn't mind renaming their base branch:
Screenshot from 2021-09-06 15-50-09

I went a step further and checked if the system does have a setting, and in that case skip the rename.  That covers two scenarios:
1. They already picked a default name they like better, so Expo doesn't need to change anything.
2. They explicitly set the old default to future-proof their....worldview? against future changes in git.  Expo doesn't need to fight them about it.  Comments contradicting this assumption are welcome.  :)

# Test Plan

I didn't exactly do this formally; again I would be willing to add tests and whatnot if folks require it.

I tested manually by monkeypatching my installed copy of expo like so:
```
// /usr/local/lib/node_modules/expo-cli/build/commands/initAsync.js ~line 553
// ...
      await (0, _spawnAsync().default)('git', ['commit', '-m', 'Created a new Expo app'], {
        cwd: root,
        stdio: 'ignore'
      });
      try {
        await (0, _spawnAsync().default)('git', ['config', '--get', 'init.defaultBranch'], {cwd: root } );
      } catch (e) {
        console.log({e})
        if (!e.stderr && !e.stdout) {
          await (0, _spawnAsync().default)('git', ['branch', '-M', 'main'], { cwd: root, stdio: 'ignore' });
        }
      }
// ...
```

Sadly, this doesn't work on git with versions before 2.30.  A git user with an older version who _does_ set the `init.defaultBranch` setting will still get a `master` branch.  (The setting will be ignored, and expo won't attempt to fix it.)  But if said user blindly sets the config value without testing that it works, pity for them.  At least they tried, and GitHub will hopefully encourage them to fix it anyway.

### DIFF
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -369,8 +369,8 @@ export async function initGitRepoAsync(
       return false;
     }
   }
+
   // not in git tree, so let's init
-    
   try {
     await spawnAsync('git', ['init', '--initial-branch', branch], { cwd: root });
     !flags.silent && Log.log('Initialized a git repository.');
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -375,19 +375,15 @@ export async function initGitRepoAsync(
     await spawnAsync('git', ['init'], { cwd: root });
     !flags.silent && Log.log('Initialized a git repository.');
     
-    // check for a default branch setting
-    const { stdout: initialBranch } = await spawnAsync('git', 
-      ['config', '--get', 'init.defaultBranch'], 
-      {
-        cwd: root,
-        stdio: ['ignore', 'pipe', 'ignore'],
-      }
-    );
-    
-    // if there is no setting, override the git default
-    // https://git-scm.com/docs/git-init#Documentation/git-init.txt--bltbranch-namegt
-    if (!initialBranch) {
-      await spawnAsync('git', ['branch', '-m', 'main'], { cwd: root, stdio: 'ignore' });
+    try {
+      // check for a default branch setting
+      await spawnAsync('git', ['config', '--get', 'init.defaultBranch'], { cwd: root });
+    } catch (e) {
+      // if there is no setting, override the git default
+      // https://git-scm.com/docs/git-init#Documentation/git-init.txt--bltbranch-namegt
+      if (!e?.stderr && !e?.stdout) {  // no error message and no setting value
+        await spawnAsync('git', ['branch', '-m', 'main'], { cwd: root, stdio: 'ignore' });
+      } 
     }
 
     if (flags.commit) {
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -374,6 +374,13 @@ export async function initGitRepoAsync(
   try {
     await spawnAsync('git', ['init'], { cwd: root });
     !flags.silent && Log.log('Initialized a git repository.');
+
+    if (flags.commit) {
+      await spawnAsync('git', ['add', '--all'], { cwd: root, stdio: 'ignore' });
+      await spawnAsync('git', ['commit', '-m', 'Created a new Expo app'], {
+        cwd: root,
+        stdio: 'ignore',
+      });
     
     try {
       // check for a default branch setting
@@ -383,15 +390,7 @@ export async function initGitRepoAsync(
       // https://git-scm.com/docs/git-init#Documentation/git-init.txt--bltbranch-namegt
       if (!e?.stderr && !e?.stdout) {  // no error message and no setting value
         await spawnAsync('git', ['branch', '-m', 'main'], { cwd: root, stdio: 'ignore' });
-      } 
-    }
-
-    if (flags.commit) {
-      await spawnAsync('git', ['add', '--all'], { cwd: root, stdio: 'ignore' });
-      await spawnAsync('git', ['commit', '-m', 'Created a new Expo app'], {
-        cwd: root,
-        stdio: 'ignore',
-      });
+      }
     }
     return true;
   } catch (e) {
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -372,7 +372,7 @@ export async function initGitRepoAsync(
 
   // not in git tree, so let's init
   try {
-    await spawnAsync('git', ['init', '--initial-branch', branch], { cwd: root });
+    await spawnAsync('git', ['init'], { cwd: root });
     !flags.silent && Log.log('Initialized a git repository.');
     
     // check for a default branch setting
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -369,11 +369,26 @@ export async function initGitRepoAsync(
       return false;
     }
   }
-
   // not in git tree, so let's init
+    
   try {
-    await spawnAsync('git', ['init'], { cwd: root });
+    await spawnAsync('git', ['init', '--initial-branch', branch], { cwd: root });
     !flags.silent && Log.log('Initialized a git repository.');
+    
+    // check for a default branch setting
+    const { stdout: initialBranch } = await spawnAsync('git', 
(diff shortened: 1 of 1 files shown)


EvanBacon:
We currently run [`git init`](https://github.com/expo/expo-cli/blob/c9db58f0f06f9ceb4d77c33788eb4cf3801d69a0/packages/expo-cli/src/commands/initAsync.ts#L375) in the project root, which effectively enables the user to make their default branch name anything.

This was changed recently because the original version forced the branch name to always be the word "main", which caused an [issue to be raised](https://github.com/expo/expo-cli/issues/3663).

The system currently in place, enables users to customize their default branch name using global git settings (as you pointed out), but it also makes `expo init` more _true_ to the `git init` command installed on the computer. It could be _unexpected behavior_ if you init a project with `git init` and get a repo with the root branch named "X", then `expo init` and get a repo with the root branch named "Y".

If the version of git that's installed on the user's machine changes the default behavior of `git init`, then `expo init` would also automatically respect that change, causing the behavior to always be consistent. CC @brentvatne

sethbattin:
@EvanBacon So i think you've pointed out a few items with the proposal, let me try to break down:
1. The prior issue of not following the setting
2. current behavior follows git system settings
3. It might be unexpected to create a different branch name

For #.1 - After my patch, we still follow the system settings.  It doesn't override `init.defaultBranch`
For #2 - same answer, unless that branch name is `master` due to lack of the setting (double unless they've explicitly set that branch name)

So I don't think either of those is a reason to choose status quo over the proposal, right?  One is a non-problem, and the other is the whole point of the patch.

---

For #.3, I would recommend weighting that particular hypothetical user confusion very weakly.  

First, `expo init` creates a git repo silently without any prompt, and as part of a very opinionated and large install process.  So if a command-runner is particular about their repo settings, they're probably not upset about _only_ the branch name (out of so many other things to be bothered by).  Anecdotally, I was certainly bothered by it for the reverse reason :)

Second, the more likely scenario is that they didn't even know what they were getting when they ran the command, or they are planning to configure git a bunch afterward anyway, e.g. set an origin and push.  And if that origin is github, then they'll get a suggestion away from `master` there too.  In my case ("hi i'm an actual, real-life, first time expo user") I was surprised by the repo even existing, and then I changed the branch name.  (my git version is apparently old.  Even though I definitely *am* very particular about my git settings, I guess I haven't updated the binary lately.)  I don't see any reason why those who prefer `master` should receive more convenience than those of us who normally click a github button first, clone second.

Third, I know it's tempting to care very little about inclusive language.  But steering the community in a positive direction should have value that outweighs a hypothetical, momentary, tiny, _unmeasurable_ piece of confusion in a giant ocean of opinionated install process.  Please don't fall back to passivity when you could be making things a little better.  Maybe ask the Black members of your team what they think?

ide:
It sounds like part of the issue is that GitHub's default suggestion is misaligned with Git's default configuration. We've thought this through in a different context because this misalignment fragments our docs: do we write "master" or "main" or is there a generalizable way to specify the developer's local branch?

Expo CLI currently uses Git to initialize new projects and allows developers to use any Git remote they want, whether it be GitHub, Bitbucket, or another server like one inside their company. When Expo CLI tells a developer it's making a new Git repository, the intent of that command is to leave the developer's file system in a state as if they had run `git init` and is as thin of an abstraction as we can make it. For this reason, Expo CLI honors the developer's Git settings - or lack of settings - over GitHub's if we need to make a choice. This might change over time as the Git ecosystem evolves.

I don't disagree with the spirit of the change but think it belongs upstream as a PR to Git itself and the behavior will propagate down to Expo CLI. Most of our active, major repos including our largest one internally use "main" as the default branch. The Expo team uses mostly "main" for Expo's own repos and uses developers' settings for  developers' repos.

sethbattin:
@ide I'm not sure I'd characterize GitHub's UX as a misalignment.  That defacto leader organization made a choice to contribute in the smallest way to improving git.  Expo could make a similar choice, and it would be much simpler because expo isn't a git service.  Github didn't wait for Linus to change his library, and I would argue they have more duty to system git than Expo does.

Docs should be a simple change as well; just write `main` in them, and accept this patch to the init script to initialize that way.  Problem solved!  I don't think it is feasible for your docs to somehow be "generalizable" to match any developer's preferred branch name; you'll always have to pick one and go with it.  So pick `main`.  If you have specific places where you refer to `master` and those would need updates, I would be happy to run that find/replace in my branch here.

As for changing upstream git instead of your own codebase, I am skeptical both of the plan and its objectivity.  If GitHub's choice doesn't count as change over time in the ecosystem, I think the standards might be impossibly high.  Mr Torvalds might accept your PR, but he probably won't break backwards compat either; he's something of a stickler about that.  So unless you can say what would count as enough of an ecosystem change, I would encourage you to not wait for that unknown event to happen elsewhere.  Make the choice yourself, in your own project.  That is the core of my proposal; that Expo contribute to improvement, rather than wait for others to do so.

As I said in the original, I was surprised that expo created a repo at all.  I did check the docs before I posted this PR.  ([It's not mentioned here, at least](https://docs.expo.dev/get-started/create-a-new-app/)).  As a user, I would not prefer a more complicated init process that isn't even referenced in the quickstart.  Rather, expo already makes opinionated choices and deliberately hides the complications, and this change wouldn't alter that at all.  It would just be a different branch name at the end.  By all means if you have a verbose installation process, add a branch name option there.  I didn't see those code paths in `initAsync` yet.

And I repeat: if it's fair to ask folks to change their branch name after running `init` in order to _not_ use `master` (because you didn't warn them ahead of time), then it's equally fair to expect folks who desperately want to preserve old language to change their branch names to reverse it.  No matter how you decide to proceed, the impact of your change is _at most_ one git command in the worst case.  Dedicated git users will have several other commands to run.  It is equivocation to cite it as some kind of disruption of git as a concept; that's simply not the case.  Also keep in mind that the core team runs this command far more often than the average user does; I am probably giving it more attention than your average library user.

Same suggestion as last post, please don't take my opinions on their own.  I am not personally affected by the usage of outdated tech terms, because I'm white.  Instead, you should ask the Black members of your team if they think the thinness of your system git wrapper is more important than avoiding the use of hostile language.

ide:
The CLI's intent right now is to honor each developer's configuration including their choice to use Git's default. I think it is important to acknowledge that Expo CLI is looking to use the developer's chosen language in this context.

We don't object to telling developers they can change their default branch with `git config --global init.defaultBranch <name>` and suggesting "main" as an option in the docs. We are supportive of suggesting inclusive language here. I think it would be a good idea also to communicate that `expo init` creates a new Git repository in the docs.

If there's a simple way to do so, we would like to have branch references in our docs be generalizable to accommodate both "main" and "master" as well as "trunk" or other default branches developers choose for their own reasons. There are some ways we could do this by just writing `<branch-name>` as a placeholder in the docs or writing scripts that look up the initial branch name.

I want to be extremely clear the people working on Expo support a positive and inclusive environment in the community, in our software, and on our team. Our [guidelines](https://expo.dev/guidelines) explain the standards we expect for community conduct. On our team, one of our most important internal values we talk about is psychological safety ([the #1 factor in Google's study on their most effective teams](https://rework.withgoogle.com/print/guides/5721312655835136/)). For the topic at hand, we support explaining to developers how to choose inclusive language like "main" for their default Git branches if they haven't done so already.

To that point, would you be interested in revising this PR to make a change to the docs instead that adds a short note under the terminal commands in "Initializing the project" that explains that `expo init` will create a local Git repository inside the project, and to change the name of the initial Git branch (e.g., to "main"), you can run `git config --global init.defaultBranch <name>`?

sethbattin:
Sure, that seems feasible.  It's certainly more proactive than I was suggesting.  I'll poke around the repos and make a draft.

Thanks for discussing with me, folks.

EvanBacon:
Stale, feel free to reopen after more changes have been made

PR #3831, opened 2021-09-08
[xdl][config] Implement naive manifest signing for EAS update manifests

wschurman:
# Why

This is the simplest solution towards getting EAS update manifests working in development in Expo Go and the dev client. This doesn't require client changes as the client currently assumes that EAS update manifests are signed using the same strategy (which they are on the server as well).

Closes ENG-1756.

# How

The way this works is by looking for the EAS project ID in the config, and then manufacturing a manifest at the time it is served with the values from the EAS server for `manifest.extra.eas.projectId` and `manifest.extra.scopeKey` if an EAS project ID was set in the config.

If the EAS project ID is not set in the config, the anonymous setting is used which allows access to most scoped modules but not EAS services like push notifications or auth session.

# Test Plan

This is tested in tandem with the necessary EAS server implementation: https://github.com/expo/universe/pull/8119 (landed) and https://github.com/expo/universe/pull/8150

1. Terminal window 1: Start a www dev server with dependent PR patched in.
2. Terminal window 2: `yarn start` this repo
3. Terminal window 3: 
    a. `export EXPO_LOCAL=1`
    b. Create a new project: `expo init blah`, 
    c. `EXPO_DEBUG=true ~/expo-cli/node_modules/.bin/expo start`
4. In insomnia/curl, load `http://localhost:19000/update-manifest-experimental?platform=android` with and without the `expo-accept-signature` header. Ensure that when `expo-accept-signature` header is present it creates an app on the server idempotently.
2. Load `exp://localhost:19000/update-manifest-experimental?platform=android` in locally-running Expo Go android client, ensure it loads. Breakpoint on the signature verification code and ensure the signature verification succeeds.

### DIFF
--- a/packages/config/src/Config.types.ts
+++ b/packages/config/src/Config.types.ts
@@ -78,21 +78,33 @@ export type ExpoGoConfig = {
   };
 };
 
-export type ExpoAppManifest = ExpoConfig &
+export type EASConfig = {
+  projectId?: string;
+};
+
+export type ClientScopingConfig = {
+  scopeKey?: string;
+};
+
+export type ExpoClientConfig = ExpoConfig & {
+  id?: string;
+  releaseId?: string;
+  revisionId?: string;
+  bundleUrl?: string;
+  hostUri?: string;
+  publishedTime?: string;
+};
+
+export type ExpoAppManifest = ExpoClientConfig &
+  EASConfig &
   Partial<ExpoGoConfig> & {
     sdkVersion: string;
     bundledAssets?: string[];
     isKernel?: boolean;
     kernel?: { androidManifestPath?: string; iosManifestPath?: string };
     assetUrlOverride?: string;
-    publishedTime?: string;
     commitTime?: string;
-    releaseId?: string;
-    revisionId?: string;
     env?: Record<string, any>;
-    bundleUrl?: string;
-    hostUri?: string;
-    id?: string;
   };
 
 export interface ExpoUpdatesManifestAsset {
@@ -109,7 +121,11 @@ export interface ExpoUpdatesManifest {
   launchAsset: ExpoUpdatesManifestAsset;
   assets: ExpoUpdatesManifestAsset[];
   metadata: { [key: string]: string };
-  extra: { [key: string]: any };
+  extra: ClientScopingConfig & {
+    expoClient?: ExpoClientConfig;
+    expoGo?: ExpoGoConfig;
+    eas?: EASConfig;
+  };
 }
 
 export type Hook = {
--- a/packages/xdl/src/ProjectAssets.ts
+++ b/packages/xdl/src/ProjectAssets.ts
@@ -404,8 +404,16 @@ export async function resolveAndCollectExpoUpdatesManifestAssets(
   projectRoot: string,
   exp: ExpoConfig,
   urlResolver: (path: string) => string
-): Promise<{ url: string; hash: string; key: string; contentType: string }[]> {
-  const manifestAssets: { url: string; hash: string; key: string; contentType: string }[] = [];
+): Promise<
+  { url: string; hash: string; key: string; contentType: string; fileExtension: string }[]
+> {
+  const manifestAssets: {
+    url: string;
+    hash: string;
+    key: string;
+    contentType: string;
+    fileExtension: string;
+  }[] = [];
   await resolveExpoUpdatesManifestAssets({
     projectRoot,
     manifest: exp,
@@ -414,11 +422,13 @@ export async function resolveAndCollectExpoUpdatesManifestAssets(
       const contents = await fs.readFile(absolutePath);
       // Expo Updates spec dictates that this hash is sha256
       const hash = crypto.createHash('sha256').update(contents).digest('hex');
+      const contentType = mime.getType(absolutePath) ?? 'application/octet-stream';
       manifestAssets.push({
         url: urlResolver(assetPath),
         hash,
         key: assetPath,
-        contentType: mime.getType(absolutePath) ?? 'application/octet-stream',
+        contentType,
+        fileExtension: path.extname(absolutePath),
       });
       return assetPath;
     },
--- a/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
+++ b/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
@@ -64,7 +64,9 @@ async function shouldUseAnonymousManifestAsync(
 
 async function getScopeKeyForProjectIdAsync(projectId: string): Promise<string> {
   const user = await UserManager.ensureLoggedInAsync();
-  const project = await ApiV2.clientForUser(user).getAsync(`projects/${projectId}`);
+  const project = await ApiV2.clientForUser(user).getAsync(
+    `projects/${encodeURIComponent(projectId)}`
+  );
   return project.scopeKey;
 }
 
--- a/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
+++ b/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
@@ -1,17 +1,24 @@
 import { ExpoUpdatesManifest, getConfig } from '@expo/config';
 import { Updates } from '@expo/config-plugins';
+import { JSONObject } from '@expo/json-file';
 import express from 'express';
 import http from 'http';
+import nullthrows from 'nullthrows';
 import { parse } from 'url';
 import { v4 as uuidv4 } from 'uuid';
 
 import {
   Analytics,
+  ANONYMOUS_USERNAME,
+  ApiV2,
   Config,
+  ConnectionStatus,
   ProjectAssets,
   ProjectUtils,
   resolveEntryPoint,
   UrlUtils,
+  UserManager,
+  UserSettings,
 } from '../internal';
 import {
   getBundleUrlAsync,
@@ -34,14 +41,51 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): 'a
   return stringifiedPlatform as 'android' | 'ios';
 }
 
+/**
+ * Whether an anonymous scope key should be used. It should be used when:
+ * 1. Offline
+ * 2. Not logged-in
+ * 3. No EAS project ID in config
+ */
+async function shouldUseAnonymousManifestAsync(
+  easProjectId: string | undefined | null
+): Promise<boolean> {
+  if (!easProjectId || ConnectionStatus.isOffline()) {
+    return true;
+  }
+
+  const currentSession = await UserManager.getSessionAsync();
+  if (!currentSession) {
+    return true;
+  }
+
+  return false;
+}
+
+async function getScopeKeyForProjectIdAsync(projectId: string): Promise<string> {
+  const user = await UserManager.ensureLoggedInAsync();
+  const project = await ApiV2.clientForUser(user).getAsync(`projects/${projectId}`);
+  return project.scopeKey;
+}
+
+async function signManifestAsync(manifest: ExpoUpdatesManifest): Promise<string> {
+  const user = await UserManager.ensureLoggedInAsync();
+  const { signature } = await ApiV2.clientForUser(user).postAsync('manifest/eas/sign', {
+    manifest: (manifest as any) as JSONObject,
+  });
+  return signature;
+}
+
 export async function getManifestResponseAsync({
   projectRoot,
   platform,
   host,
+  acceptSignature,
 }: {
   projectRoot: string;
   platform: 'android' | 'ios';
   host?: string;
+  acceptSignature: boolean;
 }): Promise<{
   body: ExpoUpdatesManifest;
   headers: Map<string, number | string | readonly string[]>;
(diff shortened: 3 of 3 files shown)


packages/config/src/Config.types.ts
```diff
@@ -78,21 +78,33 @@ export type ExpoGoConfig = {
...
+  id?: string;
+  releaseId?: string;
+  revisionId?: string;
+  bundleUrl?: string;
+  hostUri?: string;
+  publishedTime?: string;
+};
+
+export type ExpoAppManifest = ExpoClientConfig &
+  EASConfig &
```
ide, line 99: I could be mistaken but do we ever want to include the EAS config in classic manifests? I suppose it doesn't hurt but wonder if it's needed.

packages/xdl/src/ProjectAssets.ts
```diff
@@ -414,11 +422,13 @@ export async function resolveAndCollectExpoUpdatesManifestAssets(
...
       // Expo Updates spec dictates that this hash is sha256
       const hash = crypto.createHash('sha256').update(contents).digest('hex');
+      const contentType = mime.getType(absolutePath) ?? 'application/octet-stream';
       manifestAssets.push({
         url: urlResolver(assetPath),
         hash,
         key: assetPath,
-        contentType: mime.getType(absolutePath) ?? 'application/octet-stream',
+        contentType,
+        fileExtension: `.${mime.getExtension(contentType) ?? ''}`,
```
ide, line 431: Use the file extension from the filename in `absolutePath` to mimic the server's logic. This way we honor the extension used on disk.

If there is no file extension, omit the property (just use undefined here I think). If the file ends in ".", the file extension should be just ".". A string that does not begin with "." is invalid here and the client will prefix it with ".", so an empty string will be treated as if "." were specified. https://docs.expo.dev/technical-specs/expo-updates-0/#manifest-response-headers

yarn.lock
```diff
@@ -9777,10 +9777,10 @@ he@1.2.0, he@^1.2.0:
...
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
-hermes-engine@~0.5.0:
-  version "0.5.1"
-  resolved "https://registry.yarnpkg.com/hermes-engine/-/hermes-engine-0.5.1.tgz#601115e4b1e0a17d9aa91243b96277de4e926e09"
-  integrity sha512-hLwqh8dejHayjlpvZY40e1aDCDvyP98cWx/L5DhAjSJLH8g4z9Tp08D7y4+3vErDsncPOdf1bxm+zUWpx0/Fxg==
+hermes-engine@0.0.0, hermes-engine@~0.5.0:
+  version "0.0.0"
+  resolved "https://registry.yarnpkg.com/hermes-engine/-/hermes-engine-0.0.0.tgz#6a65954646b5e32c87aa998dee16152c0c904cd6"
+  integrity sha512-q5DP4aUe6LnfMaLsxFP1cCY5qA0Ca5Qm2JQ/OgKi3sTfPpXth79AQ7vViXh/RRML53EpokDewMLJmI31RioBAA==
```
ide, line 9783: This doesn't look right / relevant to this PR. If in doubt, revert the yarn.lock changes.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
   return String(platform);
 }
 
+async function shouldUseAnonymousManifestAsync(): Promise<boolean> {
+  const currentSession = await UserManager.getSessionAsync();
+  return !currentSession || ConnectionStatus.isOffline();
```
ide, line 41: nit: might be faster to immediately return false when offline

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
...
+  expoConfig: ExpoConfig
+): Promise<{
+  scopeKey: string;
+  projectId: string | null;
+}> {
+  const shouldUseAnonymousManifest = await shouldUseAnonymousManifestAsync();
+  if (shouldUseAnonymousManifest) {
+    const userAnonymousIdentifier = await UserSettings.getAnonymousIdentifierAsync();
+    return {
+      scopeKey: `@${ANONYMOUS_USERNAME}/${expoConfig.slug}-${userAnonymousIdentifier}`,
```
ide, line 54: I suspect we'll want to change this in the future not to follow the `@account/project` format and possibly to move away from using the slug (if we drop that field). But I don't think that should stop this PR and there shouldn't be high expectations for scope keys to be to be consistent for anonymous projects, especially if the anonymous identifier can change from computer to computer.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
...
+      projectId: null,
+    };
+  }
+
+  const user = await UserManager.ensureLoggedInAsync();
+  const username = await UserManager.getCurrentUsernameAsync();
```
ide, line 60: Is there no user.username field?

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
...
+  const accountName = expoConfig.owner ?? username;
+  const projectName = expoConfig.slug;
+
+  try {
+    const projects = await ApiV2.clientForUser(user).getAsync('projects', {
+      experienceName: `@${accountName}/${projectName}`,
+    });
+    const project = projects[0];
```
ide, line 68: Unrelated to this PR but slightly unexpected that a lookup by experience name returns an array

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
...
+    return {
+      scopeKey: project.scopeKey,
+      projectId: project.id,
+    };
```
ide, line 72: I expect in the future we'll change this to read the EAS project ID from a local config file, whether that's expoConfig (generated from app.config.js/app.json) or eas.json.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -29,14 +36,80 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): st
...
+    if (!(e instanceof ApiV2Error)) {
+      throw e;
+    }
+
+    if (e.code !== 'EXPERIENCE_NOT_FOUND') {
+      throw e;
+    }
+  }
+
+  const { id, scopeKey } = await ApiV2.clientForUser(user).postAsync('projects', {
```
ide, line 83: Should this be GET instead of POST? Also surprised this doesn't return an array like getAsync('projects') does.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -96,19 +171,31 @@ export async function getManifestResponseAsync({
       key: mainModuleName,
       contentType: 'application/javascript',
       url: bundleUrl,
+      fileExtension: '.fake', // shouldn't need this but android crashes without it
```
ide, line 174: We intentionally omit fileExtension from the launchAsset. We should look into this because I'd expect it to work and further more, how were updates working at all before we introduced fileExtension? Maybe it's a difference between development and production updates?

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -96,19 +171,31 @@ export async function getManifestResponseAsync({
...
       },
       expoGo: expoGoConfig,
+      scopeKey,
     },
   };
 
+  if (acceptSignature) {
+    const shouldUseAnonymousManifest = await shouldUseAnonymousManifestAsync();
+    const manifestSignature = shouldUseAnonymousManifest
+      ? 'UNSIGNED'
```
ide, line 194: Do we need this (as opposed to just omitting the header)?

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -131,6 +218,7 @@ export function getManifestHandler(projectRoot: string) {
         projectRoot,
         host: req.headers.host,
         platform: getPlatformFromRequest(req),
+        acceptSignature: req.headers['expo-accept-signature'] !== undefined,
```
ide, line 221: I'm not sure `expo-accept-signature:` (empty string) should count as enabling signatures. This looks OK for now but we probably will want to revisit it.

We didn't add much about signing to the Expo Updates spec and so I don't think the definition of a valid boolean value was ever written down in our spec. Interestingly, I can't seem to find common HTTP headers that have boolean values. Perhaps this should be a version string...

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
jkhales: Looking into this

jkhales (approved):
There's a bit of code split here in that we publish/build from `eas-cli` but develop on `expo-cli`.

Is it worth making this callable via `eas start`?

Or is development always going to be the domain of expo-cli?

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: Yeah, that's definitely one option. How we'll support this bridge between eas-cli and expo-cli is still fairly murky.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: The reason it is this way is that we may support filtering by other (non-unique) properties than `experienceName` in the future. This endpoint is designed to be a general REST search method, so returning an array allows us to be fairly future-proof.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: There is but `getCurrentUsernameAsync` gets the username from a different place, so it's very confusing.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: yep, agreed. The key here though is that i _think_ there's code in the client that looks explicitly for this: https://github.com/expo/expo/blob/master/android/expoview/src/main/java/host/exp/exponent/ExponentManifest.kt#L144

It actually seems like any `id` that starts with `@anonymous/` is considered verified, so that is independent of scopeKey. Hmm.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: Probably not. The client currently looks for it but treats it the same as omitting it it seems: https://github.com/expo/expo/blob/master/packages/expo-updates/android/src/main/java/expo/modules/updates/loader/FileDownloader.java#L130

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
wschurman: For classic update served in expo-cli, it just does a simple boolean check against the value of the header: https://github.com/expo/expo-cli/blob/master/packages/xdl/src/start/ManifestHandler.ts#L364

The type says it can be `string | string[] | undefined`. I can change this to a simple boolean check.

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
jkhales:
```suggestion
  const [user, username] = await Promise.all([UserManager.ensureLoggedInAsync(), UserManager.getCurrentUsernameAsync()]);
```

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
ide: In that case, I think we should omit it so that we use fewer ways to accomplish the same task (of telling the server not to sign the manifest)

ide:
To respond to JJ's comment about code organization, these are a few guidelines:

- There are classic manifests and Expo Updates manifests (the ones documented in the spec, and we sometimes casually call them modern manifests). "EAS manifest" is not a concept, though EAS can serve Expo Updates manifests.
- Over time, Expo CLI will server only Expo Updates manifests to newer SDK versions. I imagine code like `if (sdkVersion < X) { serveClassicManifest } else { serveManifest }`.
- Getting EAS to sign a manifest with keys accepted by Expo Go is an EAS task. However, I think the pragmatic thing to do is not to require EAS CLI to use Expo Go, and for Expo CLI to contact EAS for signing as this PR does.
- All other EAS-related tasks like publishing to EAS Update probably belong in EAS CLI, which is how things currently work.
- EAS Update needs to sign manifests on the fly, so when publishing to EAS Update I don't think EAS needs to sign anything.
- Using developers' own keys for signing should happen in Expo CLI as that is independent from EAS.

wschurman:
Updated this with a new approach: Get EAS project ID from app config. If it's there, fetch the scope key and use that in the generated and signed manifest. If it's not there or the cli is offline or the user isn't logged-in, use the anonymous identifier (will require a client change to check the scopeKey for this value instead of `id`).

packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -34,14 +41,51 @@ function getPlatformFromRequest(req: express.Request | http.IncomingMessage): 'a
...
+  if (!currentSession) {
+    return true;
+  }
+
+  return false;
+}
+
+async function getScopeKeyForProjectIdAsync(projectId: string): Promise<string> {
+  const user = await UserManager.ensureLoggedInAsync();
+  const project = await ApiV2.clientForUser(user).getAsync(`projects/${projectId}`);
```
ide, line 67:
```suggestion
  const project = await ApiV2.clientForUser(user).getAsync(`projects/${encodeURIComponent(projectId)}`);
```

ide (approved):

PR #3834, opened 2021-09-08
[next-adapter] update Babel config

axeldelafosse:
# Why

The current Babel config isn't working correctly with latest Next.js version

# How

- Check if loader name is `next-babel-turbo-loader` to detect web
- Add `next/babel` preset before `babel-preset-expo`

# Test Plan

The adapter should work correctly when using the latest Next.js version


EvanBacon (approved):
looks fine

PR #3839, opened 2021-09-09
[config-plugins] new appDelegateHeader mod/plugin

kbrandwijk:
# Why

Config plugins offer functionality out of the box to modify the appDelegate file, but not the appDelegate.h header file.

# How

The implementation is 100% identical to the one for appDelegate.

# Test Plan

The PR includes automated tests that mimic the existing appDelegate tests to provide the same test focus and coverage. I was slightly surprised that `withIosBaseMods-test.ts` didn't have any coverage for the existing appDelegate implementation, so I didn't extend that one to include tests for appDelegateHeader.

### DIFF
--- a/packages/config-plugins/src/Plugin.types.ts
+++ b/packages/config-plugins/src/Plugin.types.ts
@@ -160,6 +160,10 @@ export interface ModConfig {
      */
     appDelegate?: Mod<AppDelegateProjectFile>;
     /**
+     * Modify the `ios/<name>/AppDelegate.h` as a string (dangerous)
+     */
+    appDelegateHeader?: Mod<AppDelegateProjectFile>;
+    /**
      * Modify the `ios/Podfile.properties.json` as key-value pairs
      */
     podfileProperties?: Mod<Record<string, string>>;
--- a/packages/config-plugins/src/ios/Paths.ts
+++ b/packages/config-plugins/src/ios/Paths.ts
@@ -16,20 +16,28 @@ interface ProjectFile<L extends string = string> {
 export type AppDelegateProjectFile = ProjectFile<'objc' | 'swift'>;
 
 export function getAppDelegateFilePath(projectRoot: string): string {
-  const [using, ...extra] = globSync('ios/*/AppDelegate.@(m|swift)', {
+  return getFilePath(projectRoot, 'ios/*/AppDelegate.@(m|swift)', 'AppDelegate', 'app-delegate')
+}
+
+export function getAppDelegateHeaderFilePath(projectRoot: string): string {
+  return getFilePath(projectRoot, 'ios/*/AppDelegate.h', 'AppDelegate.h', 'app-delegate-header')
+}
+
+function getFilePath(projectRoot: string, globPattern: string, fileName: string, warningTag: string): string {
+  const [using, ...extra] = globSync(globPattern, {
     absolute: true,
     cwd: projectRoot,
     ignore: ignoredPaths,
   });
 
   if (!using) {
-    throw new UnexpectedError(`Could not locate a valid AppDelegate at root: "${projectRoot}"`);
+    throw new UnexpectedError(`Could not locate a valid ${fileName} at root: "${projectRoot}"`);
   }
 
   if (extra.length) {
     warnMultipleFiles({
-      tag: 'app-delegate',
-      fileName: 'AppDelegate',
+      tag: warningTag,
+      fileName,
       projectRoot,
       using,
       extra,
@@ -43,6 +51,7 @@ function getLanguage(filePath: string): 'objc' | 'swift' {
   const extension = path.extname(filePath);
   switch (extension) {
     case '.m':
+    case '.h':
       return 'objc';
     case '.swift':
       return 'swift';
@@ -61,6 +70,11 @@ export function getFileInfo(filePath: string) {
 
 export function getAppDelegate(projectRoot: string): AppDelegateProjectFile {
   const filePath = getAppDelegateFilePath(projectRoot);
+  return getFileInfo(filePath);
+}
+
+export function getAppDelegateHeader(projectRoot: string): AppDelegateProjectFile {
+  const filePath = getAppDelegateHeaderFilePath(projectRoot);
   return getFileInfo(filePath);
 }
 
--- a/packages/config-plugins/src/ios/Paths.ts
+++ b/packages/config-plugins/src/ios/Paths.ts
@@ -24,9 +24,9 @@ export function getAppDelegateHeaderFilePath(projectRoot: string): string {
 }
 
 function getFilePath(
-  projectRoot: string, 
-  globPattern: string, 
-  fileName: string, 
+  projectRoot: string,
+  globPattern: string,
+  fileName: string,
   warningTag: string
 ): string {
   const [using, ...extra] = globSync(globPattern, {
--- a/packages/config-plugins/src/ios/Paths.ts
+++ b/packages/config-plugins/src/ios/Paths.ts
@@ -16,14 +16,19 @@ interface ProjectFile<L extends string = string> {
 export type AppDelegateProjectFile = ProjectFile<'objc' | 'swift'>;
 
 export function getAppDelegateFilePath(projectRoot: string): string {
-  return getFilePath(projectRoot, 'ios/*/AppDelegate.@(m|swift)', 'AppDelegate', 'app-delegate')
+  return getFilePath(projectRoot, 'ios/*/AppDelegate.@(m|swift)', 'AppDelegate', 'app-delegate');
 }
 
 export function getAppDelegateHeaderFilePath(projectRoot: string): string {
-  return getFilePath(projectRoot, 'ios/*/AppDelegate.h', 'AppDelegate.h', 'app-delegate-header')
+  return getFilePath(projectRoot, 'ios/*/AppDelegate.h', 'AppDelegate.h', 'app-delegate-header');
 }
 
-function getFilePath(projectRoot: string, globPattern: string, fileName: string, warningTag: string): string {
+function getFilePath(
+  projectRoot: string, 
+  globPattern: string, 
+  fileName: string, 
+  warningTag: string
+): string {
   const [using, ...extra] = globSync(globPattern, {
     absolute: true,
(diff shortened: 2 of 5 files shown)


kbrandwijk:
The expo-cli test that is failing is, as far as I know, something that is totally untouched by my changes.

EvanBacon (changes requested):
I don't really like the idea of having a mod that could potentially not have an input file (Swift project). The `appDelegate` mod is dangerous and only here as a temporary workaround until `expo-modules-core` is released in SDK 43, it also supports the case where Swift is being used as the input file. I'd recommend just using the `withDangerousMod` instead. 

Config plugins should basically only be used for static modifications: Info.plist, gradle.properties, etc. Dynamic modifications are problematic because they won't age well when the input template file changes.

What exactly are you trying to change in the `AppDelegate.h` that cannot be changed in the `AppDelegate.m`?

kbrandwijk:
Our usecase was adding a protocol to the interface definition for AppDelegate that lives in `AppDelegate.h`. `withDangerousMod` doesn't offer you the functionality to 'find' the file for you, and allow for a similar approach to how AppDelegate.m is being edited right now. I don't think it should be the hammer that makes everything look like a nail.

I would be more than surprised if `expo-modules-core` would completely deprecate all manual editing of AppDelegate.m, but I can't be sure of that of course. 

Regardless of our specific usecase, I don't think there's anything _against_ exposing as many files as possible that are part of the native code generation for easy access in config plugins, because the current workaround doesn't exactly win a beauty contest either:

```js
const test = configPlugins.IOSConfig.Paths.getAppDelegate(config.modRequest.projectRoot)
let testPath = test.path.toString().replace(`AppDelegate.m`, `AppDelegate.h`)
```

I agree with you that config plugins should mostly be about static modifications, but I don't think that idea will prove sustainable across a wide range of integrations.

EvanBacon:
@kbrandwijk could you please provide a before/after for the transform you want to achieve, I'd like to see if it falls under the expo-modules-core umbrella, otherwise we may want to revisit some part of autolinking/expo-modules-core system to ensure it covers more use cases.

# Template

## Before

```h
#import <Foundation/Foundation.h>
#import <EXUpdates/EXUpdatesAppController.h>
#import <React/RCTBridgeDelegate.h>
#import <UIKit/UIKit.h>

#import <UMCore/UMAppDelegateWrapper.h>

@interface AppDelegate : UMAppDelegateWrapper <RCTBridgeDelegate, EXUpdatesAppControllerDelegate>

@end
```

## After


```diff
#import <Foundation/Foundation.h>
#import <EXUpdates/EXUpdatesAppController.h>
#import <React/RCTBridgeDelegate.h>
#import <UIKit/UIKit.h>

#import <UMCore/UMAppDelegateWrapper.h>

@interface AppDelegate : UMAppDelegateWrapper <RCTBridgeDelegate, EXUpdatesAppControllerDelegate>

@end
```

kbrandwijk:
For example:

## After
```objc
#import <Foundation/Foundation.h>
#import <EXUpdates/EXUpdatesAppController.h>
#import <React/RCTBridgeDelegate.h>
#import <UIKit/UIKit.h>
#import <UserNotifications/UserNotifications.h>

#import <UMCore/UMAppDelegateWrapper.h>

@interface AppDelegate : UMAppDelegateWrapper <RCTBridgeDelegate, EXUpdatesAppControllerDelegate, UNUserNotificationCenterDelegate>

@end
```

And a few other variations like that for other protocols.

EvanBacon:
@kbrandwijk I see, thanks for sharing!

## Background

For the ultimate "happy path" projects in Expo, we implement a couple things to make complex native runtime generation work: Expo config plugins, prebuilding, autolinking (the original Expo version), and expo-modules-core (AppDelegate proxy, MainActivity proxy). Together, these can safely generate complex native runtimes with advanced best practices (like monorepos, xcframeworks, prebuilt binaries, etc). For this reason, I'm a bit defensive about adding more features to improve regexes and dynamic code modifications in config plugins, as this falls under the expo-modules-core umbrella. 

In Expo, we would implement the modification you're trying to accomplish (safely) using the AppDelegate proxy, [`expo-notifications` example](https://github.com/expo/expo/blob/e1f0111a47e1e22ce135751f07ba2b1c1ee365e9/packages/expo-notifications/ios/EXNotifications/Notifications/EXNotificationCenterDelegate.h#L18), but this isn't well documented right now, and monorepo support won't be very easily attainable until SDK 43 (i.e. wrapping the upstream module and adding the proxy via your project). So your best way forward right now would be config plugins.

## Ideas

Something that concerns me a bit is that base mods should always map to a physical file (except for dangerous mod), and have a file path. In the case of a Swift project, there would be no file path for AppDelegate header, resulting in a new case across the spec. It'd be much nicer if there were different types of mod-compilers that ran for different types of projects (i.e. why I made config plugins originally).

### Option 1

One approach would be to add base mods dynamically based on the template, this one has the potential downside of versioned mismatches. For instance, say v1 of config-plugins uses `ios.foobar` but v2 changes it to `ios.coolerFoobar`, we'd want to warn users that older config plugins are attempting to access the older namespace, which would be hard to do if some config plugins were accessing a namespace that is simply not there due to dynamic mods being added to support the template (however, this may be unavoidable in the future).

### Option 2

An option that seems like it would be quick, and prevent any future architecture issues, would be adding a system to extend the base mods in user space. You would effectively be able to add extra base mods, for any file you want to manage yourself.

We could do something like this:

`custom.ts`
```ts
import {
  ConfigPlugin,
  IOSConfig,
  Mod,
  withMod,
  BaseMods,
} from "@expo/config-plugins";
import fs from "fs";

export function withCustomBaseMods(config) {
  return BaseMods.withGeneratedBaseMods<"appDelegateHeader">(config, {
    platform: "ios",
    providers: {
      // Append a custom rule to supply AppDelegate header data to mods on `mods.ios.appDelegateHeader`
      appDelegateHeader:
        BaseMods.provider<IOSConfig.Paths.AppDelegateProjectFile>({
          getFilePath(props) {
            const {
              modRequest: { projectRoot },
            } = props;
            const filePath =
              IOSConfig.Paths.getAppDelegateFilePath(projectRoot);
            // Replace the .m with a .h
            if (filePath.endsWith(".m")) {
              return filePath.substr(0, filePath.lastIndexOf(".")) + ".h";
            }
            // Possibly a Swift project...
            throw new Error(
              `Could not locate a valid AppDelegate.h at root: "${projectRoot}"`
            );
          },
          async read(filePath) {
            return IOSConfig.Paths.getFileInfo(filePath);
          },
          async write(filePath: string, { modResults: { contents } }) {
            await fs.promises.writeFile(filePath, contents);
          },
        }),
    },
  });
}

/**
 * Provides the AppDelegate header file for modification.
 *
 * @param config
 * @param action
 */
export const withAppDelegateHeader: ConfigPlugin<
  Mod<IOSConfig.Paths.AppDelegateProjectFile>
> = (config, action) => {
  return withMod(config, {
    platform: "ios",
    mod: "appDelegateHeader",
    action,
  });
};

export const withCustomHeaderInterface = (config) => {
  return withAppDelegateHeader(config, (config) => {
    console.log("modify header:", config.modResults);
    return config;
  });
};
```

`app.config.ts`
```ts
// Required for external files using TS
require("ts-node/register");

import {
  withCustomBaseMods,
  withCustomHeaderInterface,
} from "./custom.ts";

export default ({ config }) => {
  if (!config.plugins) config.plugins = [];
  config.plugins.push(
    withCustomHeaderInterface,
    // Base mods MUST be last
    withCustomBaseMods
  );
  return config;
};
```

I'll open a PR for this and we can determine from there, if that seems like enough customizability for now, sorry if this isn't quite what you had in mind, I'm happy to discuss more potential options as well!

kbrandwijk:
@EvanBacon First of all, thank you for your in-depth look into this.

Extending base mods would be a great option, although I feel that would work best 'across projects' where we need something similar in multiple places. For a single project, extending base mods first, in order to get a slightly cleaner plugin, would probably not be worth it. For comparison, this is the current workaround, which basically just abuses the current AppDelegate mod, and has to mimic a few things manually that the base code already does:

```js
const withAppDelegateH = (config) => {
    return configPlugins.withDangerousMod(config, [
        "ios",
        async (config) => {            
            // all path options: https://github.com/expo/expo-cli/blob/master/packages/config-plugins/src/android/Paths.ts
            const test = configPlugins.IOSConfig.Paths.getAppDelegate(config.modRequest.projectRoot)

            let testPath = test.path.toString().replace(`AppDelegate.m`, `AppDelegate.h`)

            let contents = fs.readFileSync(testPath, 'utf8')

            if (!contents.includes("UNUserNotificationCenterDelegate")) {
                contents = contents.replace("@interface AppDelegate : UMAppDelegateWrapper <RCTBridgeDelegate, EXUpdatesAppControllerDelegate>",
                "@interface AppDelegate : UMAppDelegateWrapper <RCTBridgeDelegate, EXUpdatesAppControllerDelegate, UNUserNotificationCenterDelegate>"
                )
            }

            if (!contents.includes("#import <UserNotifications/UserNotifications.h>")) {
                contents = contents.replace("#import <UIKit/UIKit.h>", "#import <UIKit/UIKit.h>\n#import <UserNotifications/UserNotifications.h>")
            }

            await fs.writeFile(testPath, contents);
      
        return config;
        },
    ]);
};
``` 

That's why it initially seemed to be low hanging fruit to add this functionality in. At the same time, if there's a way to publish and share mods similar to how we could share entire plugins now, that would be helpful (although the moment that's possible, I'm pretty sure the appDelegateHeader mod will be the first community mod I'm putting out there (pending a better alternative of course), so effectively we would end up in the same situation, except that it requires a community package to use, instead of being part of the defaults... I like your solution though, because there's probably more usecases then just this one where allowing mod-level community effort in will prove useful.

In either scenario, I think the most important thing to keep in mind, and the most important thing I'm trying to keep an eye on, is not getting locked in. I know there's been major effort around that (especially with config plugins) and it's come to the point that most of my projects can stay in managed mode now, which is great.
At the same time, it widens the gap. The further you take things in managed mode, the bigger the problem if you ultimately hit the wall and need to 'eject'. So whatever solution you end up with, I urge you to keep in mind that you keep these 'escapes' available.

PR #3840, opened 2021-09-10
[xdl] Add support for TypeScript webpack config

DannyNemer:
Extends the webpack integration to load `webpack.config.ts` if `webpack.config.js` does not exist in the project root. Webpack has built-in TypeScript support for its configuration: https://webpack.js.org/configuration/configuration-languages/.

# Why

To take advantage of TypeScript's type checking in the webpack config file.

# How

I added a check for a file named `webpack.config.ts` in the project root, which falls back to `webpack.config.js` if the TypeScript version does not exist. If using the TypeScript version, we use `ts-node` to [patch future calls of `require()`](https://github.com/TypeStrong/ts-node#programmatic) to transpile the TypeScript file when loaded.

# Test Plan

I confirmed this PR works locally by using a modified installation of `expo-cli` with this change. The screenshot below shows that `expo build:web` correctly loaded `webpack.config.ts` in my project:

Screen Shot 2021-09-09 at 6 08 51 PM

### DIFF
--- a/packages/xdl/package.json
+++ b/packages/xdl/package.json
@@ -90,6 +90,7 @@
     "terminal-link": "^2.1.1",
     "text-table": "^0.2.0",
     "tree-kill": "1.2.2",
+    "ts-node": "^10.2.1",
     "url-join": "4.0.0",
     "uuid": "^8.0.0",
     "webpack": "4.43.0",
--- a/packages/xdl/src/Webpack.ts
+++ b/packages/xdl/src/Webpack.ts
@@ -14,6 +14,7 @@ import http from 'http';
 import * as path from 'path';
 import formatWebpackMessages from 'react-dev-utils/formatWebpackMessages';
 import openBrowser from 'react-dev-utils/openBrowser';
+import resolveFrom from 'resolve-from';
 import webpack from 'webpack';
 import WebpackDevServer from 'webpack-dev-server';
 
@@ -628,10 +629,18 @@ async function loadConfigAsync(
   argv?: string[]
 ): Promise<WebpackConfiguration> {
   setMode(env.mode);
-  // Check if the project has a webpack.config.js in the root.
-  const projectWebpackConfig = path.resolve(env.projectRoot, 'webpack.config.js');
+
+  // Check if the project has a webpack.config.{ts,js} in the root.
+  let projectWebpackConfig = resolveFrom.silent(env.projectRoot, './webpack.config.ts');
+  if (projectWebpackConfig) {
+    // Patch `require()` to transpile TypeScript when loading the file below.
+    require('ts-node/register');
+  } else {
+    projectWebpackConfig = resolveFrom.silent(env.projectRoot, './webpack.config.js');
+  }
+
   let config: WebpackConfiguration;
-  if (fs.existsSync(projectWebpackConfig)) {
+  if (projectWebpackConfig) {
     const webpackConfig = require(projectWebpackConfig);
     if (typeof webpackConfig === 'function') {
       config = await webpackConfig(env, argv);
--- a/yarn.lock
+++ b/yarn.lock
@@ -1372,6 +1372,18 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@cspotcode/source-map-consumer@0.8.0":
+  version "0.8.0"
+  resolved "https://registry.yarnpkg.com/@cspotcode/source-map-consumer/-/source-map-consumer-0.8.0.tgz#33bf4b7b39c178821606f669bbc447a6a629786b"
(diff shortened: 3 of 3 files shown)


DannyNemer:
Though I believe this change works, I am closing the PR because I am struggling to test this locally and do not think it warrants more time. E.g., how can I point an existing project to my fork instead of the installation in `node_modules`?

packages/xdl/src/Webpack.ts
```diff
@@ -628,8 +628,10 @@ async function loadConfigAsync(
   argv?: string[]
 ): Promise<WebpackConfiguration> {
   setMode(env.mode);
-  // Check if the project has a webpack.config.js in the root.
-  const projectWebpackConfig = path.resolve(env.projectRoot, 'webpack.config.js');
+  // Check if the project has a webpack.config.{ts,js} in the root.
+  const projectWebpackConfig = ["ts", "js"]
+    .map(ext => path.resolve(env.projectRoot, `webpack.config.${ext}`))
+    .find(filePath => fs.existsSync(filePath));
```
EvanBacon, line 634: Maybe just do something like this (a bit more readable and less complicated): 
```ts
const projectWebpackConfig = resolveFrom.silent(env.projectRoot, './webpack.config.ts') || resolveFrom.silent(env.projectRoot, './webpack.config.js');
```

packages/xdl/src/Webpack.ts
```diff
@@ -628,8 +628,10 @@ async function loadConfigAsync(
...
   let config: WebpackConfiguration;
   if (fs.existsSync(projectWebpackConfig)) {
     const webpackConfig = require(projectWebpackConfig);
```
EvanBacon, line 637: This line won't support TypeScript, perhaps try testing it with some types in the file.

DannyNemer: You are right. The following is the error I encounter when types exist in `webpack.config.ts`. I am happy to close this PR unless you see a solution?

image

packages/xdl/src/Webpack.ts
DannyNemer: Done. Thank you!

packages/xdl/src/Webpack.ts
DannyNemer: I extended the code to [use `ts-node` to patch `require()`](https://github.com/TypeStrong/ts-node#programmatic) to transpile TypeScript if using the TypeScript version. This avoids the syntax error shown above.

packages/xdl/package.json
```diff
@@ -90,6 +90,7 @@
     "terminal-link": "^2.1.1",
     "text-table": "^0.2.0",
     "tree-kill": "1.2.2",
+    "ts-node": "^10.2.1",
```
EvanBacon, line 93: keep in mind the bundle size increase by including [`ts-node`](https://packagephobia.com/result?p=ts-node), this isn't _too_ big relative to expo-cli, but it also depends on the user having typescript globally installed on their computer (which we aren't checking for currently), or locally packaged in the CLI ([typescript size](https://packagephobia.com/result?p=typescript) (non-typescript users would be installing a lot of code that they don't need)). It's also unclear whether this only supports the main `webpack.config.ts` file (i.e. all imports must be transpiled some other way), or what the `tsconfig.json` resolution would be since ts-node would be executed from a higher context.

Not being able to customize `tsconfig.json` could be a problem for some users, especially when new language features come out. IIRC `ts-node` is also quite strict by default, and slow to transpile, which is why they've been exploring [swc support](https://github.com/TypeStrong/ts-node#bundled-swc-integration), have you measured the net time increase (just in general, nothing fancy)?

---

Seems like maybe most of this could be better off, by living in user-space for now, and reusing the existing `tsconfig.json` and `typescript` modules that a TypeScript user has installed in their project. 

They could then have a TypeScript Webpack config with:

`webpack.config.js`
```js
require('ts-node/register');
module.exports = require('./webpack.config.ts')
```

`webpack.config.ts`
```ts
import createExpoWebpackConfigAsync from "@expo/webpack-config/webpack";
import { Arguments, Environment } from "@expo/webpack-config/webpack/types";

module.exports = async function (env: Environment, argv: Arguments) {
  const config = await createExpoWebpackConfigAsync(env, argv);
  // Customize the config before returning it.
  return config;
};
```

DannyNemer: Thank you for your thorough response, Evan!

I did not formally benchmark the performance impact of this change, but there is no noticeable difference in compile time on my project when switching from my fork and the latest stable version of `expo-cli`.

I agree with your cost-benefit analysis of adding `ts-node` as a project dependency. And your proposed alternative solution addresses wanting to leverage TypeScript for `webpack.config.*`.

I am comfortable with closing this PR if alright with you.

EvanBacon: Ok, I'll add documentation on how to setup `webpack.config.ts` locally (as described above) so it's a bit easier for users to discover in the future. Thanks for raising this question and helping us get it resolved!

PR #3841, opened 2021-09-10
Added patches system to expo/metro-config

EvanBacon:
# Why

In order to drop the react-native fork, we need certain patches applied to react-native. We can eventually upstream these features, but in the short term we need changes quickly. In the future if we need changes added, we can utilize this system to apply them.

## Patches

- https://github.com/expo/react-native/commit/a80b16c7a0233a46a1eb82fbe274344489312670

# How

The config will check `expo/patches` in the project to find module replacements. These replacements must work relative to the patch location (i.e. these aren't proper patches that get moved into the expected location). 

The patches are then used with a custom `resolver.resolveRequest` to ensure the functionality is as close to Metro as possible.

## Considerations

- This system will have to be ported to Webpack and any other bundler we want to implement.
- Users cannot use patch-package on the files we patch.
- Users can no longer implement a custom `resolver.resolveRequest` in their local config. This is probably not a huge concern since the feature is pretty cryptic. On GitHub, it seemed like some people were using it to support React Native Windows in older versions of Metro.

# Test Plan

- In a project: `touch node_modules/expo/patches/react-native/Libraries/Utilities/Appearance.js`
- Add the following (notice the imports have been updated):
```js
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 * @format
 * @flow strict-local
 */

 'use strict';

 import EventEmitter from 'react-native/Libraries/vendor/emitter/EventEmitter';
 import NativeEventEmitter from 'react-native/Libraries/EventEmitter/NativeEventEmitter';
 import NativeAppearance, {
   type AppearancePreferences,
   type ColorSchemeName,
 } from 'react-native/Libraries/Utilities/NativeAppearance';
 import invariant from 'invariant';
 import {isAsyncDebugging} from 'react-native/Libraries/Utilities/DebugEnvironment';
 
 type AppearanceListener = (preferences: AppearancePreferences) => void;
 const eventEmitter = new EventEmitter();
 
 if (NativeAppearance) {
   const nativeEventEmitter = new NativeEventEmitter(NativeAppearance);
   nativeEventEmitter.addListener(
     'appearanceChanged',
     (newAppearance: AppearancePreferences) => {
       const {colorScheme} = newAppearance;
       invariant(
         colorScheme === 'dark' ||
           colorScheme === 'light' ||
           colorScheme == null,
         "Unrecognized color scheme. Did you mean 'dark' or 'light'?",
       );
       // Update cached value
       preferences.colorScheme = colorScheme;
       eventEmitter.emit('change', {colorScheme});
     },
   );
 }
 
 function getInitialColorScheme(): ?ColorSchemeName {
   if (__DEV__) {
     if (isAsyncDebugging) {
       // Hard code light theme when using the async debugger as
       // sync calls aren't supported
       return 'light';
     }
   }
 
   // TODO: (hramos) T52919652 Use ?ColorSchemeName once codegen supports union
   const nativeColorScheme: ?string =
     NativeAppearance == null ? null : NativeAppearance.getColorScheme() || null;
   invariant(
     nativeColorScheme === 'dark' ||
       nativeColorScheme === 'light' ||
       nativeColorScheme == null,
     "Unrecognized color scheme. Did you mean 'dark' or 'light'?",
   );
   return nativeColorScheme;
 }
 
 function getColorScheme(): ?ColorSchemeName {
   if (!preferences.colorScheme) {
     const initialColorScheme = getInitialColorScheme();
     preferences.colorScheme = initialColorScheme;
     return preferences.colorScheme;
   }
 
   return preferences.colorScheme;
 }
 
 const preferences = {
   colorScheme: null,
 };
 
 module.exports = {
   /**
    * Note: Although color scheme is available immediately, it may change at any
    * time. Any rendering logic or styles that depend on this should try to call
    * this function on every render, rather than caching the value (for example,
    * using inline styles rather than setting a value in a `StyleSheet`).
    *
    * Example: `const colorScheme = Appearance.getColorScheme();`
    *
    * @returns {?ColorSchemeName} Value for the color scheme preference.
    */
   getColorScheme,
   /**
    * Add an event handler that is fired when appearance preferences change.
    */
   addChangeListener(listener: AppearanceListener): void {
     eventEmitter.addListener('change', listener);
   },
   /**
    * Remove an event handler.
    */
   removeChangeListener(listener: AppearanceListener): void {
     eventEmitter.removeListener('change', listener);
   },
 };
```
- Running `EXPO_DEBUG=1 expod start` should log how patches are being applied:
```
[emc] using patches: [
  {
    match: /react\x2dnative\/Libraries\/Utilities\/Appearance\.js/,
    replace: '/Users/evanbacon/Documents/GitHub/lab/yolo49/node_modules/expo/patches/react-native/Libraries/Utilities/Appearance.js'
  }
]

...

[emc] Replace:  /Users/evanbacon/Documents/GitHub/lab/yolo49/node_modules/react-native/Libraries/Utilities/Appearance.js  ->  /Users/evanbacon/Documents/GitHub/lab/yolo49/node_modules/expo/patches/react-native/Libraries/Utilities/Appearance.js
[emc] Replace:  /Users/evanbacon/Documents/GitHub/lab/yolo49/node_modules/react-native/Libraries/Utilities/Appearance.js  ->  /Users/evanbacon/Documents/GitHub/lab/yolo49/node_modules/expo/patches/react-native/Libraries/Utilities/Appearance.js

```

- [ ] integration tests

### DIFF
--- a/packages/metro-config/package.json
+++ b/packages/metro-config/package.json
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "expo": "37",
+    "semver": "^7.3.5",
     "metro": "^0.59.0",
     "metro-config": "^0.59.0",
     "react-native": "~0.63.4"
--- a/packages/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/metro-config/src/ExpoMetroConfig.ts
@@ -1,13 +1,17 @@
 import { getConfig, getDefaultTarget, isLegacyImportsEnabled, ProjectTarget } from '@expo/config';
 import { getBareExtensions, getManagedExtensions } from '@expo/config/paths';
 import chalk from 'chalk';
-import { boolish } from 'getenv';
 import { Reporter } from 'metro';
 import type MetroConfig from 'metro-config';
 import path from 'path';
 import resolveFrom from 'resolve-from';
 
-export const EXPO_DEBUG = boolish('EXPO_DEBUG', false);
+import { EXPO_DEBUG } from './env';
+import { importMetroConfigFromProject } from './importMetroFromProject';
+import {
+  createModuleReplacementResolver,
+  getVendoredExpoPatches,
+} from './moduleReplacementResolver';
 
 // Import only the types here, the values will be imported from the project, at runtime.
 export const INTERNAL_CALLSITES_REGEX = new RegExp(
@@ -70,6 +74,7 @@ export function getDefaultConfig(
 ): MetroConfig.InputConfigT {
   const MetroConfig = importMetroConfigFromProject(projectRoot);
 
+  const patches = getVendoredExpoPatches(projectRoot);
   const reactNativePath = path.dirname(resolveFrom(projectRoot, 'react-native/package.json'));
 
   try {
@@ -156,6 +161,7 @@ export function getDefaultConfig(
     console.log(`- Babel config: ${babelConfigPath || 'babel-preset-expo (default)'}`);
     console.log();
   }
+
   const {
     // Remove the default reporter which metro always resolves to be the react-native-community/cli reporter.
     // This prints a giant React logo which is less accessible to users on smaller terminals.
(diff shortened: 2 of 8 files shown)


Kudo:
awesome 🔥  that is really an interesting approach!

for `Users cannot use patch-package on the files we patch.` consideration and maintain patches across react-native versions, i have some idea but not sure if metro-resolver can support file content instead of file path, or maybe we could do it by a babel plugin.

the idea is to maintain patch files and apply patches on the fly. something like:

```ts
let result = resolve(context, realModuleName, platform);
result = patch(result, '/path/to/module.patch`);
```
the patch file is generated by either `diff -u` or `git diff`.

we also need a javascript library to apply patches. i did some experiment, unfortunately, neither [jsdiff](https://github.com/kpdecker/jsdiff/blob/master/src/patch/apply.js) nor [apply in patch-package](https://github.com/ds300/patch-package/blob/master/src/patch/apply.ts) works well as gnu patch-apply tool especially for the fuzzing apply algorithm. so in the long run, we might better to build an patch-apply library. 

just to share some long-term thought and not have any meaningful items in the mean time.  😅

Issue #3843, opened 2021-09-13
run:ios -d device not found when device name contains brackets containing an integer


Lat91:
### Summary

iPhone device name (settings > General > About > Name) of the form "string (integer)" for example "iphone (2)" note the space between the string and opening bracket as well.  Reproduced with two different iphones and two different Macs with expo.  Wasted far too much time trying to figure out why my iphones couldn't be found when executing expo run:ios -d until the penny dropped and renamed the device.

Expected:
Connect iphone to mac
run: expo run:ios -d
Expect to see connected device in list of devices to run the app on, as well as the simulators.

Actual:
The device is not found and needs to be renamed in order to be found.  Any name that doesn't have a space followed by an integer in brackets seems to be ok.

Example device names used on the iphone (settings > General > About > Name) (taking values within double quotes):
"iphone (th5)" - OK
"iphone (255)" - NOK
"iphone(255)" - OK
"iphone (2)" - NOK
"iphone(2)" - OK
"Foo (2)" - NOK
"Foo(2)" - OK

### Environment

Expo CLI 4.11.0 environment info:
    System:
      OS: macOS 11.5.2
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 12.21.0 - ~/.nvm/versions/node/v12.21.0/bin/node
      npm: 7.6.0 - ~/.nvm/versions/node/v12.21.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.11.0 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
      Android SDK:
        API Levels: 28, 29, 30
        Build Tools: 29.0.2, 29.0.3, 30.0.0, 30.0.3
        System Images: android-29 | Google Play Intel x86 Atom, android-30 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 4.1 AI-201.8743.12.41.7042882
      Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~42.0.1 => 42.0.3 
      react: 16.13.1 => 16.13.1 
      react-dom: 16.13.1 => 16.13.1 
      react-native: ~0.63.4 => 0.63.4 
      react-native-web: ~0.13.12 => 0.13.18 
    npmGlobalPackages:
[...]

PR #3847, opened 2021-09-14
[expo-cli][run:ios] JS API for iOS binary install, and connected devices

EvanBacon:
# Why

To install `.app` binaries on connected iOS devices, we currently use the package `ios-deploy`, which the user must globally install before running. This new approach is pure JS and substantially faster, it's unclear how long it will last, Apple could change the underlying API, but then `ios-deploy` would probably have the same issue.

To list connected devices (required for `expo run:ios -d`) we use xctrace, which doesn't have a very reliable system for finding connected devices (https://github.com/expo/expo-cli/issues/3843). xctrace is also very slow, it takes about two seconds on fast computers, and the time varies unreliably, I've also observed that around iOS updates, xctrace can sometimes not detect the connected device unless the computer is also up to date. The new JS approach has none of these issues, it takes about 50ms to find devices, provides a JSON-type API for each device, and I've found that it tends to work more reliably than `xctrace`.

# How

The JS impl is based on the package `native-run` (based on [node-ioslib](https://www.npmjs.com/package/node-ioslib)), initially I pulled in native-run, but found that the net increase was pretty large and had many unneeded features (like android installation, or simulator utils). I've forked the relevant parts using the same file structure, included the licenses, and tested to ensure it was still working.

This feature is also much faster than `ios-deploy` because it doesn't attach lldb to the process, effectively making it faster to build onto a connected device than Xcode.

I had to roll support for delta installs myself, cannot confirm that they are working the same as before, I don't see any drastic time differences regardless.

# Test Plan

Due to the nature of this feature, I've put it behind a feature flag (`EXPO_USE_APPLE_DEVICE`) that users will have to enable for a beta period.

```
EXPO_USE_APPLE_DEVICE=1 expod run:ios --device
```

Extra logs can be enabled with `export DEBUG=expo:xdl:ios:*`, `EXPO_DEBUG=1`, and `EXPO_PROFILE=1`.

### DIFF
--- a/packages/xdl/src/apple/AppleDevice.ts
+++ b/packages/xdl/src/apple/AppleDevice.ts
@@ -69,7 +69,6 @@ export async function runOnDevice({
     await uploadApp(clientManager, appPath, destPackagePath);
 
     const installer = await clientManager.getInstallationProxyClient();
-    // TODO: Progress callback
     await installer.installApp(
       destPackagePath,
       bundleId,
--- a/packages/xdl/src/apple/AppleDevice.ts
+++ b/packages/xdl/src/apple/AppleDevice.ts
@@ -1,57 +1,23 @@
 import Debug from 'debug';
 import { readFileSync } from 'fs';
 import { boolish } from 'getenv';
-// @ts-ignore
-import type { DeviceValues, IPLookupResult } from 'native-run/dist/ios/lib';
+import * as path from 'path';
+
 import {
   AFC_STATUS,
   AFCError,
   ClientManager,
   LockdowndClient,
+  OnInstallProgressCallback,
   UsbmuxdClient,
-  // @ts-ignore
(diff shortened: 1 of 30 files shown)


packages/xdl/src/apple/AppleDevice.ts
```diff
@@ -0,0 +1,183 @@
...
+  try {
+    await mountDeveloperDiskImage(clientManager);
+
+    const packageName = path.basename(appPath);
+    const destPackagePath = path.join('PublicStaging', packageName);
+
+    await uploadApp(clientManager, appPath, destPackagePath);
+
+    const installer = await clientManager.getInstallationProxyClient();
+    // TODO: Progress callback
```
EvanBacon, line 72:
```suggestion
```

brentvatne (approved):

PR #3851, opened 2021-09-15
Add --no-install option to expo run:ios

EvanBacon:
# Why

sometimes you don't want the pods to be installed


brentvatne (approved):
ydliihrpaw971

PR #3853, opened 2021-09-16
[cli] Remove prompts to install Expo Go on upgrade

byCedric:
# Why

These questions during `expo upgrade` were made before we checked the Expo Go version when loading the app. This is a proposal to remove these questions, to make upgrade ask less questions to the user.

# How

These prompts helps users to get the right version of Expo Go:

Android | iOS
--- | ---
expo start --android | expo start --ios

# Test Plan

This is just a cleanup, no new code added.


EvanBacon (approved):
seems fine

Issue #3846, opened 2021-09-14
Android Variants - android apk with variant in wrong output folder


maitriyogin:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `expo-cli@4.11.0` for the project I'm working on.

Ok, so here's the issue, when using `expo run:android --variant devDebug` expo-cli looks for the built apk file in the wrong dir
`android/app/build/outputs/apk/devbuild`
although gradle builds it here
`android/app/build/outputs/apk/dev/build`
Also the name of the apk is `app-devbuild.apk` instead of `app-dev-build.apk`

The patch fixes this by making sure the variant is escaped correctly and the correct path and apk name is created as per gradles expectations ...

Without this, expo-cli tries to search for the apk in `android/app/build/outputs/apk/devbuild` while starting up metro and then throws the following and so ending my experience ....
`Failed to resolve APK build file in folder "<approot>/android/app/build/outputs/apk/devdebug"`

This is part 2 of the Android Variants issues.

/Stephen

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/expo-cli/build/commands/run/android/runAndroid.js b/node_modules/expo-cli/build/commands/run/android/runAndroid.js
index 59733a7..93f704d 100644
--- a/node_modules/expo-cli/build/commands/run/android/runAndroid.js
+++ b/node_modules/expo-cli/build/commands/run/android/runAndroid.js
@@ -280,9 +280,9 @@ async function resolveOptionsAsync(projectRoot, options) {
   const variant = options.variant.toLowerCase();
 
   const apkDirectory = _xdl().Android.getAPKDirectory(projectRoot);
-
-  const apkVariantDirectory = _path().default.join(apkDirectory, variant);
-
+  const variantFolder = variant.includes('debug') ? variant.replace('debug', '-debug') : variant.includes('release') ? variant.replace('release', '-release') : variant;
+  const variantFolders = variantFolder.split('-')
+  const apkVariantDirectory = _path().default.join(apkDirectory, ...variantFolders);
   return { ...options,
     port,
     device,
@@ -290,7 +290,7 @@ async function resolveOptionsAsync(projectRoot, options) {
     launchActivity: `${packageName}/${mainActivity}`,
     packageName,
     apkVariantDirectory,
-    variantFolder: variant,
+    variantFolder,
     appName: 'app'
   };
 }
[...]

Issue #3845, opened 2021-09-14
Android Variants - use package if defined in app.json or app.config.js


maitriyogin:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `xdl@59.1.0` for the project I'm working on.

I've been trying to get the dev-client working with Android Variants but have had to patch xdl to get the correct package name.

This is 1 of 2 patches for Android Variants, that I've made to allow for android product flavours. The next being in expo-cli/build/commnads/run/android/runAndroid.js which addresses the extrapolation of the variant name.

But for this one, if you set in app.json or app.config.js
```
android: {
        package: 'com.humpty.app.dev',
      },
```
expo-cli will then use the correct package based upon the variant (flavour+buildType) on the emulator.

I understand this is probably not the best solution, but with something that uses the variant flavour ( as in dev in the variant devDebug ) probably being a bit more stable.  I just kind of liked the simplicity...

/Stephen.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/xdl/build/Android.js b/node_modules/xdl/build/Android.js
index 90ca15b..2c59f59 100644
--- a/node_modules/xdl/build/Android.js
+++ b/node_modules/xdl/build/Android.js
@@ -799,13 +799,13 @@ async function openUrlAsync({
           throw new Error(`Could not find property android.package in app.config.js/app.json. This setting is required to launch the app.`);
         }
       } else {
-        applicationId = await _configPlugins().AndroidConfig.Package.getApplicationIdAsync(projectRoot);
-
+        // if the package is set in the app.json or app.config.js/ts then just use that!
+        applicationId = exp.android.package !== undefined ? exp.android.package : await _configPlugins().AndroidConfig.Package.getApplicationIdAsync(projectRoot);
+        console.log("------- ApplicationId:", applicationId);
         if (!applicationId) {
           throw new Error(`Could not find applicationId in ${_configPlugins().AndroidConfig.Paths.getAppBuildGradleFilePath(projectRoot)}`);
         }
       }
-
       clientApplicationId = applicationId;
       await ensureDevClientInstalledAsync(device, clientApplicationId);
     } else if (!isDetached) {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

Issue #3714, opened 2021-07-22
add support for appIdSuffix arg to expo run:android


Stukz:
Currently on the project I'm working on when I  run `react-native run-android --variant=devDebug --appIdSuffix=dev` the apk output file it's **android/app/build/outputs/apk/dev/debug/app-dev-debug.apk**

And If I use expo run:android 
 The command throws "**Failed to resolve APK build file in folder "/Users/$USER/../android/app/build/outputs/apk/debug**"

expo run:android --variant=devDebug 
The command throws **Failed to resolve APK build file in folder "/Users/$USER/.../android/app/build/outputs/apk/devdebug**"
@EvanBacon

EvanBacon:
After looking at this more, I think I'd rather add a `--flavor` flag which can apply all of the various settings defined in a gradle flavor. This would make the functionality very similar to the `run:ios --scheme` flag.

PR #3856, opened 2021-09-17
[expo run:android] add flavors support

EvanBacon:
# Why

Appears to resolve all `expo run:android` issues, it's unclear why we didn't make it like this originally.

- https://github.com/expo/expo-cli/issues/3846
- https://github.com/expo/expo-cli/issues/3845
- https://github.com/expo/expo-cli/issues/3714

# How

- Use forward look ahead regex split to determine which flavors are added to the build type. Use the discovered flavors to discover the correct binary output location.
- Fallback onto gradle install when we can't find the binary
- Fix opening app on device -- will split out

# Test Plan

- TBD

### DIFF
--- a/packages/expo-cli/src/commands/run/android/runAndroid.ts
+++ b/packages/expo-cli/src/commands/run/android/runAndroid.ts
@@ -14,10 +14,10 @@ import { promptToClearMalformedNativeProjectsAsync } from '../../eject/clearNati
 import { prebuildAsync } from '../../eject/prebuildAsync';
 import { installCustomExitHook } from '../../start/installExitHooks';
 import { profileMethod } from '../../utils/profileMethod';
-import { startBundlerAsync } from '../ios/startBundlerAsync';
+import { setGlobalDevClientSettingsAsync, startBundlerAsync } from '../ios/startBundlerAsync';
 import { resolvePortAsync } from '../utils/resolvePortAsync';
 import { resolveDeviceAsync } from './resolveDeviceAsync';
-import { spawnGradleAsync } from './spawnGradleAsync';
+import { assembleAsync, installAsync } from './spawnGradleAsync';
 
 type Options = {
   variant: string;
@@ -32,8 +32,9 @@ export type AndroidRunOptions = Omit<Options, 'device'> & {
   mainActivity: string;
   launchActivity: string;
   device: Android.Device;
-  variantFolder: string;
   appName: string;
+  buildType: string;
+  flavorDimensions?: string[];
 };
 
 async function resolveAndroidProjectPathAsync(projectRoot: string): Promise<string> {
@@ -101,9 +102,16 @@ async function resolveOptionsAsync(
     port = 8081;
   }
 
-  const variant = options.variant.toLowerCase();
-  const apkDirectory = Android.getAPKDirectory(projectRoot);
-  const apkVariantDirectory = path.join(apkDirectory, variant);
+  // TODO: why would this be different? Can we get the different name?
+  const appName = 'app';
+
+  const apkDirectory = path.join(projectRoot, 'android', appName, 'build', 'outputs', 'apk');
+
+  // buildDeveloperTrust -> build, developer, trust (where developer, and trust are flavors).
+  const [buildType, ...flavorDimensions] = options.variant
+    .split(/(?=[A-Z])/)
+    .map(v => v.toLowerCase());
+  const buildDirectory = path.join(apkDirectory, ...flavorDimensions, buildType);
 
   return {
     ...options,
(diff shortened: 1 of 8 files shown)


packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -101,9 +102,16 @@ async function resolveOptionsAsync(
...
-  const apkDirectory = Android.getAPKDirectory(projectRoot);
-  const apkVariantDirectory = path.join(apkDirectory, variant);
+  // TODO: why would this be different? Can we get the different name?
+  const appName = 'app';
+
+  const apkDirectory = path.join(projectRoot, 'android', appName, 'build', 'outputs', 'apk');
+
+  // buildDeveloperTrust -> build, developer, trust (where developer, and trust are flavors).
+  const [buildType, ...flavorDimensions] = options.variant
+    .split(/(?=[A-Z])/)
```
wkozyra95, line 112: could use some comment

packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -101,9 +102,16 @@ async function resolveOptionsAsync(
...
-  const variant = options.variant.toLowerCase();
-  const apkDirectory = Android.getAPKDirectory(projectRoot);
-  const apkVariantDirectory = path.join(apkDirectory, variant);
+  // TODO: why would this be different? Can we get the different name?
+  const appName = 'app';
+
+  const apkDirectory = path.join(projectRoot, 'android', appName, 'build', 'outputs', 'apk');
+
+  // buildDeveloperTrust -> build, developer, trust (where developer, and trust are flavors).
+  const [buildType, ...flavorDimensions] = options.variant
```
wkozyra95, line 111: Naming here is partially incorrect, if you have command `:app:buildFooBarRelease`
- app is an android module
- `foo` and `bar` are flavors (not flavorDimensions), but actual flavours can be named `Foo` and `Bar` we don't know that just based on the gradle command
-  flavorDimensions defines order of the flavors in the command, so if you have in your build.gradle `flavourDimensions: "dim1" "dim2"` then `foo` has to have dimension set to `dim1` and `bar` to `dim2`
- variant in this example would be `FooBarRelease`, build/bundle/assemble... is not the part of the variant
- buildType is either debug, release or some custom value

packages/expo-cli/src/commands/run/android/runAndroid.ts
wkozyra95: also, I'm not saying we should care about that case, but just FYI

if you have flavor e.g. `exaMPle` the gradle command would look like this `:app:buildExaMPleRelease`

EvanBacon: I think for the time being, the current implementation is a fair compromise.

packages/expo-cli/src/commands/run/android/runAndroid.ts
EvanBacon: renamed `flavorDimensions` to `flavors`

brentvatne (approved):

packages/expo-cli/src/commands/run/android/runAndroid.ts
```diff
@@ -101,9 +102,14 @@ async function resolveOptionsAsync(
...
-  const variant = options.variant.toLowerCase();
-  const apkDirectory = Android.getAPKDirectory(projectRoot);
-  const apkVariantDirectory = path.join(apkDirectory, variant);
+  // TODO: why would this be different? Can we get the different name?
+  const appName = 'app';
+
+  const apkDirectory = path.join(projectRoot, 'android', appName, 'build', 'outputs', 'apk');
+
+  // buildDeveloperTrust -> build, developer, trust (where developer, and trust are flavors).
+  const [buildType, ...flavors] = options.variant.split(/(?=[A-Z])/).map(v => v.toLowerCase());
```
EvanBacon, line 111:
```suggestion
  // This won't work for non-standard flavor names like "myFlavor" would be treated as "my", "flavor".
  const [buildType, ...flavors] = options.variant.split(/(?=[A-Z])/).map(v => v.toLowerCase());
```

Stukz:
hey @EvanBacon I was trying to use some flavors and launch them with `expo run:android --variant developDebug`

image

but it's giving me this issue

image

but on the other hand if I try to use `npx react-native` it only works doing `npx react-native run-android --variant developDebug --appIdSuffix dev`

not sure if `--appIdSuffix` it's a param that should be required in order to build

EvanBacon:
Seems like it, maybe we could add a naive check to automatically find the `applicationIdSuffix`. Would you be interested in opening a PR to add this? ([source lives here](https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/run/android/runAndroidAsync.ts)).

PR #3860, opened 2021-09-20
[xdl] Fix expo-updates EAS manifest assets for iOS

wschurman:
# Why

Assets in the manifest are only used in production as far as I understand. In development, these don't matter (the bundler creates a map and serves and uses them).

This also fixes the issue I created by trying to use the new asset format for local development manifest assets (splash screen URL for example). I was planning to try to use a new augmented format in the manifest for these, but it is far easier to just reuse the `${field}Url` augmentation pattern from legacy manifests since all the client libraries are already looking for this field.

# How

Revert the asset portion of https://github.com/expo/expo-cli/commit/d8eb71f0d778514349f43d78829d635fbb68f7be and replace it with a) an empty manifest.assets array, and b) the same logic as is used in classic manifests for local development.

# Test Plan

1. `yarn start` in this repo
2. In a project, `EXPO_DEBUG=true ~/expo/expo-cli/node_modules/.bin/expo start`
3. Load http://192.168.1.133:19000/update-manifest-experimental?platform=ios on ios and the android equivalent on android, ensure `Image` assets show correctly.

### DIFF
--- a/packages/xdl/src/ProjectAssets.ts
+++ b/packages/xdl/src/ProjectAssets.ts
@@ -1,7 +1,6 @@
 import { ExpoAppManifest, ExpoConfig } from '@expo/config';
 import { BundleAssetWithFileHashes, BundleOutput } from '@expo/dev-server';
 import assert from 'assert';
-import crypto from 'crypto';
 import FormData from 'form-data';
 import fs from 'fs-extra';
 import chunk from 'lodash/chunk';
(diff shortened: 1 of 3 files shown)


ide (approved):
Approving to unbreak with a question and a thought - does this still properly put assets in the manifest when publishing a modern update? And we may want to one day add this back if we can get development mode to behave more like production: load the update atomically, including assets, before running it.

wschurman:
> does this still properly put assets in the manifest when publishing a modern update?

Yep. According to @jkhales all of that code lives in `eas-cli`. The only thing it uses `expo-cli` for is building the bundle (I think?).

> And we may want to one day add this back if we can get development mode to behave more like production: load the update atomically, including assets, before running it.

Yeah, it's definitely a possibility. A lot of this was struggling to keep up with the changes to asset extension logic on iOS in the short term, but theoretically we may be able to correctly set this up in the future using asset keys and extensions more similar to production.

PR #3867, opened 2021-09-22
Fix loader bug

EvanBacon:
# Why

ora.complete is not changing the message.

# Test Plan

`expo prebuild -p ios` should log correctly:

```
✔ Updated package.json and added index.js entry point for iOS and Android
🧶 Using Yarn to install packages. Pass --npm to use npm instead.
✔ Installed JavaScript dependencies 1559ms
✔ Config synced
✔ Installed pods and initialized Xcode workspace.
```

### DIFF
--- a/packages/expo-cli/src/utils/ora.ts
+++ b/packages/expo-cli/src/utils/ora.ts
@@ -1,8 +1,17 @@
 import chalk from 'chalk';
 import program from 'commander';
-import oraReal from 'ora';
+import oraReal, { Ora } from 'ora';
 
 import Log from '../log';
+
+// eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/utils/ora.ts
```diff
@@ -60,22 +62,27 @@ export function ora(options?: oraReal.Options | string): oraReal.Ora {
     console.error = errorReal;
   };
 
-  ora.stopAndPersist = (): oraReal.Ora => {
-    origStopAndPersist();
+  spinner.start = (text: string): Ora => {
```
EvanBacon, line 65:
```suggestion
  spinner.start = (text): Ora => {
```

PR #3872, opened 2021-09-24
[metro-config] bump dependencies for SDK 43

Simek:
# Why

Since SDK 43 will be based on RN `0.64`, so I think that the `metro-config` dependencies should be bumped to match RN ones:
* https://github.com/facebook/react-native/blob/0.64-stable/package.json#L100-L103

# How

This is just an attempt of bumping the dependencies, since I'm not sure if those changes are required I have just bumped the dependencies without much testing, so in case when this is not need, I do not lose a bunch of time. 😅 

# Test Plan

Run the CI, then work on the fixes after getting approval on those changes.

### DIFF
--- a/packages/metro-config/package.json
+++ b/packages/metro-config/package.json
@@ -37,16 +37,16 @@
     "chalk": "^4.1.0",
     "debug": "^4.3.2",
     "getenv": "^1.0.0",
-    "metro-react-native-babel-transformer": "^0.59.0",
+    "metro-react-native-babel-transformer": "^0.64.0",
     "sucrase": "^3.20.0"
   },
   "devDependencies": {
     "@babel/core": "^7.15.5",
-    "expo": "37",
-    "metro": "^0.59.0",
+    "expo": "^43.0.0-alpha.3",
+    "metro": "^0.64.0",
     "metro-babel-transformer": "^0.66.2",
-    "metro-config": "^0.59.0",
-    "react-native": "~0.63.4"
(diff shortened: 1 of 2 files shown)


packages/metro-config/package.json
```diff
@@ -37,16 +37,16 @@
     "chalk": "^4.1.0",
     "debug": "^4.3.2",
     "getenv": "^1.0.0",
-    "metro-react-native-babel-transformer": "^0.59.0",
+    "metro-react-native-babel-transformer": "^0.64.0",
```
EvanBacon, line 40:
```suggestion
```

And remove from metro-config, this was added in error. We should always read from the local project and throw if a package isn't available.

Simek:
The polyfills path fail looks like expected change because of this RN commit:
* https://github.com/facebook/react-native/commit/a77f2c40d1044da63426078f356cfc7c1284a79e#diff-86b4c8f05cbd60b41e9a5b16c03721d365de1e3868f25e69d4794cc647d4a4c6

But I was not able to understand yet the issues around `getConfig`, I'm not sure about that, but they might be related to some changes in `@expo/config`?

Simek:
@EvanBacon It looks like that, at least on my machine, same tests in `metro-config` are also failing on `master`. Can you confirm that?

Simek:
It looks like the `metro-config` dependencies have been bumped already in:
* #3581

Which makes this PR no longer needed.

PR #3875, opened 2021-09-28
[xdl][expo-cli] Bump @expo/rudder-sdk-node

kgc00:
# Why

Updates the analytics client to the latest version. Will reduce the size and remove unneeded dependencies.

# How

Patched the npm package. Made small adjustments to code around the new async flush.

# Test Plan

Ensured analytics events are still being sent by running locally.

### DIFF
--- a/packages/xdl/src/Analytics.ts
+++ b/packages/xdl/src/Analytics.ts
@@ -24,9 +24,9 @@ export class AnalyticsClient {
     return this._version;
   }
 
-  public flush() {
+  public async flush() {
     if (this.rudderstackClient) {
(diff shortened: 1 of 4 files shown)


packages/xdl/src/Analytics.ts
```diff
@@ -43,7 +43,6 @@ export class AnalyticsClient {
         flushInterval: 300,
       }
     );
-    this.rudderstackClient.logger.silent = true;
```
kgc00, line 46: package is silent by default

wkozyra95 (approved):

PR #3876, opened 2021-09-30
[config-plugins] only support dynamic runtimes in managed

jkhales:
# Why

For the time being, only allow auto runtime versions in managed.

The `getRuntimeVersion` wrapper should be exported and used to compute the runtimes for `eas publish` here: https://github.com/expo/eas-cli/blob/03fa29dd5a3fae9cc926d2ba26ebc51f8b42af27/packages/eas-cli/src/commands/branch/publish.ts#L395

And for `eas build` where ever it is appropriate.

# How

Added a check for managed projects.

# Test Plan

added tests.

### DIFF
--- a/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
+++ b/packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
@@ -1,4 +1,4 @@
-import { ExpoUpdatesManifest, getConfig } from '@expo/config';
+import { ExpoUpdatesManifest, getConfig, getDefaultTarget } from '@expo/config';
 import { Updates } from '@expo/config-plugins';
 import { JSONObject } from '@expo/json-file';
 import express from 'express';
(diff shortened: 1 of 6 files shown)


packages/xdl/src/start/ExpoUpdatesManifestHandler.ts
```diff
@@ -114,9 +114,11 @@ export async function getManifestResponseAsync({
 
   const hostUri = await UrlUtils.constructHostUriAsync(projectRoot, hostname);
 
+  const isManagedProject = getDefaultTarget(projectRoot) === 'managed';
```
brentvatne, line 117: this is actually a bit problematic because eas-cli and expo-cli have different ideas of what this means. we want to move away from classifying managed/bare projects in expo-cli, so i think we shouldn't add this.

for eas-cli, **we need to know if a project has ios/android directories committed to git**. if not, then we read data and do validations differently, and we run prebuild on eas build.

because eas update is an eas service, it should follow that same logic imo.

jkhales: Ah, and here I was thinking this was solved!

Thanks for point this out.

PR #3878, opened 2021-10-01
[expo-cli] Add version check fallback for packages with exports

### LINKED ISSUE #3781: Validating dependency versions fails on Firebase SDK v9
### Summary

Expo does not currently support Firebase Javascript SDK v9 (beta). Maybe expected but seems unexpected since there is no information about unsupported versions [in the docs](https://docs.expo.dev/guides/using-firebase/). 

After installing firebase@9.0.0-beta.7 the firebase dependency is no longer recognized. [This](https://stackoverflow.com/questions/67688853/firebase-is-added-as-a-dependency-in-your-projects-package-json-but-it-doesn?rq=1) stackoverflow thread describes a similar issue. However, the suggested solution is clearly not viable and does not work for me.

`expo install firebase` currently installs version 8.2.3 of firebase, installing the latest stable version of firebase (8.9.1) causes no issues. Is there a reason why v9 does not work?

### Managed or bare workflow? If you have `ios/` or `android/` directories in your project, the answer is bare!

managed

### What platform(s) does this occur on?

iOS

### SDK Version (managed workflow only)

42

### Environment

Expo CLI 4.10.0 environment info:
    System:
      OS: macOS 11.4
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 16.3.0 - /opt/homebrew/bin/node
      Yarn: 1.22.10 - /opt/homebrew/bin/yarn
      npm: 7.15.1 - /opt/homebrew/bin/npm
      Watchman: 2021.08.02.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.10.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    IDEs:
      Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~42.0.1 => 42.0.3 
      react: 16.13.1 => 16.13.1 
      react-dom: 16.13.1 => 16.13.1 
      react-native: https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz => 0.63.2 
      react-native-web: ~0.13.12 => 0.13.18 
    Expo Workflow: managed

### Reproducible demo or steps to reproduce from a blank project

1. npm i firebase@9.0.0-beta.7
2. expo start

Output:
`"firebase" is added as a dependency in your project's package.json but it doesn't seem to be installed. Please run "yarn" or "npm install" to fix this issue.`

byCedric:
# Why

Fixes #3781 

Firebase JS SDK v9 uses a [new "exports" system](https://nodejs.org/api/modules.html#modules_exports) that defines all importable modules from this package. Right now, we don't support this when detecting the versions of installed libraries. It results in weird errors like #3781:

> "firebase" is added as a dependency in your project's package.json but it doesn't seem to be installed. Please run "yarn" or "npm install" to fix this issue.

# How

This is a wacky-workaround to extract the path from [the "ERR_PACKAGE_PATH_NOT_EXPORTED" error](https://nodejs.org/api/errors.html#errors_err_package_path_not_exported), thrown by Node. It can have two different formats:

<details>
<summary><code>$ node -e "require.resolve('firebase')"</code></summary>
<pre>
> Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/cedric/Projects/expo/expo/node_modules/firebase/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:335:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:560:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.resolve (node:internal/modules/cjs/helpers:100:19)
    at [eval]:1:9
    at Script.runInThisContext (node:vm:129:12)
    at Object.runInThisContext (node:vm:305:38) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
</pre>
</details>

<details>
<summary><code>$ node -e "require.resolve('firebase/package.json')"</code></summary>
<pre>
> Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/cedric/Projects/expo/expo/node_modules/firebase/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:335:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:560:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.resolve (node:internal/modules/cjs/helpers:100:19)
    at [eval]:1:9
    at Script.runInThisContext (node:vm:129:12)
    at Object.runInThisContext (node:vm:305:38) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
</pre>
</details>

# Test Plan

- `$ expo init . -t blank`
- `$ yarn add firebase@9.1.0`
- `$ expo start`
    - This should not error with "Firebase was not installed"

### DIFF
--- a/packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
+++ b/packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
@@ -1,5 +1,6 @@
 import { vol } from 'memfs';
 import path from 'path';
+import resolveFrom from 'resolve-from';
 
 import { validateDependenciesVersionsAsync } from '../validateDependenciesVersions';
 
@@ -76,8 +77,10 @@ describe(validateDependenciesVersionsAsync, () => {
   });
 
   it('resolves to true when installed package uses "exports"', async () => {
+    const packageJsonPath = path.join(projectRoot, 'node_modules/firebase/package.json');
+
     vol.fromJSON({
-      [path.join(projectRoot, 'node_modules/firebase/package.json')]: JSON.stringify({
+      [packageJsonPath]: JSON.stringify({
         version: '9.1.0',
         exports: {
           './analytics': {
@@ -89,6 +92,16 @@ describe(validateDependenciesVersionsAsync, () => {
           },
         },
       }),
+    });
+
+    // Manually trigger the Node import error for "exports".
+    // This isn't triggered by memfs, or our mock, that's why we need to do it manually.
+    // see: https://github.com/expo/expo-cli/pull/3878
+    (resolveFrom as jest.MockedFunction<typeof resolveFrom>).mockImplementationOnce(() => {
+      const message = `Package subpath './package.json' is not defined by "exports" in ${packageJsonPath}`;
+      const error: any = new Error(message);
+      error.code = 'ERR_PACKAGE_PATH_NOT_EXPORTED';
+      throw error;
     });
 
     const exp = {
--- a/packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
+++ b/packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
@@ -9,6 +9,7 @@ jest.mock('../bundledNativeModules', () => ({
   getBundledNativeModulesAsync: () => ({
     'expo-splash-screen': '~1.2.3',
     'expo-updates': '~2.3.4',
+    firebase: '9.1.0',
   }),
 }));
 
@@ -71,6 +72,34 @@ describe(validateDependenciesVersionsAsync, () => {
 
     await expect(validateDependenciesVersionsAsync(projectRoot, exp as any, pkg)).resolves.toBe(
       false
+    );
+  });
+
+  it('resolves to true when installed package uses "exports"', async () => {
+    vol.fromJSON({
+      [path.join(projectRoot, 'node_modules/firebase/package.json')]: JSON.stringify({
+        version: '9.1.0',
+        exports: {
+          './analytics': {
+            node: {
+              require: './analytics/dist/index.cjs.js',
+              import: './analytics/dist/index.mjs',
+            },
+            default: './analytics/dist/index.esm.js',
+          },
+        },
+      }),
+    });
+
+    const exp = {
+      sdkVersion: '43.0.0',
+    };
+    const pkg = {
+      dependencies: { firebase: '~9.1.0' },
+    };
+
+    await expect(validateDependenciesVersionsAsync(projectRoot, exp as any, pkg)).resolves.toBe(
+      true
     );
   });
 });
--- a/packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
+++ b/packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
@@ -100,8 +100,6 @@ async function getPackageVersionAsync(projectRoot: string, packageName: string):
     // include `package.json`, we have to use the error message to get the location.
     if (error.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED') {
       packageJsonPath = error.message.match(/ in (.*)$/i)?.[1];
-    } else {
-      packageJsonPath = resolveFrom.silent(projectRoot, `${packageName}/package.json`);
     }
   }
   if (!packageJsonPath) {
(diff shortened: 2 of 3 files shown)


byCedric:
@EvanBacon I'm not sure if we can test this behavior. It doesn't seem to trigger the error in a simulated volume. 🤷

packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
```diff
@@ -92,7 +92,18 @@ async function resolvePackageVersionsAsync(
...
-  const packageJsonPath = resolveFrom.silent(projectRoot, `${packageName}/package.json`);
+  let packageJsonPath: string | undefined;
+  try {
+    packageJsonPath = resolveFrom(projectRoot, `${packageName}/package.json`);
+  } catch (error: any) {
+    // This is a workaround for packages using `exports`. If this doesn't
+    // include `package.json`, we have to use the error message to get the location.
+    if (error.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED') {
+      packageJsonPath = error.message.match(/ in (.*)$/i)?.[1];
+    } else {
```
brentvatne, line 103: is this else block necessary? if `resolveFrom` throws an error then i believe `resolveFrom.silent` will return `undefined` for the same params. so we'd just be setting `packageJsonPath` to `undefined` here. i'm probably not thinking of some other case that this is meant to handle though 😅

byCedric: I somehow thought that `resolve-from` also handled another error, but that seems to be part of the lookup logic. You are right, should be able to just put null or undefined there.

byCedric:
76EE

So, if I understand our tests correctly:

```
// This
resolveFrom.silent(...)

// And this:
try {
  resolveFrom(...);
} catch (error) { }
```

Does not give us the same response. For non-silent, it exists early with "cannot find module" while silent just resolves it. Let me see if I can find out why.

byCedric:
Got it, was a partial mock. Basically what happened is that the non-silent version actually used `resolve-from`, while the silent one was using a mocked version. Only the mocked version has access to the memfs (fake) volume. 🤷

packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
```diff
@@ -73,4 +75,44 @@ describe(validateDependenciesVersionsAsync, () => {
...
+            default: './analytics/dist/index.esm.js',
+          },
+        },
+      }),
+    });
+
+    // Manually trigger the Node import error for "exports".
+    // This isn't triggered by memfs, or our mock, that's why we need to do it manually.
+    // see: https://github.com/expo/expo-cli/pull/3878
+    (resolveFrom as jest.MockedFunction<typeof resolveFrom>).mockImplementationOnce(() => {
```
byCedric, line 100: So this was a pain to implement, we have to trigger the warning manually because we mock the whole system. We use memfs and some mocked/fake version of resolve from. At least this fakes the real world now 😄  (it actually tests this fix now)

byCedric:
The CI failure on 32225ec does confirm we are actually having the same scenario IRL in the test, will commit the fix again to see if CI passes.

packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
```diff
@@ -73,4 +75,44 @@ describe(validateDependenciesVersionsAsync, () => {
...
+      const message = `Package subpath './package.json' is not defined by "exports" in ${packageJsonPath}`;
```
brentvatne, line 101: i really hope node doesn't change this error message

brentvatne (approved):

packages/expo-cli/src/commands/utils/__tests__/validateDependenciesVersions-test.ts
byCedric: 🤷

PR #3899, opened 2021-10-11
[cli][xdl][dev-client] Add interstitial disambiguation page

lukmccall:
# Why

Closes ENG-1937.

# How

Added interstitial disambiguation page when the dev-client package is installed, but `expo start` was invoked without the `dev-client` flag. In that case, we don't know what we should launch. So let's give users control and we launch the simple page, where they can choose in which app their project should be open. 

Note: this is not a final version of the disambiguation page. There is a couple of weirdness, which need to be addressed. For example, `expo-cli` always install the `Expo Go` application. We don't check if the dev-client app is installed on the device. That's why this feature is behind the `EXPO_ENABLE_INTERSTITIAL_PAGE` flag.

# Test Plan

https://user-images.githubusercontent.com/9578601/136762439-c8f1d8f9-4613-44e2-ac08-3163bbde47e4.mov


brentvatne:
@EvanBacon - any other concerns about shipping this?

packages/xdl/static/loading-page/index.html
```diff
@@ -0,0 +1,244 @@
...
...xwhF4Lnnn59tIGSsnoFy9Jc9TpXQjJgn1XOJzFP6nmuUdv7qLoCpfLo3nsgEzJP5GzNHjeK0IPJe88/Nt2wouV++1Zuvh9XQjdhT+h5+kPS04LvvPR7O7SFzI/rbyXDZrNzwXJPrF4XyFx3ydH8MeQlhr7oJziS5cf+W5Yk+WOB45uez9DJP8yZ7x24zuFf9j4eQRm1k4uWQ7ks9wjeXhuWBv7+J/zNUFJ0tB+/M2twLbq04peybN0eg9x6KPFDgK/7IQjoWT7wijCyeXQq8dyaOYlBHbi+IXkO3ox1xtj92MpD9vcyuwvepmrHkavKPROwr/05jvqhfHGX81NO7A6MIphdtb0FFMyojtRfELyHb0Y662x25G0p+3uRXYXnUz1jwN3tHoHYX/acx31Yuj8FdD4w6MLpxSuL0FHcWkjNheFL+AbEc/5mp77GYk/XmbW4HtVTdjzdPgHY3eUfifxnwPe9F7IwCgYzH1wzm6wBJSO6pHsUmROlqPLwJ+Eff8mFs57VXj4T8/2khEn2uht9nwbD7jt46j3yv+x2LzolbSRhZckupH9Sj+KVpH63FF4BKFv79iHleMLmVtbx86in6K/FH4UyyK1rHgipC4hN6Cy4X6UT0Kfx6xo/d4InAU/oebq94+dBT+NDdH4U+xcFv9MuWKPStib8GVwWhH9Sj8ZcQOyu1H4BJFH163V8vtx+UaFo7sQUfhTzPxrAt/CkP/sj95j0XISGw7jixQH7kd+ePNgh+1g3peBPrFfFtGt7P5PJufg3Qv6keRb2fBUfiX+PQ+72cYjwXLSGw79hZsHbUd+aPw1yN3jGyPwFH4t8fukpK9feQo/I3o3718Pg/waYQhDB2FvxehfcZ7C7au5Sj89dgcI5eKwFH4LxXZ83B7+8hR+CvxlTOkN5/45FH4GZ6j8DMSlz32Fmxd+1H467E5Ri4VgaPwXyqy5+H29pGj8DvxRdG/e+v05hvfPAq/Ds9R/HU0LtfuLVpfc7vwQ+a43O9H7qBui0C/6AN3Ppv7mbzN3uciNRLxo/CbbJDL+2/e+fjpzVe/dnrz7e8ehd+EJ+uOvhHIhKRzLGwbkbn+yMIeQ5ybieONw1hUHwPXWNG2npyfeXMZZ/UffURg6ywcxb6eP2/efuf05vPvnu7/x//p9OI73z39/6rYW0vPBRr9AAAAAElFTkSuQmCC"/>
```
ide, line 210: Remove the huge image string and the other placeholder info

packages/xdl/static/loading-page/index.html
```diff
@@ -0,0 +1,244 @@
...
+        </div>
+        <div class="info-box-details-record">
+          <p>Last commit</p>
+          <p>Jul 6, 2021, 4:51PM PT</p>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <div class="bottom-bar">
```
ide, line 236: I noticed this bottom bar can cover the other page content and if the viewport is not tall enough, it's not possible to scroll down to see all the content.

packages/xdl/src/Simulator.ts
```diff
@@ -766,6 +769,30 @@ export async function resolveApplicationIdAsync(projectRoot: string) {
...
+    process.env['EXPO_ENABLE_INTERSTITIAL_PAGE'] &&
+    !devClient &&
+    isDevClientPackageInstalled(projectRoot)
+  ) {
+    return UrlUtils.constructLoadingUrlAsync(projectRoot, 'localhost');
+  } else {
+    return await UrlUtils.constructDeepLinkAsync(projectRoot, {
+      hostType: 'localhost',
+      scheme,
+    }).catch(e => {
```
ide, line 787: try catch instead

packages/xdl/src/start/LoadingPageHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+          await deeplinkEndpointHandler(projectRoot, req, noCacheMiddleware(res));
+          break;
+        default:
+          next();
+      }
+    } catch (exception) {
+      res.statusCode = 520;
+      res.end(
+        JSON.stringify({
+          //@ts-ignore
```
ide, line 85:
```suggestion
          // @ts-ignore always explain why an error is being ignored
```

packages/xdl/static/loading-page/index.html
lukmccall: I'm not sure if I understand your intention here. The image string represents the actual image icon. I can remove that, however, it won't be easy to replace it with the actual `png` of the icon. I can't just add a static file, cause our server doesn't know how to serve static files. I don't think that is worth it for one icon.

packages/xdl/src/start/LoadingPageHandler.ts
```diff
@@ -80,12 +80,15 @@ export function getLoadingPageHandler(projectRoot: string) {
...
-        })
-      );
+      if (typeof exception == 'object' && exception != null) {
+        res.end(
+          JSON.stringify({
+            error: exception.toString(),
+          })
+        );
+      } else {
+        res.end('Unexpected error.');
```
ide, line 90:
```suggestion
        res.end(`Unexpected error: ${exception}`);
```

brentvatne (approved):

PR #3936, opened 2021-10-22
[expo-cli] Always upgrade expo package when prereleased

### LINKED ISSUE #3919: Upgrade between beta versions with "expo upgrade" not possible.
### Summary

Running `EXPO_BETA=1 expo upgrade` won't upgrade between beta versions (eg. beta-3 to beta-4)
You need to manually bump the version inside of package.json which doesn't give a seamless DX.

@brentvatne asked me on twitter to create an issue about that here.

byCedric:
# Why

Fixes #3919

The upgrade command doesn't upgrade the `expo` package during prerelease. It's disabled to avoid some issues at react-navigation (cc @satya164, do you know which one and/or if we can remove this check all together?). But, doing this causes issues like #3919. 

We could also remove the check, but I don't want to break the react-navigation repo. It would help installing newer patch versions though, these aren't installed atm.

# How

This adds an exception to that rule when the installed `expo` package is a prerelease (e.g. alpha/beta/rc versions).

Upgrading from `expo@43.0.0-beta.3` | Upgrading from `expo@43.0.0`
--- | ---
Screenshot 2021-10-22 at 15 15 14 |  Screenshot 2021-10-22 at 15 16 37

# Test Plan

- Initialize a new project
- Use the prerelease `expo` version
- Run `expo upgrade`
    - This should always reinstall `expo`, using the newer prerelease or stable version

### DIFF
--- a/packages/expo-cli/src/commands/info/upgradeAsync.ts
+++ b/packages/expo-cli/src/commands/info/upgradeAsync.ts
@@ -478,10 +478,16 @@ export async function upgradeAsync(
     ? `expo@next`
     : `expo@^${targetSdkVersionString}`;
 
+  const exactExpoPackageVersion = await getExactInstalledModuleVersionAsync('expo', projectRoot);
+  const isExpoPackageVersionPrerelease =
+    exactExpoPackageVersion && semver.prerelease(exactExpoPackageVersion) !== null;
+
   // Skip installing the Expo package again if it's already installed. @satya164
   // wanted this in order to work around an issue with yarn workspaces on
   // react-navigation.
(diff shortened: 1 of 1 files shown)


brentvatne:
looks like some tests were failing + there is a conflict. @byCedric - is this still relevant and if so can we land it before beta release?

byCedric:
Weird, I did not run into merge conflicts locally. Just rebasing on main did the trick.

As for the relevancy, I think this fix will be relevant for the beta period. As long as we use that special versioning and not just a major/minor/patch bump.

PR #3951, opened 2021-10-27
[config-plugins] Drop Android UserInterfaceStyle plugin in favor of template defaults

EvanBacon:
# Why

The template was already shipped with the required MainActivity and AndroidManifest requirements met, performing all updates as versioned changes via the [template now](https://github.com/expo/expo/pull/14929).

Because the features we add are specifically tied to making a feature `react-native` work, it should just exist as a requirement in the template and not a mandatory plugin that runs every time.


EvanBacon:
Closing in favor of https://github.com/expo/expo-cli/pull/3957

Issue #3937, opened 2021-10-22
Expo optimize does not make use of all CPU cores


Rohansi:
### Summary

Our app has ~800 images bundled and 99% of those are pulled from another location occasionally meaning we need to re-run expo optimize across all of them quite often where the current hash checking doesn't save us. It would be great if expo optimize made use of additional CPU cores so it can zip through large sets of images like this.

### Environment

  Expo CLI 4.12.8 environment info:
    System:
      OS: Windows 10 10.0.19043
    Binaries:
      Node: 14.17.6 - C:\Program Files\nodejs\node.EXE
      npm: 6.14.15 - C:\Program Files\nodejs\npm.CMD
    npmPackages:
      expo: ^40.0.0 => 40.0.1
      react: 16.13.1 => 16.13.1
      react-dom: 16.13.1 => 16.13.1
      react-native: https://github.com/expo/react-native/archive/sdk-40.0.1.tar.gz => 0.63.2
    Expo Workflow: managed

(I'm working on upgrading now I promise)

### Please specify your device/emulator/simulator platform, model and version

Android device, Windows dev PC

### Reproducible demo or steps to reproduce from a blank project

1. Add lots of images to your expo project
2. Run `npx expo-optimize`
3. Watch your CPU usage sit at <5% for a long time

mako-taco:
@Rohansi you can install my company's [fork](https://github.com/expo/expo-cli/pull/3956) of `expo-optimize` to get this behavior.  It is a lot faster.

kbrandwijk:
Seems implemented in https://github.com/expo/expo-cli/pull/3956.

PR #3956, opened 2021-10-29
[expo-cli] parallelize expo-optimize

mako-taco:
# Why

Addresses #3937

# How

Using `@supercharge/promise-pool`, we're able to run IO-bound tasks in parallel based on the number of available cores.

# Test Plan

I ran `expo-optimize` in my own project. It would be nice to get some fixtures and automated tests for this tool, though.


mako-taco:
image
things happening in parallel.  some logging info is lost, because the module being used to call `sharp` does not let us specify what to do with stdout.

brentvatne:
hey @mako-taco! thanks for this. it may be worth following up on the sharp api here to ensure we can control the output. you can find the sharp wrapper module here: https://github.com/expo/expo-cli/blob/master/packages/image-utils/src/sharp.ts

mako-taco:
@brentvatne I was mistaken, all the output was from this file.  I've batched it up and it looks better.

image

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: '.yarn/releases/yarn-1.23.0-20210726.1745.cjs'
```
brentvatne, line 1: can you remove this and the `.yarn/releases` file please?

packages/expo-optimize/package.json
```diff
@@ -36,6 +36,7 @@
     "@expo/config": "6.0.7",
     "@expo/image-utils": "0.3.17",
     "@expo/json-file": "8.2.33",
+    "@supercharge/promise-pool": "^1.8.0",
```
brentvatne, line 39: this dependency doesn't look particularly reliable and it's much larger than it needs to be. can you replace this with p-limit instead? https://www.npmjs.com/package/p-limit

brentvatne (changes requested):

.yarnrc.yml
mako-taco: sorry, `commit -am` got me.  you should consider checking in `.yarn/releases` and an rc file though, it makes development very difficult for those of us who have upgraded yarn to 2+.

packages/expo-optimize/src/assets.ts
```diff
@@ -140,18 +136,31 @@ function filterImages(files: string[], projectRoot: string) {
 }
 
 // Calculate SHA256 Checksum value of a file based on its contents
-function calculateHash(filePath: string): string {
-  const contents = readFileSync(filePath);
+async function calculateHash(filePath: string): Promise<string> {
+  const contents = await new Promise<Buffer>((resolve, reject) =>
```
EvanBacon, line 140: generally we use `await fs.promises.readFile()` since Node.js was dropped earlier this year.

EvanBacon (approved):
lgtm

PR #3965, opened 2021-11-01
[metro-config][exotic] Fix bug preventing babel.config.js from being used

EvanBacon:
# Why

Project `babel.config.js` was being ignored in exotic mode.


Slapbox:
Does the readme need to be updated?

> When enabled, exotic mode adds the following assumptions:
> [...]
> - .babelrc support is removed in favor of babel.config.js.

PR #3966, opened 2021-11-01
[expo-cli] Skip showing dev tools UI on startup for new users

EvanBacon:
# Why

Follow up to https://github.com/expo/expo-cli/pull/3207 (9 months ago).
Dev Tools UI has race conditions and many other issues where it hasn't kept up with Expo CLI / Terminal UI and therefore is not the most optimal thing for new users to see.
We added a PR a while ago to persist the user's current preference so this PR should only effect first time users.

# Test Plan

- My current setting should continue to work.
- Deleting `~/.expo/state.json` and rerunning `expo start` should not open the dev tools UI.


brentvatne (approved):
i agree with this but let's wait to see what others think before landing

PR #3976, opened 2021-11-04
[metro config] use project versioning

EvanBacon:
# Why

- https://github.com/expo/expo/issues/15035
- https://github.com/expo/expo/issues/15040

Use the installed version of expo/metro-config before falling back on the latest. Users can opt-in to using the latest by setting `sdkVersion: 'UNVERSIONED'` or installing the latest and using it in their `metro.config.js`. 

We also now show the version in the debug logs of `expo/metro-config` and warn when using UNVERSIONED (which can occur when using `expo-cli` in a project without `expo` installed).

I've also verified that monorepos work in the latest RN used by Expo CLI without the middleware or asset plugin.
Minimized chances of projects breaking by reverting changes unless the monorepo root can be found.
And fixed support for uuid package.


packages/xdl/src/start/startDevServerAsync.ts
```diff
@@ -66,10 +66,15 @@ export async function startDevServerAsync(
     options.maxWorkers = startOptions.maxWorkers;
   }
 
-  const { server, middleware, messageSocket } = await runMetroDevServerAsync(projectRoot, options);
   // TODO: reduce getConfig calls
   const projectConfig = getConfig(projectRoot, { skipSDKVersionRequirement: true });
 
+  // Use the unversioned metro config.
+  options.unversioned =
```
brentvatne, line 73: want to make it possible to opt in with env var also?

brentvatne: i guess opting in with metro.config.js manually is fine too

EvanBacon: maybe we'll add an env var in the future to make testing expo-cli easier, but for now this is fine.

brentvatne (approved):

PR #4001, opened 2021-11-13
Skip APNS entitlement by default unless expo-notifications is installed

EvanBacon:
# Why

Prevent extraneous entitlements from being added to projects. Developers with free Apple developer accounts who attempt to build onto a physical iOS devices will find that they don't have APNS setup by default so they won't be able to build projects. This PR should give them the ability to remove the APNS entitlement. This will be introducing a breaking change as projects will now need `expo-notifications` installed.

In a future revision, we'll also disable the entitlement if the package isn't autolinked.

# Test Plan

Tests cover all cases


packages/prebuild-config/src/plugins/unversioned/expo-notifications/withIosNotificationsEntitlement.ts
```diff
@@ -0,0 +1,33 @@
+import { ConfigPlugin, withEntitlementsPlist } from '@expo/config-plugins';
+import { ExpoConfig } from '@expo/config-types';
+import path from 'path';
+import resolveFrom from 'resolve-from';
+
+const debug = require('debug')('expo:prebuild-config:expo-notifications') as typeof console.log;
+
+export function isExpoNotificationsInstalled(projectRoot: string): string | null {
```
EvanBacon, line 8: This is already handled in versioned plugin, just remove

brentvatne (approved):

PR #4014, opened 2021-11-17
[expo-cli] install: use recommended versions from remote

Simek:
# Why

Fixes ENG-2445

# How

This PR adds fetch for the `versions` endpoint and uses the data to install the recommended version taken from `relatedPackages` for a used in the project SDK version.

I have also spotted that only length of `nativeModules` and `other` arrays is used for determining the log message content, so I have refactored them to the counters and I have also streamlined the `messages` array creation a bit.

# Test Plan

`yarn test` run in package directory and `yarn lint` run in workspace root did not yield any errors.

Also I have locally aliased the Expo CLI and have tested those changes "live".

# Preview

Screenshot 2021-11-17 at 12 48 46

Screenshot 2021-11-17 at 12 49 31


EvanBacon (changes requested):
maybe this should reuse the existing logic with `expo upgrade`. Regardless, needs tests.

Simek:
> maybe this should reuse the existing logic with `expo upgrade`.

Refs: 
* https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/commands/info/upgradeAsync.ts#L186-L223

My first idea was to use the `upgrade` code but of all the logic is not extracted to separate function and the whole logic and checks seems to be too complex for the simple install usage. 

I would rather separate getter and formatter of versions fetch for SDK and refactor the current `upgrade` code, so it will reuse the helpers. I will look at that, but I'm not sure if it would be in this PR. I also will add the requested tests soon.

packages/expo-cli/src/utils/fetchers.ts
```diff
@@ -0,0 +1,24 @@
+import { Versions } from 'xdl';
+
+export type DependencyList = Record<string, string>;
+
+export const getRemoteVersionsForSdk = async (sdkVersion?: string): Promise<DependencyList> => {
+  const { sdkVersions } = await Versions.versionsAsync({ skipCache: true });
+  if (sdkVersion && sdkVersions[sdkVersion]) {
```
EvanBacon, line 7: sdkVersion in sdkVersions

packages/expo-cli/src/utils/fetchers.ts
```diff
@@ -0,0 +1,24 @@
+import { Versions } from 'xdl';
```
EvanBacon, line 1: Change file name to reflect the only method exported from the file

EvanBacon (approved):

PR #4015, opened 2021-11-17
[workspace] bump and align Babel dependencies

Simek:
# Why

Currently many packages uses different version of Babel dependencies, this PR is an effort to clean this up.

# How

I have bumped all the `@babel` dependencies to the `7.16+` (latest minor) versions and added `@babel/core` as a dev dependency to the `next-adapter` to fix part of the warnings which appears on install:

Screenshot 2021-11-17 at 14 13 55

# Test Plan

I have run test in few packages and the result was promising, but I would like to check things again using CI.


packages/webpack-config/e2e/basic/package.json
```diff
@@ -10,7 +10,7 @@
     "resize-observer-polyfill": "^1.5.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.9.0"
+    "@babel/core": "^7.16.0"
```
EvanBacon, line 13: test fixture, this should match expo sdk project versioning

packages/webpack-config/package.json
```diff
@@ -40,7 +40,7 @@
     "url": "https://github.com/expo/expo-cli/issues"
   },
   "dependencies": {
-    "@babel/core": "7.9.0",
+    "@babel/core": "^7.16.0",
```
EvanBacon, line 43: Ensure this works with the version of babel-loader -> webpack@4

packages/expo-cli/e2e/fixtures/basic/package.json
```diff
@@ -14,7 +14,6 @@
       "react-native": "0.64.3"
     },
     "devDependencies": {
-      "@babel/core": "^7.12.9"
+      "@babel/core": "^7.16.0"
```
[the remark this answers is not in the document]
Simek: The fixtures are not part of the workspace, it was my bad changing that, sorry!

packages/webpack-config/package.json
Simek: I did not noticed any issues with Webpack 4 which is still used in this package. 

Do you have any exact test or check in mind which I should perform?

packages/webpack-config/e2e/custom-homepage/package.json
```diff
@@ -9,7 +9,7 @@
     "react-native-web": "0.17.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.12.9"
+    "@babel/core": "^7.16.0"
```
EvanBacon, line 12: This does not appear to align with any known version of Expo SDK, please revert https://github.com/expo/expo/blob/98255a180d4a4dc4dc39bb54449caaa241df7c58/templates/expo-template-bare-minimum/package.json#L22

packages/webpack-config/e2e/minimum/package.json
```diff
@@ -10,6 +10,6 @@
     "react-native-web": "0.17.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.12.9"
+    "@babel/core": "^7.16.0"
```
EvanBacon, line 13: Same as above, this is an e2e test.

packages/webpack-config/e2e/custom-homepage/package.json
Simek: Yeah, sorry for that, it looks like I have messed up the syncs. 🤦‍♂️

PR #4022, opened 2021-11-18
[xdl] Remove md extension from expo updates config fyi link

byCedric:
# Why

Expo FYI links should be without md extension, it's added by the redirect.

See https://github.com/expo/expo-fyi/pull/1

# How

Removed md extension

# Test Plan

Try opening fixed link.


jkhales (approved):
thanks!

PR #4023, opened 2021-11-19
[interstitial] Capture analytics event when user selects "custom client"

lukmccall:
# Why

Closes ENG-2268.

# How

- Added a listener that triggers every time user selects something on the interstitial page.
- Pipe `platform` value through interstitial page to obtain it in the listener. 

# Test Plan

I couldn't verify if the events are sent, but I replaced them with console logs. Everything seems to work.

### DIFF
--- a/packages/expo-cli/src/commands/start/startAsync.ts
+++ b/packages/expo-cli/src/commands/start/startAsync.ts
@@ -2,8 +2,9 @@ import { ConfigError, ExpoConfig, getConfig, isLegacyImportsEnabled } from '@exp
 import chalk from 'chalk';
 import path from 'path';
 import resolveFrom from 'resolve-from';
-import { Project, UnifiedAnalytics, UrlUtils, Versions } from 'xdl';
+import { LoadingPageHandler, Project, UnifiedAnalytics, UrlUtils, Versions } from 'xdl';
 
+import StatusEventEmitter from '../../analytics/StatusEventEmitter';
 import getDevClientProperties from '../../analytics/getDevClientProperties';
 import Log from '../../log';
 import { assertProjectHasExpoExtensionFilesAsync } from '../utils/deprecatedExtensionWarnings';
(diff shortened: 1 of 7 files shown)


esamelson (approved):

packages/expo-cli/src/commands/start/startAsync.ts
```diff
@@ -77,7 +78,29 @@ export async function actionAsync(projectRoot: string, options: NormalizedOption
...
+  LoadingPageHandler.setOnDeepLink(
+    async (projectRoot: string, isDevClient: boolean, platform: string | null) => {
+      if (!isDevClient) {
+        return;
+      }
 
+      const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+      StatusEventEmitter.once('deviceLogReceive', () => {
+        // Send the 'ready' event once the app is running in a device.
+        UnifiedAnalytics.logEvent('dev client interstitial page', {
```
tcdavis, line 90: This should still use the 'dev client start command' event

packages/expo-cli/src/commands/start/startAsync.ts
```diff
@@ -77,7 +78,29 @@ export async function actionAsync(projectRoot: string, options: NormalizedOption
...
+          status: 'ready',
+          platform,
+          ...getDevClientProperties(projectRoot, exp),
+        });
+      });
+
+      UnifiedAnalytics.logEvent('dev client interstitial page', {
```
tcdavis, line 97: same here

tcdavis (changes requested):
Sorry, we should have talked about what the event should be when scoping!

It will be much easier to look at if we use the same 'dev client start command' as we do with the flag. 
If you could add another property flow_used: interstitial | flag here and when calling the flag directly it will let us see as people switch over and track any issues that come up.

Issue #3979, opened 2021-11-05
Should error and refuse to start project with invalid name like "react-native" or "expo" or "react"


brentvatne:
### Summary

Projects with names like "react-native" will error out during bundling with not-so-useful error messages.

### Environment

```
  Expo CLI 4.12.12 environment info:
    System:
      OS: macOS 12.0.1
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
      Yarn: 1.22.10 - ~/.nvm/versions/node/v14.16.1/bin/yarn
      npm: 7.20.3 - ~/.nvm/versions/node/v14.16.1/bin/npm
      Watchman: 2021.06.07.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.10.1 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    IDEs:
      Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~43.0.0 => 43.0.1 
      react: 17.0.1 => 17.0.1 
      react-dom: 17.0.1 => 17.0.1 
      react-native: 0.64.2 => 0.64.2 
      react-native-web: 0.17.1 => 0.17.1 
    npmGlobalPackages:
      expo-cli: 4.12.12
    Expo Workflow: managed
```

### Please specify your device/emulator/simulator platform, model and version

n/a

### Reproducible demo or steps to reproduce from a blank project

Try creating a project: `expo init react-native` and running it, you will see an error like this:

<img width="1391" alt="Screenshot_2021-11-05_at_17 02 49" src="https://user-images.githubusercontent.com/90494/140544423-375115f5-bd03-4f88-9b3e-2df11098ed66.png">

Aryojaam:
I can confirm seeing this error after initializing and starting the expo server. @EvanBacon are you currently working on this issue? I would like to have a go at preventing the problem if possible.

EvanBacon:
sure you can give it a try

Aryojaam:
@EvanBacon I have created a draft PR for this problem that shows my approach for fixing this issue. Would you mind reviewing it and give me some feedback if possible? Thanks in advance :)

Aryojaam:
[...]

PR #4025, opened 2021-11-21
Forbidden project names

Aryojaam:
# Why
This pull request resolves the following issue:
https://github.com/expo/expo-cli/issues/3979

If project names such as `react-native` or `expo` is fed to the command  `expo init <project_name>`, it leads to the project being correctly initialized however crashing upon starting up. These edge cases should be caught and handled. 

# How
There is an array of blacklisted project names. Upon initializing the project, it is checked if the project name is included in this array. If so, the user is stopped from creating the project with that name.  

Here is how the result looks like:
Screenshot from 2021-11-21 15-51-02
# Test Plan
For testing using the command line simply run `node ./packages/expo-cli/bin/expo.js init react` and the error should pop up. Make sure the project is built and running in the background using `yarn build` and `yarn start` respectively.  

There is also an E2E test case for this fix that checks if the command has failed by checking if `stdout` is empty in comparison to a passing state in which the `stdout` is `Your project is ready!`. However currently a warning gets thrown on executing the test. The warning is present in the following screenshot:
expotest

### DIFF
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -64,6 +64,10 @@ function assertValidName(folderName: string) {
       `Cannot create an app named ${chalk.red(`"${folderName}"`)}. ${validation}`
     );
   }
+  const isFolderNameBlacklisted = CreateApp.isFolderNameBlacklisted(folderName);
+  if (isFolderNameBlacklisted) {
+    throw new CommandError(`Cannot create an app named ${chalk.red(`"${folderName}"`)}.`);
+  }
 }
 
 function parseOptions(command: Partial<Options>): Options {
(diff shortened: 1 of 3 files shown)


axeldelafosse:
Would be cool to avoid the term "blacklist" :)

Aryojaam:
Sure, I will refactor that :)

packages/expo-cli/src/commands/initAsync.ts
```diff
@@ -64,6 +64,10 @@ function assertValidName(folderName: string) {
       `Cannot create an app named ${chalk.red(`"${folderName}"`)}. ${validation}`
     );
   }
+  const isFolderNameForbidden = CreateApp.isFolderNameForbidden(folderName);
+  if (isFolderNameForbidden) {
+    throw new CommandError(`Cannot create an app named ${chalk.red(`"${folderName}"`)}.`);
```
EvanBacon, line 69:
```suggestion
    throw new CommandError(`Cannot create an app named ${chalk.red(`"${folderName}"`)} because it would conflict with a dependency of the same name.`);
```

EvanBacon (approved):
lgtm

Aryojaam:
Cool! Thanks! I have committed your suggestion :)

PR #4026, opened 2021-11-22
[prebuild-config] Deprecate expo-branch plugin

EvanBacon:
# Why

In the future we will be dropping `expo-branch` in favor of `react-native-branch` (after we add a config plugin to RNBranch).

# How

This PR removes the half support for expo-branch in favor of warning users to setup react-native-branch. It's unclear at this time if we'll need additional support for RNBranch.

# Test Plan

Updated tests


EvanBacon:
Closing, will reopen when versioned CLI lands

PR #4032, opened 2021-11-24
[expo-cli] check update URL is not pointing at EAS

jkhales:
# Why

We want to avoid users trying to use EAS Update from accidentally using `expo publish` 

# How

Check the update URL is not pointed at an EAS url.

# Test Plan
tried on demo repo.


EvanBacon (approved):
not familiar with this code, looks fine

PR #4043, opened 2021-12-01
[config-plugins] use new error recovery setting in config plugin for updates 0.11.x

esamelson:
# Why

expo-updates 0.11.x will have a new native config setting that better matches the description & intended behavior of app.json's `updates.checkAutomatically: 'ON_ERROR_RECOVERY'` (which previously was only implemented in Expo Go-derived shell apps). We should map to this new config setting when it's available.

# How

Read expo-updates version number from package.json, and if it can be found and is >= 0.11.0, then use the new native setting; otherwise, default to the old `NEVER` setting (which doesn't have a perfect analog in app.json).

# Test Plan

Added a few test cases, all pass. Also tested manually with `expo prebuild --clean --no-install` / local `@expo/config-plugins`:
✅ expo-updates@0.10.15 `checkAutomatically: 'ON_ERROR_RECOVERY'` -> `NEVER`
✅ expo-updates@0.11.0 `checkAutomatically: 'ON_ERROR_RECOVERY'` -> `ERROR_RECOVERY_ONLY` (modified version number in node_modules/expo-updates/package.json)
✅ expo-updates@0.10.15 `checkAutomatically: 'ON_LOAD'` -> `ALWAYS`
✅ expo-updates@0.11.0 `checkAutomatically: 'ON_LOAD'` -> `ALWAYS`

### DIFF
--- a/packages/config-plugins/src/ios/Updates.ts
+++ b/packages/config-plugins/src/ios/Updates.ts
@@ -1,10 +1,16 @@
 import * as path from 'path';
 import resolveFrom from 'resolve-from';
+import semver from 'semver';
 import xcode from 'xcode';
 
 import { ConfigPlugin } from '../Plugin.types';
 import { withExpoPlist } from '../plugins/ios-plugins';
-import { ExpoConfigUpdates, getRuntimeVersionNullable, getUpdateUrl } from '../utils/Updates';
+import {
+  ExpoConfigUpdates,
+  getExpoUpdatesPackageVersion,
+  getRuntimeVersionNullable,
+  getUpdateUrl,
+} from '../utils/Updates';
 import { ExpoPlist } from './IosConfig.types';
 
 const CREATE_MANIFEST_IOS_PATH = 'expo-updates/scripts/create-manifest-ios.sh';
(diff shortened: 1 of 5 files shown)


packages/config-plugins/src/ios/Updates.ts
```diff
@@ -48,20 +59,27 @@ export const withUpdates: ConfigPlugin<{ expoUsername: string | null }> = (
   { expoUsername }
 ) => {
   return withExpoPlist(config, config => {
-    config.modResults = setUpdatesConfig(config, config.modResults, expoUsername);
+    const expoUpdatesPackageVersion = getExpoUpdatesPackageVersion(config.modRequest.projectRoot);
```
jkhales, line 62: Since we are getting the `projectRoot` from an attribute of `config`, can we move `getExpoUpdatesPackageVersion` inside of `getUpdatesCheckOnLaunch`?

jkhales (approved):
😎

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -1,14 +1,30 @@
...
 import { AndroidConfig, IOSConfig } from '..';
+import { resolvePackageRootFolder } from './resolvePackageRootFolder';
 
 export type ExpoConfigUpdates = Pick<
   ExpoConfig,
   'sdkVersion' | 'owner' | 'runtimeVersion' | 'updates' | 'slug'
 >;
 
+export function getExpoUpdatesPackageVersion(projectRoot: string): string | null {
+  const expoUpdatesRootFolder = resolvePackageRootFolder(projectRoot, 'expo-updates');
```
EvanBacon, line 16: `resolvePackageRootFolder` will be deleted.

```suggestion
  let expoUpdatesRootFolder = resolveFrom.silent(projectRoot, 'expo-updates/package.json');
  if (expoUpdatesRootFolder) expoUpdatesRootFolder = path.dirname(expoUpdatesRootFolder);
```

EvanBacon (approved):
mostly looks fine, we should investigate finishing a flow where expo works without expo-updates, then we can remove the unversioned copy and by extension, version checks.

packages/config-plugins/src/ios/Updates.ts
esamelson: Having it as a separate parameter makes the unit tests a lot nicer since we can easily check the different cases without having to mock the filesystem.

PR #4045, opened 2021-12-01
[config-plugins] add missing sdk-runtime-versions dep

jkhales:
# Why

`@expo/sdk-runtime-version` is used here: https://github.com/expo/expo-cli/blob/main/packages/config-plugins/src/utils/Updates.ts#L2

But was missing from our dependencies.

### DIFF
--- a/packages/config-plugins/package.json
+++ b/packages/config-plugins/package.json
@@ -35,6 +35,7 @@
     "@expo/config-types": "^43.0.1",
     "@expo/json-file": "8.2.34",
     "@expo/plist": "0.0.16",
+    "@expo/sdk-runtime-versions": "^1.0.0",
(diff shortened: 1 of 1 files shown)


packages/config-plugins/src/ios/ProvisioningProfile.ts
```diff
@@ -33,21 +33,30 @@ export function setProvisioningProfileForPbxproj(
     : findFirstNativeTarget(project);
   const [nativeTargetId, nativeTarget] = nativeTargetEntry;
 
+  const quotedAppleTeamId = ensureQuotes(appleTeamId);
```
jkhales, line 36: seems like the inclusion of these files are from some weird main/master rebase issue. reverting...

dsokal (approved):

PR #4049, opened 2021-12-03
[xdl] anonymize events

kgc00:
# Why

We'd like to remove any personal data from incoming events.

# How

Removed properties such as username and project root from events.

Edit: will make a follow-up PR to handle Project Warning at some point.

# Test Plan

fired some events, made sure they did not contain the data.

### DIFF
--- a/packages/xdl/src/User.ts
+++ b/packages/xdl/src/User.ts
@@ -403,7 +403,6 @@ export class UserManagerInstance {
     if (this._currentUser && !this._currentUser?.accessToken) {
       Analytics.logEvent('Logout', {
         userId: this._currentUser.userId,
-        username: this._currentUser.username,
         currentConnection: this._currentUser.currentConnection,
       });
     }
@@ -495,7 +494,6 @@ export class UserManagerInstance {
         Analytics.logEvent('Login', {
           userId: user.userId,
           currentConnection: user.currentConnection,
-          username: user.username,
         });
       }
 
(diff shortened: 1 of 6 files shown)


brentvatne (approved):

wkozyra95:
I don't have anything against those changes, but I also do not see any reason to make them. Events still contain userId which can be used to access all that info you are redacting. The only difference is that those values won't be available directly from analytics. If you want to anonymize  data, userId(and any other id should be hashed)

> I'm also removing the stack trace in Project Warning events (do we need it?), but if we want to preserve for some reason, lemme know and i can rework this

I don't see any changes like that in this PR, so I assume you are talking about error events in eas-cli? If that is the case I'll respond to your DM in slack

wkozyra95 (approved):

kgc00:
> I don't have anything against those changes, but I also do not see any reason to make them. Events still contain userId which can be used to access all that info you are redacting. The only difference is that those values won't be available directly from analytics. If you want to anonymize data, userId(and any other id should be hashed)
> 
> > I'm also removing the stack trace in Project Warning events (do we need it?), but if we want to preserve for some reason, lemme know and i can rework this
> 
> I don't see any changes like that in this PR, so I assume you are talking about error events in eas-cli? If that is the case I'll respond to your DM in slack

The change was removing the entire Project Warning event - sorry, my description was a bit out of date and inaccurate. it seemed useless to include a project warning event if it contained no data (e.g. warning, project root).

PR #4054, opened 2021-12-07
[expo-cli] Server modern manifest type with EAS Update updates.url

jonsamp:
# Why

Inspired by https://github.com/expo/expo-cli/pull/3890. Closes ENG-2607.

If a developer has an `updates.url` that matches "https://u.expo.dev" or "https://staging-u.expo.dev", and they also have not specified a `--force-manifest-type` flag when running `expo start`, we should serve the modern manifest format.

This makes is so that developers previewing EAS Update do not have to run `expo start --force-manifest-type=expo-updates` when running their project locally.

# How

Removes logic that detected the presence of a project id, and adds logic to choose the `"expo-updates"` format when no format is present and the URL matches the EAS Update URL pattern.

# Test Plan

Make sure that you can run:

- `expo start --force-manifest-type=expo-updates` and that you get the new manifest format
- `expo start --force-manifest-type=classic` and that you get the classic manifest format
- `expo start` and that you get the classic manifest format
- `expo start` after running `eas update:configure` and that you get the modern manifest format

### DIFF
--- a/packages/expo-cli/src/commands/start/parseStartOptions.ts
+++ b/packages/expo-cli/src/commands/start/parseStartOptions.ts
@@ -183,7 +183,11 @@ export function parseStartOptions(
         ? 'expo-updates'
         : undefined;
   } else {
-    startOpts.forceManifestType = 'classic';
+    const easUpdatesUrlRegex = /^https:\/\/(staging-)?u\.expo\.dev/;
+    const updatesUrl = exp.updates?.url ?? '';
+    const isEasUpdatesUrl = easUpdatesUrlRegex.test(updatesUrl);
+
+    startOpts.forceManifestType = isEasUpdatesUrl ? 'expo-updates' : 'classic';
   }
 
   if (isLegacyImportsEnabled(exp)) {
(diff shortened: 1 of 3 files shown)


packages/expo-cli/src/commands/start/parseStartOptions.ts
```diff
@@ -183,7 +183,11 @@ export function parseStartOptions(
         ? 'expo-updates'
         : undefined;
   } else {
-    startOpts.forceManifestType = 'classic';
+    const easUpdatesUrlRegex = /^https:\/\/(staging-)?u\.expo\.dev/;
+    const updatesUrl = exp.updates?.url ?? '';
+    const isEasUpdatesUrl = easUpdatesUrlRegex.test(updatesUrl);
```
jkhales, line 188:
```suggestion
    const updatesUrl = exp.updates?.url;
    const isEasUpdatesUrl = updatesUrl && easUpdatesUrlRegex.test(updatesUrl);
```

jkhales (approved):
😎

PR #4057, opened 2021-12-09
[schemer] replace few Lodash methods with native code

Simek:
# Why

This is first PR from the upcoming series in which I would like to focus on removing Lodash methods from the CLI codebase.

The idea is to keep the changeset small so the potential issues can be easily caught or even revert if needed.

# How

Replace `fill` and `flatten` methods by the native counterparts.

# Test Plan

`tsc` compilation and `fieldPathToSchemaPath` tests were successful.


EvanBacon (approved):
lgtm

PR #4058, opened 2021-12-10
[next-adapter] fix incorrect detection of webpack 5

Simek:
# Why

I was working on the `with-nextjs` Expo example update to use Next 10+, when I spotted that besides Next was using Webpack 4 for the compilation, Next was failing with an Internal Error pointing out to the code guarded by `isWebpack5` flag.

Screenshot 2021-12-10 at 13 12 56

After an investigation in code, it looks like the `webpack5` flag was set incorrectly a level up, in `next-adapter`, when there was no `webpack5` entry in `args`.

Screenshot 2021-12-10 at 13 12 36

# How

This small change fixes the incorrect Webpack 5 detection and allows in the future update of `with-nextjs` example (after the new version of package including fix will be released).

# Test Plan

I have init app using `npx create-react-native-app -t with-nextjs`, bumped the Next dependency to `~10.2.3` and `next-adapter` to latest version, then I have copied in the updated build files of `next-adapter` to test app `node_modules` which fixed the Internal Server Error.

### DIFF
--- a/packages/next-adapter/src/withExpo.ts
+++ b/packages/next-adapter/src/withExpo.ts
@@ -19,7 +19,7 @@ export default function withExpo(nextConfig: any = {}): any {
         },
         {
           supportsFontLoading: false,
-          webpack5: webpack5 !== false,
+          webpack5: webpack5 && webpack5 !== false,
         }
       );
       // Use original public path


brentvatne (approved):

marklawlor:
@brentvatne This PR did the opposite of its intention and actually broke `withExpo` by causing it to default to `webpack5: undefined` instead of `webpack5: true`

Previously, if webpack5 was not declared in the `next.config.js` the variable `webpack5 === undefined`. As `undefined !== false` it would pass `webpack5: true` to withUniModules.

Now, if webpack5 is not declared in the `next.config.js` the variable `webpack5 === undefined`. But now `(webpack5 && webpack !== false) === undefined` so it will pass `webpack5: undefined` to the `withUnimodules` function.

@Simek Your issue was caused by providing a custom webpack4 configuration in your `next.config.js`.

Reverting this change will fix 
- https://github.com/expo/expo-cli/issues/4066
- https://github.com/expo/expo-cli/issues/4066

PR #4065, opened 2021-12-13
[xdl] adds message scrubber

kgc00:
# Why

Added a message scrubber to keep things like system paths and account names out of the logger warnings that we collect to monitor and ensure CLI health.

# How

Created a `MessageScrubber` class which contains a `scrubMethod` function. Went with a class-based approach because there's a bunch of logic that I wanted to keep together, and there's a number of private variables and RegExps that I wanted to cache for subsequent usage.

lmk if there are any performance concerns. it's a pretty beefy regex and i may be able to trim down the # of steps if folks feel it would be a good idea.

# Test Plan

Added a bunch of tests to make sure file paths get removed, account names get removed, general messages do not get altered.

### DIFF
--- a/packages/xdl/src/project/MessageScrubber.ts
+++ b/packages/xdl/src/project/MessageScrubber.ts
@@ -1,0 +1,41 @@
+class MessageScrubber {
+  private matchANSIEscapeCodesRegex: RegExp;
+  private matchFilePathRegex: RegExp;
+  private matchNewLineRegex: RegExp;
+  private matchAccountNamesRegex: RegExp;
+
+  constructor() {
+    const matchANSIEscapeCodes = `\\x1B|\\[[0-9]{1,2}m`;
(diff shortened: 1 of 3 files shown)


wkozyra95 (commented):
I don't think this is a good approach

It does not fully fix the problem:
- it's very likely that there will be some errors that we missed here
- there is no way to update it if it's done client-side

I'm assuming that the goal here is to make sure that we are not keeping any data after someone has deleted their account to comply with GDPR and other privacy laws. If that is the case I think the approach here should be different, 
- normal analytics event should not contain any generic values/messages that were taken from cli logs, it should consist only of known values we are explicitly defining
- all the generic messages/warnings/errors basically all the data that can potentially contain any PII should go-to destination with a short retention time e.g. 30 days. Using sentry for that would probably be easiest, plus they already have a mechanism to remove sensitive info

PR #4071, opened 2021-12-14
[expo-cli] `expo config --json`

andrejpavlovic:
# Why

I switched from using `app.json` to `app.config.ts` for configuring my Expo instance. Previously in my build scripts I would import `app.json`  to get a few relevant values such as `owner` and `slug`, but it's not too clear to me how I can do this now.

I was hoping to retrieve the config values by running `expo config` in the command line and parsing the output as JSON, but the output is not valid JSON and is also prefixed with timestamps when running inside a script:

```javascript
// test.js
console.log(require('child_process').execSync(`expo config`,{
  encoding: 'utf-8',
}))

// output when executing `node test.js`
[16:41:02]
[16:41:02] {
  name: 'test',
  slug: 'test',
...
```

# How

To make it easier to process output from `expo config` I've added support for `--json` flag. The flag will output the config values as a valid JSON on a single line without a prefix, making it easy to consume in other processes.

```javascript
// test.js
console.log(require('child_process').execSync(`expo config --json`,{
  encoding: 'utf-8',
}))

// output when executing `node test.js`
{"_internal":{"isDebug":false,"projectRoot":"\some\root\","dynamicConfigPath":"\some\root\app.config.ts","staticConfigPath":null,"packageJsonPath" ...
```

# Test Plan

I've ran and compared the output of the following command to ensure the data is the same, and that the values differ only by formatting:
`expo config`  <=> `expo config --json`
`expo config --full` <=> `expo config --json --full`


andrejpavlovic:
Two workaround I've thought about for now:
1. Instead of parsing the config, set the values in the config using environment values. This way we can directly access the environment values instead of parsing the config. This may not be an option in every case.

1. Or try to parse it anyway in a very unsafe way:
```javascript
const manifest = require('child_process').execSync(`expo config`,{
  encoding: 'utf-8',
});
const manifestWithoutPrefixes = manifest.split('\n').map(line => line.replace(/^\[.*?\]\s*/, '')).filter(line => !!line).join('\n')
const manifestObject = eval('`' + manifestWithoutPrefixes + '`');
```

brentvatne:
you can use `getConfig` from `@expo/config` for this - it's ultimately what `expo config` runs to produce the output. https://github.com/expo/expo-cli/blob/c24263b4df40c4d592ff0c5b7b904f261f9a949a/packages/expo-cli/src/commands/info/config/configAsync.ts

packages/expo-cli/src/commands/info/config/configAsync.ts
```diff
@@ -49,7 +50,13 @@ export async function actionAsync(projectRoot: string, options: Options) {
...
-  logConfig(options.full ? config : config.exp);
-  Log.log();
+  const configOutput = options.full ? config : config.exp;
+
+  if (!options.json) {
+    Log.log();
+    logConfig(configOutput);
+    Log.log();
+  } else {
+    Log.nested(JSON.stringify(configOutput));
```
EvanBacon, line 60:
```suggestion
    Log.nested(JSON.stringify(configOutput, null, 2));
```

EvanBacon (approved):

packages/expo-cli/src/commands/info/config/configAsync.ts
andrejpavlovic: I was thinking about maybe formatting the output the way you suggested, but that's what we already get without passing in `--json`. Compact json format made more sense to me.

Issue #4051, opened 2021-12-05
yarn pod-install not retrying with repo-update when it could resolve problem by updating


mikehardy:
### Summary

When pod repo is out of date and pod install fails with string `pod repo update` in pod error, pod-install does not detect it and just hard fails instead of attempting to fix.

I'm aware of the work here - I took a quick spelunk through the code to see if I could snipe it -
https://github.com/expo/expo-cli/pull/3399/files#diff-81dde81e701f92cc29f402f4c34fc9d752a1dafcf20741a87034333f5b2b6130R207

But for some reason stdout and stderr were empty strings

### Environment

macos big sur
pod 1.11.2
yarn 1.22.17
all else as from the template cloned out

### Please specify your device/emulator/simulator platform, model and version

it's a CLI thing, just pod install

### Error output

You can see empty stdout/stderr strings, so there's nothing to look in to find `pod repo update` string, even though the error from cocoapods is clearly showing it on console prior to the failure.

Looks like a failure to capture streams somehow?

```
$ pod-install --repo-update
Scanning for pods...
1.11.2
> pod install
RNFBAnalytics: Using default Firebase/Analytics with Ad Ids. May require App Tracking Transparency. Not allowed for Kids apps.
RNFBAnalytics: You may set variable `$RNFirebaseAnalyticsWithoutAdIdSupport=true` in Podfile to use analytics without ad ids.
Using firebase.json from '/Users/mike/work/Invertase/react-native-firebase-authentication-example/template/firebase.json'
Adding a custom script phase for Pod RNFBApp: [RNFB] Core Configuration
Adding a custom script phase for Pod RNFBCrashlytics: [RNFB] Crashlytics Configuration
Auto-linking React Native modules for target `ProjectName`: RNAppleAuthentication, RNCMaskedView, RNFBAnalytics, RNFBApp, RNFBAuth, RNFBCrashlytics, RNFBPerf, RNGestureHandler, RNGoogleSignin, RNLocalize, RNScreens, RNVectorIcons, react-native-fbsdk-next, react-native-pager-view, and react-native-safe-area-context
Analyzing dependencies
Fetching podspec for `DoubleConversion` from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`
Fetching podspec for `RCT-Folly` from `../node_modules/react-native/third-party-podspecs/RCT-Folly.podspec`
[...]

PR #4084, opened 2021-12-21
Fix `pod repo update` not automatically running

EvanBacon:
# Why

- Resolve https://github.com/expo/expo-cli/issues/4051

# How

When the 'silent' feature was added it broke the `output` collection of forked processes, meaning we'd match for 'pod repo update' against an empty string. This is a bit tricky since the code in Expo CLI is often silent but the usage in `npx pod-install` is verbose.

This PR also adds support for automatically running `pod update [package]` which can often be more useful than `pod repo update` since it will fix broken versions. If `pod update X` doesn't work, we'll fallback on `pod repo update` like before.

An example of this new feature working would be to open your `Podfile.lock`:

```diff
PODS:
-  - boost-for-react-native (1.63.0)
+  - boost-for-react-native (420.63.0)
  - DoubleConversion (1.1.6)
  - EXApplication (4.0.0):
```

Running `pod-install` (or similar commands) will automatically fix this version without running `pod repo update` (which wouldn't work).

# Test Plan

Added many more tests to hopefully prevent this from breaking again.

### DIFF
--- a/packages/package-manager/src/CocoaPodsPackageManager.ts
+++ b/packages/package-manager/src/CocoaPodsPackageManager.ts
@@ -360,16 +360,24 @@ export class CocoaPodsPackageManager implements PackageManager {
     if (!this.silent) {
       console.log(`> pod ${args.join(' ')}`);
     }
-    const promise = spawnAsync('pod', [...args], {
-      // Add the cwd and other options to the spawn options.
-      ...this.options,
-      // We use pipe by default instead of inherit so that we can capture stderr/stdout and process it for errors.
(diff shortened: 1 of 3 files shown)


Kudo (approved):
LGTM ✅  although it's hard to review but the tests look promising 🔥

mikehardy:
I'm excited by this! The way react-native-firebase handles firebase-ios-sdk pod versions means that a repo update and/or pod update is frequently needed, and pod-install should be handy now that it's repaired. Thanks for fixing it

PR #4087, opened 2021-12-21
[xdl] persist dev client ids locally, use to start dev sessions

esamelson:
# Why

closes ENG-1987, closes ENG-1988

In order to make it easier for dev client users to find their development sessions before logging in, we're adding an installation ID flow similar to snack's device ID flow. We generate and persist an ID on the client, and pass it to expo-cli in the first manifest request (via QR code or other method) so expo-cli can persist it locally and include it in future calls to `development-sessions/notify-alive`. This PR adds those two pieces of functionality to expo-cli.

# How

- read a device ID from a manifest request and persist to disk (in `.expo/devices.json` as of now -- but definitely open to renaming/moving this)
- prune old IDs so there is a max of 10 persisted at any one time
- also prune any IDs that haven't been used in 30 days or longer
- start a development session for each device ID if there are any when starting sessions

# Test Plan

Added unit tests for storage and pruning behavior. Also tested manually with dev client from master and local www:
- [x] request manifest with `expo-dev-client-id` header in request -> id is saved to project's `.expo/devices.json` file
- [x] run `expo start --dev-client` with multiple ids in `.expo/devices.json` -> open http://localhost:3000/--/api/v2/development-sessions?deviceId=${id} for each id to verify a session has been started for each one

### DIFF
--- a/packages/xdl/src/DevSession.ts
+++ b/packages/xdl/src/DevSession.ts
@@ -1,6 +1,8 @@
 import { ExpoConfig } from '@expo/config-types';
 import os from 'os';
+import { URLSearchParams } from 'url';
 
+import { ProjectSettings } from '.';
 import {
   ApiV2 as ApiV2Client,
   ConnectionStatus,
@@ -25,12 +27,12 @@ export async function startSession(
   }
 
   if (!ConnectionStatus.isOffline() && keepUpdating) {
-    // TODO(anp) if the user has configured device ids, then notify for those too
     const authSession = await UserManager.getSessionAsync();
+    const { devices } = await ProjectSettings.readDevicesInfoAsync(projectRoot);
 
-    if (!authSession) {
+    if (!authSession && (!devices || !devices.length)) {
       // NOTE(brentvatne) let's just bail out in this case for now
-      // throw new Error('development sessions can only be initiated for logged in users');
+      // throw new Error('development sessions can only be initiated for logged in users or with a device ID');
       return;
     }
 
@@ -44,8 +46,17 @@ export async function startSession(
         throw new Error(`Unsupported platform: ${platform}`);
       }
 
+      let queryString = '';
+      if (devices) {
+        const searchParams = new URLSearchParams();
+        devices.forEach(device => {
+          searchParams.append('deviceId', device.installationId);
+        });
+        queryString = `?${searchParams.toString()}`;
+      }
+
       const apiClient = ApiV2Client.clientForUser(authSession);
-      await apiClient.postAsync('development-sessions/notify-alive', {
+      await apiClient.postAsync(`development-sessions/notify-alive${queryString}`, {
         data: {
           session: {
             description: `${exp.name} on ${os.hostname()}`,
--- a/packages/xdl/src/DevSession.ts
+++ b/packages/xdl/src/DevSession.ts
@@ -2,11 +2,11 @@ import { ExpoConfig } from '@expo/config-types';
 import os from 'os';
 import { URLSearchParams } from 'url';
 
-import { ProjectSettings } from '.';
 import {
   ApiV2 as ApiV2Client,
   ConnectionStatus,
   Logger as logger,
+  ProjectSettings,
   UrlUtils,
   UserManager,
 } from './internal';
--- a/packages/xdl/src/DevSession.ts
+++ b/packages/xdl/src/DevSession.ts
@@ -28,7 +28,7 @@ export async function startSession(
 
   if (!ConnectionStatus.isOffline() && keepUpdating) {
     const authSession = await UserManager.getSessionAsync();
-    const { devices } = await ProjectSettings.readDevicesInfoAsync(projectRoot);
+    const { devices } = await ProjectSettings.getDevicesInfoAsync(projectRoot);
 
     if (!authSession && !devices?.length) {
       // NOTE(brentvatne) let's just bail out in this case for now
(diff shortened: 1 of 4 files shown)


packages/xdl/src/DevSession.ts
```diff
@@ -1,6 +1,8 @@
 import { ExpoConfig } from '@expo/config-types';
 import os from 'os';
+import { URLSearchParams } from 'url';
 
+import { ProjectSettings } from '.';
```
EvanBacon, line 5: I think these need to be imported from './internal' or something like that, not entirely sure why.

packages/xdl/src/ProjectSettings.ts
```diff
@@ -135,6 +148,64 @@ export async function setPackagerInfoAsync(
...
+  try {
+    return await devicesJsonFile(projectRoot).mergeAsync(json, {
+      cantReadFileDefault: { devices: [] },
+    });
+  } catch (e) {
+    return await devicesJsonFile(projectRoot).writeAsync(json);
+  }
+}
+
+const MILLISECONDS_IN_30_DAYS = 30 * 24 * 60 * 60 * 1000;
```
EvanBacon, line 182: Move to top

packages/xdl/src/ProjectSettings.ts
```diff
@@ -165,6 +236,10 @@ The "packager-info.json" file contains port numbers and process PIDs that are us
 
 The "settings.json" file contains the server configuration that is used to serve the application manifest.
 
+> What does the "devices.json" file contain?
```
EvanBacon, line 239:
```suggestion
> What is `devices.json`?
```

packages/xdl/src/DevSession.ts
```diff
@@ -25,12 +27,12 @@ export async function startSession(
   }
 
   if (!ConnectionStatus.isOffline() && keepUpdating) {
-    // TODO(anp) if the user has configured device ids, then notify for those too
     const authSession = await UserManager.getSessionAsync();
+    const { devices } = await ProjectSettings.readDevicesInfoAsync(projectRoot);
 
-    if (!authSession) {
+    if (!authSession && (!devices || !devices.length)) {
```
EvanBacon, line 33:
```suggestion
    if (!authSession && !devices?.length) {
```

packages/xdl/src/ProjectSettings.ts
```diff
@@ -135,6 +148,64 @@ export async function setPackagerInfoAsync(
...
+  projectRoot: string,
+  json: DevicesInfo
+): Promise<DevicesInfo> {
+  devicesInfo = json;
+
+  try {
+    return await devicesJsonFile(projectRoot).mergeAsync(json, {
+      cantReadFileDefault: { devices: [] },
+    });
+  } catch (e) {
```
EvanBacon, line 177:
```suggestion
  } catch {
```

packages/xdl/src/ProjectSettings.ts
```diff
@@ -135,6 +148,64 @@ export async function setPackagerInfoAsync(
...
+export async function readDevicesInfoAsync(projectRoot: string): Promise<DevicesInfo> {
+  if (devicesInfo) {
+    return devicesInfo;
+  }
+
+  try {
+    return await devicesJsonFile(projectRoot).readAsync({
+      cantReadFileDefault: { devices: [] },
+    });
+  } catch (e) {
```
EvanBacon, line 162:
```suggestion
  } catch {
```

packages/xdl/src/ProjectSettings.ts
```diff
@@ -165,6 +237,10 @@ The "packager-info.json" file contains port numbers and process PIDs that are us
 
 The "settings.json" file contains the server configuration that is used to serve the application manifest.
 
+> What is \`devices.json\`?
```
brentvatne, line 240:
```suggestion
> What does the "devices.json" file contain?
```

for consistency with the other headings

packages/xdl/src/ProjectSettings.ts
```diff
@@ -135,6 +150,63 @@ export async function setPackagerInfoAsync(
...
+  const currentTime = new Date().getTime();
+  const newDeviceIds = typeof deviceIds === 'string' ? [deviceIds] : deviceIds;
+
+  const { devices } = await readDevicesInfoAsync(projectRoot);
+  const newDevicesJson = devices
+    .filter(device => {
+      if (newDeviceIds.includes(device.installationId)) {
+        return false;
+      }
+      // delete any devices that haven't been used to open this project in 30 days
```
brentvatne, line 197: not sure intentional but it looks like this would only run when a manifest is requested. i think what would happen here is if it has been a few months since you worked on your project and go back and run `expo start` then we'd start dev sessions for all old devices and then shortly afterwards delete them, when the first manifest is loaded

brentvatne (approved):

packages/xdl/src/ProjectSettings.ts
esamelson: @EvanBacon [suggested this change](https://github.com/expo/expo-cli/pull/4087#discussion_r779807142). Maybe we should make all the headers consistent with this new one?

packages/xdl/src/ProjectSettings.ts
esamelson: That's a fair point. I don't think it's a huge deal to start sessions for old devices once in that case, but on the other hand it's not hard to filter out old devices in the `read` function as well, so I'll go ahead and make that change.

PR #4112, opened 2022-01-03
[expo-cli] Remove instance of got

EvanBacon:
# Why

Prep to move into `expo` package.

# How

# Test Plan

### DIFF
--- a/packages/expo-cli/src/commands/eject/Github.ts
+++ b/packages/expo-cli/src/commands/eject/Github.ts
@@ -26,8 +26,12 @@ type RepoInfo = {
 };
 
 async function isUrlOk(url: string): Promise<boolean> {
-  const res = await fetch(url);
-  return res.status === 200;
+  try {
+    const res = await fetch(url);
+    return res.status === 200;
+  } catch {
+    return false;
+  }
 }
 
 async function getRepoInfo(url: any, examplePath?: string): Promise<RepoInfo | undefined> {
--- a/packages/expo-cli/src/commands/eject/Github.ts
+++ b/packages/expo-cli/src/commands/eject/Github.ts
@@ -1,6 +1,7 @@
 import chalk from 'chalk';
 import * as fs from 'fs-extra';
 import got from 'got';
+import fetch from 'node-fetch';
 import { Ora } from 'ora';
 import path from 'path';
 import { Stream } from 'stream';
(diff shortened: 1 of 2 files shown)


packages/expo-cli/src/commands/utils/validateApplicationId.ts
```diff
@@ -33,8 +33,7 @@ export async function getBundleIdWarningAsync(bundleId: string): Promise<string
 
   const url = `http://itunes.apple.com/lookup?bundleId=${bundleId}`;
   try {
-    const response = await got(url);
-    const json = JSON.parse(response.body?.trim());
+    const json = await fetch(url).then(value => value.json());
```
kbrandwijk, line 36: I'm not sure if this is a common pattern throughout the codebase, but I'm not a fan of mixing async/await with then callbacks. I would prefer (even with the extra variable assignment):
```js
const response = await fetch(url)
const json = await response.json()
```

packages/expo-cli/src/commands/eject/Github.ts
```diff
@@ -25,8 +26,8 @@ type RepoInfo = {
 };
 
 async function isUrlOk(url: string): Promise<boolean> {
-  const res = await got(url).catch(e => e);
-  return res.statusCode === 200;
+  const res = await fetch(url);
```
kbrandwijk, line 29: The old `.catch(e => e)` would catch an eventual error and just make it the return value, causing the condition in the next line to fail, and the function would return `false` even in case of an exception. The new implementation doesn't do that, so in case of an exception, the function will throw an Unhandled Promise Rejection, bubbling all the way to a rather undescriptive error here: https://github.com/expo/expo-cli/blob/64eef52a6557cb66b9191819aa55ed96238a41b7/packages/expo-cli/src/commands/eject/createNativeProjectsFromTemplateAsync.ts#L148.

kbrandwijk (commented):
`node-fetch` also supports streams, is there a compelling reason to keep using `got` here? https://github.com/expo/expo-cli/blob/main/packages/expo-cli/src/commands/eject/Github.ts#L148

EvanBacon:
@kbrandwijk we remove the last instance in this much larger PR https://github.com/expo/expo-cli/pull/4088

packages/expo-cli/src/commands/utils/validateApplicationId.ts
EvanBacon:
```suggestion
    const response = await fetch(url);
    const json = await response.json();
```

PR #4118, opened 2022-01-05
[expo-cli] Add environment variables for prebuild template and skipDependencyUpdate parameters

kbrandwijk:
# Why

Resolves https://github.com/expo/eas-cli/issues/887.
> `expo prebuild` allows a custom template. You can subsequently build this with `eas build` without issues. Therefore, it would be useful for managed workflow projects, to be able to specify the template directly in eas.json, instead of having to prebuild locally.

# How

Instead of using eas.json, which might clutter the config, and would also require changes outside of `expo-cli`, this initial implementation introduced environment variables `EXPO_PREBUILD_TEMPLATE` and (through a stretch goal of the original issue) `EXPO_PREBUILD_SKIP_DEPENDENCY_UPDATE` that are alternatively observed when the command-line parameter is not provided.

# Test Plan

[ ] TBD (only tested locally by setting the ENV vars)

### DIFF
--- a/packages/expo-cli/src/commands/eject/prebuildAppAsync.ts
+++ b/packages/expo-cli/src/commands/eject/prebuildAppAsync.ts
@@ -50,6 +50,13 @@ export async function prebuildAsync(
   const { exp, pkg } = await ensureConfigAsync({ projectRoot, platforms });
   const tempDir = temporary.directory();
 
+  const template = options.template || process.env.EXPO_PREBUILD_TEMPLATE;
+  const skipDependencyUpdate =
+    options.skipDependencyUpdate ||
+    (process.env.EXPO_PREBUILD_SKIP_DEPENDENCY_UPDATE
+      ? process.env.EXPO_PREBUILD_SKIP_DEPENDENCY_UPDATE.split(',')
+      : undefined);
+
   const {
     hasNewProjectFiles,
     needsPodInstall,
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/eject/prebuildAppAsync.ts
```diff
@@ -50,6 +50,13 @@ export async function prebuildAsync(
   const { exp, pkg } = await ensureConfigAsync({ projectRoot, platforms });
   const tempDir = temporary.directory();
 
+  const template = options.template || process.env.EXPO_PREBUILD_TEMPLATE;
+  const skipDependencyUpdate =
+    options.skipDependencyUpdate ||
+    (process.env.EXPO_PREBUILD_SKIP_DEPENDENCY_UPDATE
```
EvanBacon, line 56: Adding two env variables seems a bit verbose, we want to minimize the amount of options/input users have because it increases the maintenance cost of CLI.

Would be better if a user could change the prebuild script or add prebuild flags imo.

kbrandwijk: They mimic the existing parameters though, so we're not really _adding_ options here.

If I understand your alternative approach correctly, you would have something similar to `gradleCommand` in your eas.json, for `prebuildCommand`, and you can pass in all existing parameters that way? While I like that idea, the footprint for that change seems a bit bigger. Within the context of `expo-cli`, there's no concept of eas build config, for starters. Also see @wkozyra95's comment on the issue: https://github.com/expo/eas-cli/issues/887#issuecomment-1005895174

PR #4121, opened 2022-01-06
[cli][init] Skip git init in monorepo

EvanBacon:
# Why

- Remove flags from git init function since it's only used once.
- Provide better logs.
- Auto skip git init in a git repo to better support monorepos. It's easy enough to manually set one up later.
- Feature split out of https://github.com/expo/expo-cli/pull/4088

# Test Plan

- `expo init my-app && cd my-app && expo init inner-app` -- `my-app/inner-app/.git` won't exist.

### DIFF
--- a/packages/expo-cli/src/commands/initAsync.ts
+++ b/packages/expo-cli/src/commands/initAsync.ts
@@ -343,7 +343,7 @@ export async function actionAsync(incomingProjectRoot: string, command: Partial<
   try {
     // check if git is installed
     // check if inside git repo
-    await initGitRepoAsync(projectPath, { silent: true, commit: true });
+    await initGitRepoAsync(projectPath);
   } catch {
     // todo: check if git is installed, bail out
   }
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/initAsync.ts
```diff
@@ -361,37 +361,38 @@ async function installNodeDependenciesAsync(projectRoot: string, packageManager:
...
-    if (flags.commit) {
-      await spawnAsync('git', ['add', '--all'], { cwd: root, stdio: 'ignore' });
-      await spawnAsync('git', ['commit', '-m', 'Created a new Expo app'], {
-        cwd: root,
-        stdio: 'ignore',
-      });
-    }
+    Log.debug('Initialized a git repository.');
+
+    await spawnAsync('git', ['add', '--all'], { cwd: root, stdio: 'ignore' });
```
kbrandwijk, line 388: If you are already in a repo (in case of a monorepo), this part gets skipped, but wouldn't it still be nice to commit the changes `expo init` makes (albeit to the 'parent' repo)?

kbrandwijk (commented):
The comment above the calling method is no longer accurate because of these changes (https://github.com/expo/expo-cli/pull/4121/files#diff-3f1a11b31c10c72303bafe2f6c933108c06b1664979c20392af50ed7cd5f6d12R340)

PR #4122, opened 2022-01-06
[config-plugins] Add code signing configuration to updates config plugin

wschurman:
# Why

Configuration of expo-updates code signing adds the certificate and metadata in the app.config (see https://github.com/expo/universe/pull/8847).

This PR adds code to the updates config plugin to transfer that info to the manifest and ios plist in the correct format.

# How

- Factor out common code and tests.
- Add new code and tests for the new fields.

Depends on https://github.com/expo/expo/pull/15682. Depends on https://github.com/expo/universe/pull/8847.

# Test Plan

Run tests.

1. In test app, add two new configuration fields to `app.json` updates field.
2. In `src/plugins/unversioned/expo-updates.ts`, comment out static plugin and use fallback
3. `EXPO_DEBUG=true ~/expo/expo-cli/node_modules/.bin/expo prebuild  --clean --no-install -p ios`, and check Expo.plist for values. Ensure they are the correct format.
4. `EXPO_DEBUG=true ~/expo/expo-cli/node_modules/.bin/expo prebuild  --clean --no-install -p android`, and check AndroidManifest.xml for values. Ensure they are the correct format.

`~/expo/expo-cli/packages/expo-cli/bin/expo.js config --type=public` in a project with these fields to ensure they are removed from the public config (so they're not embedded in manifests).

`~/expo/expo-cli/packages/expo-cli/bin/expo.js build:web` in a project with these fields to ensure they are not present in webpack config (though not sure if this is the right way to test it since it seems like most fields aren't included by default).


EvanBacon:
Thoughts on maybe putting this in the EAS config and embedding that in the build?

Ensure that this value doesn't show up in 'expo config --type public' (unless that's the desired behavior).

Finally, ensure this value is deleted in the Webpack config so it isn't exposed in web projects.

wschurman:
Thanks for the review!

> Thoughts on maybe putting this in the EAS config and embedding that in the build?

Unfortunately this isn't an EAS-only feature, though I'm not as familiar with how we typically handle configuration for features that are fully open-source (this codesigning feature) that also have an EAS implementation. 

So far, it looks like all updates configuration in this category is added to app.json and then transferred to the native configuration files via this config plugin, but maybe there's an argument that says that to use the feature you need to either a) be using EAS and put the config in eas.json and then we'll put it in the native config files for you, or b) put it in the native config files manually. Thoughts? I think for now the way this is written makes the most sense for consistency with the other similar fields, but could see an argument for changing it.

> Ensure that this value doesn't show up in 'expo config --type public' (unless that's the desired behavior).

Yep, we definitely don't want this in the config. I'll look into removing it (no idea how but I'll figure it out)

> Finally, ensure this value is deleted in the Webpack config so it isn't exposed in web projects.

Ditto, will do.

packages/config-plugins/src/ios/Updates.ts
```diff
@@ -71,6 +52,7 @@ export const withUpdates: ConfigPlugin<{ expoUsername: string | null }> = (
 };
 
 export function setUpdatesConfig(
+  projectRoot: string,
```
jkhales, line 55: Make sure to mention this in the Changelog as it breaks: https://github.com/expo/eas-cli/blob/23237210d1485ce5b1afeb6146a5d0f8e2f4475a/packages/eas-cli/src/build/ios/UpdatesModule.ts#L17

packages/config/src/Config.ts
```diff
@@ -152,6 +152,13 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}):
       configWithDefaultValues.exp.currentFullName = getFullName(configWithDefaultValues.exp);
       // @ts-ignore: originalFullName not on type yet.
       configWithDefaultValues.exp.originalFullName = getFullName(configWithDefaultValues.exp);
+
+      if (configWithDefaultValues.exp.updates?.codeSigningCertificate) {
+        delete configWithDefaultValues.exp.updates.codeSigningCertificate;
+      }
+      if (configWithDefaultValues.exp.updates?.codeSigningMetadata) {
+        delete configWithDefaultValues.exp.updates.codeSigningMetadata;
+      }
```
jkhales, line 161: I see that checking before deleting is the convention in this file, but this is cleaner.
```suggestion
      delete configWithDefaultValues.exp.updates?.codeSigningCertificate;
      delete configWithDefaultValues.exp.updates?.codeSigningMetadata;
```

jkhales (approved):
✨

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -125,3 +127,67 @@ export function getRuntimeVersion(
...
+export function getUpdatesCodeSigningMetadataStringified(
+  config: Pick<ExpoConfigUpdates, 'updates'>
+): string | undefined {
+  const metadata = getUpdatesCodeSigningMetadata(config);
+  if (!metadata) {
+    return undefined;
+  }
+
+  // replace double quotes with single quotes for android manifest
+  return JSON.stringify(metadata).replace(/"/g, "'");
```
ide, line 192: This seems particularly fragile and an abstraction violation. When a string goes inside AndroidManifest.xml, shouldn't it be XML-encoded? And when the Android manifest is read, an XML decoder should get back the original JSON string.

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -125,3 +127,67 @@ export function getRuntimeVersion(
...
+  projectRoot: string,
+  config: Pick<ExpoConfigUpdates, 'updates'>
+): string | undefined {
+  const codeSigningCertificatePath = config.updates?.codeSigningCertificate;
+  if (!codeSigningCertificatePath) {
+    return undefined;
+  }
+
+  const finalPath = path.join(projectRoot, codeSigningCertificatePath);
+  if (!finalPath || !fs.existsSync(finalPath)) {
```
ide, line 169: Reading the code, it seems that finalPath is always defined

packages/config-plugins/src/utils/Updates.ts
```diff
@@ -125,3 +127,67 @@ export function getRuntimeVersion(
...
+    throw new Error(`File not found at \`updates.codeSigningCertificate\` path: ${finalPath}`);
+  }
+
+  // need to format it as a single line string with `\n` characters actually in the string
+  return fs.readFileSync(finalPath, 'utf8').replace(/\n/g, '\\n');
```
ide, line 174: This should be the job of the code that is storing/transmitting the certificate, since it knows the escaping/encoding rules of the storage/transmission format.

ide (changes requested):

wschurman:
Testing how android parses these was a surprisingly difficult (multi-day) task. But, the end result is @ide was correct, we don't need to substitute newlines or quotes in the values in `AndroidManifest.xml`.

I'm working on a writeup of the testing instructions for that, but that is independent of this PR since it is a super complex set of steps.

packages/config-plugins/src/ios/Updates.ts
```diff
@@ -1,6 +1,5 @@
-import * as path from 'path';
+import * as Updates from 'path';
```
ide, line 1: This doesn't look intentional - revert

ide (approved):

packages/config-plugins/src/ios/Updates.ts
wschurman: lol how did that happen

PR #4132, opened 2022-01-10
[xdl] Fix various ngrok bugs

EvanBacon:
# Why

- Fix "pressing `i` in TerminalUI doesn't open ngrok URL".
- Prevent opening multiple ngrok servers when the new dev server API is in use.
- Fix "Tunnel URL not found (it might not be ready yet), falling back to LAN URL." warning on `rm -rf .expo && expo start --tunnel`
- Add better inline documentation about what's going on.

# Test Plan

- `rm -rf .expo && expo start --tunnel` then open on iOS and Android.
  - Ensure Android doesn't crash (possible when port is missing).
  - Ensure iOS opens using ngrok URL and not localhost URL.

### DIFF
--- a/packages/xdl/src/UrlUtils.ts
+++ b/packages/xdl/src/UrlUtils.ts
@@ -427,12 +427,7 @@ function joinURLComponents({
   const validPort = port ?? '80';
   const validProtocol = protocol ? `${protocol}://` : '';
 
-  let output = `${validProtocol}${hostname}`;
-  if (validPort) {
-    output += `:${validPort}`;
-  }
-
-  return output;
+  return `${validProtocol}${hostname}:${validPort}`;
 }
 
 export function stripJSExtension(entryPoint: string): string {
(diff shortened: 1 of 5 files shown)


packages/xdl/src/UrlUtils.ts
```diff
@@ -419,11 +420,19 @@ function joinURLComponents({
...
+  // Android HMR breaks without this port 80.
+  // This is because Android React Native WebSocket implementation is not spec compliant and fails without a port:
+  // `E unknown:ReactNative: java.lang.IllegalArgumentException: Invalid URL port: "-1"`
+  // Invoked first in `metro-runtime/src/modules/HMRClient.js`
   const validPort = port ?? '80';
   const validProtocol = protocol ? `${protocol}://` : '';
 
-  return `${validProtocol}${hostname}:${validPort}`;
+  let output = `${validProtocol}${hostname}`;
+  if (validPort) {
```
kbrandwijk, line 431: This can never be `false`, right, with it being defined as `port ?? '80'`?

kbrandwijk (approved):

PR #4141, opened 2022-01-13
[expo-cli][docs] update broken flags in docs generation

kbrandwijk:
# Why

The current introspection script breaks formatting when there's multiple pipe characters in the flags
image

# How

Replace _all_ occurences instead of only the first. Properly escape them instead of replacing them with a less appealing unicode character.

# Test Plan

Generated json and markdown locally to verify.


brentvatne:
@kbrandwijk - when you get a chance can you re-generate docs with this so we can deploy fix?

PR #4143, opened 2022-01-14
[Interstitial page] Add missing information about project

lukmccall:
# Why

Pretty straightforward PR - just adding more information about application to interstitial page.

# How

Added information like:
- app name
- runtime version
- path

# Test Plan

Go to `localhost:19000/_expo/loading?platform=<android | ios>` and check if everything is render correctly.

### DIFF
--- a/packages/xdl/src/start/LoadingPageHandler.ts
+++ b/packages/xdl/src/start/LoadingPageHandler.ts
@@ -1,3 +1,5 @@
+import { ExpoConfig, getConfig, getNameFromConfig } from '@expo/config';
+import { getRuntimeVersionNullable } from '@expo/config-plugins/build/utils/Updates';
 import express from 'express';
 import { readFile } from 'fs-extra';
 import http from 'http';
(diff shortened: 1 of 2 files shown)


brentvatne (approved):

packages/xdl/src/start/LoadingPageHandler.ts
```diff
@@ -21,14 +23,22 @@ export function setOnDeepLink(listener: OnDeepLinkListener) {
...
 
   return null;
 }
 
+function getRuntimeVersion(exp: ExpoConfig, platform: 'android' | 'ios' | null) {
+  if (!platform) {
+    return 'Undetected';
+  }
+
+  return getRuntimeVersionNullable(exp, platform) ?? 'Undetected';
```
esamelson, line 39: could we show the SDK version (and have the page say "SDK version") instead of "Undetected" if the runtime version is null?

esamelson (approved):

packages/xdl/src/start/LoadingPageHandler.ts
lukmccall: Will do it in the separate PR ;)

PR #4144, opened 2022-01-14
[Interstitial page] Ensure that development build is installed

lukmccall:
# Why

Close ENG-2211.

# How

Before opening a deep link to the interstitial page, we need to check if everything is installed. So I've added a logic that will install the `Expo Go` and check if the development build is available. If not we will fall back to the expo go. 

# Test Plan

Test scenarios:
- app without dev-client 
- check if the expo go will be installed if not present 
- check what happens if the development build isn't present 
- check what happens if both development build and expo go isn't available

### DIFF
--- a/packages/xdl/src/Android.ts
+++ b/packages/xdl/src/Android.ts
@@ -378,13 +378,7 @@ async function ensureDevClientInstalledAsync(device: Device, applicationId: stri
 }
 
 async function isDevClientInstalledAsync(device: Device, applicationId: string): Promise<boolean> {
-  try {
-    await ensureDevClientInstalledAsync(device, applicationId);
-  } catch (e) {
-    return false;
-  }
-
-  return true;
+  return await isInstalledAsync(device, applicationId);
 }
 
 async function getExpoVersionAsync(device: Device): Promise<string | null> {
@@ -776,7 +770,7 @@ async function openUrlAsync({
       clientApplicationId = await getClientApplicationId();
       await ensureDevClientInstalledAsync(device, clientApplicationId);
     } else if (
-      process.env['EXPO_ENABLE_INTERSTITIAL_PAGE'] &&
+      Env.isInterstitiaLPageEnabled() &&
       !devClient &&
       isDevClientPackageInstalled(projectRoot)
     ) {
@@ -797,9 +791,9 @@ async function openUrlAsync({
         // Everything is installed, we can present the interstitial page.
         clientApplicationId = ''; // it will open browser
       } else {
-        // The development build ins't available. So let's fallback to Expo Go.
+        // The development build isn't available. So let's fall back to Expo Go.
         Logger.global.warn(
-          `\u203A The 'expo-dev-client' package is installed, but the development build isn't available.\nYour app will be open in Expo Go instead. If you want to use the development build, please install it on the simulator first.\n${learnMore(
+          `\u203A The 'expo-dev-client' package is installed, but a development build isn't available.\nYour app will open in Expo Go instead. If you want to use the development build, please install it on the simulator first.\n${learnMore(
             'https://docs.expo.dev/clients/distribution-for-ios/#building-for-ios'
           )}`
         );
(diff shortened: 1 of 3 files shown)


packages/xdl/src/Android.ts
```diff
@@ -377,6 +377,16 @@ async function ensureDevClientInstalledAsync(device: Device, applicationId: stri
   }
 }
 
+async function isDevClientInstalledAsync(device: Device, applicationId: string): Promise<boolean> {
```
EvanBacon, line 380: Use `isInstalledAsync`

packages/xdl/src/Simulator.ts
```diff
@@ -696,6 +730,18 @@ async function assertDevClientInstalledAsync(
   }
 }
 
+async function isDevClientInstalledAsync(
+  simulator: Pick<SimControl.SimulatorDevice, 'udid' | 'name'>,
+  bundleIdentifier: string
+): Promise<boolean> {
+  try {
+    await profileMethod(assertDevClientInstalledAsync)(simulator, bundleIdentifier);
```
EvanBacon, line 738: Use the internal method directly instead of wrapping the same method.

packages/xdl/src/Simulator.ts
```diff
@@ -617,18 +617,52 @@ async function openUrlInSimulatorSafeAsync({
...
-        exp
+        exp as ExpoConfig
       );
       await profileMethod(assertDevClientInstalledAsync)(simulator, bundleIdentifier);
       if (!skipNativeLogs) {
         // stream logs before opening the client.
         await streamLogsAsync({ udid: simulator.udid, bundleIdentifier });
       }
+    } else if (
+      process.env['EXPO_ENABLE_INTERSTITIAL_PAGE'] &&
```
EvanBacon, line 628: Extract this out into a global constant and use getenv.boolish

packages/xdl/src/Android.ts
```diff
@@ -711,55 +721,100 @@ async function openUrlAsync({
   let installedExpo = false;
   let clientApplicationId = 'host.exp.exponent';
 
+  const installExpoIfNeeded = async (device: Device) => {
```
EvanBacon, line 724: Maybe add some form of automated testing since this logic was pretty fragile.

packages/xdl/src/Simulator.ts
```diff
@@ -617,18 +617,52 @@ async function openUrlInSimulatorSafeAsync({
...
+
+      const isDevClientInstalled = await isDevClientInstalledAsync(
+        simulator,
+        devClientBundlerIdentifier
+      );
+      if (isDevClientInstalled) {
+        // Everything is installed, we can present the interstitial page.
+        bundleIdentifier = ''; // it will open browser.
+      } else {
+        // The development build ins't available. So let's fallback to Expo Go.
```
brentvatne, line 646:
```suggestion
        // The development build isn't available. So let's fall back to Expo Go.
```

packages/xdl/src/Simulator.ts
```diff
@@ -617,18 +617,52 @@ async function openUrlInSimulatorSafeAsync({
...
+        Logger.global.warn(
+          `\u203A The 'expo-dev-client' package is installed, but the development build isn't available.\nYour app will be open in Expo Go instead. If you want to use the development build, please install it on the simulator first.\n${learnMore(
```
brentvatne, line 648:
```suggestion
          `\u203A The 'expo-dev-client' package is installed, but a development build isn't available.\nYour app will be open in Expo Go instead. If you want to use the development build, please install it on the simulator first.\n${learnMore(
```

brentvatne (approved):

packages/xdl/src/Simulator.ts
brentvatne: (same thing above in the android code)

packages/xdl/src/Simulator.ts
brentvatne: (same thing above in the android code)

esamelson:
```suggestion
          `\u203A The 'expo-dev-client' package is installed, but a development build isn't available.\nYour app will be opened in Expo Go instead. If you want to use a development build, please install it on the simulator first.\n${learnMore(
```

packages/xdl/src/Android.ts
lukmccall: I'll do it in a separate PR.

PR #4145, opened 2022-01-14
[Interstitial page] Add option to remember user choice

lukmccall:
# Why

Closes ENG-2210.

# How

Added the `forceExecutionEnvironment` flag that determined where we should open a deep link if the dev-client was installed but the `--dev-client` flag wasn't provided. You can clear that variable using the `-c` option. 

# Test Plan

Test scenarios:

- app without dev-client
- check what happens if a user chooses the development builds 
- check what happens if a user chooses the Expo Go
- check if a user can clear the `forceExecutionEnvironment` flag

### DIFF
--- a/packages/xdl/src/ProjectSettings.ts
+++ b/packages/xdl/src/ProjectSettings.ts
@@ -14,6 +14,7 @@ export type ProjectSettings = {
   minify: boolean;
   urlRandomness: string | null;
   https: boolean;
+  forceExecutionEnvironment: 'expo-go' | 'expo-dev-client' | null;
 };
 export type Settings = ProjectSettings;
 
@@ -27,6 +28,7 @@ const projectSettingsDefaults: ProjectSettings = {
   minify: false,
   urlRandomness: null,
   https: false,
+  forceExecutionEnvironment: null,
 };
 
 type PackagerInfo = {
--- a/packages/expo-cli/src/commands/start/startAsync.ts
+++ b/packages/expo-cli/src/commands/start/startAsync.ts
@@ -2,7 +2,14 @@ import { ConfigError, ExpoConfig, getConfig, isLegacyImportsEnabled } from '@exp
 import chalk from 'chalk';
 import path from 'path';
 import resolveFrom from 'resolve-from';
-import { LoadingPageHandler, Project, UnifiedAnalytics, UrlUtils, Versions } from 'xdl';
+import {
+  LoadingPageHandler,
+  Project,
+  ProjectSettings,
+  UnifiedAnalytics,
+  UrlUtils,
+  Versions,
+} from 'xdl';
 
 import StatusEventEmitter from '../../analytics/StatusEventEmitter';
 import getDevClientProperties from '../../analytics/getDevClientProperties';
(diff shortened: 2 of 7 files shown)


packages/xdl/src/ProjectSettings.ts
```diff
@@ -14,6 +14,7 @@ export type ProjectSettings = {
   minify: boolean;
   urlRandomness: string | null;
   https: boolean;
+  forceExecutionEnvironment: 'expo-go' | 'expo-dev-client' | null;
```
EvanBacon, line 17: add comments explaining what this, and everything else in this PR are doing.

EvanBacon (changes requested):
`expo start -c` is clearly marked for resetting the bundler cache, I don't know if it makes much sense to combine purposes here. I think it'd actually be quite upsetting to clear the bundler cache and have to choose the interstitial every time, or vise versa.

brentvatne:
i agree about not overloading `-c` to reset the preference selected here.

i'm also curious to chat through this approach for saving preferences :) in the past we've found that saving expo-cli options related to the dev server between runs (eg: no-dev, localhost/lan/tunnel preference) has cause confusion for users because they forgot that they set a preference previously. maybe another solution here is to keep the `--dev-client` flag and just have users pass that flag in if they want to have this same effect?

lukmccall:
> i agree about not overloading `-c` to reset the preference selected here.

I put the reset logic there, cause I didn't find a better place and I didn't want to add another flag.

> i'm also curious to chat through this approach for saving preferences :) in the past we've found that saving expo-cli options related to the dev server between runs (eg: no-dev, localhost/lan/tunnel preference) has cause confusion for users because they forgot that they set a preference previously (eg: if you ran `expo start --no-dev` then closed the server and came back the next day and ran `expo start`, you might not be expecting the `--no-dev` flag to be defaulting to `true` but it used to). maybe another solution here is to keep the `--dev-client` flag and just have users pass that flag in if they want to have this same effect?

I don't have a strong option on that topic. However, I think it'll be very nice to save somehow user choice in that case. Also, I don't think that users will switch between the `Expo Go` and the `development build` often. It seems to be a set once and forget scenario, cause you won't change it anyway. cc @tcdavis

tcdavis:
I think we do want the option to save the choice. The point of the interstitial page generally is to avoid the need to know and remember to pass the --dev-client flag without forcing developers who still want to use Go to switch.

Having the interstitial come up every time I run expo start after installing expo-dev-client would get pretty annoying though.  This change gives the user the option to explicitly make a durable choice rather than having to make the selection every time. 

I think the closest analogue we have currently is the devtools toggle rather than something like remembering the --no-dev flag when you use it. Given that the user is explicitly making the choice, I wouldn't expect the need to undo that choice to get a ton of use.  Something like `expo start --restore-defaults` that would clear any user settings generally makes sense to me but happy to defer to @EvanBacon or @brentvatne if you'd prefer something else.

brentvatne:
discussed with @tcdavis and i believe the plan is to put persistence of the user selection on hold, and instead explain in the interstitial that it is possible to use the `--dev-client` flag to skip it. this should help us proceed with shipping the interstitial feature and improving the success rate of onboarding without having to add anything else new that we may end up wanting to change or roll back.

Issue #2221, opened 2020-06-03
Feature (& PR offer) - dependency update script


actuallymentor:
## Description

Running `expo install` installs a compatible dependency version. Right now I rely on [ncu](https://www.npmjs.com/package/npm-check-updates) to check for dependency updates, but that does not take into account expo-compatible updates.

Ideally we would add an `expo update` command that runs `expo install` for every line in the `package.json` dependency trees.

Optionally we can add an exclusion `-x` marker to skip certain packages.

## PR offer

I'm happy to write the code for this. If this is a desirable feature let me know and I'll take some time for it.

brentvatne:
hey @actuallymentor! thank for you the issue.

currently `expo upgrade` handles this, so we could reuse some code from there. i think what we may want is something like `expo install --fix`. i'm afraid that `update` will clash too much with `upgrade`

if you want to work on that it would be great. sorry for not getting back to you until now!

actuallymentor:
Thanks for the response!

Could we do something in `expo-cli/packages/expo-cli/src/commands/upgrade.ts` like:

1. `if( currentSdkVersionString == latestSdkVersion.version && process.argv.includes( '--dependencies' ) )`
2. Skip SDK code part of `upgradeAsync`
3. Update the relevant end-log entries (e.g. "updated SDK" etc)
4. Leave the rest of the script as is

That would allow for the relatively intuitive `expo upgrade --dependencies`.

actuallymentor:
Shall I draft a PR?

![johntravolta](https://user-images.githubusercontent.com/9071382/89023506-f66ee400-d323-11ea-8758-ea47b7136680.gif)

brentvatne:
hello! i think i lost this in my sea of github notifications. usually the best way to move something forward is to just open a pr and discuss there.

i would rather this functionality live on install than upgrade. a common use case here is if you run `npm install` and get the wrong version of a package, then we can just tell people to run `expo install --some-flag` to ensure all of their dependencies are compatible with the sdk version.

so, something like `expo install --fix` or `expo install --fix-sdk-dependencies` or `expo install:sync-sdk-dependencies` would be a good fit for this

EvanBacon:
I think this is better suited for `expo doctor`

brentvatne:
@EvanBacon - maybe `expo doctor --fix-dependencies`?

PR #4153, opened 2022-01-20
feat: new --fix-dependencies option for doctor

kbrandwijk:
# Why

- https://github.com/expo/expo-cli/issues/2221

# How

Added option to `doctor` command, internally using the `install` command for the installation.

# Test Plan

Tested manually.

### DIFF
--- a/packages/expo-cli/src/commands/info/doctor/index.ts
+++ b/packages/expo-cli/src/commands/info/doctor/index.ts
@@ -8,7 +8,7 @@ export default function (program: Command) {
       .command('doctor')
       .description('Diagnose issues with the project')
       .helpGroup('info')
-      .option('--fix-dependencies', 'Fix dependencies versions'),
+      .option('--fix-dependencies', 'Fix incompatible dependency versions'),
     () => import('./doctorAsync')
   );
 }
--- a/packages/expo-cli/src/commands/info/doctor/index.ts
+++ b/packages/expo-cli/src/commands/info/doctor/index.ts
@@ -5,9 +5,10 @@ import { applyAsyncActionProjectDir } from '../../utils/applyAsyncAction';
 export default function (program: Command) {
   applyAsyncActionProjectDir(
     program
-      .command('doctor [path]')
+      .command('doctor')
       .description('Diagnose issues with the project')
-      .helpGroup('info'),
+      .helpGroup('info')
+      .option('--fix-dependencies', 'Fix dependencies versions'),
     () => import('./doctorAsync')
   );
 }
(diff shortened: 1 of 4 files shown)


kbrandwijk:
/changelog-entry new-feature new `--fix-dependencies` option for `doctor`

EvanBacon:
Getting:
```
The "path" argument must be of type string. Received an instance of Command
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Command
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:119:11)
    at Object.resolve (node:path:1098:7)
    at /Users/evanbacon/Documents/GitHub/cli/packages/expo-cli/src/exp.ts:382:26
    at Command.<anonymous> (/Users/evanbacon/Documents/GitHub/cli/packages/expo-cli/src/exp.ts:347:13)
```

packages/expo-cli/src/commands/info/doctor/index.ts
```diff
@@ -5,7 +5,7 @@ import { applyAsyncActionProjectDir } from '../../utils/applyAsyncAction';
 export default function (program: Command) {
   applyAsyncActionProjectDir(
     program
-      .command('doctor')
+      .command('doctor [path]')
```
EvanBacon, line 8: This is a breaking change as you can no longer run `expod doctor ./my-app --fix-dependencies`

kbrandwijk: I couldn't get that to work anyway, so I assumed it didn't. But it's restored now (and still not working for me)

packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
```diff
@@ -8,12 +8,14 @@ import { Versions } from 'xdl';
 
 import CommandError from '../../CommandError';
 import Log from '../../log';
+import { actionAsync } from '../installAsync';
```
EvanBacon, line 11: Generally you shouldn't use the `actionAsync` function directly since it's meant to be used from commander, using the internal functions would potentially be better for this use case.

kbrandwijk: I've looked at that `actionAsync` method, and it happens to do exactly what it needs to do here, as we're doing the _exact_ equivalent of calling `expo install package1 package2 package3` here. `actionAsync` directly contains the entire implementation flow for doing that, using the internal functions directly would mean duplicating/copying that entire logic, or extracting it into another exported function that's not `actionAsync`, but `doTheActualWorkAsync`, which is more of a semantic workaround than much else.

EvanBacon (approved):

PR #4154, opened 2022-01-20
[expo-cli] deprecate eject command

kbrandwijk:
# Why

- https://linear.app/expo/issue/ENG-2768/drop-expo-eject

# How

Remove the main command, it's implementation, and cleanup orphans. Update some phrasing throughout to not use the term 'eject' anymore.

# Test Plan

Remaining tests still pass, related commands (like prebuild) still work, etc.


brentvatne (approved):

packages/expo-cli/src/exp.ts
```diff
@@ -709,7 +709,9 @@ async function runAsync(programName: string) {
     trackUsage(program.commands); // must be after register commands
 
     program.on('command:detach', () => {
-      Log.warn('To eject your project to ExpoKit (previously "detach"), use `expo eject`.');
+      Log.warn(
+        'Ejecting your project to ExpoKit (previously "detach") is deprecated, use `expo prebuild`.'
```
EvanBacon, line 713: This should stay, most everything else should go. Ideally we'd just deprecate by not bringing eject over to versioned CLI and simply deprecating the entire CLI. 

We should leave expo eject in place for now so users have a way to access it for legacy reasons. Would be good to add a console log to the command saying that eject is deprecated in favor of prebuild though.

EvanBacon (changes requested):
see comment

packages/expo-cli/src/exp.ts
kbrandwijk: Half of this PR is actually getting rid of the command. So you're saying now:
- Leave the original log warning as is for `detach`
- Leave the `eject` command working, but only add a similar log warning to it.
Why would you want to keep the deprecated command intact, if the whole point is that we're cleaning up because it's deprecated?

kbrandwijk: @EvanBacon Could you comment on the intention, so that this one can be wrapped up?

EvanBacon: Please revert deleting `expo eject`. We need to add a warning and indication of deprecation before deleting code.

PR #4160, opened 2022-01-25
Rename `@expo/api` to `@expo/dev-api`

EvanBacon:
# Why

The name `@expo/api` is a bit too broad and possibly valuable for other projects. The name `api` isn't great since this package is primarily tied to the CLI (can't be used in other JS runtimes).

## Other Options

- @brentvatne recommended `@expo/expo-api` which seems a bit repetitive but is also fine.
- `@expo/cli-api` could also work but it sort of implies that you use it to interact with the CLI.


kbrandwijk:
Ultimately, what does this package do? Is dev-api a temporary name?

EvanBacon:
@kbrandwijk no, we're looking for a more permanent name for an internal package to use in expo/expo.

byCedric:
Some things on my mind:
- `@expo/core` (or even eas core) - since it's the "core" functionality of interacting with our backend
- `@expo/sdk` - more in line with snack-sdk and basically the "SDK for Expo/EAS" stuff
- `@expo/client` (or even eas client) - seems more in [line with vercel](https://github.com/vercel/vercel/tree/main/packages/client)

API or dev-api seems not too bad to me tbh. We could also combine stuff, like core-api or sdk-api? (although the latter seems weird)

PR #4164, opened 2022-01-26
[config-plugins] avoid duplicate CFBundleURLTypes

giautm:
# Why

- `appendScheme` insert the same schema twice if I call it twice. It should insert only once

# How

# Test Plan

- Exists test

### DIFF
--- a/packages/config-plugins/src/ios/Scheme.ts
+++ b/packages/config-plugins/src/ios/Scheme.ts
@@ -40,18 +40,20 @@ export function setScheme(
   };
 }
 
-// TODO: update this to be idempotent!
 export function appendScheme(scheme: string | null, infoPlist: InfoPlist): InfoPlist {
   if (!scheme) {
     return infoPlist;
   }
 
-  const existingSchemes = infoPlist.CFBundleURLTypes;
+  const existingSchemes = infoPlist.CFBundleURLTypes ?? [];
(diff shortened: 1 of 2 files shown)


packages/config-plugins/src/ios/Scheme.ts
```diff
@@ -40,18 +40,20 @@ export function setScheme(
...
 }
 
-// TODO: update this to be idempotent!
 export function appendScheme(scheme: string | null, infoPlist: InfoPlist): InfoPlist {
   if (!scheme) {
     return infoPlist;
   }
 
-  const existingSchemes = infoPlist.CFBundleURLTypes;
+  const existingSchemes = infoPlist.CFBundleURLTypes ?? [];
```
EvanBacon, line 48: Update the tests at `ios/__tests__/Scheme-test.ts` to stress this.

giautm:
hey, @EvanBacon I added the test, but it failed by a network issue. Can you rerun it for me?
```
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
```

EvanBacon (approved):
Excellent work!

PR #4169, opened 2022-02-01
[expo-cli][code-signing-certificates] Add package and basic self-signed certificate generation

wschurman:
# Why

Part of code signing will be helping users generate, manage, and use code signing certificates (and their private keys). This PR adds a basic set of commands to expo-cli:
- `expo codesigning:generate` - assistant to help the user generate a public/private RSA key pair and certificate valid for expo-updates code signing: https://github.com/expo/expo/blob/main/packages/expo-updates/README.md#configuration
- `expo codesigning:configure` - validates the generated certificate (or a user-generated certificate) and updates app.json with the configuration necessary for the updates config plugin.

After running these setup steps:
a. (if using EAS) `eas publish` will see the code signing field in app.json and require that the user supplies their private key. It will then use that private key to sign the publish and upload the signature info to the server.
b. (if self-hosting modern manifest or classic manifest) The developer will put their generated private key on their server and sign the manifests as they are about to be served.
c. (classic Expo updates on Expo server) Code signing will not work with this, and we may want to add a check that the code signing fields aren't present in app.json when publishing.

# How

This PR is structured as follows:
- Bulk of certificate and key code and tests of the code is in a new package, `@expo/code-signing-certificates`. This is to make it clearer where the code should live, but also because we may need to use the package elsewhere in the future.
- Add two new commands and tests for them.

# Test Plan

Run tests. Wait for CI.


packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
+import assert from 'assert';
+import { md, pki as PKI, random, util } from 'node-forge';
+
+// a hexString is considered negative if it's most significant bit is 1
+// because serial numbers use ones' complement notation
+// this RFC in section 4.1.2.2 requires serial numbers to be positive
+// http://www.ietf.org/rfc/rfc5280.txt
+function toPositiveHex(hexString: string) {
+  let mostSiginficativeHexAsInt = parseInt(hexString[0], 16);
```
jkhales, line 9:
```suggestion
  let mostSiginficantHexAsInt = parseInt(hexString[0], 16);
```

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  if (mostSiginficativeHexAsInt < 8) {
+    return hexString;
+  }
+  mostSiginficativeHexAsInt -= 8;
+  return mostSiginficativeHexAsInt.toString() + hexString.substring(1);
```
jkhales, line 14:
```suggestion
  return mostSiginficativeHexAsInt.toString(16) + hexString.substring(1);
```

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+// a hexString is considered negative if it's most significant bit is 1
+// because serial numbers use ones' complement notation
+// this RFC in section 4.1.2.2 requires serial numbers to be positive
+// http://www.ietf.org/rfc/rfc5280.txt
+function toPositiveHex(hexString: string) {
+  let mostSiginficativeHexAsInt = parseInt(hexString[0], 16);
+  if (mostSiginficativeHexAsInt < 8) {
+    return hexString;
+  }
+  mostSiginficativeHexAsInt -= 8;
```
jkhales, line 13:
```suggestion
  if (mostSiginficativeHexAsInt > 8) {
    return hexString;
  }
  mostSiginficativeHexAsInt += 8;
```

otherwise wouldn't the case of `9` return a negative hex?

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  validityNotBefore,
+  validityNotAfter,
+  commonName,
+}: GenerateParameters): PKI.Certificate {
+  const cert = PKI.createCertificate();
+  cert.publicKey = publicKey;
+  cert.serialNumber = toPositiveHex(util.bytesToHex(random.getBytesSync(9)));
+
+  assert(
+    validityNotAfter >= validityNotBefore,
```
jkhales, line 77: Do we want the weak inequality? This creates an invalid cert.

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  cert.validity.notAfter = validityNotAfter;
+
+  const attrs = [
+    {
+      name: 'commonName',
+      value: commonName,
+    },
+  ];
+  cert.setSubject(attrs);
+  cert.setIssuer(attrs);
```
jkhales, line 90: Is this the standard values for self signed?

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ *
+ * @param certificate
+ */
+export function validateSelfSignedCertificate(certificate: PKI.Certificate) {
+  if (certificate.issuer.hash !== certificate.subject.hash) {
+    throw new Error(
+      'Certificate issuer has does not match subject hash, certificate may not be self-signed.'
```
jkhales, line 124:
```suggestion
      'Certificate issuer hash does not match subject hash, certificate may not be self-signed.'
```
Also a bit confused, the function name says "validate self signed" the error message seems to demand that it is not?

packages/code-signing-certificates/src/index.ts
```diff
@@ -0,0 +1,144 @@
...
+      dataEncipherment: false,
+    },
+    {
+      name: 'extKeyUsage',
+      critical: true,
+      serverAuth: false,
+      clientAuth: false,
+      codeSigning: true,
+      emailProtection: false,
+      timeStamping: false,
```
jkhales, line 109: Are these "standard" extensions? Or is there some important configuration happening here?

wschurman:
Going to pull this from review. Lots of changes coming.

packages/code-signing-certificates/src/index.ts
wschurman: Anything less than hex 8 (8) has a most significant digit of 0 when represented in binary (verified in https://www.rapidtables.com/convert/number/hex-to-binary.html), so if the most significant digit is 0-8, the number is fine and it should be returned verbatim. When it is greater than 8, we subtract 8 from the first hex digit (making it be in the range 0-8) so that the number has a 0 most significant digit when represented in binary.

packages/code-signing-certificates/src/index.ts
wschurman: It's standard for subject and issuer to be the same. That being said, the only required attribute seems to be commonName, but it is common for other fields to be specified. In this case it doesn't matter, but depending on where we go with expo go certificates and child certificates, we may use these fields more.

packages/code-signing-certificates/src/index.ts
wschurman: Haha, english is bad and should feel bad. I meant "may" more as the "expressing possibility" version of the word than the "expressing permission" version. Will update to be clearer.

packages/code-signing-certificates/src/index.ts
wschurman: The configuration that is important is the codeSigning extKeyUsage which is specified in https://github.com/expo/expo/blob/main/packages/expo-updates/README.md#configuration.

packages/expo-cli/src/commands/codesigning/configureCodeSigningAsync.ts
```diff
@@ -31,16 +31,26 @@ export async function actionAsync(projectRoot: string, { input }: Options) {
   validateSelfSignedCertificate(certificate, keyPair);
 
   const { exp } = getConfig(projectRoot, { skipSDKVersionRequirement: true });
+
+  const fields: ExpoConfig['updates'] = {
+    codeSigningCertificate: `./${input}/certificate.pem`,
```
ide, line 36: IMO it might be better to normalize `input` by doing
```suggestion
    codeSigningCertificate: `${path.relative(projectRoot, inputDir)}/certificate.pem`,
```

packages/expo-cli/src/commands/codesigning/configure.ts
```diff
@@ -0,0 +1,14 @@
+import type { Command } from 'commander';
+
+import { applyAsyncActionProjectDir } from '../utils/applyAsyncAction';
+
+export default function (program: Command) {
+  applyAsyncActionProjectDir(
+    program
+      .command('codesigning:configure [path]')
+      .description('Configure and validate expo-updates code signing for this project')
```
ide, line 9: Just to confirm, this says expo-updates (the library) and not Expo Updates (the spec) because it configures app.json for expo-updates?

packages/expo-cli/src/commands/codesigning/configureCodeSigningAsync.ts
```diff
@@ -0,0 +1,58 @@
...
+    codeSigningMetadata: {
+      keyid: 'main',
+      alg: 'rsa-v1_5-sha256',
+    },
+  };
+  await attemptModification(
+    projectRoot,
+    {
+      updates: {
+        ...(exp.updates || {}),
```
ide, line 46:
```suggestion
        ...exp.updates,
```
It's fine if it's null/undefined

packages/expo-cli/src/commands/codesigning/configureCodeSigningAsync.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export async function actionAsync(projectRoot: string, { input }: Options) {
+  assert(typeof input === 'string', '--input must be a string');
+
+  const inputDir = path.resolve(projectRoot, input);
+
+  const [certificatePEM, privateKeyPEM, publicKeyPEM] = (
+    await Promise.all(
+      ['certificate.pem', 'private-key.pem', 'public-key.pem'].map(fname =>
+        fs.readFile(`${inputDir}/${fname}`)
```
ide, line 24:
```suggestion
        fs.readFile(path.join(inputDir, fname), 'utf8')
```

Also the synchronous file methods often outperform the async ones except when working with large files. Don't think it matters either way here.

packages/expo-cli/src/commands/codesigning/configureCodeSigningAsync.ts
```diff
@@ -0,0 +1,58 @@
...
+      )
+    )
+  ).map(buffer => buffer.toString());
```
ide, line 27:
```suggestion
  );
```

ide (approved):
Separate from the code in this PR, we should discuss with the team what the flow looks like. Instead of having to generate and then configure, it might be nice just to configure and have it generate if needed.

jkhales (approved):
exciting PR!

Also did you mean to include the empty `verify.ts` file?

packages/code-signing-certificates/__tests__/utils-test.ts
```diff
@@ -0,0 +1,13 @@
+import { toPositiveHex } from '../src/utils';
+
+describe(toPositiveHex, () => {
+  test.each([
+    ['646a60245faa852fe108ecbbd018dfc8', '646a60245faa852fe108ecbbd018dfc8'],
+    ['eb668fcee52ce9ebd5461a2e71be1269', '6b668fcee52ce9ebd5461a2e71be1269'], // "negative" hex string
+    ['6', '6'],
+    ['e', '6'], // "negative" hex string
+    ['9', '1'], // "negative" hex string
```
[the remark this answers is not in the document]
ide: 9 = 0b1001
Make 0b1001 positive = 0b0001 = 1

wschurman:
> Separate from the code in this PR, we should discuss with the team what the flow looks like. Instead of having to generate and then configure, it might be nice just to configure and have it generate if needed.

I originally had a single command, but separation into two ended up being cleaner because the set of arguments that are needed for generation are very different from those to configure. For example, if the developer has their own key, they shouldn't need to worry about a `commonName` argument (since that's used for generation). I have a feeling we'll be adding more arguments to generation as well of the same nature.

ide:
We might want to have an interactive prompt that asks for info about generating a certificate when configuring one. The idea is that people (not machines) will run these commands in most cases.

One other thing is that @EvanBacon is moving the CLI code into the expo package, out of expo-cli. So the couple of files that call into the codesigning package will need to move to the expo package eventually.

wschurman:
> We might want to have an interactive prompt that asks for info about generating a certificate when configuring one. The idea is that people (not machines) will run these commands in most cases.

That's a good point. I could see a machine running the configuration command if the developer wanted to switch out certificates automatically (in CI for example), but definitely not the generation command.

> One other thing is that @EvanBacon is moving the CLI code into the expo package, out of expo-cli. So the couple of files that call into the codesigning package will need to move to the expo package eventually.

I'm not too familiar with this effort, but I assume it is to use `npx`. I think it'll be fine to copy them over once this has settled.

ide:
Yep the idea is to enable `npx expo`/`yarn expo`, which will be super cheap to run because `expo` is already installed in each project so `npx` knows to use the local copy.

packages/expo-cli/src/commands/codesigning/configure.ts
wschurman: Correct, expo-updates prescribes the certificate format and fields in app.json for use by the config plugin to configure the native configuration.

The spec doesn't actually prescribe anything but the use of certificates and private keys to sign manifests. i.e. it doesn't specify how to get the certificate into the app or whether the expo-updates library needs to be used.

yarn.lock
```diff
@@ -9501,10 +9501,10 @@ he@1.2.0, he@^1.2.0:
   resolved "https://registry.yarnpkg.com/he/-/he-1.2.0.tgz#84ae65fa7eafb165fddb61566ae14baf05664f0f"
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
-hermes-engine@~0.7.0:
-  version "0.7.2"
-  resolved "https://registry.yarnpkg.com/hermes-engine/-/hermes-engine-0.7.2.tgz#303cd99d23f68e708b223aec2d49d5872985388b"
-  integrity sha512-E2DkRaO97gwL98LPhgfkMqhHiNsrAjIfEk3wWYn2Y31xdkdWn0572H7RnVcGujMJVqZNJvtknxlpsUb8Wzc3KA==
+hermes-engine@0.0.0, hermes-engine@~0.7.0:
```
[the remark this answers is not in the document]
wschurman: No idea. But when I run `yarn` on clean master it applies the same change. Looks like https://github.com/expo/expo-cli/commit/d04554d3a65596ef731258e581969b67ccd2a375 removed this entry.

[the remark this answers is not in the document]
wschurman: https://github.com/expo/expo-cli/pull/4184

wschurman:
Talked to @EvanBacon. Requested changes:
- Move these commands into the expo-updates package (in its own CLI) since these are specific for configuring that library. We can declare the CLI to have a bin so that within a project that has `expo-updates` installed we can run `yarn expo-updates codesigning:configure` (or something like that). I still need to look up how to accomplish this.
- Move the `@expo/code-signing-certificates` package either into its own repo or into the eas-cli repo.

PR #4170, opened 2022-02-01
[chore] update packages readme files

Simek:
# Why

Lately I have spotted that not all Readme files includes the correct package names in header, also in some cases, the Readme files looks a bit empty.

# How

This PR updates most of the packages Readme files using the common header template which can be found in few of the packages already.

I have also added missing languages to the few code block and tweaked the main description in few cases to better fit with the header.

# Test Plan

I have used the live preview, when editing markdown files and will test this on GitHub, after creating this PR.

### DIFF
--- a/packages/osascript/README.md
+++ b/packages/osascript/README.md
@@ -3,7 +3,7 @@
 👋 Welcome to <br><code>@expo/osascript</code>
 </h1>
 
-<p align="center">Tools for running an osascripts in Node.</p>
+<p align="center">Tools for running osascripts in Node.</p>
 
 <p align="center">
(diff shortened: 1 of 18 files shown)


EvanBacon (approved):
ok

packages/osascript/README.md
```diff
@@ -3,7 +3,7 @@
 👋 Welcome to <br><code>@expo/osascript</code>
 </h1>
 
-<p align="center">A library for reading and writing JSON files.</p>
+<p align="center">Tools for running an osascripts in Node.</p>
```
kbrandwijk, line 6: `an osascript` or `osascripts`

kbrandwijk (changes requested):

PR #4174, opened 2022-02-02
[cli][package-manager] documentation and proper handling of passing unparsed options to `expo install`

kbrandwijk:
# Why

Resolves https://github.com/expo/expo-cli/issues/2135.

# How

Extended help options and made it so that parameters are not counted as packages, and are passed separately to the package manager so that both `yarn` and `npm` use them properly, extended the package managers to accept extra parameters.

# Test Plan

Ran the commands:
- `expo install --help`
- `expo install expo-updates -- --verbose`
- `expo install expo-updates --yarn -- --verbose`
- `expo install expo-updates --npm -- --loglevel verbose`


packages/package-manager/src/NodePackageManagers.ts
```diff
@@ -96,19 +96,28 @@ export class NpmPackageManager implements PackageManager {
     await this._runAsync(['install', '--global', ...names]);
   }
 
-  async addAsync(...names: string[]) {
+  async addWithParametersAsync(names: string[], parameters: string[] = []) {
     if (!names.length) return this.installAsync();
```
EvanBacon, line 100: In this case, wouldn't you want to still use parameters?

kbrandwijk: Yes, you would. Good catch.

EvanBacon (approved):

EvanBacon:
Would be useful to add a test or two.

EvanBacon (approved):

PR #4179, opened 2022-02-02
[expo-cli] check expo-updates is installed before a publish

jkhales:
# Why

As of sdk 44 we no longer include expo-updates.

# How

Prevent accidental publishes.

# Test Plan

Tested in demo repo.

### DIFF
--- a/packages/expo-cli/src/commands/publish/publishAsync.ts
+++ b/packages/expo-cli/src/commands/publish/publishAsync.ts
@@ -6,15 +6,18 @@ import {
   PackageJSONConfig,
   ProjectTarget,
 } from '@expo/config';
+import spawnAsync from '@expo/spawn-async';
 import chalk from 'chalk';
 import fs from 'fs';
 import { Ora } from 'ora';
 import path from 'path';
+import resolveFrom from 'resolve-from';
 import { Project, UserManager } from 'xdl';
 
 import CommandError, { ErrorCodes } from '../../CommandError';
 import Log from '../../log';
 import { logNewSection } from '../../utils/ora';
+import { confirmAsync } from '../../utils/prompts';
 import * as TerminalLink from '../utils/TerminalLink';
 import { formatNamedWarning } from '../utils/logConfigWarnings';
 import * as sendTo from '../utils/sendTo';
@@ -41,6 +44,7 @@ export async function actionAsync(
   const { exp, pkg } = getConfig(projectRoot, {
     skipSDKVersionRequirement: true,
   });
+  await confirmExpoUpdatesInstalledAsync(projectRoot);
   assertUpdateURLCorrectlyConfigured(exp);
   const { sdkVersion, runtimeVersion } = exp;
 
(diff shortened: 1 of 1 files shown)


packages/config/src/Config.ts
```diff
@@ -140,9 +140,7 @@ export function getConfig(projectRoot: string, options: GetConfigOptions = {}):
       if (configWithDefaultValues.exp.hooks) {
         delete configWithDefaultValues.exp.hooks;
       }
-      if (configWithDefaultValues.exp.ios?.config) {
-        delete configWithDefaultValues.exp.ios.config;
-      }
+      delete configWithDefaultValues.exp.ios?.config;
```
brentvatne, line 143: looks like unrelated change

packages/expo-cli/src/commands/publish/publishAsync.ts
```diff
@@ -150,6 +154,25 @@ function assertValidReleaseChannel(releaseChannel?: string): void {
...
+  const packageJsonPath = resolveFrom.silent(projectDir, './package.json');
+  if (packageJsonPath) {
+    const expoPackageJson = JsonFile.read(packageJsonPath, { json5: true });
+    if (expoPackageJson.dependencies && (expoPackageJson.dependencies as any)['expo-updates']) {
+      return;
+    }
+  }
+
+  throw new Error(
+    `Expo CLI is not installed in this project. Please run "expo install expo-updates".`
```
brentvatne, line 172: - i believe we should warn and prompt instead of hard fail here to account for one of the possible strange edge cases, so we don't block anybody.
- i suspect you meant to write "expo-updates" and not "Expo CLI"
- i realize now that we don't have a docs page with info about installing expo-updates which accounts for both managed and bare projects. in managed apps people just need to run `expo install expo-updates` but in bare apps we may want to link them to [this docs page](https://docs.expo.dev/bare/installing-updates/)

packages/expo-cli/src/commands/publish/publishAsync.ts
```diff
@@ -150,6 +154,25 @@ function assertValidReleaseChannel(releaseChannel?: string): void {
   }
 }
 
+function assertExpoUpdatesInstalled(projectDir: string, sdkVersion?: string): void {
+  // before sdk 44, expo-update was included in with the expo module
+  if (sdkVersion && semver.lt(sdkVersion, '44.0.0')) {
```
brentvatne, line 159: if someone is using expo-updates prior to sdk 44 in a bare project they would have react-native-unimodules instead of the expo package, and in that case their sdkVersion would be undefined. I'm not sure why someone would run `expo publish` in that context, though.

i'd prefer to skip any validation we're doing here in the context that the sdkVersion is undefined for now because we're specifically trying to address sdk 44+ defaulting to no expo-updates in this pr.

packages/expo-cli/src/commands/publish/publishAsync.ts
```diff
@@ -150,6 +154,25 @@ function assertValidReleaseChannel(releaseChannel?: string): void {
...
+    return;
+  }
+
+  const packageJsonPath = resolveFrom.silent(projectDir, './package.json');
+  if (packageJsonPath) {
+    const expoPackageJson = JsonFile.read(packageJsonPath, { json5: true });
+    if (expoPackageJson.dependencies && (expoPackageJson.dependencies as any)['expo-updates']) {
+      return;
+    }
+  }
```
brentvatne, line 169: we have existing code for handling reading the package.json - in the `actionAsync` function above the `pkg` is returned from `getConfig`, for example. you could pass that pkg into this function instead of projectDir.

also, checking `dependencies` may not be a completely dependable way of determining whether the library is installed -- maybe it's accidentally installed as a dev dependency, in which case it would still work fine. there may be some edge cases in which the user is depending on expo-updates via a transitive dep (unlikely but with enough users is bound to happen) or they use something like nx for their monorepo, which i believe centralizes dependencies in a monorepo root package.json. see next comment for follow up on this point.

brentvatne (changes requested):

packages/expo-cli/src/commands/publish/publishAsync.ts
kbrandwijk: Behavior should be similar to how we ask about installing expo-dev-client (on a managed project only) when eas build kicks off a build with `developmentClient: true`. There's a detection there for managed/bare as well with behavior like you describe.

kbrandwijk: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/build/utils/devClient.ts#L16

packages/expo-cli/src/commands/publish/publishAsync.ts
kbrandwijk: Referring to `expo-dev-client` detecting in eas, that's simply done by finding `expo-dev-client/package.json` beneath the project root (so in node_modules): https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/build/utils/devClient.ts#L91-L99

packages/expo-cli/src/commands/publish/publishAsync.ts
```diff
@@ -150,6 +154,25 @@ function assertValidReleaseChannel(releaseChannel?: string): void {
...
+
+  throw new Error(
```
EvanBacon, line 171:
```suggestion
  throw new CommandError(
```
Use ⌘ error so it logs pretty in the CLI

packages/expo-cli/src/commands/publish/publishAsync.ts
```diff
@@ -150,6 +154,25 @@ function assertValidReleaseChannel(releaseChannel?: string): void {
...
+function assertExpoUpdatesInstalled(projectDir: string, sdkVersion?: string): void {
+  // before sdk 44, expo-update was included in with the expo module
+  if (sdkVersion && semver.lt(sdkVersion, '44.0.0')) {
+    return;
+  }
+
+  const packageJsonPath = resolveFrom.silent(projectDir, './package.json');
+  if (packageJsonPath) {
+    const expoPackageJson = JsonFile.read(packageJsonPath, { json5: true });
+    if (expoPackageJson.dependencies && (expoPackageJson.dependencies as any)['expo-updates']) {
```
EvanBacon, line 166:
```suggestion
    if ((expoPackageJson.dependencies as any)?.['expo-updates']) {
```
Or something of this nature

packages/expo-cli/src/commands/publish/publishAsync.ts
EvanBacon, line 159: Probably want to use Versions from xdl since it accounts for `UNVERSIONED` and asserts when the semver is invalid in a controlled way.

jkhales:
> i believe we should warn and prompt instead of hard fail here to account for one of the possible strange edge cases, so we don't block anybody.

Followed the confirmAsync pattern @kbrandwijk suggested.

brentvatne (approved):

PR #4183, opened 2022-02-04
[Interstitial page] Improving user experience

lukmccall:
# Why

Adjusting a couple of things according to the feedback https://www.notion.so/expo/Interstitial-page-testing-c4ee6e71e936460c8df786410d33e3fe

# How

- Replaced the link in the warning with the new one. 
- Detected the platform from the user-agent header if the get parameter wasn't provided 
- `Development App` -> `Development Build`
- Update QR code to point to the interstitial page if needed. 
- The deep links generated by the interstitial page aren't local anymore. 

# Test Plan

Test scenarios:

- app without dev-client
- check if the expo go will be installed if not present
- check what happens if the development build isn't present
- check what happens if both development build and expo go isn't available
- test QR code

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.ts
+++ b/packages/expo-cli/src/commands/start/TerminalUI.ts
@@ -5,6 +5,8 @@ import chalk from 'chalk';
 import wrapAnsi from 'wrap-ansi';
 import {
   Android,
+  Env,
+  isDevClientPackageInstalled,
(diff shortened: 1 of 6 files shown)


tcdavis (approved):
Looking good to me :)
I think we can still improve things around steering people away from needing to see the interstitial at all or when they don't have a build installed already, but we can handle that later

PR #4184, opened 2022-02-04
[expo-cli] Run `yarn`

wschurman:
# Why

Seems the lockfile is out of date. Looks like d04554d may have accidentally removed the hermes-engine 0.0.0 resolution. Or maybe there's a bug in yarn or something.

Either way, this PR just is a clean checkout that runs `yarn`.

# How

`yarn`

# Test Plan

`yarn`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9501,10 +9501,10 @@ he@1.2.0, he@^1.2.0:
   resolved "https://registry.yarnpkg.com/he/-/he-1.2.0.tgz#84ae65fa7eafb165fddb61566ae14baf05664f0f"
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
-hermes-engine@~0.7.0:
-  version "0.7.2"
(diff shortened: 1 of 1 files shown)


ide:
I don't see a problem but don't understand why the lockfile is the way it is, deferring to Evan.

yarn.lock
```diff
@@ -9501,10 +9501,10 @@ he@1.2.0, he@^1.2.0:
   resolved "https://registry.yarnpkg.com/he/-/he-1.2.0.tgz#84ae65fa7eafb165fddb61566ae14baf05664f0f"
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
-hermes-engine@~0.7.0:
-  version "0.7.2"
-  resolved "https://registry.yarnpkg.com/hermes-engine/-/hermes-engine-0.7.2.tgz#303cd99d23f68e708b223aec2d49d5872985388b"
-  integrity sha512-E2DkRaO97gwL98LPhgfkMqhHiNsrAjIfEk3wWYn2Y31xdkdWn0572H7RnVcGujMJVqZNJvtknxlpsUb8Wzc3KA==
+hermes-engine@0.0.0, hermes-engine@~0.7.0:
```
EvanBacon, line 9504: This changes unreliably since we use a yarn resolution in the root, ideally we'd just have 0.0.0 since it's an empty package and we don't perform native builds in this repo.

PR #4187, opened 2022-02-04
[expo-cli] Publish command correctly passes max-workers option through to metro

scottbartell:
# Why

```
expo publish --max-workers 1
``` 

In the publish command the `max-workers` option is not being passed through to metro causing it to ignore the option and always use the default.

This is causing our CI build to run out of memory when attempting to publish even when specifying to use `--max-workers 1`

# Test Plan

This area of the code did not appear to be tested so I took a stab at adding a test for at least part of it. Happy to add more tests or change or remove it.

### DIFF
--- a/packages/xdl/src/project/__tests__/createBundlesAsync-test.ts
+++ b/packages/xdl/src/project/__tests__/createBundlesAsync-test.ts
@@ -1,0 +1,61 @@
+import { getConfig, isLegacyImportsEnabled, ProjectTarget } from '@expo/config';
+import { bundleAsync } from '@expo/dev-server';
+
+import { getLogger } from '../ProjectUtils';
+import { createBundlesAsync } from '../createBundlesAsync';
+
+jest.mock('@expo/dev-server');
+jest.mock('@expo/config');
+jest.mock('../../tools/resolveEntryPoint');
+jest.mock('../ProjectUtils');
+
+describe(createBundlesAsync, () => {
+  const expoConfig = {
+    name: 'my-app',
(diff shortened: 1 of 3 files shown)


EvanBacon (approved):
lgtm, nice work on the test

packages/xdl/src/project/__tests__/createBundlesAsync-test.ts
```diff
@@ -0,0 +1,61 @@
+import { getConfig, isLegacyImportsEnabled, ProjectTarget } from '@expo/config';
+import { bundleAsync } from '@expo/dev-server';
+
+import { getLogger } from '../ProjectUtils';
+import { createBundlesAsync } from '../createBundlesAsync';
+
+jest.mock('@expo/dev-server');
+jest.mock('@expo/config');
+jest.mock('../../tools/resolveEntryPoint');
+jest.mock('../../project/projectUtils');
```
kbrandwijk, line 10: Automated tests are failing because this path cannot be found.

kbrandwijk (changes requested):

packages/xdl/src/project/__tests__/createBundlesAsync-test.ts
scottbartell: I think this was related to a case sensitivity issue- it should be `ProjectUtils`. (It was passing for me locally with the improper case because MacOS is case insensitive while linux/CI is not)

kbrandwijk (approved):

EvanBacon (approved):

Issue #4185, opened 2022-02-04
Can't start web in production mode with in v5.0.3


CarlosHdez:
### Summary

After upgrading the `expo-cli` version from 4.7.1 to 5.0.3 I can't run `expo start --web --no-dev`.
I've created a project from scratch and the issue is present there.

 

### Environment

  Expo CLI 5.0.3 environment info:
    System:
      OS: macOS 11.6
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 12.22.6 - ~/.nvm/versions/node/v12.22.6/bin/node
      Yarn: 1.22.15 - ~/.nvm/versions/node/v12.22.6/bin/yarn
      npm: 6.14.15 - ~/.nvm/versions/node/v12.22.6/bin/npm
    Managers:
      CocoaPods: 1.11.2 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
    IDEs:
      Android Studio: 2020.3 AI-203.7717.56.2031.7678000
      Xcode: 13.2.1/13C100 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~44.0.0 => 44.0.6
      react: 17.0.1 => 17.0.1
      react-dom: 17.0.1 => 17.0.1
      react-native: 0.64.3 => 0.64.3
      react-native-web: 0.17.1 => 0.17.1
    npmGlobalPackages:
      expo-cli: 5.0.3
    Expo Workflow: managed

### Please specify your device/emulator/simulator platform, model and version

web

### Error output

```
Developer tools running on http://localhost:19002
Opening developer tools in the browser...
Starting Webpack on port 19006 in production mode.
Cannot read property 'before' of undefined
TypeError: Cannot read property 'before' of undefined
    at Object.startAsync (/Users/user/.nvm/versions/node/v12.22.6/lib/node_modules/expo-cli/node_modules/xdl/src/Webpack.ts:273:44)
    at startAsync (/Users/user/.nvm/versions/node/v12.22.6/lib/node_modules/expo-cli/node_modules/xdl/src/start/startAsync.ts:64:5)
 ```

### Reproducible demo or steps to reproduce from a blank project

1. `expo init test --npm`
2. `cd test`
3. `expo start --web --no-dev` or `expo start:web --no-dev`

brentvatne:
[...]

PR #4188, opened 2022-02-05
Fix running `expo web --no-dev`

EvanBacon:
# Why

- fix https://github.com/expo/expo-cli/issues/4185

# How

- `devServer.before` isn't defined in production Webpack configs, account for this.

# Test Plan

- `expo web --no-dev` should work as expected.


kbrandwijk (commented):
Would using `devServer.setupMiddlewares`, which doesn't depend on the existence of `before`, nor requires manually injecting something into that method, be a cleaner option?

EvanBacon:
@kbrandwijk this all changes after Webpack 5 upgrade, just need the temporary fix.

kbrandwijk (approved):

PR #4209, opened 2022-02-11
[Interstitial page] Try to detect if a deep link succeed

lukmccall:
# Why

Currently, if someone scans the QR code, we can't detect if the device can handle a deep link to `Expo Go` or `Development Build`. So we're trying to detect if the redirection was successful. If not we display a message. 

# How

When the user clicks, we trigger a timeout. After some time, we check if the document became inactive (if so, we know that the redirection was successful). This flow gets more complicated on iOS because the user will see a modal. In that case, we need to fire the timeout after the document will be focused. 

# Test Plan

- manual tests with Android device and iOS simulator.

### DIFF
--- a/packages/xdl/static/loading-page/index.html
+++ b/packages/xdl/static/loading-page/index.html
@@ -270,10 +270,10 @@
 
   const showAlert = (isExpoGo) => {
     if (isExpoGo) {
-      alertElement.innerHTML = 'It seems that deep-link into your application failed. Make sure that you have the <b>Expo Go</b> installed.' + 
-      '<br><a href="https://docs.expo.dev/get-started/installation/#2-expo-go-app-for-ios-and">Learn more.</a>';
+      alertElement.innerHTML = 'Unable to open in Expo Go. Please install <b>Expo Go</b> to continue.' + 
+      '<br><a href="https://expo.dev/client">Learn more.</a>';
     } else {
-      alertElement.innerHTML = 'It seems that deep-link into your application failed. Make sure that you have the <b>Development Build</b> installed.' + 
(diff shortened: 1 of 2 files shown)


packages/xdl/static/loading-page/index.html
```diff
@@ -248,6 +266,24 @@
   </div>
   
   <script>
+  const alertElement = document.getElementById("alert");
+
+  const showAlert = (isExpoGo) => {
+    if (isExpoGo) {
+      alertElement.innerHTML = 'It seems that deep-link into your application failed. Make sure that you have the <b>Expo Go</b> installed.' +
```
tcdavis, line 273:
```suggestion
      alertElement.innerHTML = 'Unable to open in Expo Go. Please install <b>Expo Go</b> to continue.' + 
```
and link to https://expo.dev/client

packages/xdl/static/loading-page/index.html
```diff
@@ -248,6 +266,24 @@
...
   
   <script>
+  const alertElement = document.getElementById("alert");
+
+  const showAlert = (isExpoGo) => {
+    if (isExpoGo) {
+      alertElement.innerHTML = 'It seems that deep-link into your application failed. Make sure that you have the <b>Expo Go</b> installed.' + 
+      '<br><a href="https://docs.expo.dev/get-started/installation/#2-expo-go-app-for-ios-and">Learn more.</a>';
+    } else {
+      alertElement.innerHTML = 'It seems that deep-link into your application failed. Make sure that you have the <b>Development Build</b> installed.' +
```
tcdavis, line 276:
```suggestion
      alertElement.innerHTML = 'Unable to open in Development Build with [scheme]. Please build and install a compatible <b>Development Build</b> to continue.' + 
```

tcdavis (changes requested):
Yay, this is much nicer 🙏

tcdavis (approved):

PR #4210, opened 2022-02-12
[tests] fix tests

kbrandwijk:
# Why

There are a few tests that do not run locally because of hardcoded paths, or things that are not mocked. This PR is to make sure that all tests pass locally, regardless of local dev environment.

# How

- [X] webpack-config tests were failing on WSL, I mocked the WSL detection for the tests
- [X] config e2e test was failing due to hardcoded path in snapshot, switched to an inline snapshot with a dynamically injected project root.
- [X] lots of similar path issues in other tests, see individual commits for details

# Test Plan

Run all automated tests locally and see them pass. Optionally find people on different platforms to do the same.


packages/config/src/__tests__/ConfigParsing-test.ts
```diff
@@ -136,9 +136,9 @@ describe(getConfig, () => {
       expect(exp.foo).toBe('bar+value');
       expect(exp.name).toBe('rewrote+ts-config-test');
       expect(exp._internal).toStrictEqual({
-        dynamicConfigPath: 'ts/app.config.ts',
+        dynamicConfigPath: `ts${path.sep}app.config.ts`,
```
EvanBacon, line 139: I think there's a `path.*` method for parsing this, would look a bit prettier.

kbrandwijk: `slash` is elegant 👍🏻

EvanBacon:
Would you be interested in fixing and landing this?

kbrandwijk:
Closing this because there's been too many changes, with modules taken out of the repo, and other refactorings, and limited value to these improvements.

PR #4214, opened 2022-02-15
[Interstitial page] Fix QR on web UI & fallback to SDK version if the runtime version wasn't defined

lukmccall:
# Why

Fixes that QR code on web UI pointed to the `Expo Go` directly instead of to the interstitial page.
Fallback to the SDK version if the runtime version wasn't defined. 

# How

- Extended the dev graphQL server to serve Interstitial page URL if available. 

# Test Plan

- runs the local version of the CLI and confirms that the web QR code points to the correct place.
- checks if the SDK version was detected when the runtime version isn't available.


tcdavis (approved):
LGTM 🙏

esamelson (approved):
🔥

PR #4218, opened 2022-02-16
[expo-cli][xdl] remove env flag for interstitial page

esamelson:
# Why

we are ready to try releasing this feature to production

# How

~Remove the `Env.isInterstitiaLPageEnabled()` method and all uses of it~

Removing the function entirely will conflict with https://github.com/expo/expo-cli/pull/4214 (which adds another call to it), so just return true for now, and we can clean it up later on

# Test Plan

Ran local `expo start` and scanned QR code on Android device:
✅ in a project with expo-dev-client installed, scanning the QR code opened the interstitial page
✅ in a project without expo-dev-client installed, scanning the QR opened the project in Expo Go
✅ in a project with expo-dev-client, running `expo start --dev-client` and then scanning the QR code opened the project in the dev client (skipping the interstitial page)

### DIFF
--- a/packages/expo-cli/src/commands/start/TerminalUI.ts
+++ b/packages/expo-cli/src/commands/start/TerminalUI.ts
@@ -5,6 +5,7 @@ import chalk from 'chalk';
 import wrapAnsi from 'wrap-ansi';
 import {
   Android,
+  Env,
   isDevClientPackageInstalled,
   Project,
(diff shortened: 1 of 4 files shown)


tcdavis (approved):
Looking good! Thanks for making it easy to land alongside https://github.com/expo/expo-cli/pull/4214

PR #4223, opened 2022-02-18
[expo-cli] use `npm pack` instead of `fetch` for template download

kbrandwijk:
# Why

If you are using a private NPM repository, that is set up for all packages (so not only scoped ones), then `expo init` and `expo prebuild` fail while trying to download the template tarball. This is because `npm view` is used to get the tarball URL, but a normal `fetch` request is used for the actual download, which does not respect `.npmrc` settings.
This issue occurs both locally, and on EAS.

# How

- Switch template download to use `npm pack`, which respects .npmrc settings
- Preserve the functionality for using template cache

# Test Plan

- Configured a free AWS CodeArtifact repository that I set up to use the npm store
- Configured npm globally to use this repository
- Ran `expo init` to verify the error happening
- Ran `expo init` with the proposed changes and verified success
- Went into the newly created project
- Ran `expo prebuild` to verify the error happening
- Ran `expo prebuild with the proposed changes and verified success


packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -53,10 +53,22 @@ function applyKnownNpmPackageNameRules(name: string): string | null {
...
+  cwd: string | undefined = undefined,
+  ...props: string[]
+): Promise<JSONValue> {
+  return npmCommandAsync(['pack', packageName, ...props, '--json'], { cwd });
+}
+
+async function npmCommandAsync(cmd: string[], spawnOpts: SpawnOptions = {}): Promise<JSONValue> {
   const cmdString = `npm ${cmd.join(' ')}`;
   Log.debug('Run:', cmdString);
+  const results = (await spawnAsync('npm', cmd, spawnOpts)).stdout?.trim();
```
EvanBacon, line 70: If this function always parses as json then it should also append the `--json` flag, otherwise there's an untyped requirement to make the function work.

packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -101,10 +113,30 @@ export async function downloadAndExtractNpmModuleAsync(
...
 ): Promise<void> {
-  const url = await getNpmUrlAsync(npmName);
-
-  Log.debug('Fetch from URL:', url);
-  await extractNpmTarballFromUrlAsync(url, props);
+  Log.debug(`Looking for tarball for ${npmName} in ${getCacheFilePath()}...`);
+  try {
+    const cachePath = getCacheFilePath();
+    // Perform dry-run to get actual filename for resolved version
+    const filename = (((await npmPackAsync(
```
EvanBacon, line 120: My understanding is that this command is used `Create a tarball from a package` I don't see how this is analogous to the fetch command.

packages/expo-cli/e2e/__tests__/__snapshots__/export-test.ts.snap
```diff
@@ -6,7 +6,7 @@ Array [
   "dist/assetmap.json (size ignored)",
   "dist/assets (dir)",
   "dist/bundles (dir)",
-  "dist/bundles/android-XXX.js (838 kB)",
+  "dist/bundles/android-XXX.js (837 kB)",
```
EvanBacon, line 9:
```suggestion
  "dist/bundles/android-XXX.js (838 kB)",
```

packages/expo-cli/e2e/__tests__/__snapshots__/export-test.ts.snap
```diff
@@ -55,7 +55,7 @@ Array [
...
```
EvanBacon, line 58:
```suggestion
  "dist/bundles/android-XXX.js (838 kB)",
```

packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -69,42 +73,37 @@ export async function npmViewAsync(...props: string[]): Promise<JSONValue> {
...
+    if (!(await fs.stat(cacheFilename).catch(() => null))?.isFile() ?? false) {
+      Log.debug(`Downloading tarball for ${npmName} to ${cachePath}...`);
+      await npmPackAsync(npmName, cachePath);
+    }
+    await extractLocalNpmTarballAsync(cacheFilename, {
+      cwd: props.cwd,
+      name: npmName,
+    });
+  } catch (error) {
+    Log.error('Error downloading and extracting template package');
```
EvanBacon, line 105:
```suggestion
    Log.error('Error downloading and extracting template package:', error);
```
This could provide more context.

packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -69,42 +73,37 @@ export async function npmViewAsync(...props: string[]): Promise<JSONValue> {
...
+    const cacheFilename = path.join(cachePath, filename);
+
+    // TODO: This cache does not expire, but neither does the FileCache at the top of this file.
+    if (!(await fs.stat(cacheFilename).catch(() => null))?.isFile() ?? false) {
+      Log.debug(`Downloading tarball for ${npmName} to ${cachePath}...`);
+      await npmPackAsync(npmName, cachePath);
+    }
+    await extractLocalNpmTarballAsync(cacheFilename, {
+      cwd: props.cwd,
+      name: npmName,
```
EvanBacon, line 102: Wrong `name` field, the previous version passed the props directly. This `name` will be used to rewrite the template name, by passing in the template name it does breaks the assumption.

packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -69,42 +73,37 @@ export async function npmViewAsync(...props: string[]): Promise<JSONValue> {
...
   npmName: string,
   props: ExtractProps
 ): Promise<void> {
-  const url = await getNpmUrlAsync(npmName);
-
-  Log.debug('Fetch from URL:', url);
-  await extractNpmTarballFromUrlAsync(url, props);
+  Log.debug(`Looking for tarball for ${npmName} in ${getCacheFilePath()}...`);
+  try {
+    const cachePath = getCacheFilePath();
```
wkozyra95, line 85: you need to ensure that `cachePath` exists, that is why those tests are failing

wkozyra95 (changes requested):

packages/expo-cli/src/commands/utils/npm.ts
kbrandwijk: Doh! Thanks!

packages/expo-cli/src/commands/utils/npm.ts
```diff
@@ -69,42 +73,38 @@ export async function npmViewAsync(...props: string[]): Promise<JSONValue> {
...
```
wkozyra95, line 85: I have some concerns about this location:
- old way is eventually cleaning old entries in cache, your approach is leaving them forever
- `cachePath` is used for `FileSystemCache` already, adding other files won't break anything, but mixing different stuff like that is not a good approach.

I would recommend putting that in tmp instead

wkozyra95 (approved):

packages/expo-cli/src/commands/utils/npm.ts
kbrandwijk: Good point, but I suggest we make that change when we port this over to versioned cli.

Issue #4227, opened 2022-02-22
[expo-cli] pick wrong scheme to open application on Android


giautm:
### Summary

When I config the `react-native-fbsdk-next` with Expo. I found that Expo-CLI always picks the longest scheme to open the app. This behavior is not correct, because the FB scheme has an `android:host` with the value `applinks`. So, It does not match the requested URL `expo-development-client`, then the application can't load the JS bundle.

The correct behavior is: **Expo CLI should pick the scheme with the empty host or it's much exactly matched to `expo-development-client`**

https://github.com/expo/expo-cli/blob/c73681f811e254bfaf40e1459b62cfd3e9a15167/packages/expo-cli/src/commands/run/android/runAndroid.ts#L164-L173

https://github.com/expo/expo-cli/blob/c73681f811e254bfaf40e1459b62cfd3e9a15167/packages/expo-cli/src/commands/run/utils/schemes.ts#L25-L34

```xml
      <intent-filter>
        <action android:name="android.intent.action.VIEW"/>
        <category android:name="android.intent.category.DEFAULT"/>
        <category android:name="android.intent.category.BROWSABLE"/>
        <data android:scheme="baconapp"/>
        <data android:scheme="com.bacon.app"/>
        <data android:scheme="fb4812712715480611" android:host="applinks" /> <!-- Expo CLI should not pick this scheme to open the app -->
      </intent-filter>
```

### Environment

Not related

### Please specify your device/emulator/simulator platform, model and version

Android

### Reproducible demo or steps to reproduce from a blank project

Create a blank project, then edit `android/app/src/main/AndroidManifest.xml` to config the `intent-filter` with the custom host.

giautm:
cc @EvanBacon

PR #4228, opened 2022-02-22
[expo-cli]: pick correct scheme for open android app

giautm:
Expo CLI should pick the scheme with the empty host
or it's much exactly matched to `expo-development-client`

# Why

close https://github.com/expo/expo-cli/issues/4227

# How

# Test Plan

- Unittests

### DIFF
--- a/packages/config-plugins/src/android/Scheme.ts
+++ b/packages/config-plugins/src/android/Scheme.ts
@@ -82,11 +82,11 @@ function propertiesFromIntentFilter(intentFilter: any): IntentFilterProps {
   const categories = intentFilter?.category?.map((data: any) => data?.$?.['android:name']) ?? [];
   const data =
     intentFilter?.data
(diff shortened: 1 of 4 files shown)


EvanBacon (approved):
LGTM, it's a bit unclear if the dev client config plugin has always added the `android:host` though (and this change must be backwards compat for now). CC @lukmccall to verify this works against older versions of dev client.

giautm:
> LGTM, it's a bit unclear if the dev client config plugin has always added the `android:host` though (and this change must be backwards compat for now). CC @lukmccall to verify this works against older versions of dev client.

The dev client does not need to add the `android:host`, and this change was backward compat.

lukmccall:
Yes, I agree with @giautm. This should work fine with previous dev-client versions ;)

lukmccall (approved):

Issue #3998, opened 2021-11-12
EAS Android Build (expo prebuild?) fails if app name contains an apostrophe


dacruz21:
### Summary

If the app name (as defined by the name key in app.json) contains an apostrophe (e.g. `"name": "Let's Break Expo"`), `expo prebuild` will not properly escape the apostrophe when setting the app_name string in strings.xml. This causes EAS builds (`eas build -p android`) to fail.

### Environment

```
  Expo CLI 4.12.1 environment info:
    System:
      OS: Linux 5.14 Arch Linux
      Shell: 5.8 - /usr/bin/zsh
    Binaries:
      Node: 16.11.1 - /usr/bin/node
      Yarn: 1.22.17 - /usr/bin/yarn
      npm: 8.1.3 - /usr/bin/npm
    IDEs:
      Android Studio: AI-203.7717.56.2031.7784292
    npmPackages:
      expo: ^42.0.0 => 42.0.5
      react: 16.13.1 => 16.13.1
      react-dom: 16.13.1 => 16.13.1
      react-native: 0.63.4 => 0.63.4
      react-native-web: ~0.13.12 => 0.13.18
    Expo Workflow: managed
```

### Please specify your device/emulator/simulator platform, model and version

n/a

### Error output

```
[stderr] FAILURE: Build failed with an exception.

[stderr] * What went wrong:

[stderr] Execution failed for task ':app:mergeReleaseResources'.

[stderr] > A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade

[stderr]    > Android resource compilation failed

[stderr]      /build/workingdir/build/mobile/android/app/src/main/res/values/strings.xml:3:3-46: AAPT: error: unescaped apostrophe in string

[stderr]          

[stderr]      /build/workingdir/build/mobile/android/app/src/main/res/values/strings.xml:3:3-46: AAPT: error: not a valid string.

[stderr]          

[stderr]      /build/workingdir/build/mobile/android/app/build/intermediates/incremental/mergeReleaseResources/merged.dir/values/values.xml: AAPT: error: file failed to compile.
```

### Reproducible demo or steps to reproduce from a blank project

- `expo init`
- Edit app.json, set app name to "Let's Break Expo"
- `eas build -p android`
- `Run gradlew` step in EAS will fail with the above error

dacruz21:
[...]

PR #4229, opened 2022-02-22
[config-plugins] Unescape values when reading strings.xml

wkozyra95:
# Why

reading and writing string resources is escaping characters twice.

# How

This is not a proper fix, the correct approach would be to remove escaping from `writeXMLAsync`, but there is probably a bunch of places that rely on this behavior.

this PR is attempting to revert changes that `escapeAndroidString` will do on write

# Test Plan

added test
verified with local builds by applying a similar patch in eas-build code directly
https://github.com/expo/eas-build/pull/76

### DIFF
--- a/packages/config-plugins/src/utils/__tests__/XML-test.ts
+++ b/packages/config-plugins/src/utils/__tests__/XML-test.ts
@@ -1,52 +1,90 @@
+import fs from 'fs-extra';
 import { vol } from 'memfs';
 
 import { buildResourceItem, readResourcesXMLAsync } from '../../android/Resources';
 import { setStringItem } from '../../android/Strings';
-import { escapeAndroidString, format } from '../XML';
+import { escapeAndroidString, format, unescapeAndroidString, writeXMLAsync } from '../XML';
 
 jest.mock('fs');
 
 export const sampleStringsXML = `
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <resources>
-  <string name="app_name">expo &amp;bo&lt;y&gt;&apos;</string>
+  <string name="app_name">exp\\'o &amp;bo&lt;y&gt;&apos;</string>
 </resources>`;
 
-beforeAll(async () => {
-  const directoryJSON = {
-    './android/app/src/main/res/values/strings.xml': sampleStringsXML,
-  };
-  vol.fromJSON(directoryJSON, '/app');
+describe(readResourcesXMLAsync, () => {
+  beforeAll(async () => {
+    const directoryJSON = {
+      './android/app/src/main/res/values/strings.xml': sampleStringsXML,
+    };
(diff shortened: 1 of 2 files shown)


dsokal (approved):

packages/config-plugins/src/utils/__tests__/XML-test.ts
```diff
@@ -1,52 +1,90 @@
...
-import { escapeAndroidString, format } from '../XML';
+import { escapeAndroidString, format, unescapeAndroidString, writeXMLAsync } from '../XML';
 
 jest.mock('fs');
 
 export const sampleStringsXML = `
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <resources>
-  <string name="app_name">expo &amp;bo&lt;y&gt;&apos;</string>
+  <string name="app_name">exp\\'o &amp;bo&lt;y&gt;&apos;</string>
```
EvanBacon, line 13: Why reduce the complexity of the input test string?

wkozyra95: what do you mean reduce? adding `\'` complicates this example

If you are talking about why I didn't use this in that other test, then it's because test would not pass:
- `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` was removed when reading and writing
- there was sth wrong with `&amp;bo&lt;y&gt;&apos;` part but I don't remember exactly.

As I mentioned in a PR description this is not a proper fix, I currently don't have the bandwidth to work on that, so I'm just patching this case. 

btw we have a similar change in eas-build code directly and I know this is a hacky solution, so it is fine from our side if you don't want to merge it.

EvanBacon: Yes I meant to have the original test be complicated as a way to ensure a series of issues ([including this](https://github.com/expo/expo-cli/issues/3998)) would be fixed, if you change the functionality and the tests break then it potentially means the fixes are broken too.

wkozyra95: Ok I'm still not sure what you want me to do here, do you want me to leave `sampleStringsXML` as it was?

> series of issues (https://github.com/expo/expo-cli/issues/3998)

this PR is a fix for that exact case, if I read XML using `readXMLAsync` and write it using `writeXMLAsync` the file will be malformed if the name contained an apostrophe. (caused by this code https://github.com/expo/expo-cli/blob/main/packages/config-plugins/src/utils/XML.ts#L52)

> if you change the functionality and the tests break then it potentially means the fixes are broken too.

my changes did not break this test, I added additional `\'` just to complicate it more(if that is a problem I can remove it)

The part I mentioned that did not work is if I use this example in the new test I added that reads and writes XML(using `readXMLAsync/writeXMLAsync`), but this probably never worked.

EvanBacon (approved):

PR #4238, opened 2022-02-28
[config-plugin] append scheme REVERSED_CLIENT_ID from GoogleService-Info.plist

giautm:
# Why

While writing a new plug-in for Google Sign In, I noticed that `ios.config.googleSignIn.reservedClientId` is used to add a new scheme to the iOS configuration.

We can do this automatically by reading the `REVERSED_CLIENT_ID` field from the `GoogleService-Info.plist` file

# How

- Update `withGoogle` plugin to read the field from the `GoogleService-Info.plist` file.
- We still respect the old `ios.config.googleSignIn.reservedClientId` config for backward compatible.
- *Next step*: Update the Google SignIn module to remove mention about `ios.config.googleSignIn.reservedClientId`.

# Test Plan

- Unit test

### DIFF
--- a/packages/config-plugins/src/ios/Google.ts
+++ b/packages/config-plugins/src/ios/Google.ts
@@ -6,13 +6,18 @@ import path from 'path';
 import { XcodeProject } from 'xcode';
 
 import { ConfigPlugin, ModProps } from '../Plugin.types';
-import { createInfoPlistPlugin, withXcodeProject } from '../plugins/ios-plugins';
+import { withInfoPlist, withXcodeProject } from '../plugins/ios-plugins';
 import { InfoPlist } from './IosConfig.types';
 import { getSourceRoot } from './Paths';
 import { appendScheme } from './Scheme';
 import { addResourceFileToGroup, getProjectName } from './utils/Xcodeproj';
 
-export const withGoogle = createInfoPlistPlugin(setGoogleConfig, 'withGoogle');
+export const withGoogle: ConfigPlugin = config => {
+  return withInfoPlist(config, config => {
+    config.modResults = setGoogleConfig(config, config.modResults, config.modRequest);
+    return config;
+  });
+};
 
 export const withGoogleServicesFile: ConfigPlugin = config => {
   return withXcodeProject(config, config => {
--- a/packages/config-plugins/src/ios/Google.ts
+++ b/packages/config-plugins/src/ios/Google.ts
@@ -33,14 +33,10 @@ function readGoogleServicesInfoPlist(
   relativePath: string,
   { projectRoot }: { projectRoot: string }
 ) {
-  try {
-    const googleServiceFilePath = path.resolve(projectRoot, relativePath);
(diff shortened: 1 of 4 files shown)


giautm:
cc @EvanBacon for review.

PS: I can't request the review from the UI because I don't have repo access.

packages/config-plugins/src/plugins/ios-plugins.ts
```diff
@@ -17,7 +21,7 @@ type MutateInfoPlistAction = (expo: ExpoConfig, infoPlist: InfoPlist) => InfoPli
 export function createInfoPlistPlugin(action: MutateInfoPlistAction, name?: string): ConfigPlugin {
   const withUnknown: ConfigPlugin = config =>
     withInfoPlist(config, async config => {
-      config.modResults = await action(config, config.modResults);
+      config.modResults = await action(config, config.modResults, config.modRequest);
```
EvanBacon, line 24: This changes the symmetry of the other methods, should just use `withInfoPlist` directly where needed.

packages/config-plugins/src/ios/Google.ts
```diff
@@ -22,8 +24,40 @@ export const withGoogleServicesFile: ConfigPlugin = config => {
...
+  relativePath: string,
+  { introspect, projectRoot }: { introspect: boolean; projectRoot: string }
+) {
+  try {
+    const googleServiceFilePath = path.resolve(projectRoot, relativePath);
+    const contents = fs.readFileSync(googleServiceFilePath, 'utf8');
+    assert(contents, 'GoogleService-Info.plist is empty');
+    return plist.parse(contents);
+  } catch (error) {
+    if (introspect) {
```
EvanBacon, line 37: Not sure if this is needed, if a value is defined and not found then maybe it should warn/throw and return empty. Skipping on introspect is mostly to cover the cases where a file would exist from a template, for example if you read the Info.plist or entitlements file. The Google Services file should always exist since it isn't part of the template cloning step.

packages/config-plugins/src/plugins/ios-plugins.ts
```diff
@@ -7,7 +7,10 @@ import type { ExpoPlist, InfoPlist } from '../ios/IosConfig.types';
 import type { AppDelegateProjectFile } from '../ios/Paths';
 import { withMod } from './withMod';
 
-type MutateInfoPlistAction = (expo: ExpoConfig, infoPlist: InfoPlist) => InfoPlist;
+type MutateInfoPlistAction = (
+  expo: ExpoConfig,
+  infoPlist: InfoPlist
+) => Promise<InfoPlist> | InfoPlist;
```
giautm, line 13: correct the type definition for the action function, it supports async but the type said no.

giautm:
@EvanBacon I updated the pull request and the CI was green.

Can you make a new release for the `@expo/config-plugins` package? The upstream PR depended on this update.

EvanBacon (approved):

PR #4243, opened 2022-03-03
[config-plugins] Added support for `AppDelegate.mm`

EvanBacon:
# Why

- React Native 68 uses `AppDelegate.mm` instead of `AppDelegate.m` which means the existing prebuild check would assert that vital project files are missing and the Google Maps dangerous plugin would fail.

# How

- Add `objcpp` which is returned when the AppDelegate file extension is `.mm`.

# Test Plan

- In a local project, rename `AppDelegate.m` to `AppDelegate.mm` and rerun `expo prebuild -p ios --no-install` -- no error should be thrown.
- It's unclear what the new template file will look like so we don't have a continuous test in place for the Maps plugin.
- Added a FS unit test.

### DIFF
--- a/packages/config-plugins/src/ios/Maps.ts
+++ b/packages/config-plugins/src/ios/Maps.ts
@@ -180,7 +180,7 @@ const withMapsCocoaPods: ConfigPlugin<{ useGoogleMaps: boolean }> = (config, { u
 
 const withGoogleMapsAppDelegate: ConfigPlugin<{ apiKey: string | null }> = (config, { apiKey }) => {
   return withAppDelegate(config, config => {
-    if (['objc', 'cxx'].includes(config.modResults.language)) {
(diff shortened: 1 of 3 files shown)


kbrandwijk:
Apple calls it 'Objective-C' and 'Objective-C++'. We call the first `objc`, so for the latter `objcpp` would be the most accurate, but I can live with `cpp` as well.

Kudo:
yep, i prefer `objcpp` as well.
reference from pbxproj 

 /* AppDelegate.mm */ = {isa = PBXFileReference; fileEncoding = 4; `lastKnownFileType = sourcecode.cpp.objcpp;` name = AppDelegate.mm; path = RN068/AppDelegate.mm; sourceTree = "<group>"; };

it is `sourcecode.c.objc` for .m files.

Kudo (approved):
awesome! thanks for the change.

PR #4249, opened 2022-03-08
[expo-cli] Move `expo eject` to the deprecated help group

EvanBacon:
Screen Shot 2022-03-08 at 12 21 10 PM

# Why

- Update the description to reflect that users should move to expo prebuild instead.
- We want to deprecate `expo eject` but we can do this naturally by waiting until the `expo-cli` package is deprecated in favor of `npx expo`.
- Waiting until `npx expo` is in beta ready adding a proper notice that would detect the SDK version and prompt the user to use `npx expo prebuild` instead.


kbrandwijk:
Does this replace https://github.com/expo/expo-cli/pull/4154?

PR #4251, opened 2022-03-09
bump webpack-config's react-dev-utils

f4z3k4s:
# Why

Fixes a security risk with immer, more info: https://nvd.nist.gov/vuln/detail/CVE-2021-23436

# How

Dependabot threw me critical security warnings about immer, that came from this package

# Test Plan

I've tried the CLI and worked without issues.


packages/webpack-config/package.json
```diff
@@ -61,7 +61,7 @@
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "pnp-webpack-plugin": "^1.5.0",
     "postcss-safe-parser": "^4.0.2",
-    "react-dev-utils": "~11.0.1",
+    "react-dev-utils": "12.0.0",
```
EvanBacon, line 64: From what I can tell, this version is for Webpack 5, whereas the config uses Webpack 4 meaning this change would probably be incompatible. Have you tried using this locally?

f4z3k4s: Yes, I did try it and it seemed to work fine. I just took a look in the commits of `react-dev-utils` and indeed see that it's intended for Webpack 5 so I guess it's better stay with the current version for now. Thanks.

Issue #4066, opened 2021-12-14
Issue using turborepo and @expo/next-adapter


reggi:
### Summary

The issue is that when using a monorepo such as turborepo or any yarn workspace, for some reason this next-js plugin and it's webpack plugins are giving me an error. This only happens however, when the project is not at the root and webpack is being hoisted outside of the nextjs project package directory. There must be a piece of code specific to looking into `./pages` directory even though i'm passing in ` { projectRoot: __dirname }`. I have tested this and gotten it working when the modules are not hoisted out of the directory via yarn workspace, and it has worked for me. It's only failing within the monorepo framework.

### Environment

Expo CLI 4.13.0 environment info:
    System:
      OS: macOS 11.4
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 17.1.0 - ~/.nvm/versions/node/v17.1.0/bin/node
      Yarn: 1.22.17 - ~/.nvm/versions/node/v17.1.0/bin/yarn
      npm: 8.1.2 - ~/.nvm/versions/node/v17.1.0/bin/npm
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.0.1, iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0
    IDEs:
      Xcode: 13.1/13A1030d - /usr/bin/xcodebuild
    npmGlobalPackages:
      expo-cli: 4.13.0
    Expo Workflow: managed

### Please specify your device/emulator/simulator platform, model and version

n/a

### Error output

ready - started server on 0.0.0.0:3000, url: http://localhost:3000
info  - Disabled SWC as replacement for Babel because of custom Babel configuration "babel.config.js" https://nextjs.org/docs/messages/swc-disabled
info  - automatically enabled Fast Refresh for 1 custom loader
info  - Using external babel configuration from /Users/thomasreggi/Desktop/turbo/y/apps/nextjs/babel.config.js
Warning: useTransformReactJsxExperimental has been renamed to useTransformReactJSXExperimental (capitalized JSX) in react-native@0.64.0
info  - automatically enabled Fast Refresh for 1 custom loader
error - unhandledRejection: HookWebpackError: Cannot read properties of undefined (reading 'replace')
HookWebpackError: Cannot read properties of undefined (reading 'replace')
    at makeWebpackError (/Users/thomasreggi/Desktop/turbo/y/node_modules/next/dist/compiled/webpack/bundle5.js:47168:9)
[...]

PR #4254, opened 2022-03-10
[next-adapter] revert webpack5 detection

marklawlor:
# Why

Quoting comment from https://github.com/expo/expo-cli/pull/4058

> This PR did the opposite of its intention and actually broke `withExpo` by causing it to default to `webpack5: undefined` instead of `webpack5: true`
> 
> The idea of the previous logic was to default to passing `webpack5: true` to `withUnimodules` if `next.config.js` had a truthy value or it wasn't declared (undefined)
> 
> It should only pass `false` to `withUnimodules` if the user explicitly set `webpack5: false` in their `next.config.js`
> 
> ```
> const webpack5 = undefined
> const previous = webpack5 !== false // previous === true
> const current = webpack5 && webpack5 !== false // current === undefined
> ```
> 

Reverting this change will fix [Issue using turborepo and @expo/next-adapter #4066](https://github.com/expo/expo-cli/issues/4066)

### DIFF
--- a/packages/next-adapter/src/withExpo.ts
+++ b/packages/next-adapter/src/withExpo.ts
@@ -19,7 +19,7 @@ export default function withExpo(nextConfig: any = {}): any {
         },
         {
           supportsFontLoading: false,
-          webpack5: webpack5 && webpack5 !== false,
+          webpack5: webpack5 !== false,
         }
       );
       // Use original public path


Simek:
Hello @marklawlor, thank you for the report & explanation! 👍 However, I'm afraid that just reverting this change is not a proper solution, because it will break project based on Webpack 4 (where old code yields invalid `webpack5: true`). 

So the problem seems to be in logic or with the code which passes the flag. In the other words, I think that te best solution will be to guarantee that if Webpack 5 is used, the flag is correctly set to `true`.

Simek:
The other idea which I can think of is to have a shift to the `@expo/next-adapter` versioning and start to matching the package version to the Next package (some other Next related libs do this), so we can have more clear separation for the different version Next support for the user.

To make this work, the current version should be re-published as `11.0.0`, and then your change could be a part of `12.0.0`, since the version in which the flag has been removed on the Next side. This approach might reduce the amount of logic code needed and allows us to skip the versioned logic, but it's a quite big change form the package philosophy stand point and skipping so many version might confuse some in the community, if not communicated properly.

marklawlor:
When this package had a peer of NextJS 10 - it defaulted to Webpack4. Now that it targets NextJS 11 it defaults to Webpack5.

This matches the functionality of NextJS and their migration webpack5. As per their docs, if you use NextJS 11 and wish to continue to use webpack 4 you need to explicit opt out using `webpack5: false` https://nextjs.org/docs/messages/webpack5.

I don't understand how this change would be breaking for webpack 4 projects?

Simek:
> When this package had a peer of NextJS 10 - it defaulted to Webpack4. Now that it targets NextJS 11 it defaults to Webpack5.

As I stated in the comment before, one of the parts of this issue is that we do not link `next-adapter` with the Next version, which means that latest version should work in all cases (of Next and Webpack version mix), and this is not possible without the Next version detection, because of how Next handled this flag over time (see the `undefined` case below).

> > In Next.js 11 webpack 5 was enabled by default for all applications. You could still opt-out and use webpack 4 to help with backwards compatibility using webpack5: false in next.config.js

The issue here is that the value of this flag means something else between few versions of Next. 

In short, Next =< 10 `undefined === Webpack4`, Next 11+ `undefined === Webpack5`

> I don't understand how this change would be breaking for webpack 4 projects?

The details are in the PR which you want to revert. 😅

marklawlor:
@Simek The `@expo/next-adapter` package already did a major version jump when it changed the peer dependency to Next 11.

marklawlor:
@EvanBacon  @brentvatne Can a reviewer be assigned to this issue please?

The failing test is due to a network issue during `npm install`, but I am unable to rerun it.

brentvatne (approved):
will do a major version bump along with this

PR #4260, opened 2022-03-17
[config-plugins] use relative path for locales

giautm:
# Why

Close: https://github.com/expo/expo/issues/15441

If the user run `expo prebuild -p ios` in their local and submit a new build with `eas build`. Then, the absolute path to locales files becomes wrong.

# How

- This PR fixes the issue by using a relative path for them.


EvanBacon (approved):
Would be good to test against this so it doesn't regress. Otherwise thanks for getting this!

MollieRick:
@EvanBacon Hi Evan, when can we expect the update with this fix? :-)

PR #4261, opened 2022-03-18
[cli] update supported Node version warning to reflect current LTS schema

kbrandwijk:
# Why

The Node.js versions warning doesn't reflect the current version scheme (https://nodejs.org/en/about/releases/).

# How

Changed the warning copy.

# Test Plan

- Select different node version with `nvm`
- View warning

### DIFF
--- a/packages/expo-cli/bin/expo.js
+++ b/packages/expo-cli/bin/expo.js
@@ -13,9 +13,8 @@ const minor = parseInt(match[2], 10);
 
 const supportedVersions =
   'expo-cli supports following Node.js versions:\n' +
-  '* >=12.13.0 <13.0.0 (Maintenance LTS)\n' +
-  '* >=14.0.0 <15.0.0 (Active LTS)\n' +
-  '* >=15.0.0 <17.0.0 (Current Release)\n';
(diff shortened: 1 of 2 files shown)


packages/expo-cli/bin/expo.js
```diff
@@ -13,9 +13,8 @@ const minor = parseInt(match[2], 10);
 
 const supportedVersions =
   'expo-cli supports following Node.js versions:\n' +
-  '* >=12.13.0 <13.0.0 (Maintenance LTS)\n' +
-  '* >=14.0.0 <15.0.0 (Active LTS)\n' +
-  '* >=15.0.0 <17.0.0 (Current Release)\n';
+  '* >=12.13.0 <15.0.0 (Maintenance LTS)\n' +
```
EvanBacon, line 16:
```suggestion
  '* >=12.13.0 <16.0.0 (Maintenance LTS)\n' +
```

kbrandwijk: Node 15 is considered 'current', not 'Maintenance LTS', that's why I excluded 15, and I didn't want to list it separately because we advise LTS version.

EvanBacon (approved):

PR #4270, opened 2022-03-22
[config-plugins] Introduce rule-based build properties config-plugin

Kudo:
# Why

to support more customized native build properties, like android `compileSdkVersion` and `targetSdkVersion`. we need an extensible way for what we did in the `jsEngine` property (reading a value from source config and merge it into properties file).
close ENG-4021

# How

refactor original dedicated `JsEngine` plugin to generic `BuildProperties` plugin.  it is rule based and easy for extension. a rule example for `jsEngine` to Podfile.properties.json is like this:

```ts
export const withJsEnginePodfileProps: ConfigPlugin = config => {
  const configToPropertyRules = [
    {
      // property name in `Podfile.properties.json`
      propName: 'expo.jsEngine',
      // passing `config` and get the property value
      propValueGetter: config => config.ios?.jsEngine ?? config.jsEngine ?? 'jsc',
    },
  ];
  return withBuildPodfileProps(config, { configToPropertyRules });
};
```

the source config can be either from expo config or from other custom config passing through config-plugin props.

in the `expo-build-properties` config-plugin, that would just be a thin layer and all in rules. e.g. https://github.com/expo/expo/blob/%40kudo/expo-build-properties/packages/expo-build-properties/plugin/src/android.ts

# Test Plan

```
 PASS   @expo/config-plugins  src/android/__tests__/BuildProperties-test.ts
  withJsEngineGradleProps
    ✓ set the property from shared `jsEngine` config (4 ms)
    ✓ set the property from platform override `jsEngine`
    ✓ overwrite the property if an old property is existed (2 ms)
  updateAndroidBuildPropertiesFromConfig
    ✓ should respect `propValueGetter` order
  updateAndroidBuildProperty
    ✓ should merge properties (2 ms)
    ✓ should keep original property when `value` is null by default
    ✓ should remove original property when `value` is null when `removePropWhenValueIsNull` is true

 PASS   @expo/config-plugins  src/ios/__tests__/BuildProperties-test.ts
  withJsEnginePodfileProps
    ✓ set the property from shared `jsEngine` config (6 ms)
    ✓ set the property from platform override `jsEngine`
    ✓ overwrite the property if an old property is existed
  updateIosBuildPropertiesFromConfig
    ✓ should respect `propValueGetter` order
  updateIosBuildProperty
    ✓ should merge properties
    ✓ should keep original property when `value` is null by default (1 ms)
    ✓ should remove original property when `value` is null when `removePropWhenValueIsNull` is true
```


kbrandwijk:
What's the philosophy behind using strings for `configFields` if a (typed) config object is passed in to the function. In other words, why is `configFields: ['ios.jsEngine', 'jsEngine'], defaultValue: 'jsc'` better than `value: config.ios?.jsEngine ?? config.jsEngine ?? 'jsc'` with intellisense, type safety, etc.?

Kudo:
> What's the philosophy behind using strings for `configFields` if a (typed) config object is passed in to the function. In other words, why is `configFields: ['ios.jsEngine', 'jsEngine'], defaultValue: 'jsc'` better than `value: config.ios?.jsEngine ?? config.jsEngine ?? 'jsc'` with intellisense, type safety, etc.?

awesome! i really like the idea. i've applied the pr with this approach - `propValueGetter: (config: BuildPropertiesConfig) => string | null;`. thanks for sharing this great idea 💛

packages/config-plugins/src/utils/BuildProperties.types.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Source config can be either expo config or generic config
+ */
+export type BuildPropertiesConfig = ExpoConfig | Record<string, any>;
+
+export interface ConfigToPropertyRuleType<SourceConfigType extends BuildPropertiesConfig> {
+  // Property name in `android/gradle.properties` or `ios/Podfile.properties.json`
+  propName: string;
+
+  // Passing config and get the property value
```
EvanBacon, line 27:
```suggestion
  /** Passing config and get the property value */
```

packages/config-plugins/src/utils/BuildProperties.types.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+
+/**
+ * Source config can be either expo config or generic config
+ */
+export type BuildPropertiesConfig = ExpoConfig | Record<string, any>;
+
+export interface ConfigToPropertyRuleType<SourceConfigType extends BuildPropertiesConfig> {
+  // Property name in `android/gradle.properties` or `ios/Podfile.properties.json`
```
EvanBacon, line 24:
```suggestion
  /** Property name in `android/gradle.properties` or `ios/Podfile.properties.json` */
```

packages/config-plugins/src/utils/BuildProperties.types.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * ConfigPlugin extended type for `withBuildPodfileProps` or `withBuildGradleProps` which accepting SourceConfigType generics
+ */
+export type BuildPropertiesConfigPlugin = <SourceConfigType extends BuildPropertiesConfig>(
+  config: ExpoConfig,
+  props: {
+    configToPropertyRules: ConfigToPropertyRuleType<SourceConfigType>[];
+    sourceConfig?: SourceConfigType;
+  }
+) => ExpoConfig;
```
EvanBacon, line 40: Perhaps there needs to be a different name for this type of function since config plugins should, in theory, be fully serializable (i.e. props cannot be functions). This is more of a factory-type method like [this](https://github.com/expo/expo-cli/blob/139a09fbc04e601b0f3fe6d37ad1adedf27b1159/packages/config-plugins/src/plugins/ios-plugins.ts#L20) since it generates functionality and cannot be used in a fully JSON config.

packages/config-plugins/src/utils/BuildProperties.types.ts
```diff
@@ -0,0 +1,40 @@
+import type { ExpoConfig } from '@expo/config-types';
+
+/**
+ * Rule to transform from config to build properties
+ *
+ * @example
+ * ```
```
EvanBacon, line 7:
```suggestion
 * ```ts
```

EvanBacon (approved):

PR #4281, opened 2022-03-29
[cli] Cap Expo CLI at Node 16

ide:
# Why

It doesn't work with Node 17 right now. Make this clear in `engines`.

# Test Plan

Verified that `>=12 <=16` matches numbers like `12.4.5` and `16.1.2`.

### DIFF
--- a/packages/expo-cli/package.json
+++ b/packages/expo-cli/package.json
@@ -39,7 +39,7 @@
   },
   "homepage": "https://github.com/expo/expo-cli/tree/main/packages/expo-cli#readme",
   "engines": {
-    "node": ">=12"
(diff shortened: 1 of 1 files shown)


kbrandwijk:
We already warn about Node versions when you _run_ the CLI. I it the CLI itself that doesn't support it, or only the upstream cert/webpack/etc issues that are causing some things to break without a workaround?

EvanBacon (approved):

ide:
This is for tools such as Volta. I'm not sure if Volta honors engines in application packages specifically but adding this is harmless.

SimenB:
Node 18 is now LTS - it should definitely be added to the `engines` range

PR #4295, opened 2022-04-12
ci: add stale bot from expo/expo

EvanBacon:
# Why

We have a lot of issues which do not apply anymore and should be closed.


kbrandwijk:
Too bad this wasn't up for review a bit longer. On eas-cli, we added the exception for untriaged issues, I'd like to see that same thing here.

PR #4303, opened 2022-04-14
[ci] Update workflows with concurrency and latest actions

byCedric:
# Why

The workflows were still using Node 12 and 14, older action versions and were missing concurrency.

# How

See commits

# Test Plan

Test if CI still works

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['12', '14']
+        node: ['14', '16']
     name: Build with Node ${{ matrix.node }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


kbrandwijk:
Did we drop Node 12 support for the existing expo-cli already?

.github/workflows/test.yml
```diff
@@ -89,24 +102,30 @@ jobs:
...
+      - name: 🗂 Download builds
+        uses: actions/cache@v2
         with:
           path: '*'
           key: v2-${{ github.sha }}-${{ matrix.node }}
-      - name: Install puppeteer
+      - name: 🏗 Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
```
kbrandwijk, line 113:
```suggestion
          node-version: ${{ matrix.node }}
```

byCedric: No, this one is intentional. The puppeteer job is a separate unrelated one to the normal testing. We don't have a test on multiple node versions for that now, and I don't think it would add much value if we add it.

EvanBacon (approved):

byCedric:
Superseded by https://github.com/expo/expo-cli/pull/4399

PR #4305, opened 2022-04-15
[DO NOT MERGE] Fix websocket controls on RN 0.68

lukmccall:
# Why

**This PR isn't a ready fix, because I don't really know how this change should be introduced to the CLI.** 

Fixes expo-cli terminal commands does not recognise the device:

```
warn No apps connected. Sending "reload" to all React Native apps failed. Make sure your app is running in the simulator or on a phone connected via USB.
```

# How

RN 0.67 introduced a different way of handling WebSocket connections. You can read more here -> https://github.com/react-native-community/cli/pull/1560. 

TL;DR
WebSockets started using a shared server. They are created with the `noServer` flag right now and have to be attached to the `upgrade` event like here https://github.com/websockets/ws#multiple-servers-sharing-a-single-https-server.
However, by doing that we are losing the backwards compatibility.

# Test Plan

- run project with RN 0.68 ✅
- a project with RN 0.64.X doesn't work with that version of CIL

### DIFF
--- a/packages/dev-server/package.json
+++ b/packages/dev-server/package.json
@@ -38,7 +38,7 @@
     "@expo/bunyan": "4.0.0",
     "@expo/metro-config": "0.3.13",
     "@expo/osascript": "2.0.32",
-    "@react-native-community/cli-server-api": "^5.0.1",
+    "@react-native-community/cli-server-api": "^7.0.3",
     "body-parser": "1.19.0",
     "chalk": "^4.0.0",
     "connect": "^3.7.0",
--- a/packages/dev-server/src/MetroDevServer.ts
+++ b/packages/dev-server/src/MetroDevServer.ts
@@ -76,6 +76,7 @@ export async function runMetroDevServerAsync(
   server: http.Server;
   middleware: any;
   messageSocket: MessageSocket;
+  debuggerProxyEndpoint: MessageSocket;
 }> {
   const Metro = importMetroFromProject(projectRoot);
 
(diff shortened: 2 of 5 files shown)


packages/dev-server/package.json
```diff
@@ -38,7 +38,7 @@
     "@expo/bunyan": "4.0.0",
     "@expo/metro-config": "0.3.13",
     "@expo/osascript": "2.0.32",
-    "@react-native-community/cli-server-api": "^5.0.1",
+    "@react-native-community/cli-server-api": "^7.0.3",
```
EvanBacon, line 41: We should drop this dependency and import it from the project if available like we do for metro.

packages/dev-server/src/MetroDevServer.ts
```diff
@@ -85,7 +86,13 @@ export async function runMetroDevServerAsync(
...
   const metroConfig = await ExpoMetroConfig.loadAsync(projectRoot, { reporter, ...options });
 
-  const { middleware, attachToServer } = createDevServerMiddleware({
+  const {
+    middleware,
+    debuggerProxyEndpoint,
+    messageSocketEndpoint,
+    eventsSocketEndpoint,
+    websocketEndpoints,
+  } = createDevServerMiddleware({
```
EvanBacon, line 95: We may need to check which features are available based on the version installed and implement the new functionality only when the new features are available. Ultimately we'll drop this in SDK 46 when we drop the unversioned CLI.

EvanBacon:
Closing in favor of https://github.com/expo/expo-cli/pull/4306

PR #4308, opened 2022-04-15
[schemer] Update user-facing error message for `not` fields

wschurman:
# Why

Previously, an error in a negated field (one surrounded by the `not` schema composition keyword: https://json-schema.org/understanding-json-schema/reference/combining.html#not) would just read as "should NOT be valid" without any additional context (https://github.com/ajv-validator/ajv/blob/master/lib/vocabularies/applicator/not.ts#L35)

# How

We have a precedent for custom error message functions for particular types of fields. This adds one for `not`.

Similar to how we have `regexHuman` (which maybe should have been called `patternHuman`?), we now have `notHuman` (which sounds weird and I'm open to other names).

Note that putting `not` inside a field still results in a weird error message since that field is now considered by `ajv` to be a `not`. So something like the following will be considered a "not" instead of a "pattern" and will have a "not" error message (constructed with `notHuman`) even though it evaluates the pattern and the not:
```
{
  "type": "string",
  "pattern": "^[a-zA-Z\\d][a-zA-Z\\d._+()-]{0,254}$",
  "meta": {
    "regexHuman": "String beginning with an alphanumeric character followed by any combination of alphanumeric character, \"_\", \"+\", \".\",\"(\", \")\", or \"-\". Examples: \"1.0.3a+\", \"1.0.0\", \"._+()-0a1\", \"0\"."
  },
  "not": {
    "pattern": "^\\d+\\.\\d*0$",
    "meta": {
      "regexHuman": "Decimal ending in a 0."
    }
  }
},
```

The proposed solution to the above is to surround it with an `allOf`:
```
{
  "allOf": [
    {
      "type": "string",
      "pattern": "^[a-zA-Z\\d][a-zA-Z\\d._+()-]{0,254}$",
      "meta": {
        "regexHuman": "String beginning with an alphanumeric character followed by any combination of alphanumeric character, \"_\", \"+\", \".\",\"(\", \")\", or \"-\". Examples: \"1.0.3a+\", \"1.0.0\", \"._+()-0a1\", \"0\"."
      },
    },
    {
      "not": {
        "pattern": "^\\d+\\.\\d*0$",
        "meta": {
          "regexHuman": "Decimal ending in a 0."
        }
      }
    }
  ]
}
```
... so that the resulting error message is more correct.

# Test Plan

Run test, look at resulting snapshot.


ide (approved):

ide (commented):
One thing I find counterintuitive is how notHuman is defined. I would expect the field to be a description explaining that the pattern is rejected. For example instead of:
```
{
  "not": {
    "pattern": "^\\d+\\.\\d*0$"
  },
  "meta": {
    "notHuman": "A decimal ending in a 0."
  }
}
```
I would expect:
```
"notHuman": "Not a decimal ending in a 0."
```
or for the positive description to describe the pattern
```
{
  "not": {
    "pattern": "^\\d+\\.\\d*0$",
    "regexHuman": "A decimal ending in a 0."
  }
}
```

The reason I think intuitiveness is important here is because if we get it wrong, the error message will say the exact opposite of what it should.

wschurman:
Updated to be `"notHuman": "Not a decimal ending in a 0."`. 

Unfortunately we can't take the `regexHuman` from inside the `not` since then we'd have to enumerate all the types of subfields of not when generating the message (there can be other negations other than "pattern"). The way ajv works is a `not` field is the type of field for message generation and it doesn't compose the fields it contains for the purposes of error messages.

PR #4311, opened 2022-04-16
[RFC][package-manager] Add support for pnpm

byCedric:
# Why

There are a few package managers that people are using, mainly `yarn` and `npm`. Currently, Yarn is in a weird state where we still recommend yarn v1 as a result of the discussion around yarn v2 and plug-n-play issues with React Native. Also, npm v7 decided to enforce the peer dependencies more strictly, which may cause install issues for some parts of React Native.

We probably need to move on beyond yarn v1 at some point. This is just a proposal to consider opt-in pnpm support in tooling like `expo install`, `expo prebuild`, and `expo upgrade`. 

This is also the first step to get support in EAS build. We can also add an option to let users decide what install command should be used, like Vercel does. But adding "native" support is, IMHO, better for pnpm.

## Usage statistics

`pnpm` has been growing exponentially in usage. Mainly driven by popular monorepo tooling, like turborepo, [which kind of recommends to use pnpm for performance](https://twitter.com/jaredpalmer/status/1506634649921064966). Or the `actions/setup-node`'s `cache` property [supporting `pnpm` officially](https://github.com/actions/setup-node#caching-packages-dependencies)

<details><summary>npm downloads in past years</summary>

image

</details>

<details><summary>yarn downloads in past years</summary>

image

</details>

<details><summary>pnpm downloads in past years</summary>

image

</details>

## My own tests with pnpm

Turns out Expo SDK 44 is also working with `.npmrc` `node-linker=hoisted` setting. The `--shamefully-hoist` should be something from the past.

~~A while ago, [I tested pnpm to check support for React Native](https://github.com/byCedric/eas-monorepo-example/pull/19). Turns out that we only need the `--shamefully-hoist` to let pnpm behave more like the traditional node module resolution behavior, something React Native expects.~~

It's also _a lot_ faster with installing modules, especially in larger monorepos.

- [With pnpm](https://github.com/byCedric/eas-monorepo-example/actions/runs/1352322142)  → **1m 33s** (using yarn to execute workspace commands)
- [With yarn](https://github.com/byCedric/eas-monorepo-example/actions/runs/1318138790) → **5m 58s**

That's a whopping 3.8x improvement.

# How

- Created `PnpmPackageManager` class.
- Added `findWorkspaceRoot` with support for either `npm`, `yarn`, or `pnpm`
- Added `resolvePackageManager` with support for either `npm`, `yarn`, or `pnpm`
- ~~Marked `isUsingYarn` as deprecated, as it only resolves to `yarn` or `npm`~~ → Reverted

## Want to try it out?

To try it out, build the CLI from this branch and run:

```bash
# create the project
$ expo init ./expo-with-pnpm
$ cd ./expo-with-pnpm

# remove node_modules and yarn or npm lockfiles
$ npx rimraf node_modules yarn.lock package-lock.json

# create `.npmrc` file with hoisted node-linker
# see: https://pnpm.io/npmrc#node-linker
$ echo "node-linker=hoisted" > .npmrc

# change the entrypoint to default bare project entrypoint
# don't forget to update `package.json` main, set it to `index.js`
$ echo "import { registerRootComponent } from 'expo';
import App from './App';

registerRootComponent(App);" > index.js


# install the project
$ pnpm install
```

## ~~Rationale for each command~~ → Reverted

Per command, I think these are the reasons to add or not-add it.

1. `$ expo install`  → Expo install fetches the right version of each dependency and adds it to the project. We want this behavior to be similar for pnpm users. Without it, they are highly likely to install the wrong version and encounter non-descriptive errors.
2. `$ expo upgrade` → Similar to `expo install`. We want to this work as best as we can to handle most of the complexity added by us handling dependency versions.
3. `$ expo prebuild|eject` → This one is actually optional. It's mostly for users who want to use this locally, without the hassle. We can also skip installing dependencies.
4. `$ expo init` → Because you'll need additional settings that we don't want to add to the template, it doesn't make sense to add the `--pnpm` flag here. What happens is that the user gets something "broken out of the box". Users who intentionally switch should also be opting in for pnpm and the additional setup complexity.

Also, point 3 could be improved here. It uses `yarn` as default, and only falls back to `npm` whenever `yarn` is not installed. We should be using the same detection as we do for `expo install` and `expo upgrade`, which is based on the lockfiles and not if one of the package manager isn't installed.

## ~~Note about `--shamefully-hoist` workaround~~

Turns out Expo SDK 44 is also working with `.npmrc` `node-linker=hoisted` setting. The `--shamefully-hoist` should be something from the past.

<details><summary>old issue</summary>

When the repo is installed _without_ this flag, you could run into this error when installing packages using the flag:

image

I could try and catch that error, and force a full reinstall of the repo. But for now, I think just surfacing this error might be enough.

</details>

# Next Steps

Because we only supported `yarn` or `npm`, we have these type literals spread out through the codebase. It would be great if we can move that over to `NodePackageManager` instead (or similar literals), to improve support. The same goes for `find-yarn-workspace-root` package, which is not sufficient anymore for pnpm.

This change also isn't done by just updating the package manager, we have to gradually add support for this in all of the systems. Including `expo-cli`, `@expo/cli`, and EAS build. But at least it's a start 😄 

- Add support for `pnpm --offline`

# Test Plan

- Added tests to the new methods
- ~~Following commands should work with pnpm~~ → reverted!


ide:
Tangential to this PR, I would be curious to see benchmarks and general insights into usability with npm 8 + lockfiles + workspaces. I have heard that npm 8 is quite fast (no benchmarks to show). It is also included with Node and many people who use npm don't ever install it directly.

byCedric:
@ide the results are in!

## Test project

I repurposed my example monorepo ([eas-monorepo-example](https://github.com/bycedric/eas-monorepo-example)) and tested the installs with(out) caching, on both Linux and MacOS.

## Results

tool       | os    | cache? | install time (seconds)
---        | ---   | ---    | ---
`npm@7`    | linux | ✅     | [15s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434801?check_suite_focus=true#step:7:1)
`npm@7`    | macos | ✅     | [34s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434936?check_suite_focus=true#step:7:1)
`npm@7`    | linux | ❌     | [30s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434857?check_suite_focus=true#step:7:1)
`npm@7`    | macos | ❌     | [46s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435018?check_suite_focus=true#step:7:1)
           |       |        |
`npm@8`    | linux | ✅     | [15s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434212?check_suite_focus=true#step:7:1)
`npm@8`    | macos | ✅     | [28s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434365?check_suite_focus=true#step:7:1)
`npm@8`    | linux | ❌     | [30s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434289?check_suite_focus=true#step:7:1)
`npm@8`    | macos | ❌     | [65s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434433?check_suite_focus=true#step:7:1)
           |       |        |
`pnpm@6`   | linux | ✅     | [5s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434547?check_suite_focus=true#step:7:1)
`pnpm@6`   | macos | ✅     | [25s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083621463?check_suite_focus=true#step:7:1)
`pnpm@6`   | linux | ❌     | [18s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434602?check_suite_focus=true#step:7:1)
`pnpm@6`   | macos | ❌     | [50s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083434723?check_suite_focus=true#step:7:1)
           |       |        |
`pnpm@7`   | linux | ✅     | [6s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435088?check_suite_focus=true#step:7:1)
`pnpm@7`   | macos | ✅     | [22s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435216?check_suite_focus=true#step:7:1)
`pnpm@7`   | linux | ❌     | [19s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435155?check_suite_focus=true#step:7:1)
`pnpm@7`   | macos | ❌     | [48s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435283?check_suite_focus=true#step:7:1)
           |       |        |
`yarn@1`   | linux | ✅     | [30s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435365?check_suite_focus=true#step:8:1)
`yarn@1`   | macos | ✅     | [52s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435475?check_suite_focus=true#step:8:1)
`yarn@1`   | linux | ❌     | [68s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435427?check_suite_focus=true#step:8:1)
`yarn@1`   | macos | ❌     | [93s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435535?check_suite_focus=true#step:8:1)
           |       |        |
`yarn@2+`  | linux | ✅     | [48s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435597?check_suite_focus=true#step:8:1)
`yarn@2+`  | macos | ✅     | [81s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435693?check_suite_focus=true#step:8:1)
`yarn@2+`  | linux | ❌     | [130s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435633?check_suite_focus=true#step:8:1)
`yarn@2+`  | macos | ❌     | [136s](https://github.com/byCedric/expo-monorepo-benchmark/runs/6083435754?check_suite_focus=true#step:8:1)

packages/package-manager/src/NodePackageManagers.ts
```diff
@@ -328,24 +319,121 @@ export class YarnPackageManager implements PackageManager {
   }
 }
 
-export type CreateForProjectOptions = {
-  npm?: boolean;
-  yarn?: boolean;
+export class PnpmPackageManager implements PackageManager {
```
EvanBacon, line 322: Split this into a new file

packages/package-manager/src/NodePackageManagers.ts
```diff
@@ -328,24 +319,121 @@ export class YarnPackageManager implements PackageManager {
...
+  options: SpawnOptions;
+  private log: Logger;
+
+  constructor({ cwd, log, silent }: { cwd: string; log?: Logger; silent?: boolean }) {
```
EvanBacon, line 326: Add tests for this module

packages/package-manager/src/NodePackageManagers.ts
```diff
@@ -28,19 +31,6 @@ const yarnPeerDependencyWarningPattern = new RegExp(
   'g'
 );
 
-/**
- * Returns true if the project is using yarn, false if the project is using npm.
- *
- * @param projectRoot
- */
-export function isUsingYarn(projectRoot: string): boolean {
```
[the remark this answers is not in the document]
byCedric: No it doesn't, I just moved the function to the `nodeWorkspace` file. see: https://github.com/expo/expo-cli/pull/4311/files#diff-88977d86b1eb9675ab803f035c0c9a76381422bc0fd46d915958c7519417bc69R100-R102 Can revert though.

packages/expo-cli/src/commands/info/upgrade.ts
```diff
@@ -10,7 +10,8 @@ export default function (program: Command) {
       .description('Upgrade the project packages and config for the given SDK version')
       .helpGroup('info')
       .option('--npm', 'Use npm to install dependencies. (default when package-lock.json exists)')
-      .option('--yarn', 'Use Yarn to install dependencies. (default when yarn.lock exists)'),
+      .option('--yarn', 'Use Yarn to install dependencies. (default when yarn.lock exists)')
+      .option('--pnpm', 'Use pnpm to install dependencies. (default when pnpm-lock.yaml exists)'),
```
EvanBacon, line 14: The package manager change is fine but we should be adding new features to the versioned CLI so we don't have to wrangle issues across tools. The expo-cli package should not have new features added to it.

byCedric:
Also, I want to stress that this RFC/PR is intended to "get expo out of the way" that prevent users from using pnpm. It's not an RFC to switch Expo repos over to pnpm.

With this change, we can add support for pnpm in eas build so people with `pnpm-lock.yaml` files can install without issues.

packages/package-manager/src/NodePackageManagers.ts
byCedric: Done

packages/package-manager/src/NodePackageManagers.ts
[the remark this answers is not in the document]
byCedric: Reverted!

packages/package-manager/src/NodePackageManagers.ts
byCedric: Done!

packages/expo-cli/src/commands/info/upgrade.ts
byCedric: Reverted!

byCedric:
Also, Twitter's own mjackson did a poll on workspaces (yarn vs npm). In the replies, you can see a decent amount of people mentioning pnpm https://twitter.com/mjackson/status/1519356560769622016

EvanBacon (approved):

PR #4338, opened 2022-04-29
feat: create create-expo-app package

EvanBacon:
# Why

As part of the new versioned CLI, we'll need a package for initializing expo apps. This package is a fork of our package `create-react-native-app` but with most of the features stripped out. We will incrementally decide how we want to handle templates and examples after this has landed.

- Resolves ENG-4832


packages/create-expo-app/src/Update.ts
```diff
@@ -0,0 +1,29 @@
+import chalk from 'chalk';
+import { execSync } from 'child_process';
+import checkForUpdate from 'update-check';
+
+export function shouldUseYarn(): boolean {
```
byCedric, line 5: Any reason why we shouldn't use `@expo/package-manager`'s `isUsingYarn` instead? Or the package manager that was used when invoking this library?

packages/create-expo-app/src/dir.ts
```diff
@@ -0,0 +1,28 @@
...
+  // System
+  '.DS_Store',
+  'Thumbs.db',
+  // Git
+  '.git',
+  '.gitattributes',
+  '.gitignore',
+  // Project
+  '.npmignore',
+  '.travis.yml',
```
byCedric, line 14: tbh, with the [huge security breach from both Heroku and Travis CI](https://github.blog/2022-04-15-security-alert-stolen-oauth-user-tokens/), I'm not too keen to actively support Travis CI.

byCedric (approved):
Left a few comments, but its looking food so far

packages/create-expo-app/src/dir.ts
EvanBacon, line 14:
```suggestion
```

packages/create-expo-app/src/Update.ts
EvanBacon: `isUsingYarn` is for projects with a `yarn.lock`, `shouldUseYarn` detects if yarn is installed globally.

PR #4341, opened 2022-05-04
[config-plugins] Handle quoted build configuration

wkozyra95:
# Why

If someone creates build configuration with space the middle e.g. `Release production` value in pbxproj will be in quotes which is not handled correctly in util functions that search for build configuration by name

# How

trim quotes when comparing

# Test Plan

- reproduce the issue described above (`eas build` fails with ` Error: Build configuration 'Release' does not exist in list with id '13B07F931A680F5B00A75B9A'`)
- yarn link config-plugins with above changes to eas-cli
- run build on eas (build works)


kbrandwijk (approved):

EvanBacon:
Which config plugin uses `setProvisioningProfileForPbxproj`?

PR #4343, opened 2022-05-05
[expo-cli] warn for outdated `@expo/config-plugins` in deps' deps

kbrandwijk:
# Why

SDK 45 needs @expo/config-plugins@^4.1.0. Because the error when using older version is quite cryptic, an explicit warning is added to `expo doctor`

# How

The required version is hardcoded for now. The warning doesn't show which packages actually use which versions (like npm/yarn why) because that test is complex.

# Test Plan

Tested manually.
Combined output:
WindowsTerminal_HjvllCM3Gy
Solo output:
WindowsTerminal_TWw0XQJ5W2

### DIFF
--- a/packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
+++ b/packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
@@ -46,7 +46,11 @@ export async function validateDependenciesVersionsAsync(
   // find incorrect dependencies by comparing the actual package versions with the bundled native module version ranges
   const incorrectDeps = findIncorrectDependencies(packageVersions, bundledNativeModules);
 
+  let issuesFound = false;
+
   if (incorrectDeps.length > 0) {
+    issuesFound = true;
+
     Log.warn('Some dependencies are incompatible with the installed expo package version:');
     incorrectDeps.forEach(({ packageName, expectedVersionOrRange, actualVersion }) => {
       Log.warn(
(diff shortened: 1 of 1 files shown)


packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
```diff
@@ -71,10 +75,39 @@ export async function validateDependenciesVersionsAsync(
           )}`
       );
     }
+  }
 
-    return false;
+  if (Versions.gteSdkVersion(exp, '45.0.0')) {
```
brentvatne, line 80: let's do this for exactly 45.0.0

packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
```diff
@@ -71,10 +75,39 @@ export async function validateDependenciesVersionsAsync(
...
+    try {
+      /* This will throw if the dependency looked for is not installed,
+       * but that doesn't apply here, so that error is ignored */
+      const packageVersionConfigPlugins = await resolvePackageVersionsAsync(projectRoot, [
```
brentvatne, line 84: this will only find a single version of the dep, right? if you have libraries in your project that each depend on a different version of `@expo/config-plugins` that may not be surfaced here

kbrandwijk: Yes, it will loop through the flat list of packages in node_modules, so if will only find one. Still, if there _is_ one with the wrong version there, and it warns about it, then the next step is to use `yarn why` or `npm why` which will turn up everything.

packages/expo-cli/src/commands/utils/validateDependenciesVersions.ts
kbrandwijk: xdl doesn't expose an exact compare, and I don't want to duplicate the rest of the logic from that function (check for valid sdkVersion in json, compare, handle errors etc.) for such a rudimentary check.
```suggestion
  if (Versions.gteSdkVersion(exp, '45.0.0') && !Versions.gteSdkVersion(exp, '46.0.0')) {
```
Would this be an acceptable alternative?

brentvatne: yeah that's fine - although i would have a slight preference for extending `Versions` to add `Versions.equalsSdkVersion`

PR #4346, opened 2022-05-06
[package-manager] Separate and test yarn and npm package managers

byCedric:
# Why

This should add some missing tests for the `YarnPackageManager` and `NpmPackageManager`, and make it on par with the `PnpmPackageManager`.

# How

- Split the package managers outside the `NodePackageManager` file into their own
- Copied and modified the tests from `PnpmPackageManager` for both `yarn` and `npm`

# Test Plan

- See tests

### DIFF
--- a/packages/package-manager/src/NpmPackageManager.ts
+++ b/packages/package-manager/src/NpmPackageManager.ts
@@ -1,0 +1,179 @@
+import JsonFile from '@expo/json-file';
+import spawnAsync, { SpawnOptions } from '@expo/spawn-async';
+import ansiRegex from 'ansi-regex';
+import assert from 'assert';
+import fs from 'fs';
+import npmPackageArg from 'npm-package-arg';
+import path from 'path';
+import rimraf from 'rimraf';
+import split from 'split';
+import { Transform } from 'stream';
+
+import { DISABLE_ADS_ENV } from './NodePackageManagers';
+import { Logger, PackageManager } from './PackageManager';
+
+const ansi = `(?:${ansiRegex().source})*`;
+const npmPeerDependencyWarningPattern = new RegExp(
+  `${ansi}npm${ansi} ${ansi}WARN${ansi}.+You must install peer dependencies yourself\\.\n`,
+  'g'
+);
+
+/** Exposed for testing */
+export class NpmStderrTransform extends Transform {
+  _transform(
+    chunk: Buffer,
+    encoding: string,
+    callback: (error?: Error | null, data?: any) => void
+  ) {
+    this.push(chunk.toString().replace(npmPeerDependencyWarningPattern, ''));
+    callback();
+  }
+}
+
+export class NpmPackageManager implements PackageManager {
+  options: SpawnOptions;
+
+  private log: Logger;
+
+  constructor({ cwd, log, silent }: { cwd: string; log?: Logger; silent?: boolean }) {
+    this.log = log || console.log;
+    this.options = {
+      env: {
+        ...process.env,
+        ...DISABLE_ADS_ENV,
+      },
+      cwd,
+      ...(silent
+        ? { ignoreStdio: true }
+        : {
+            stdio: ['inherit', 'inherit', 'pipe'],
+          }),
+    };
+  }
+
+  get name() {
+    return 'npm';
+  }
+
+  async installAsync(parameters: string[] = []) {
+    await this._runAsync(['install', ...parameters]);
+  }
+
+  async addGlobalAsync(...names: string[]) {
+    if (!names.length) return this.installAsync();
+    await this._runAsync(['install', '--global', ...names]);
+  }
+
+  async addWithParametersAsync(names: string[], parameters: string[] = []) {
+    if (!names.length) return this.installAsync(parameters);
+
+    const { versioned, unversioned } = this._parseSpecs(names);
+    if (versioned.length) {
+      await this._patchAsync(versioned, 'dependencies');
+      await this.installAsync(parameters);
+    }
+    if (unversioned.length) {
+      await this._runAsync([
+        'install',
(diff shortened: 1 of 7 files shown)


byCedric:
This doesn't change the API in any way, it just adds more test coverage and makes the npm/yarn more consistent with how we implemented pnpm right now.

packages/package-manager/src/NpmPackageManager.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (promise.child.stderr && !this.options.ignoreStdio) {
+      promise.child.stderr
+        .pipe(split(/\r?\n/, (line: string) => line + '\n'))
+        .pipe(new NpmStderrTransform())
+        .pipe(process.stderr);
+    }
+    return promise;
+  }
+
+  private _parseSpecs(names: string[]) {
```
wkozyra95, line 149: we shouldn't use `_` prefix for private functions

packages/package-manager/src/YarnPackageManager.ts
```diff
@@ -0,0 +1,139 @@
...
+  private log: Logger;
+
+  constructor({ cwd, log, silent }: { cwd: string; log?: Logger; silent?: boolean }) {
+    this.log = log || console.log;
+    this.options = {
+      env: {
+        ...process.env,
+        ...DISABLE_ADS_ENV,
+      },
+      cwd,
```
wkozyra95, line 42: it's not clear to me what cwd is, if I saw an interface like this (and I know that this package supports workspaces), I would expect that if cwd is in a subproject most command will be executed there, but e.g. yarn install would run in root.

packages/package-manager/src/NpmPackageManager.ts
```diff
@@ -0,0 +1,179 @@
+import JsonFile from '@expo/json-file';
+import spawnAsync, { SpawnOptions } from '@expo/spawn-async';
+import ansiRegex from 'ansi-regex';
+import assert from 'assert';
+import fs from 'fs';
+import npmPackageArg from 'npm-package-arg';
+import path from 'path';
+import rimraf from 'rimraf';
```
wkozyra95, line 8: we are trying to get rid of `fs-exra` in favor of `fs/promise`, but we are still using `rimraf`?

packages/package-manager/src/NpmPackageManager.ts
```diff
@@ -0,0 +1,179 @@
...
+  async getConfigAsync(key: string) {
+    const { stdout } = await spawnAsync('npm', ['config', 'get', key], { stdio: 'pipe' });
+    return stdout.trim();
+  }
+
+  async removeLockfileAsync() {
+    assert(this.options.cwd, 'cwd required for NpmPackageManager.removeLockfileAsync');
+    const lockfilePath = path.join(this.options.cwd, 'package-lock.json');
+    if (fs.existsSync(lockfilePath)) {
+      rimraf.sync(lockfilePath);
```
wkozyra95, line 120: existsSync does not have a good async equivalent in fs/promise, but why `rimraf` call is sync inside async function?

wkozyra95 (approved):
LGTM

All comments I had refer to code that was not introduced in this PR(just moved between files), so feel free to ignore them

packages/package-manager/src/NpmPackageManager.ts
byCedric: Fixed in https://github.com/expo/expo/pull/18576/files#diff-ed6baab8aa32f4cba63e7ad6287f12167c9c3b62b771c6574f0e0b77eceeaf7eR89, using only `private` and without the underscore.

packages/package-manager/src/YarnPackageManager.ts
byCedric: I think, but correct me if I'm wrong, I addressed this in https://github.com/expo/expo/pull/18576/files#diff-674041de5e2d6dc1344c60e176ff815c71ad2c76c1ac823e2c99b77022813c4fR5-R19. There is only 1 `PackageManagerOptions`, which inherits the `SpawnOptions` (with additional `silent` and `log` options). I think that should make it a bit more clear.

Besides that, there is also the `packageManager.workspaceRoot()` which returns a package manager instance of the same type (`npm|pnpm|yarn`), only if there was a workspace root detected. You can use that to run install from root:

```
// Compact version
await (packageManager.workspaceRoot() ?? packageManager).installAsync()

// Example with logging
const workspaceManager = packageManager.workspaceRoot();
if (workspaceManager) {
  console.log(`Running install in workspace root ${workspaceManager.options.cwd}`);
  await workspaceManager.installAsync();
} else {
  console.log(`Running install in project ${packageManager.options.cwd}`);
  await packageManager.installAsync();
}
```

packages/package-manager/src/NpmPackageManager.ts
byCedric: Yeah, I think @dsokal mentioned getting rid of rimraf. We can definitely swap `rimraf` for just `fs` and use async methods. Can follow up in a later PR after https://github.com/expo/expo/pull/18576

packages/package-manager/src/NpmPackageManager.ts
byCedric: Same as https://github.com/expo/expo-cli/pull/4346#discussion_r1008043056:
> Yeah, I think @dsokal mentioned getting rid of rimraf. We can definitely swap rimraf for just fs and use async methods. Can follow up in a later PR after https://github.com/expo/expo/pull/18576

We can definitely do that in a follow-up PR :). For now, I left in the same as it was, but it's an async method, we can swap it out.

byCedric:
Closing in favor of https://github.com/expo/expo/pull/18576

PR #4347, opened 2022-05-06
[package-manager] Add custom spawn option for node package managers

byCedric:
# Why

> Note, this is just a proposal. With this PR, we could move EAS and classic turtle back to use `@expo/package-manager` _and_ add support for pnpm.

In EAS we need to pipe the logs in a different way, [using a different logger](https://github.com/expo/eas-build/blob/f1abc21a3da8049a241edfc902e378ac8523f081/packages/build-tools/src/utils/project.ts#L116-L120) - [`@expo/turtle-spawn`](https://github.com/expo/eas-build/tree/f1abc21a3da8049a241edfc902e378ac8523f081/packages/turtle-spawn). This allows us to use that logger with just the package manager. It also should inherit the annoying peer-dep log filtering.

# How

- Added spawner option to the constructors
- Added tests to validate the new spawner for `installAsync`

# Test Plan

- See tests

### DIFF
--- a/packages/package-manager/src/NodePackageManagers.ts
+++ b/packages/package-manager/src/NodePackageManagers.ts
@@ -1,3 +1,4 @@
+import spawnAsync from '@expo/spawn-async';
 import { existsSync } from 'fs';
 import path from 'path';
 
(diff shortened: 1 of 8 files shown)


wkozyra95 (commented):
Let's maybe hold on with those changes and figure out first how the overall API should look to implement everything EAS Build needs.

In particular, I don't think the spawn function needs to be passed to it, instead, we could pass SpawnOptions end return SpawResult. There are few other issues we already  talked via slack, but this PR is not the best place to discuss them, I suggest creating some linear issue or notion doc

PR #4348, opened 2022-05-06
[package-manager] Disable frozen lockfiles for pnpm in CI

byCedric:
# Why

This is required in EAS to run installAsync after running prebuild. Some packages might have changed, and thus changes the lockfile.

[Pnpm automatically enables this on CI environments.](https://pnpm.io/cli/install#--frozen-lockfile)

I think we can also just run this every time instead of just on CI environments though.

# How

- Added `--no-frozen-lockfile` for CI only
- Added test

# Test Plan

See test

### DIFF
--- a/packages/package-manager/src/PnpmPackageManager.ts
+++ b/packages/package-manager/src/PnpmPackageManager.ts
@@ -66,6 +66,12 @@ export class PnpmPackageManager {
   }
 
   async installAsync() {
+    if (this.options.env?.['CI']) {
+      // When running in EAS, we need to disable this to run install after prebuild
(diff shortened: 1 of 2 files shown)


kbrandwijk:
Don't we run `prebuild` with `--no-install` on EAS?

byCedric:
> Don't we run `prebuild` with `--no-install` on EAS?

Yes, I'm mostly talking in GitHub Actions environments or other CIs. If something works locally, like `expo install <whatever>` I would expect it to work in CI as well. :)

This all has been moved to https://github.com/expo/expo/pull/18576, closing this.

PR #4352, opened 2022-05-07
[config-plugins] Remove warning when permissions config is empty when using blockedPermissions

brentvatne:
# Why

https://github.com/expo/expo/issues/17380

[The `expo-image-picker` config plugin uses `withBlockedPermissions`](https://github.com/expo/expo/blob/091d1acebb5cdaacaf45decc1fb179ce6770d082/packages/expo-image-picker/plugin/src/withImagePicker.ts#L56-L62), but it's possible that at the time that this plugin runs the `permissions` config is empty - depending on the order the plugins are executed. So we shouldn't show this warning in this case.

# How

Remove the warning

# Test Plan

Make the change manually in `node_modules` and verify there is no warning in [this repro](https://github.com/expo/expo/issues/17380#issuecomment-1120195369).

# Open questions

Does it matter that we might have a permission listed in both the `permissions` and `blockedPermission` key? Should we update the permissions config plugin to skip adding a permission if it's listed in `blockedPermissions`?

### DIFF
--- a/packages/config-plugins/src/android/Permissions.ts
+++ b/packages/config-plugins/src/android/Permissions.ts
@@ -29,10 +29,6 @@ export const withBlockedPermissions: ConfigPlugin<string[] | string> = (config, 
     Boolean
   );
 
-  if (!resolvedPermissions.length) {
-    WarningAggregator.addWarningAndroid('block-permissions', 'No permissions provided');
-  }
-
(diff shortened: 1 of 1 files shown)


kbrandwijk (approved):
Approved, except for the bot's comments...

PR #4365, opened 2022-05-11
Add support for detecting Gemfile and using ruby bundler for cocoapods

### LINKED ISSUE #2206: Support Gemfile for pod-install
Running npx pod-install should check for the presence of a Gemfile and run `bundle exec pod install`

cltnschlosser:
# Why

Resolves #2206

Different developers are ended up with different cocoapods versions which causes problems with Podfile.lock being generated differently.

# Test Plan

~I need help with this. I'm not an experienced JS developer, I come from an iOS background. I can't seem to get anything to use my local `package-manager` package. I can point a project to my local `pod-install` package, but even that doesn't use the local `package-manager` (even when the dependency is overridden to a local path).~
- Tested against local project which uses a Gemfile.
- Added a few unit tests.

### DIFF
--- a/packages/package-manager/src/CocoaPodsPackageManager.ts
+++ b/packages/package-manager/src/CocoaPodsPackageManager.ts
@@ -29,12 +29,18 @@ export function extractMissingDependencyError(errorOutput: string): [string, str
   return null;
 }
 
+interface CocoaPodsPackageManagerProps {
+  cwd: string;
+  silent?: boolean;
+  spawnOptions?: SpawnOptions;
+  isNonInteractive?: boolean;
+}
+
 export class CocoaPodsPackageManager implements PackageManager {
   options: SpawnOptions;
 
-  private silent: boolean;
+  silent: boolean;
   private isNonInteractive: boolean;
-  private shouldUseRubyBundler: boolean;
 
   static getPodProjectRoot(projectRoot: string): string | null {
     if (CocoaPodsPackageManager.isUsingPods(projectRoot)) return projectRoot;
@@ -49,6 +55,7 @@ export class CocoaPodsPackageManager implements PackageManager {
     return existsSync(path.join(projectRoot, 'Podfile'));
   }
 
+  // Visible for testing
   static isUsingRubyBundler(projectRoot: string): boolean {
     return existsSync(path.join(projectRoot, 'Gemfile'));
   }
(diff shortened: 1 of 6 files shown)


EvanBacon:
Looks good so far, thanks for the contribution!

To test locally, you're going to want to [setup the repo](https://github.com/expo/expo-cli/blob/main/CONTRIBUTING.md#setting-up-the-repository-for-development). You can then create an alias or run the bin directly:
```sh
# .zshrc
alias pod-installd='/Users/colton/Documents/GitHub/expo-cli/packages/pod-install/build/index.js'
```

The pod-install package will automatically pull in the latest changes but they need to be built (`yarn start` in the root).

I also recommend setting up some jest tests in `packages/package-manager/src/__tests__/CocoaPodsPackageManager-test.ts` -- test with `yarn test --watch` in `packages/package-manager/`.

packages/package-manager/src/CocoaPodsPackageManager.ts
```diff
@@ -47,6 +47,25 @@ export class CocoaPodsPackageManager implements PackageManager {
     return existsSync(path.join(projectRoot, 'Podfile'));
   }
 
+  static isUsingRubyBundler(): boolean {
+    return existsSync('Gemfile');
```
EvanBacon, line 51: Be sure to pass in the projectRoot and join the path components (like we do for `isUsingPods`), if this file is supposed to live in the user home directory, then use `os.homedir()`.

packages/package-manager/src/CocoaPodsPackageManager.ts
```diff
@@ -91,6 +116,38 @@ export class CocoaPodsPackageManager implements PackageManager {
...
+            chalk.red(
+              `\u203A Failed to install CocoaPods CLI with Bundler. Make sure cocoapods is in your Gemfile`
+            )
+          );
+        throw new CocoaPodsError(
+          '`bundle install` appeared to succeed but CocoaPods CLI not found in PATH and unable to link.',
+          'NO_CLI'
+        );
+      }
+    }
```
EvanBacon, line 149: Move this to a different function.

cltnschlosser:
Thanks, it was `yarn start/build` that I was missing. I was doing `yarn install`. I resolved your comments and added tests. I had to do a lot of shuffling to pass in the projectRoot, but I think the code is generally in better shape than before.

packages/package-manager/src/CocoaPodsPackageManager.ts
```diff
@@ -315,6 +346,22 @@ export class CocoaPodsPackageManager implements PackageManager {
     }
   }
 
+  // Exposed for testing
+  spawnPodCommandAsync(
+    args?: readonly string[],
+    options?: SpawnOptions
+  ): SpawnPromise<SpawnResult> {
+    if (this.shouldUseRubyBundler) {
```
EvanBacon, line 354: This makes me think that RubyBundler should be some sort of subclass or different class. If you need to pass in a boolean that changes core functionality of most features then it is basically a different class.

cltnschlosser: Hmm, I maybe can do that (Still new to TS), but all this is really doing is changing executions from `pod <args>` to `bundle exec pod <args>`. At the command line level it's super simple.

cltnschlosser: I finally got back to this. Removed the boolean flag, create a subclass, and created a "factory" static `create` method for returning the appropriate instance.

cltnschlosser:
@EvanBacon Can you take another look at this when you get a chance? Thanks!

hardworker:
Is there any chance to merge these changes? Looks like this PR is a little bit stuck...
I'm ready to provide help if needed :)

cltnschlosser:
I think this all moved to the expo repo, since this repo is deprecated. Feel free to port these changes :)

cltnschlosser:
https://github.com/expo/expo/pull/24048 This was moved to expo repo

PR #4368, opened 2022-05-12
Rewrite android icon config plugin

EvanBacon:
# Why

Reported issue of jimp working differently than sharp for non-adaptive icons. This appears to be a breaking change between Android versions in SDK 44 and SDK 45.

- This PR rewrites the module to share more code across image generation.
- Adds and removes `android:roundIcon` from the main application in the `AndroidManifest.xml` based on the type of icon used.
- Prevent generating the round icons when non-adaptive icons are enabled.
- Now deletes round icons between runs to make testing easier.

# Test Plan

- `rm -rf .expo android && expo run:android` with `EXPO_IMAGE_UTILS_NO_SHARP` enabled/disabled -- should use small image.
- Add `expo.android.adaptiveIcon.foregroundImage` to the app.json and do it again -- should use filled image.


packages/image-utils/src/Image.ts
```diff
@@ -79,7 +79,8 @@ async function resizeAsync(imageOptions: ImageOptions): Promise<Buffer> {
           backgroundColor && backgroundColor !== 'transparent' ? backgroundColor : 'none'
         }" /></svg>`
       );
-      sharpBuffer.composite([{ input: mask, blend: 'dest-over' }]);
+
+      sharpBuffer.composite([{ input: mask, blend: 'dest-in' }]);
```
brentvatne, line 83: what's the purpose of this change?

brentvatne (approved):

packages/image-utils/src/Image.ts
EvanBacon: This is the more dangerous change, it causes. the circle to actually be applied, without this the image is still square. But the change isn't perfect, in some cases I observed transparent images being created.

Issue #4381, opened 2022-05-17
ios: can't disable device orientation "upside down" / can't set Team ID with dev-client


hirbod:
### Summary

Looks like expo-cli will always toggle "Upside Down" when orientation is set to Portrait. While correct in general, we shoule be able to disable this. Providing the infoPlist settings will be overwritten afaik.

https://github.com/expo/expo-cli/blob/20f71f7d6fd526b776bfbe88a9dd34c6a02486c3/packages/config-plugins/src/ios/Orientation.ts#L40

Please make it possible to specify fixed orientations (Portrai, Upside Down, Landscape Left, Landscape Right).
Also, please add the possibility to set a Team ID, so `expo prebuild --clean` wont reset those settings. 

### Environment

  expo-env-info 1.0.3 environment info:
    System:
      OS: macOS 12.3.1
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 14.19.2 - /usr/local/opt/node@14/bin/node
      Yarn: 1.22.17 - /usr/local/bin/yarn
      npm: 6.14.17 - /usr/local/opt/node@14/bin/npm
      Watchman: 2022.03.21.00 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.11.2 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    IDEs:
      Android Studio: 2021.2 AI-212.5712.43.2112.8512546
      Xcode: 13.4/13F17a - /usr/bin/xcodebuild
    npmGlobalPackages:
      eas-cli: 0.51.0
      expo-cli: 5.4.4
    Expo Workflow: bare

### Please specify your device/emulator/simulator platform, model and version

iOS, all devices

### Reproducible demo or steps to reproduce from a blank project

No repro

PR #4383, opened 2022-05-17
[config plugins] allow overwriting Info.plist values

EvanBacon:
# Why

- [User request](https://twitter.com/nightstomp/status/1526478440773427201).

# How

- Save the original config and pass it around in mods under `modRawConfig`. Mods can observe this config to determine if the user had manually defined the Info.plist key that they would have otherwise modified. If the abstract field is defined (i.e. there's a conflict) then a warning will be thrown.
- We mostly only need this part of the spec for internal use since other plugins run in user space and can be modified.

# Test Plan

- Added tests.


byCedric (approved):
shipit

hirbod:
Fixes https://github.com/expo/expo-cli/issues/4381

Issue #4355, opened 2022-05-09
expo run:android uses yarn even if there is a package-lock.json and no yarn.lock


wodin:
### Summary

As Brent says [here](https://forums.expo.dev/t/android-failure-for-managed-workflows-due-to-npm/63576/8?u=wodin):

> it appears that the default logic when running expo run:android defaults to using yarn if you have it installed, rather than checking for lockfile as it should

### Environment

```
  expo-env-info 1.0.3 environment info:
    System:
      OS: Linux 5.15 Ubuntu 22.04 LTS 22.04 LTS (Jammy Jellyfish)
      Shell: 5.1.16 - /bin/bash
    Binaries:
      Node: 16.13.2 - ~/bin/node
      Yarn: 1.22.18 - ~/.npm-global/bin/yarn
      npm: 8.9.0 - ~/.npm-global/bin/npm
    npmPackages:
      expo: ~45.0.0 => 45.0.1 
      react: 17.0.2 => 17.0.2 
      react-dom: 17.0.2 => 17.0.2 
      react-native: 0.68.1 => 0.68.1 
      react-native-web: 0.17.7 => 0.17.7 
    npmGlobalPackages:
      eas-cli: 0.52.0
      expo-cli: 5.4.3
    Expo Workflow: managed
```

### Please specify your device/emulator/simulator platform, model and version

N/A

### Error output

🧶 Using Yarn to install packages. Pass --npm to use npm instead.

### Reproducible demo or steps to reproduce from a blank project

```
/tmp$ expo init blah -t blank --npm
✔ Downloaded template.
📦 Using npm to install packages.
✔ Installed JavaScript dependencies.

✅ Your project is ready!

To run your project, navigate to the directory and run one of the following npm commands.

- cd blah
- npm start # you can open iOS, Android, or web from here, or run them directly with the commands below.
- npm run android
- npm run ios # requires an iOS device or macOS for access to an iOS simulator
- npm run web
/tmp$ cd blah
/tmp/blah$ ls -l
total 359
-rw-r--r--   1 michael michael    454 1985-10-26 10:15 App.js
-rw-r--r--   1 michael michael    675 2022-05-09 13:09 app.json
drwxrwxr-x   2 michael michael      6 2022-05-09 13:09 assets
-rw-r--r--   1 michael michael    107 1985-10-26 10:15 babel.config.js
drwxrwxr-x 593 michael michael    594 2022-05-09 13:10 node_modules
[...]

PR #4387, opened 2022-05-21
fix(package-manager): should not use yarn when npm lockfile exists

ger88555:
# Why

- This fixes [#4355](https://github.com/expo/expo-cli/issues/4355#issue-1229575049)

As Brent says [here](https://forums.expo.dev/t/android-failure-for-managed-workflows-due-to-npm/63576/8?u=wodin):

>it appears that the default logic when running expo run:android defaults to using yarn if you have it installed, rather than checking for lockfile as it should

# How

So now Yarn is still favored over NPM, but if there's already a `package-lock.json` in the project root, NPM will be used.

# Test Plan

- See tests

### DIFF
--- a/packages/package-manager/src/utils/__tests__/shouldUseYarn-test.ts
+++ b/packages/package-manager/src/utils/__tests__/shouldUseYarn-test.ts
@@ -1,6 +1,12 @@
+import { isUsingNpm } from '../nodeWorkspaces';
 import shouldUseYarn from '../shouldUseYarn';
 
+export const asMock = <T extends (...args: any[]) => any>(fn: T): jest.MockedFunction<T> =>
+  fn as jest.MockedFunction<T>;
+
 beforeAll(() => {
+  jest.mock('../nodeWorkspaces', () => ({ isUsingNpm: jest.fn() }));
+
   jest.mock('child_process', () => {
     return {
       execSync: () => {
@@ -8,13 +14,6 @@ beforeAll(() => {
           throw new Error('failed');
         }
         return 'something';
-      },
-    };
-  });
-  jest.mock('../nodeWorkspaces', () => {
-    return {
-      isUsingNpm: () => {
-        return !!process.env.TEST_NPM_LOCKFILE_EXISTS;
       },
     };
   });
@@ -44,6 +43,6 @@ it(`returns false if yarn -v throws an error`, () => {
 
 it(`returns false if npm lockfile exists`, () => {
   // Use project with npm lockfile
-  process.env.TEST_NPM_LOCKFILE_EXISTS = 'true';
+  asMock(isUsingNpm).mockClear().mockReturnValueOnce(true);
   expect(shouldUseYarn()).toBe(false);
 });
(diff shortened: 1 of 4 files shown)


kbrandwijk:
There's an existing function `resolvePackageManager` that already takes lockfiles into account to decide which package manager to use. https://github.com/expo/expo-cli/blob/a40a58cf72eea85b2811d3e336430d71f53edfad/packages/package-manager/src/utils/nodeWorkspaces.ts#L86-L113

ger88555:
Thank you @kbrandwijk

Now `shouldUseYarn` is using `resolvePackageManager` to look for the presence of an npm lockfile.

---

Although this new function:

https://github.com/ger88555/expo-cli/blob/3e21b68526073ec9217d5adbd58d96ae932707b2/packages/package-manager/src/utils/shouldUseYarn.ts#L19-L24

is somewhat similar to

https://github.com/expo/expo-cli/blob/a40a58cf72eea85b2811d3e336430d71f53edfad/packages/package-manager/src/NodePackageManagers.ts#L34-L45

save the arguments and the package manager it's testing for.

Would `isUsingYarn` do the same as `resolvePackageManager(projectRoot, 'yarn') !== null` and with the same steps?

Would it be a good idea to move both similar functions into `utils/nodeWorkspace.ts`?

ger88555:
Take a look at the last commit, that's how having both `isUsingYarn` and `isUsingNpm` on `utils/nodeWorkspaces.ts` would look like.

Let me know if I should revert it.

packages/package-manager/src/utils/shouldUseYarn.ts
```diff
@@ -1,11 +1,19 @@
...
+
 export default function shouldUseYarn(): boolean {
   try {
     if (process.env.npm_config_user_agent?.startsWith('yarn')) {
       return true;
     }
+
     execSync('yarnpkg --version', { stdio: 'ignore' });
+
+    if (isUsingNpm(process.cwd())) {
```
EvanBacon, line 13: Why verify that yarn is installed just to throw the result away and check for the lock file?  Maybe move this before the execSync call. Also since you're here, move that and the `if (process...` outside of the try/catch block since the try is for the execSync but it might suppress useful warnings from the other two statements.

packages/package-manager/src/utils/__tests__/shouldUseYarn-test.ts
```diff
@@ -11,6 +11,13 @@ beforeAll(() => {
...
     };
   });
+  jest.mock('../nodeWorkspaces', () => {
+    return {
+      isUsingNpm: () => {
+        return !!process.env.TEST_NPM_LOCKFILE_EXISTS;
+      },
+    };
+  });
 });
```
EvanBacon, line 21:
```suggestion
  jest.mock('../nodeWorkspaces', () => ({ isUsingNpm: jest.fn() }))
});
```
Then import `isUsingNpm` and do something like this:



```ts
import { isUsingNpm } from '../path/to/thing';

const asMock = <T extends (...args: any[]) => any>(fn: T): jest.MockedFunction<T> =>
  fn as jest.MockedFunction<T>;

it(`does foobar`, () => {
  // Mock what should be returned for the test case.
  asMock(isUsingNpm).mockClear().mockReturnValueOnce(true);
})


```

ger88555:
Thank you @EvanBacon 

Just pushed those suggestions, let me know if one of them wasn't carried out well, or if another one becomes apparent.

EvanBacon (approved):

PR #4407, opened 2022-06-02
[config] Include automatically EAS Build config plugin

wkozyra95:
# Why

General case: Sometimes we might want to modify(or maybe just read) manifest values when building on eas
Specific case: Specific issue we want to resolve here is to allow us to override android.versionCode and ios.buildNumber. To support autoincrementing versions we need a way to update versions in the project, if someone is using app.conifg.js we can't modify the file, but if we use the config plugin we can handle all cases.

# How

If EAS_BUILD_CONFIG_PLUGIN_PATH is set add value of that env as location of the plugin you want to inject

# Test Plan

`expo config` with example plugin

### DIFF
--- a/packages/config/src/plugins/withConfigPlugins.ts
+++ b/packages/config/src/plugins/withConfigPlugins.ts
@@ -9,7 +9,14 @@ import { serializeAfterStaticPlugins } from '../Serialize';
  * @param projectRoot
  */
 export const withConfigPlugins: ConfigPlugin<boolean> = (config, skipPlugins) => {
-  // @ts-ignore: plugins not on config type yet -- TODO
+  if (process.env.EAS_BUILD_CONFIG_PLUGIN_PATH) {
+    const easPlugin = process.env.EAS_BUILD_CONFIG_PLUGIN_PATH;
+    if (!Array.isArray(config.plugins) || !config.plugins?.length) {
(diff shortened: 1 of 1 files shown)


dsokal (approved):

EvanBacon (changes requested):
As it stands this is extremely dangerous, adding a global variable to inject any modification to any config read is not something we should add to the `expo/config` package. We need to reliably predict the in/out of the config at any time.

The `expo/config` package is the lowest common denominator for all Expo tooling, this change could potentially be applied to any command or context that Expo CLI could be used in.

wkozyra95:
> adding a global variable to inject any modification to any config read is not something we should add to the expo/config package

I agree with the statement, but I fail to see where is the risk in this case?  Are you afraid that someone will have this env set by accident?

Is the objection here that I am adding a plugin based on env or that I'm adding it here? 

The goal here is to inject the plugin automatically, it does not really matter to me if that plugin is in `config.plugins` list or not, all I want is for that plugin to get executed. Is there a better place to add it?
If I understand correctly we already have a bunch of default plugins that are always added, would it be better if I add that logic wherever those defaults are added?
 
 >  this change could potentially be applied to any command or context that Expo CLI could be used.
 
almost any change in config plugins and config affects the results of the getConfig and we already have logic there that depends on envs.

PR #4422, opened 2022-06-13
Make Metro start with IPv4 address

### LINKED ISSUE #17843: Hermes debugging websocket URL generates as IPv6, causes confusion/incompatibility in some environments
### Summary

Using the Hermes debugger with Flipper is... tough. A number of issues in the Expo, React Native and Flipper projects demonstrate users have difficulty connecting to the Hermes debugger through Metro. (E.g. https://github.com/expo/expo/issues/15436#issuecomment-986075733, https://github.com/facebook/flipper/issues/2615)

After a bit of sleuthing, it appears to me this could be, in part, due to Expo CLI calling `Metro.runServer()` without explicitly setting the `host` option. This results in the `HttpServer` starting in IPv6-capable environments listening on `::`, which is then used by the inspector proxy to generate a Websocket URL on `ws://[::1]`. If you're running Expo/Metro inside of a Dockerized environment, for instance, and don't forward ports on IPv6, this can lead to a very WTF condition when trying to connect the Hermes debugger.

### What platform(s) does this occur on?

Android, iOS

### SDK Version

45

### Environment

```  expo-env-info 1.0.3 environment info:
    System:
      OS: Linux 5.17 Debian GNU/Linux 10 (buster) 10 (buster)
      Shell: 5.0.3 - /bin/bash
    Binaries:
      Node: 16.13.1 - /tmp/yarn--1655147751222-0.8076740935194806/node
      Yarn: 1.22.15 - /tmp/yarn--1655147751222-0.8076740935194806/yarn
      npm: 8.1.2 - /usr/local/bin/npm
    npmPackages:
      expo: ^45.0.0 => 45.0.5 
      react: 17.0.2 => 17.0.2 
      react-dom: 17.0.2 => 17.0.2 
      react-native: 0.68.2 => 0.68.2 
      react-native-web: 0.17.7 => 0.17.7 
    Expo Workflow: bare
```

### Reproducible demo

Start a Hermes Expo project. Navigate to `http://localhost:8081` and observe `webSocketDebuggerUrl` is IPv6:

```json
[
  {
    "id": "0-3",
    "description": "dating.kinksters.meet",
    "title": "Hermes React Native",
    "faviconUrl": "https://reactjs.org/favicon.ico",
    "devtoolsFrontendUrl": "chrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=%5B%3A%3A1%5D%3A8081%2Finspector%2Fdebug%3Fdevice%3D0%26page%3D3",
    "type": "node",
    "webSocketDebuggerUrl": "ws://[::1]:8081/inspector/debug?device=0&page=3",
    "vm": "Hermes"
  },
  ...
]
```

This requires Metro be accessible over IPv6.

if the IP is set explicitly to IPv4 when constructing the metro server, the URL is then `ws://localhost:8081/...`.

bradjones1:
# Why

Fixes https://github.com/expo/expo/issues/17843

# How

Starting Metro with an IPv4 address ensures that when it generates a websocket URL, it is also IPv4. E.g., helpful in containerized environments.

# Test Plan

Load `http://localhost:8081/json` when Metro is running and `webSocketDebuggerUrl` should be `ws://localhost...`

### DIFF
--- a/packages/dev-server/src/MetroDevServer.ts
+++ b/packages/dev-server/src/MetroDevServer.ts
@@ -111,6 +111,8 @@ export async function runMetroDevServerAsync(
   const server = await Metro.runServer(metroConfig, {
     hmrEnabled: true,
     websocketEndpoints,
+    // Ensure Metro starts with an IPv4 address for maximum interoperability.
+    host: "0.0.0.0",
   });
 
   if (attachToServer) {


bradjones1:
Related https://github.com/expo/expo-cli/pull/4394

Kudo:
heya! i checked the rn-cli project and it doesn't specify `host`. in the case when the websocket url is `ws://[::1]:8081/inspector/debug?device=1&page=-1` flipper can still work. could you share more reproducible problem where makes flipper not working?

bradjones1:
@Kudo 

> flipper can still work. could you share more reproducible problem where makes flipper not working?

Thanks for the reply. My issue is when the networking is not all on the same interface, e.g. I run Metro in a container. In that case the port forwards for Metro are IPv4, and it's non-trivial to add IPv6 forwards, as well. The linked issue goes into my thought process a little more. It seems that some less networking-savvy users are also tripped up by a potential mix of networking stacks.

So yes, it should "just work," but the Metro inspector proxy also [does some weirdness here that doesn't help](https://github.com/facebook/metro/blob/fdc4ef17230ee093cfd302ddc8c59a88019b1f02/packages/metro-inspector-proxy/src/InspectorProxy.js#L98-L103). If the stack is IPv6, it converts `[::]` to `[::1]`, but then if it's IPv4, you get a hostname (`localhost`) which is not inherently IPv4, right?

To be fair, my solution might not be the right one, but it does seem that Metro juggles networking stacks/addresses in a way that is confusing at the very least. I'm thinking perhaps we can hedge against that, given Expo runs what I'd call an opinionated implementation of the bundler.

Kudo:
@bradjones1 thank to share the insight. in the case, how do you think if creating a pr for metro? or at least i'm thinking to add a cli command argument like `expo start --metro-server-host 0.0.0.0` is more flexible.

bradjones1:
I think a PR on Metro wouldn't be necessary as it already provides the option to set the host... so it's tools like `expo-cli` that likely need to implement that.

Making it an option on the start command seems reasonable enough to me. Thanks for the feedback.

Kudo:
for metro, i mean the [weirdness you mentioned](https://github.com/facebook/metro/blob/fdc4ef17230ee093cfd302ddc8c59a88019b1f02/packages/metro-inspector-proxy/src/InspectorProxy.js#L98-L103)

bradjones1:
@kbrandwijk That's different; I am also using this setting but they do not interact/this is different.

bradjones1:
@Kudo Ah ok, yes good point... I can open an issue other there referencing this as an example of "localhost ambiguity," I suppose. Thanks for the clarification.

mgscreativa:
I can confirm that this solves the issue on iOS, but on Android I still get 

    No compatible apps connected. JavaScript Debugging can only be used with the Hermes engine

bradjones1:
@mgscreativa Confirm you're running Hermes on Android?

mgscreativa:
Hi @bradjones1!  I confirm that I'm running Hermes on both Android and iOS. I've put this code on App.tsx:

```
  const isHermes = () => !!global.HermesInternal;
  console.log('isHermesEnabled:', isHermes());
```

And, in both platforms, I get  `LOG  isHermesEnabled: true`

byCedric:
Hi all! React Native (and Expo) has now swapped over to new debugging infrastructure. This involves a lot of new changes around the inspector proxy and what tools you can use. It now supports core features of the Chrome DevTools protocol.

If you think we should still move forward with this PR, please open a new PR on [github.com/expo/expo](https://github.com/expo/expo).

PR #4436, opened 2022-06-22
fix license

EvanBacon:
# Why

# How

# Test Plan


kbrandwijk:
Might as well update the year to say 2022 as well while you're in there...

PR #4439, opened 2022-06-23
fix: fix auto TypeScript version check

makhataibar:
# Why

Ensure we check against the package.json for versions.


EvanBacon (approved):

makhataibar:
@EvanBacon when will be this fix in new expo-cli version?

PR #4441, opened 2022-06-23
[prebuild-config] Remove the dependency on expo-modules-autolinking

tsapeta:
# Why

`expo-modules-autolinking` being a dependency of `@expo/prebuild-config` is the most common reason of duplicated versions of the autolinking installed in the project. To avoid such issues the `expo` package must be the only one that depends on `expo-modules-autolinking`.

# How

- Instead of importing the autolinking statically, I use Node's resolution to find the path to `expo-modules-autolinking` that is installed in the project where the prebuild command is run
- Removed `expo-modules-autolinking` from the dependencies

# Test Plan

- Tested `dexpo prebuild` on a fresh project and confirmed that the results are the same as before (`searchPaths` and `platformPaths`)
- Ensured that `require.resolve` and especially `paths` option is available in all versions of Node that we support (it's available as of v8.9.0 so yes)

### DIFF
--- a/packages/prebuild-config/package.json
+++ b/packages/prebuild-config/package.json
@@ -46,6 +46,9 @@
     "semver": "7.3.2",
     "xml2js": "0.4.23"
   },
+  "peerDependencies": {
+    "expo-modules-autolinking": ">=0.8.1"
+  },
   "publishConfig": {
     "access": "public"
   }
--- a/packages/prebuild-config/package.json
+++ b/packages/prebuild-config/package.json
@@ -41,7 +41,6 @@
     "@expo/image-utils": "0.3.20",
     "@expo/json-file": "8.2.36",
     "debug": "^4.3.1",
-    "expo-modules-autolinking": "0.8.1",
     "fs-extra": "^9.0.0",
     "resolve-from": "^5.0.0",
     "semver": "7.3.2",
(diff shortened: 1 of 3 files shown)


packages/prebuild-config/src/getAutolinkedPackages.ts
```diff
@@ -64,3 +61,19 @@ export function shouldSkipAutoPlugin(
...
+/**
+ * Imports the `expo-modules-autolinking` package installed in the project at the given path.
+ */
+function importExpoModulesAutolinking(projectRoot: string) {
+  try {
+    const resolvedAutolinkingPath = require.resolve('expo-modules-autolinking/build/autolinking', {
+      paths: [projectRoot],
+    });
+    return require(resolvedAutolinkingPath);
+  } catch (error) {
```
EvanBacon, line 74:
```suggestion
  } catch {
```

packages/prebuild-config/package.json
```diff
@@ -41,7 +41,6 @@
     "@expo/image-utils": "0.3.20",
     "@expo/json-file": "8.2.36",
     "debug": "^4.3.1",
-    "expo-modules-autolinking": "0.8.1",
```
EvanBacon, line 44: Maybe move this to peer deps so the supported version is known?

packages/prebuild-config/src/getAutolinkedPackages.ts
```diff
@@ -64,3 +61,19 @@ export function shouldSkipAutoPlugin(
   }
   return false;
 }
+
+/**
+ * Imports the `expo-modules-autolinking` package installed in the project at the given path.
+ */
+function importExpoModulesAutolinking(projectRoot: string) {
```
EvanBacon, line 68: Add return types

EvanBacon (approved):

PR #4448, opened 2022-06-29
[doctor] Validate that projects do not include illegal packages

brentvatne:
# Why

Any SDK 45+ project that includes `@unimodules/core`, `@unimodules/react-native-adapter`, and `react-native-unimodules` is likely not going to build. We should let people know about this and `doctor` is a good spot.

# How

I adapted our existing checks for deep dependencies to let us check for the existence of a package that should not exist. I changed the output slightly for this to make sense and to remove some (I think) unnecessary logs. See below for some examples.

# Test Plan

https://user-images.githubusercontent.com/90494/176566254-808fa54d-c983-4f7b-aa6c-d4649ce706e6.mov

https://user-images.githubusercontent.com/90494/176566264-f4766dff-000d-4ab3-a0fd-e5ef80331fb0.mov

https://user-images.githubusercontent.com/90494/176566279-0f0bc8b9-debf-4e07-bc79-1b2cfeb2d4e7.mov

### DIFF
--- a/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
+++ b/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
@@ -32,7 +32,23 @@ async function validateSupportPackagesAsync(sdkVersion: string): Promise<boolean
       }
     }
   }
+
+  // None of these packages should exist in any modern project
+  const illegalPackages = [
+    '@unimodules/core',
+    '@unimodules/react-native-adapter',
+    'react-native-unimodules',
+  ];
+
+  for (const pkg of illegalPackages) {
+    const isPackageAbsent = await warnAboutDeepDependenciesAsync({ name: pkg });
+    if (!isPackageAbsent) {
+      allPackagesValid = false;
+    }
+  }
+
   Log.newLine();
+
   return allPackagesValid;
 }
 
--- a/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
+++ b/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
@@ -32,24 +32,27 @@ async function validateSupportPackagesAsync(sdkVersion: string): Promise<boolean
       }
     }
   }
+  return allPackagesValid;
+}
 
-  // None of these packages should exist in any modern project
+// Ensures that a set of packages
+async function validateIllegalPackagesAsync(): Promise<boolean> {
   const illegalPackages = [
     '@unimodules/core',
     '@unimodules/react-native-adapter',
     'react-native-unimodules',
   ];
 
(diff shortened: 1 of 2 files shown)


kbrandwijk:
It's a bit arbitrary if we should or should not show logs if nothing is wrong. For the other 3 packages we check, we confirm that they satisfy semver, for this new check, we only report if something's wrong.

I think I personally like this new approach better, but I'd like to make it consistent, and remove the additional logging for succesful semver checks. That way, if nothing's wrong, you only see `Didn't find any issues with the project.` which I think is a nice DX.

Maybe we can hide the successful logs behind a debug flag, in which case we should probably have them for these checks as well? So that in debug mode, you can actually see all the checks it's doing.

packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
```diff
@@ -32,7 +32,23 @@ async function validateSupportPackagesAsync(sdkVersion: string): Promise<boolean
...
+    '@unimodules/core',
+    '@unimodules/react-native-adapter',
+    'react-native-unimodules',
+  ];
+
+  for (const pkg of illegalPackages) {
+    const isPackageAbsent = await warnAboutDeepDependenciesAsync({ name: pkg });
+    if (!isPackageAbsent) {
+      allPackagesValid = false;
+    }
```
kbrandwijk, line 47: We want them to be absent, right? So how does the fact that a package is absent toggle `allPackagesValid` to `false`? This is probably a naming issue, and not a logic issue.

kbrandwijk (changes requested):

brentvatne:
> It's a bit arbitrary if we should or should not show logs if nothing is wrong. For the other 3 packages we check, we confirm that they satisfy semver, for this new check, we only report if something's wrong.
> 
> I think I personally like this new approach better, but I'd like to make it consistent, and remove the additional logging for succesful semver checks. That way, if nothing's wrong, you only see `Didn't find any issues with the project.` which I think is a nice DX.
> 
> Maybe we can hide the successful logs behind a debug flag, in which case we should probably have them for these checks as well? So that in debug mode, you can actually see all the checks it's doing.

makes sense, updated accordingly

packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
brentvatne: i clarified in naming in latest commit

kbrandwijk (approved):

brentvatne:
merged this so we have it in place when @kbrandwijk extracts the `expo-doctor` command. we need to follow up on experimenting with parallelizing as evan suggested

PR #4449, opened 2022-06-30
Deleted config packages

EvanBacon:
# Why

Migrate the config packages to expo/expo so they can be more easily modified with the new versioned CLI. [Reference](https://github.com/expo/expo/pull/17958).

# How

- Deleted source for `@expo/config`, `@expo/config-plugins`, `@expo/prebuild-config`, `@expo/config-types`.
- Left READMEs in place with info about where the code moved to.

# Test Plan

- Tests should continue to pass


brentvatne:
(i merged because i wanted to get this out of expo-cli repo asap because it's confusing to have the code in multiple places)

PR #4456, opened 2022-07-05
add better warning when prebuilding in non-interactive mode

wkozyra95:
# Why

currently, if someone has a managed project with an ios directory, or is writing eas secrets to ios directory e.g.

```
echo "$IOS_GOOGLE_SERVICES_BASE64" | base64 -d > ios/GoogleService-Info.plist
```
then only warning they are getting that ios directory is deleted is log

```
- Clearing ios
✔ Cleared ios code
```

# How

print working when rebuilding in non-interactive mode

# Test Plan

### DIFF
--- a/packages/expo-cli/src/commands/eject/clearNativeFolder.ts
+++ b/packages/expo-cli/src/commands/eject/clearNativeFolder.ts
@@ -104,6 +104,9 @@ export async function promptToClearMalformedNativeProjectsAsync(
       initial: true,
     }))
   ) {
+    if (isNonInteractive()) {
+      Log.warn(`${message}, project files will be cleared and reinitialized.`);
+    }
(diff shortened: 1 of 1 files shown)


dsokal (approved):

packages/expo-cli/src/commands/eject/clearNativeFolder.ts
```diff
@@ -104,6 +104,9 @@ export async function promptToClearMalformedNativeProjectsAsync(
       initial: true,
     }))
   ) {
+    if (isNonInteractive()) {
```
EvanBacon, line 107: It's unclear if any changes landed here will be published, I recommend opening here https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/prebuild/clearNativeFolder.ts

PR #4465, opened 2022-07-13
Deleted metro packages

EvanBacon:
# Why

- Delete packages that are moved in https://github.com/expo/expo/pull/18221

# How

- Followed same formula as https://github.com/expo/expo-cli/pull/4449

# Test Plan


EvanBacon:
Unblocked by https://github.com/expo/expo/pull/18221

PR #4471, opened 2022-07-15
[install-expo-modules] Add SDK 46 support

Kudo:
# Why

close ENG-5724

# How

- add react-native 0.69 with sdk 46 version info
- add missing `BuildConfig.IS_NEW_ARCHITECTURE_ENABLED` migration in MainActivity. this is also available from sdk 45. it's just me forgot to add it from template.
- skip the prompt for ios deployment target update if the requirement is already meet.

# Test Plan

- ci  and unit tests passed
- `npx install-expo-modules` on a react-native 0.68 project
- `npx install-expo-modules` on a react-native 0.69 project

### DIFF
--- a/packages/install-expo-modules/src/utils/expoVersionMappings.ts
+++ b/packages/install-expo-modules/src/utils/expoVersionMappings.ts
@@ -10,6 +10,11 @@ export interface VersionInfo {
 export const ExpoVersionMappings: VersionInfo[] = [
   // Please keep sdk versions in sorted order (latest sdk first)
   {
+    expoSdkVersion: '46.0.0',
+    iosDeploymentTarget: '12.4',
+    reactNativeVersionRange: '>= 0.69.0',
+  },
+  {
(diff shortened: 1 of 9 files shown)


packages/install-expo-modules/src/utils/expoVersionMappings.ts
```diff
@@ -9,6 +9,11 @@ export interface VersionInfo {
 
 export const ExpoVersionMappings: VersionInfo[] = [
   // Please keep sdk versions in sorted order (latest sdk first)
+  {
+    expoSdkVersion: '46.0.0',
+    iosDeploymentTarget: '12.4',
+    reactNativeVersionRange: '>= 0.69.0',
```
Kudo, line 15: i think sdk 46 may also support react-native 0.68 or lower versions. i'm not clear whether we should aggressively promote latest sdk version or keep to use sdk 45 where we are confident supported.

brentvatne (approved):

PR #4481, opened 2022-07-25
chore(create-expo-app): drop all URL caching in favor of temporary cache

EvanBacon:
# Why

- resolve ENG-5816
- drop unimplemented `node-fetch` features.
- prevent caching local file path templates
- support `EXPO_NO_CACHE`
- use tmpdir cache to reset caches when the computer restarts or runs out of memory.


packages/create-expo-app/src/utils/npm.ts
```diff
@@ -18,15 +18,16 @@ type ExtractProps = {
...
 // @ts-ignore
 const pipeline = promisify(Stream.pipeline);
 
-const cachedFetch = createFetch({
-  cacheDirectory: 'template-cache',
-  // Set no timeout on the templates since they're versioned already.
-});
+function getTemporaryCacheFilePath(subdir: string = 'template-cache') {
+  // This is cleared when the device restarts
+  return path.join(os.tmpdir(), '.create-expo-app', subdir);
```
byCedric, line 29: Not sure if relevant, but if we want to support the [XDG base spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), we probably also need to check for `$XDG_CACHE_HOME`

packages/create-expo-app/src/utils/npm.ts
```diff
@@ -244,32 +234,49 @@ function isNpmPackageInfo(item: any): item is NpmPackageInfo {
...
 
   await fs.promises.mkdir(cachePath, { recursive: true });
 
-  const info = await getNpmInfoAsync(npmName);
+  const info = await getNpmInfoAsync(npmName, cachePath);
 
+  const cacheFilename = path.join(cachePath, info.filename);
   try {
-    const cacheFilename = path.join(cachePath, info.filename);
+    // TODO: This cache does not expire.
```
byCedric, line 260: Is this TODO important to fix now?

packages/create-expo-app/package.json
```diff
@@ -36,21 +36,17 @@
...
     "@types/minipass": "^3.1.2",
     "@types/node": "^12.6.8",
     "@types/prompts": "2.0.8",
     "@types/tar": "^6.1.1",
     "arg": "^5.0.1",
-    "cacache": "^15.3.0",
     "chalk": "^4.0.0",
     "debug": "^4.3.1",
     "execa": "^4.0.1",
-    "form-data": "^2.3.2",
```
byCedric, line 53: Love it, drop it

byCedric (approved):
LGTM, left 2 comments

packages/create-expo-app/src/utils/npm.ts
EvanBacon: possibly, if we want remote templates to eventually expire without the computer restarting.

packages/create-expo-app/src/utils/npm.ts
EvanBacon: Unclear if this applies, since we don't have an example of how XDG overlaps with the tmpdir folder, we should skip. Easy enough to add in the future when we have more clarity.

packages/create-expo-app/src/utils/npm.ts
kbrandwijk: Restarting the computer doesn't clear the cache (on Windows).

PR #4505, opened 2022-08-09
feat(cli): add migration logs when available

EvanBacon:
# Why

- resolve ENG-5940

# How

- When the user is running a replaced command in an SDK 46 project that doesn't have the local CLI disabled, warn that they should migrate.

Screen Shot 2022-08-09 at 8 20 55 PM

# Test Plan

- `expod start` in an SDK 46 project -- warning.
- `expod config` in an SDK 44 project -- no warning.
- `expo init` anywhere -- warning.


brentvatne (approved):
i updated the message slightly, this looks perfect. we can possibly add a link to your upcoming blog post as well before shipping or just ship this and add it later

brentvatne:
let's just ship it and add link later

PR #4519, opened 2022-08-16
Update README.md

EvanBacon:
# Why

Make it clear that this repo is no longer the home of the latest Expo CLI, redirect traffic to the new location.

- Demo https://github.com/expo/expo-cli/blob/@evanbacon/readme/redirect/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,32 +1,6 @@
-<!-- Title -->
+# Assorted Expo Dev Tools
 
-<p align="center">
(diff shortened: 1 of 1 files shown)


Simek (approved):
LGTM, but I have one question - shouldn't we remove `expo-cli` from the packages list, or it is too early call?

EvanBacon:
@Simek I think it's a bit too early to start dropping things like `expo-cli` at the moment

PR #4536, opened 2022-08-31
[CEA] bump and cleanup dependencies

Simek:
# Why

Lately, when checking the CEA code I have spotted that Node related dependencies are outdated, and this simple bump escalated a bit.

# How

This PR bump most of the dependencies (`node-fetch` v3 is ESM, `ora` or `chalk` deserve separate PRs) and removes the `execa` used only in test, which has been replaced with already present `@expo/spawn-async`. 

Additionally, I have spotted that we only have `fs-extra` types defined, but not the package. Again, it looks like this has been only used in tests, and that probably why this was even working (via shadow dependency - defined somewhere in the workspace by other package). All usages of `fs-extra` has been replaced with `fs`.

# Test Plan

Package builds successfully, I have run bot normal and E2E test locally and there were no failures.


brentvatne (approved):
test failures appear to be caused by https://github.com/expo/universe/pull/10442 cc @EvanBacon

PR #4538, opened 2022-09-01
Handle multiple query parameters in Android URIs

Minishlink:
# Why

Currently, `npx uri-scheme open --android "myscheme://test?varA=1&varB=2"` launch the app with a deeplink `myscheme://test?varA=1` instead of `myscheme://test?varA=1&varB=2`

# How

Apparently adb expects the character `&` to be escaped.

# Test Plan

Tested with the current version and `npx uri-scheme open --android "myscheme://test?varA=1\&varB=2"`


EvanBacon (approved):
thanks

PR #4550, opened 2022-09-16
[cli] update script generating docs content

Simek:
# Why

Currently generated code do not display correctly, after the update to MDX v2.

# How

This PR replaces the raw HTML output with the dedicated docs component.

# Test Plan

TODO


EvanBacon (changes requested):
This script is no longer in use. We don't doc gen the CLI anymore.

PR #4554, opened 2022-09-21
[install-expo-modules] Replace xcode with xcparse

Kudo:
# Why

`simple-plist@1.3.1` is a transitive dependency of the xcode package. it has [an incompatible issue](https://github.com/wollardj/simple-plist/issues/58) with ncc. the issue is pending for a while without publishing newer version as latest dist-tag. that blocks us from updating install-expo-modules.

# How

replace xcode with the awesome xcparse. this pr introduces a `withXCParseXcodeProject` config-plugin internally.

# Test Plan

- ✅ unit test passed
- ✅ manual test on a react-native 0.68 project
- ✅ manual test on a react-native 0.69 project

### DIFF
--- a/packages/install-expo-modules/package.json
+++ b/packages/install-expo-modules/package.json
@@ -47,7 +47,8 @@
     "glob": "7.1.6",
     "prompts": "^2.3.2",
     "resolve-from": "^5.0.0",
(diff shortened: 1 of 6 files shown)


EvanBacon (approved):
looks great!

Issue #4553, opened 2022-09-19
[next-adapter] unable to resolve dependency tree


philiplindberg:
### Summary

I'm trying to use Next.js with my existing Expo app as per [this guide](https://docs.expo.dev/guides/using-nextjs/). 
However, I'm getting the following error when running `npm i --save-dev @expo/next-adapter`:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: expo-next@1.0.0
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   react@"18.0.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^17.0.2" from next@11.1.4
npm ERR! node_modules/next
npm ERR!   peer next@"^11" from @expo/next-adapter@4.0.12
npm ERR!   node_modules/@expo/next-adapter
npm ERR!     dev @expo/next-adapter@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See /Users/philiplindberg/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/philiplindberg/.npm/_logs/2022-09-19T15_12_42_884Z-debug-0.log
```

### Environment

  expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 12.6
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 18.7.0 - /opt/homebrew/bin/node
      Yarn: 1.22.17 - /opt/homebrew/bin/yarn
      npm: 8.15.0 - /opt/homebrew/bin/npm
    Managers:
      CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.4, iOS 16.0, macOS 12.3, tvOS 16.0, watchOS 9.0
    IDEs:
      Android Studio: 2021.2 AI-212.5712.43.2112.8815526
      Xcode: 14.0/14A309 - /usr/bin/xcodebuild
    npmPackages:
      @expo/webpack-config: ^0.17.0 => 0.17.2
      expo: ~46.0.9 => 46.0.9
[...]

Issue #4548, opened 2022-09-14
@expo/next-adapter unexpected property, projectRoot


linojon:
### Summary

Trying the create a new expo project with next.js, i followed the steps here: o	https://github.com/expo/expo/blob/master/docs/pages/guides/using-nextjs.md#add-nextjs-to-expo-projects as follows:

```
$ npx create-expo-app -t expo-template-blank-typescript
cd to the folder
$ yarn add @expo/next-adapter
$ yarn next-expo
$ yarn next dev
```

it tries to start with the following message, and browsing to http://localhost:3000 loads nothing.

```
  - The root value has an unexpected property, projectRoot, which is not in the list of allowed properties (amp, analyticsId, assetPrefix, basePath, cleanDistDir, compiler, compress,
 crossOrigin, devIndicators, distDir, env, eslint, excludeDefaultMomentLocales, experimental, exportPathMap, future, generateBuildId, generateEtags, headers, httpAgentOptions, i18n, 
images, onDemandEntries, optimizeFonts, output, outputFileTracing, pageExtensions, poweredByHeader, productionBrowserSourceMaps, publicRuntimeConfig, reactStrictMode, redirects, rewr
ites, sassOptions, serverRuntimeConfig, staticPageGenerationTimeout, swcMinify, trailingSlash, typescript, useFileSystemPublicRoutes, webpack).
```

### Environment

  expo-env-info 1.0.5 environment info:
    System:
      OS: Windows 10 10.0.22000
    Binaries:
      Node: 16.17.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.4.0.0 AI-183.6156.11.34.5522156
    npmPackages:
      expo: ~46.0.9 => 46.0.10
      react: 18.0.0 => 18.0.0
      react-dom: 18.0.0 => 18.0.0
      react-native: 0.69.5 => 0.69.5
      react-native-web: ~0.18.7 => 0.18.9
    Expo Workflow: managed

### Please specify your device/emulator/simulator platform, model and version

any

### Error output

see above

### Reproducible demo or steps to reproduce from a blank project

see above

the-programmer-alex:
Get the same, but when trying to add Expo to an existing Nextjs project.

chrisarts:
[...]

Issue #4523, opened 2022-08-21
Nextjs Warning: "root value has unexpected property" with @expo/next-adapter


ericvicenti:
### Summary

Hi friends, this is low-priority because nothing is broken. But there is log spew from next now whenever I run with @expo/next-adapter and its a little embarrassing. Pretty sure this was caused by nextjs adding the warning recently

See the "error output" for the warning. Again, this is not an exception or a breakage

Side note: I realized that `yarn next-expo` is confused about typescript and created .js files alongside my .tsx files. But this may be due to my monorepo setup and I don't have time to isolate that issue. Also I had to add back `next-transpile-modules` to my `next.config.js` in order get things working in the monorepo after `yarn next-expo` did its thing 🤷‍♂️ these are perhaps more severe issues for people trying to set up a next+expo+ts repo

Thank youuu, kind Expo people 😍

### Environment

latest expo things

```
    "expo": "^46.0.8",
    "@expo/config-types": "^46.0.2",
    "@expo/next-adapter": "^4.0.12",
```

latest next things

```
    "next": "12.2.5",
    "next-compose-plugins": "^2.2.1",
    "next-transpile-modules": "^9.0.0",
```

### Please specify your device/emulator/simulator platform, model and version

macOS node 16.17.0

### Error output

```
$ yw bug-expo-next-adapter dev
yarn workspace v1.22.19
yarn run v1.22.19
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
warn  - Invalid next.config.js options detected: 
  - The root value has an unexpected property, webpackDevMiddleware, which is not in the list of allowed properties (amp, analyticsId, assetPrefix, basePath, cleanDistDir, compiler, compress, crossOrigin, devIndicators, distDir, env, eslint, excludeDefaultMomentLocales, experimental, exportPathMap, future, generateBuildId, generateEtags, headers, httpAgentOptions, i18n, images, onDemandEntries, optimizeFonts, output, outputFileTracing, pageExtensions, poweredByHeader, productionBrowserSourceMaps, publicRuntimeConfig, reactStrictMode, redirects, rewrites, sassOptions, serverRuntimeConfig, staticPageGenerationTimeout, swcMinify, trailingSlash, typescript, useFileSystemPublicRoutes, webpack).
[...]

Issue #4509, opened 2022-08-11
[next-js] Error: Unable to resolve module ./index from <PROJECT_ROOT>


oVerde:
### Summary

The project can't even start, as the build phase starts it crash asking for a missing module that its name is unknown.

### Environment

```
 expo-env-info 1.0.5 environment info:
    System:
      OS: Linux 5.19 Pop!_OS 22.04 LTS
      Shell: 5.9 - /home/linuxbrew/.linuxbrew/bin/zsh
    Binaries:
      Node: 16.16.0 - ~/.volta/tools/image/node/16.16.0/bin/node
      Yarn: 3.2.2 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
      npm: 8.12.1 - ~/.volta/tools/image/npm/8.12.1/bin/npm
      Watchman: 2022.07.04.00 - /home/linuxbrew/.linuxbrew/bin/watchman
    SDKs:
      Android SDK:
        API Levels: 26, 30, 31, 32, 33
        Build Tools: 26.0.0, 29.0.2, 30.0.2, 30.0.3, 33.0.0, 33.0.0
        System Images: android-30 | Google Play Intel x86 Atom_64
    npmPackages:
      expo: ~46.0.3 => 46.0.3 
      react: 18.0.0 => 18.0.0 
      react-native: 0.69.4 => 0.69.4 
    npmGlobalPackages:
      expo-cli: 5.5.1
    Expo Workflow: managed
```

### Please specify your device/emulator/simulator platform, model and version

ExpoGO

### Error output

```
Logs for your project will appear below. Press Ctrl+C to exit.
Started Metro Bundler
Error: Unable to resolve module ./index from /home/daniloverde/conduit/Conduit/.: 

None of these files exist:
  * index(.native|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json)
  * index/index(.native|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json)
    at ModuleResolver.resolveDependency (/home/daniloverde/conduit/Conduit/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:136:15)
[...]

PR #4582, opened 2022-10-28
feat(next)!: rewrite package

EvanBacon:
# Why

Too much bloat involved with the default setup script. Users should either bootstrap from a template (Next.js and Expo both have templates), or they should manually copy the couple files into their projects. 

Since Next.js changes so often, this will help keep projects a bit more dynamic and capable of upgrading.

- replaces https://github.com/expo/expo-cli/pull/3691 https://github.com/expo/expo-cli/pull/3036
- fixes https://github.com/expo/expo-cli/issues/4584
- fixes https://github.com/expo/expo-cli/issues/4553 
- fixes https://github.com/expo/expo-cli/issues/4548 
- fixes https://github.com/expo/expo-cli/issues/4523 
- fixes https://github.com/expo/expo-cli/issues/4509
- Need to update the docs and articles to reflect the new system.
- Made `npx next-expo` log a deprecation notice.

# Test Plan

- Made a Next.js 13 project and followed setup instructions


byCedric:
Do we also need to include the `react-native-web` babel plugin? It seems to do:
- a rewrite from `react-native` to `react-native-web`
  → _We don't need this since Webpack / Turbopack? already takes care of it._
- dropping `react-native-web` modules that aren't used in the bundle output
  → This seems like a nice addition. Maybe just documenting adding this is already enough, but would be nice if we can include it easier somehow.

nandorojo:
> Do we also need to include the [react-native-web babel plugin](https://github.com/necolas/react-native-web/tree/master/packages/babel-plugin-react-native-web)? It seems to do:

Probably not, since Next.js is moving away from Babel towards SWC / ESM.

PR #4587, opened 2022-11-02
[xdl] Add feature gates

wschurman:
# Why

This is the legacy CLI equivalent of https://github.com/expo/eas-cli/pull/1475. This uses the new www API to fetch the feature gates since there isn't a graphql client available in this CLI.

This will be used to roll out features to encourage users to move to EAS update before the deadline.

# How

Add similar code, update to be in the style of this repo and with the `internal` module stuff.

# Test Plan

Run test.

### DIFF
--- a/packages/xdl/src/Env.ts
+++ b/packages/xdl/src/Env.ts
@@ -15,6 +15,13 @@ export function isLocal(): boolean {
   return getenv.boolish('EXPO_LOCAL', false);
 }
 
+export function getFeatureGateOverrides(): { enable: string[]; disable: string[] } {
+  return {
+    enable: getenv.array('EXPO_FG_ENABLE'),
+    disable: getenv.array('EXPO_FG_DISABLE'),
+  };
+}
+
 // TODO: remove this function once all related PRs have landed and there's no chance for conflict
 export function isInterstitiaLPageEnabled(): boolean {
   return true;
--- a/packages/xdl/src/gating/FeatureGating.ts
+++ b/packages/xdl/src/gating/FeatureGating.ts
@@ -1,0 +1,72 @@
+import {
+  featureGateDefaultValueWhenNoServerValue,
+  FeatureGateEnvOverrides,
+  FeatureGateKey,
+  FeatureGateTestOverrides,
+} from '../internal';
+
+export default class FeatureGating {
+  constructor(
+    private readonly serverValues: { [key: string]: boolean },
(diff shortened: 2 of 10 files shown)


packages/xdl/src/Env.ts
```diff
@@ -15,6 +15,13 @@ export function isLocal(): boolean {
   return getenv.boolish('EXPO_LOCAL', false);
 }
 
+export function getFeatureGateOverrides(): { enable: string[]; disable: string[] } {
+  return {
+    enable: getenv.array('EXPO_FG_ENABLE'),
```
EvanBacon, line 20: Traditionally our naming format for env variables in the CLI has been more like `EXPO_FG` and `EXPO_NO_FG` [ref](https://docs.expo.dev/workflow/expo-cli/#environment-variables)

packages/xdl/src/gating/FeatureGating.ts
```diff
@@ -0,0 +1,72 @@
...
+} from '../internal';
+
+export default class FeatureGating {
+  constructor(
+    private readonly serverValues: { [key: string]: boolean },
+    private readonly envOverrides: FeatureGateEnvOverrides
+  ) {}
+
+  public isEnabled(experimentKey: FeatureGateKey): boolean {
+    if (process.env.NODE_ENV === 'test' && FeatureGateTestOverrides.isOverridden(experimentKey)) {
```
EvanBacon, line 15: This seems like an odd thing to ship to production. Couldn't the same be achieved using a jest mock on this function?

packages/xdl/src/Env.ts
wschurman: Yep, that makes sense for booleans, but this is to match the EAS cli `EAS_FG_ENABLE=test1,test2,test3` style which seemed best for string arrays.

packages/xdl/src/gating/FeatureGating.ts
wschurman: In the CLIs yep, I think it could. I think I kept it this way so that all of our feature gating infra was the same across all our codebases (this is the way the test mocking works in www, website, eas-cli).

wschurman:
ping

As far as the inline comments go, I think that consistency across all our codebases is fairly important for this.

wschurman:
ping 2

EvanBacon (approved):

PR #3763, opened 2021-08-12
[webpack] Upgrade to Webpack 5

EvanBacon:
# Why

CRA (the base config) has just merged support for Webpack v5 (https://github.com/facebook/create-react-app/pull/11201), meaning we can start upgrading.

- Resolve ENG-3881
- Fix https://github.com/expo/expo-cli/issues/2881
- Fix https://github.com/expo/expo-cli/issues/4575

# How

- Upgraded packages in expo/webpack-config and xdl to webpack v5.
- Upgraded plugins to use new syntax.
- Upgraded config and dev server schema.
- Removed unused features.

## Prerequisites

I added a few features that should be split out and merged separately:

- [x] Unified logging with Metro https://github.com/expo/expo-cli/pull/3766
- [x] Update web beta warning https://github.com/expo/expo-cli/pull/3766
- [x] Determine if we want to roll out Webpack 5 behind a feature flag / ship two versions of webpack to users. Not doing this, instead, `npx expo` enforces that the Webpack config must be installed locally, meaning users will opt-in manually.

# Test Plan

- `expo start -w`
- `expo start:web`
  - [x] ~~Error overlay should show and work (currently doesn't, seems related to CRA) -- Pending https://github.com/facebook/create-react-app/pull/11799~~ Never merged.
  - [x] ~~Babel caching should work~~
  - [x] Font loading and expo/vector-icons should work
  - [x] ~~Video and image loading / importing like metro (interop)~~ Moved to [Metro for web](https://docs.expo.dev/guides/customizing-metro/#web-support).
  - [ ] NCL should bundle
  - [x] Debug logging should work
  - [x] Auto resize observer adding
  - [x] HTML loading
  - [ ] Eslint
  - [ ] TypeScript
  - [ ] AVIF support
  - [x] Packages with CSS
- `expo build:web`
  - [x] EXPO_WEB_DEBUG web debugging optimizations
  - [x] WEB_HOST, WEB_PORT
  - [ ] No icons should warn
  - [x] PWA Asset generation
  - [x] PWA overrides

PR #4593, opened 2022-11-03
update babel version to allow for new features

goatandsheep:
# Why

My library is having some issues while running in expo environment as expo-cli has locked in the version of Babel in a webpack environment https://github.com/goatandsheep/react-native-dotenv/issues/75

I am working directly with the Babel team on this one.

# How

Updated babel and babel-loader to minimum versions needed to enable the new `addExternalDependencies` feature in Babel

# Test Plan

Run an expo sample app with `@expo/webpack-config` and watch it recompile every time you make Javascript changes


packages/webpack-config/package.json
```diff
@@ -40,8 +40,8 @@
     "url": "https://github.com/expo/expo-cli/issues"
   },
   "dependencies": {
-    "@babel/core": "7.9.0",
-    "babel-loader": "8.1.0",
+    "@babel/core": "^7.12.0",
+    "babel-loader": "^8.1.0",
```
nicolo-ribaudo, line 44: If preferred, the latest versions are `^7.20.2` and `^8.3.0`.

goatandsheep:
@EvanBacon I noticed the issue template has been updated to say

> Open Expo CLI changes here -> https://github.com/expo/expo/tree/main/packages/@expo/cli

This PR is for `webpack-config` which I can't find in the above repo. I recommend if this repo is deprecated to add that flag to the README so people know. Also, npm packages still point to this repo so I advise you update the repo field in the libraries before the next pushes.

EvanBacon (changes requested):
Run `yarn` to update the lockfile so the tests run

goatandsheep:
Hi @EvanBacon thank you for reviewing this PR. I have run every build I thought may update the `yarn.lock` and they all fail:

* `~/yarn`
* `~/npx lerna run prepare --stream`
* `~/yarn postinstall`
* `~/packages/webpack-config/yarn`
* `~/packages/webpack-config/npm i`
* `~/packages/webpack-config/npx tsc`

The only thing that changed the lockfile was deleting and recreating. If this is not good, I can revert the latest commit

trajano:
babel-loader is at 9.1.2 right now would that be a better version to use or would it fail react-native-dotenv

goatandsheep:
Anything after 8.1 should work but it's not really fail it's just that the new features make the library caching work much better

trajano:
Juts tried with babel-loader 9.1.2 it react-native-dotenv actually does not work on the web with it for me.  But that's a bug for another repo. @goatandsheep

goatandsheep:
@EvanBacon thank you for your review. Contrary to the git warnings, the requested changes have been made. Please re-review

goatandsheep:
@EvanBacon just updated from upstream. Thank you for keeping this repo up to date.

goatandsheep:
Okay I'm not sure why tests are passing after PR #3763 but I realized the resolved babel versions in yarn are correct and I can't figure out how to do this properly so I give up. If a contributor wants to fix this that would be great, but otherwise I'm no longer going to be following this.

elitenas:
This pr is much needed but it slows down the react-native-web refresh times.

https://github.com/babel/babel-loader/compare/v8.2.5...v8.3.0

Works fine in 8.2.5 but 8.3.0... it is considerably slower

goatandsheep:
@elitenas once this is merged there won't be a need for API.cache(false) so the slowdowns won't occur

elitenas:
@goatandsheep we need that line there because otherwise it says:

```
The project's Babel config is invalid: Caching was left unconfigured. Babel's plugins, presets, and .babelrc.js files can be configured

for various types of caching, using the first param of their handler functions:

module.exports = function(api) {
  // The API exposes the following:

  // Cache the returned value forever and don't call this function again.
  api.cache(true);

  // Don't cache at all. Not recommended because it will be very slow.
  api.cache(false);

  // Cached based on the value of some function. If this function returns a value different from
  // a previously-encountered value, the plugins will re-evaluate.
  var env = api.cache(() => process.env.NODE_ENV);

  // If testing for a specific env, we recommend specifics to avoid instantiating a plugin for
  // any possible NODE_ENV value that might come up during plugin execution.
  var isProd = api.cache(() => process.env.NODE_ENV === "production");

  // .cache(fn) will perform a linear search though instances to find the matching plugin based
  // based on previous instantiated plugins. If you want to recreate the plugin and discard the
  // previous instance whenever something changes, you may use:
  var isProd = api.cache.invalidate(() => process.env.NODE_ENV === "production");

  // Note, we also expose the following more-verbose versions of the above examples:
  api.cache.forever(); // api.cache(true)
  api.cache.never();   // api.cache(false)
  api.cache.using(fn); // api.cache(fn)

  // Return the value that will be cached.
  return { };
};
Error: Caching was left unconfigured. Babel's plugins, presets, and .babelrc.js files can be configured
for various types of caching, using the first param of their handler functions:

module.exports = function(api) {
  // The API exposes the following:

  // Cache the returned value forever and don't call this function again.
  api.cache(true);

  // Don't cache at all. Not recommended because it will be very slow.
  api.cache(false);

  // Cached based on the value of some function. If this function returns a value different from
  // a previously-encountered value, the plugins will re-evaluate.
  var env = api.cache(() => process.env.NODE_ENV);

  // If testing for a specific env, we recommend specifics to avoid instantiating a plugin for
  // any possible NODE_ENV value that might come up during plugin execution.
  var isProd = api.cache(() => process.env.NODE_ENV === "production");

  // .cache(fn) will perform a linear search though instances to find the matching plugin based
  // based on previous instantiated plugins. If you want to recreate the plugin and discard the
  // previous instance whenever something changes, you may use:
  var isProd = api.cache.invalidate(() => process.env.NODE_ENV === "production");

  // Note, we also expose the following more-verbose versions of the above examples:
  api.cache.forever(); // api.cache(true)
  api.cache.never();   // api.cache(false)
  api.cache.using(fn); // api.cache(fn)

  // Return the value that will be cached.
  return { };
};

```

elitenas:
`api.cache(true); ` also results in slow refresh times (~45 secs)

elitenas:
This seems to be resolved by upgrading to expo 48 and babel-loader: 9.x.x

I do believe this change needs to be merged in to fixing babel caching issues @EvanBacon

elenitaex5:
Do we know something about this issue?? Expo cli is not working opening web at all. Error comes from webpack

elenitaex5:
@goatandsheep there's a way to "fixing" this until this PR is accepted? or a forked project? My web is not working. I don't know when your PR is gonna be accepted.

elenitaex5:
@EvanBacon there's a huge issue using Expo webpack with Dotenv. Please review @goatandsheep PR.

goatandsheep:
I'm happy to add anyone as a contributor to the fork

EvanBacon (approved):

elenitaex5:
What about this? @EvanBacon I know you include it into main, but there's a new version of expo-cli and webpack which include this new babel dependencies update?

goatandsheep:
Yes the new version should include this. I will retest and update the react native dotenv Lib and docs to make sure it's working sometime this week

elenitaex5:
``

> Yes the new version should include this. I will retest and update the react native dotenv Lib and docs to make sure it's working sometime this week

The problem is **expo version** is still working with an old @expo/webpack-config. If I try to install newest version an error is shown:

`  
Some dependencies are incompatible with the installed expo version:
@expo/webpack-config@18.0.4 - expected version: ^0.17.2
`

PR #4596, opened 2022-11-04
fix(cli): 'expo not installed' error message

samuelastech:
For beginners like me, this error message would save an extra google search.


EvanBacon (changes requested):
`expo-cli` is deprecated and moved to [`@expo/cli`](https://github.com/expo/expo/tree/main/packages/%40expo/cli). The CLI is now installed with `expo` making this error impossible to obtain.

PR #4625, opened 2023-01-17
[doctor] Add warning when explicitly using `androidStatusBar.translucent`

gabrieldonadel:
# Why

When explicitly adding `androidStatusBar.translucent` `true` inside `app.json` EAS builds add `<item name="android:windowTranslucentStatus">true</item>` to `android/app/src/main/res/values/styles.xml` which when used with `softwareKeyboardLayoutMode` `resize` don't combine to the expected keyboard behavior.

 
<details>
   <summary>Example</summary>
<table>
    <tr><th>With explicit translucent `true`</th><th>Without translucent</th></tr>
    <tr>
        <td><video src="https://user-images.githubusercontent.com/11707729/213022005-68cacde6-6c16-44b4-b674-4e14058a978b.mp4"></td>
        <td><video src="https://user-images.githubusercontent.com/11707729/213021976-73254bd9-0dd9-40b3-98c0-d8e397c8643c.mp4"></td>
    </tr> 
</table>    
</details>

 
Unfortunately, this is a platform quirk and there isn't much we can do about it, but given that we let people set both of these values directly in `app.json` we can detect if they are set to incompatible values and warn people about the need of using a KeyboardAvoidingView in such cases

Closes ENG-4613
 
# How

Check ExpoConfig values to detect if they are set to incompatible values and if so show a warning

# Test Plan

Running `expod doctor` in a project with `"softwareKeyboardLayoutMode": "resize"` and `"androidStatusBar"."translucent": true`

image

Running `expod doctor` in a project without `"androidStatusBar"."translucent": true` or  `"softwareKeyboardLayoutMode": "resize"

image

### DIFF
--- a/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
+++ b/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
@@ -37,8 +37,9 @@ async function validateSupportPackagesAsync(sdkVersion: string): Promise<boolean
 
 function validateAndroidExpoConfig(exp: ExpoConfig) {
   if (
-    exp.android?.softwareKeyboardLayoutMode === 'resize' &&
-    exp.androidStatusBar?.translucent === true
+    exp.androidStatusBar?.translucent === true &&
+    (!exp.android?.softwareKeyboardLayoutMode ||
+      exp.android?.softwareKeyboardLayoutMode === 'resize')
   ) {
     Log.warn(
       'Explicitly setting `androidStatusBar.translucent` to `true` in `app.json` with `softwareKeyboardLayoutMode` `resize` may result in unexpected keyboard behavior on Android and you will have to use a KeyboardAvoidingView to manage the keyboard layout.'
--- a/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
+++ b/packages/expo-cli/src/commands/info/doctor/doctorAsync.ts
@@ -1,4 +1,4 @@
-import { getConfig } from '@expo/config';
+import { ExpoConfig, getConfig } from '@expo/config';
 import chalk from 'chalk';
 import { Doctor, Versions } from 'xdl';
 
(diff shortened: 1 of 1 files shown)


keith-kurak:
Thanks for opening this!

Does this bug only happen if the developer explicitly sets `softwareKeyboardLayoutMode = "resize"` AND `androidStatusBar.translucent = true` or can it also happen if they explicitly set `androidStatusBar.translucent = true` while leaving the default `softwareKeyboardLayoutMode` in place? I imagine it would, since `android:windowSoftInputMode="adjustResize"` is set by default. It'd be ideal if we could check that value against either whatever is explicitly set or the default, though the default is very unlikely to change before migrating to a new standalone doctor, so assuming "not set" == true without checking against the actual defaults could be fine.

Would you be able to open a corresponding PR for the docs (app.json schema) to document this? I think that should land first, so devs have a place to go to understand this better, and Aman could help us fine-tune how we describe this, since there's a good bit of nuance here. Depending on how much detail is needed there, we could consider using creating expo.fyi link to connect that discussion to the CLI warning, e.g., "These things don't work together, check out expo.fyi/android-transluent-status-bar-quirks.

gabrieldonadel:
> Does this bug only happen if the developer explicitly sets `softwareKeyboardLayoutMode = "resize"` AND `androidStatusBar.translucent = true` or can it also happen if they explicitly set `androidStatusBar.translucent = true` while leaving the default `softwareKeyboardLayoutMode` in place? I imagine it would, since `android:windowSoftInputMode="adjustResize"` is set by default. It'd be ideal if we could check that value against either whatever is explicitly set or the default, though the default is very unlikely to change before migrating to a new standalone doctor, so assuming "not set" == true without checking against the actual defaults could be fine.

 It can also happen if they explicitly set `androidStatusBar.translucent = true` while leaving the default `softwareKeyboardLayoutMode` in place. I've updated it so it also shows the warning when using the default config.

> Would you be able to open a corresponding PR for the docs (app.json schema) to document this? I think that should land first, so devs have a place to go to understand this better, and Aman could help us fine-tune how we describe this, since there's a good bit of nuance here.

Sure, I'm working on that and will link a PR when that's ready

gabrieldonadel:
@keith-kurak here is the xdl-schema PR https://github.com/expo/universe/pull/11245 necessary to update `@expo/config-types` and the app-config documentation

gabrieldonadel:
Closing this for now as discussed with Keith on Slack. Current Expo Doctor is being deprecated and I'll reopen this PR when we move this into a new standalone version

PR #4641, opened 2023-02-01
Standalone Expo Doctor

keith-kurak:
# Why
Separate Expo doctor from the global CLI into its own package.

Goals for v1:
- More-or-less feature parity with old Doctor
- No dependencies on XDL
- Lean on versioned CLI for SDK-specific checks (e.g., use `expo install` to check dependencies instead of copying that code)

More info on Doctor rewrite/ future ideas [being sketched out here](https://www.notion.so/expo/Expo-Doctor-Roadmap-4cff1da736cd44e884fabe407f6b451f).

# How
- Used `expo-env-info` as a template
- Copied relevant code from expo-cli, replacing callouts to XDL (usually for API access) with direct API calls. For now, it seemed reasonable to favor having the latest data over potential offline support.
- Replaced dependency checks with a call to `npx expo install --check`. In the future, we could enhance the versioned CLI dependency check to have more friendly output (e.g., non-interactive/ JSON support).
- Restructured in a test runner, where new tests can be created by creating a new class and adding it to a list, with a runner standardized output and catching unexpected errors.

## Schemer asset validation issue
I made a small change to schemer to fix an issue where it would not validate assets in a sub-property in app.json. The schema has definitions like `Splash.image`, and schemer tries to get the deep property with the capital "Splash" and fails. This isn't an issue in the original doctor command, I think because maybe it massages the output from the schema API to add lowercase versions of those definitions. 

However, for non-asset properties, schemer validates correctly without any massaging, possibly because it uses `ajv` for that and custom validation for assets. Supposing that asset traversal should mirror what `ajv` is doing, I tweaked it to look for the lowercase prop name. I suppose I could go all the way and make that case-insensitive.

# Test Plan
Test in-place with `npx expo-doctor [projectRoot]`

- [ ] Test global dependency checks
- [x] Test deep dependency checks for support packages / illegal packages
- [x] Test schema checks
- [x] Test dependency version check

Some early ideas for unit tests are started in there. I think the best place to focus on testing would be with the individual checks, but would appreciate feedback there.

Trying some ideas on organizing the output, separating errors from suggestions for fixing (*ahem* diagnoses and prescriptions). Feedback also much appreciated!
image

# Refactoring context
The PR can make this difficult to see, but I copied a lot of code from the doctor command in the global CLI, minimally/ safely refactoring it to remove unwanted dependencies and fit the desired output. I agree that their logic could be significantly refactored in the future, but I think that would be better to consider in separate PR's to avoid introducing bugs into checks that were already production-battle-tested. 

If there's a preferred way to highlight/ isolate this code via comments/ organization, all ears there!
## Windows CMD.exe warning
- copied from **expo-cli/src/commands/info/doctor/window**, zero changes.
## SupportPackageVersionCheck/ IllegalPackageCheck
- copied files:
  - **expo-cli/src/commands/utils/getRemoteVersionsForSdk** -> **expo-doctor/utils/getRemoteVersionsForSdkAsync**
    - switched XDL request for API utility function
    - adapted existing unit tests to changes
  - **expo-cli/src/commands/info/doctor/depedencies/explain** -> **expo-doctor/utils/explainDependencies**
    - removed inline logging, returned errors instead so they could be formatted by new code
    - added `projectRoot` param to facilitate testing
  - `XDL.Versions.gteSdkVersion` -> **expo-doctor/utils/gteSdkVersion**
## ExpoConfigSchemaCheck
- separated from `XDL.Doctor.validateWithNetworkAsync`
- copied schema validation from **xdl/src/project/Doctor** into **expo-doctor/utils/schema**, feeding schema in from API utility function
## GlobalPrereqsVersionCheck
- copied NPM, etc. checks directly from **xdl/src/project/Doctor**, changed logging statements to return statements so they could be formatted by new code
## InstalledDependencyVersionCheck
- all new code


packages/expo-doctor/src/__mocks__/@expo/spawn-async.ts
```diff
@@ -0,0 +1,12 @@
+import { SpawnResult } from '@expo/spawn-async';
```
EvanBacon, line 1: we usually put mocks in the root https://github.com/expo/expo/tree/main/packages/%40expo/cli/__mocks__

packages/expo-doctor/src/utils/log.ts
```diff
@@ -0,0 +1,257 @@
+import chalk from 'chalk';
```
EvanBacon, line 1: The new [expo cli log file](https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/log.ts) is better if you're copy/pasting. You may want to add [something like this too](https://github.com/expo/expo/blob/fd7776f7c9b00648bc3e75580a925d32bf71b1fe/packages/%40expo/cli/bin/cli.ts#L7-L12), to support EXPO_DEBUG and DEBUG.

packages/schemer/src/index.ts
```diff
@@ -3,6 +3,7 @@ import addFormats from 'ajv-formats';
 import fs from 'fs';
 import traverse from 'json-schema-traverse';
 import get from 'lodash/get';
+import lowerFirst from 'lodash/lowerFirst';
```
EvanBacon, line 6: the unwritten rule is to never introduce any extra lodash usage. Maybe this works:

```ts
function lowerFirst(str: string): string {
    return str.charAt(0).toLowerCase() + str.slice(1);
}
```

packages/schemer/src/index.ts
```diff
@@ -165,7 +166,7 @@ export default class Schemer {
         const fieldPath = schemaPointerToFieldPath(jsonPointer);
         assets.push({
           fieldPath,
-          data: get(data, fieldPath),
+          data: get(data, lowerFirst(fieldPath)),
```
EvanBacon, line 169: Is this always true?

yarn.lock
```diff
@@ -1578,6 +1578,27 @@
     xcode "^3.0.1"
     xml2js "0.4.23"
 
+"@expo/config-plugins@~5.0.3":
```
EvanBacon, line 1581: Avoid duplicating deps

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+import { RootNodePackage, VersionSpec } from './explainDependencies.types';
+
+type TargetPackage = { name: string; version?: VersionSpec };
+
+function isSpawnResult(result: any): result is SpawnResult {
+  return 'stderr' in result && 'stdout' in result && 'status' in result;
+}
+
+/** Spawn `npm explain [name] --json` and return the parsed JSON. Returns `null` if the requested package is not installed. */
+async function explainAsync(
```
EvanBacon, line 14: Might be worth adding some tests for this which mock spawn and return some static results.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+export async function actionAsync(projectRoot: string) {
+  await warnUponCmdExe();
+
+  const { exp, pkg } = getConfig(projectRoot);
+
+  // expo-doctor relies on versioned CLI, which is only available for 44+
+  if (!gteSdkVersion(exp, '44.0.0')) {
+    console.log(
+      chalk.red(
+        `expo-doctor is only supported for SDK 44 and higher. Please use 'expo-cli doctor' for older SDKs.`
```
EvanBacon, line 67:
```suggestion
        `expo-doctor supports Expo SDK +44. Use 'expo-cli doctor' for SDK 43 and lower.`
```

EvanBacon: Avoid saying "please"

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+  // add additional checks here
+  const checks = [
+    new GlobalPrereqsVersionCheck(),
+    new IllegalPackageCheck(),
+    new SupportPackageVersionCheck(),
+    new InstalledDependencyVersionCheck(),
+    new ExpoConfigSchemaCheck(),
+  ];
+
+  let foundSomeIssues = false;
```
EvanBacon, line 83:
```suggestion
  let hasIssues = false;
```

EvanBacon: or flip it and call it `isValid` or something.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  if (foundSomeIssues) {
+    console.log(
+      chalk.red(
+        `\n✖ Found one or more possible issues with the project. See above logs for issues and advice to resolve.`
+      )
+    );
+    process.exitCode = 1;
+  } else {
+    console.log(chalk.green(`\n🎉 Didn't find any issues with the project!`));
```
EvanBacon, line 101: Avoid emojis, there's always some obscure terminal that doesn't support them.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+  for (const check of checks) {
+    if (!(await runCheckAsync(check, checkParams))) {
+      foundSomeIssues = true;
+    }
+  }
+
+  if (foundSomeIssues) {
+    console.log(
+      chalk.red(
+        `\n✖ Found one or more possible issues with the project. See above logs for issues and advice to resolve.`
```
EvanBacon, line 96: I'd prefer `console.log()` above the if so you can drop the `\n` prefix for all the rest of the messages.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+  const checkParams = { exp, pkg, projectRoot };
+
+  for (const check of checks) {
+    if (!(await runCheckAsync(check, checkParams))) {
+      foundSomeIssues = true;
+    }
+  }
+
+  if (foundSomeIssues) {
+    console.log(
```
EvanBacon, line 94: If you add a custom log file then it should be used, this helps with assumptions about mocking for tests.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  } catch (e: any) {
+    if (e.code === 'ENOTFOUND') {
+      console.log(
+        chalk.red(
+          `Error: this check requires a connection to the Expo API. Please check your network connection.`
+        )
+      );
+    } else {
+      console.log(chalk.red(`Error: check could not be completed due to an unexpected error.`));
```
EvanBacon, line 52: `console.error` and show the error message

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+      }
+      if (result.advice.length) {
+        console.log(chalk.underline(chalk.green(`Advice:`)));
+        console.group();
+        for (const advice of result.advice) {
+          console.log(chalk.green(`• ${advice}`));
+        }
+        console.groupEnd();
+      }
+      return false;
```
EvanBacon, line 39: move this out of the try/catch

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+import { warnUponCmdExe } from './warnings/windows';
+
+export async function runCheckAsync(
+  check: DoctorCheck,
+  checkParams: DoctorCheckParams
+): Promise<boolean> {
+  try {
+    const ora = logNewSection(check.description);
+    const result = await check.runAsync(checkParams);
+    if (!result.isSuccessful) {
```
EvanBacon, line 21: better to inverse this so the majority of code is one indentation lower, e.g. `if (result.isSuccessful) { return true; }`

packages/expo-doctor/src/checks/checks.types.ts
```diff
@@ -0,0 +1,19 @@
+export interface DoctorCheck {
+  // description that will appear as each check is run
+  description: string;
+  runAsync: (params: DoctorCheckParams) => Promise<DoctorCheckResult>;
+}
+
+export interface DoctorCheckResult {
+  isSuccessful: boolean;
+  // many checks currently output their own issues, no need to duplicate
```
EvanBacon, line 9:
```suggestion
  /** many checks currently output their own issues, no need to duplicate */
```
and related for doc blocks, makes it easier to read around the code.

packages/expo-doctor/src/checks/__tests__/InstalledDependencyVersionCheck.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { InstalledDependencyVersionCheck } from '../InstalledDependencyVersionCheck';
+
+const asMock = <T extends (...args: any[]) => any>(fn: T): jest.MockedFunction<T> =>
+  fn as jest.MockedFunction<T>;
+
+describe(InstalledDependencyVersionCheck, () => {
+  let check;
+  beforeEach(() => {
+    check = new InstalledDependencyVersionCheck();
+  });
```
EvanBacon, line 12: Move into the describe to reduce the possible scope of access. Could also just inline `const check = ...` in each test to reduce any chance of the tests going awry.

packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts
```diff
@@ -0,0 +1,41 @@
...
+import { DoctorCheck, DoctorCheckParams, DoctorCheckResult } from './checks.types';
+
+export class SupportPackageVersionCheck implements DoctorCheck {
+  description = 'Verifying prebuild support package versions are compatible';
+
+  async runAsync({ exp, projectRoot }: DoctorCheckParams): Promise<DoctorCheckResult> {
+    const advice: string[] = [];
+    const issues: string[] = [];
+
+    if (gteSdkVersion(exp, '45.0.0')) {
```
EvanBacon, line 13: Invert this, it's much easier to read "if less than 45 then return ..." makes it very clear at a glance that this test is +45. Might even make sense to add this type of filter to the `DoctorCheck` class so there's a unified system for bailing on SDK versions.

packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      // only way to check dependencies without automatically fixing them is to use interactive prompt
+      // In the future, we should add JSON output to npx expo install, check for support, and use that instead
+      await spawnAsync('sh', ['-c', 'echo "n" | npx expo install --check'], {
+        stdio: 'pipe',
+        cwd: projectRoot,
+      });
+    } catch (error: any) {
+      if (isSpawnResult(error)) {
+        issues.push(error.stderr.trim());
```
EvanBacon, line 25: Add a debug-only log which prints the entire error

packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
```diff
@@ -0,0 +1,38 @@
...
+        advice.push(`Use npx expo install --check to review and upgrade your dependencies.`);
+      } else {
+        throw error;
```
EvanBacon, line 28: What's the difference between `issues` and assertions from within the function?

packages/expo-doctor/src/checks/IllegalPackageCheck.ts
```diff
@@ -0,0 +1,37 @@
...
+
+      for (const pkg of illegalPackages) {
+        const warning = await warnAboutDeepDependenciesAsync({ name: pkg }, projectRoot);
+        if (warning) {
+          issues.push(warning);
+        }
+      }
+
+      if (issues.length) {
+        advice.push(`Remove any 'unimodules' packages from your project.`);
```
EvanBacon, line 27: Maybe these should link to an fyi page or related post with more info.

packages/expo-doctor/src/checks/IllegalPackageCheck.ts
```diff
@@ -0,0 +1,37 @@
...
+      }
+    }
+
+    return {
+      isSuccessful: issues.length === 0,
```
EvanBacon, line 32:
```suggestion
      isSuccessful: !issues.length,
```
xit, I don't care

packages/expo-doctor/src/checks/GlobalPrereqsVersionCheck.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const watchmanCheck = await _checkWatchmanVersionAsync();
+    if (watchmanCheck) {
+      issues.push(watchmanCheck);
+    }
+
+    const npmCheck = await _checkNpmVersionAsync();
+    if (npmCheck) {
+      issues.push(npmCheck);
+    }
```
EvanBacon, line 102: Run these in parallel with Promise.all

packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
keith-kurak: You mean why could it possibly throw even if normally issues are logged? The issues would be like "we successfully completed the check, and the check found issues with your project", and throwing would be like, "something unexpectedly broke while attempting the check, let's bail". So you don't have to think about every possible error while writing a check, just the issues that your check is looking for. 

Generally this would mean writing a check wouldn't need any try/ catches, but this one breaks that pattern a bit, since `spawnAsync` throws on exit code !== 0 (e.g., the issues we're looking for), but in theory could also throw for another reason. Could definitely at least clarify in a comment.

packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts
keith-kurak: really like the idea of integrating the filter.

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+export async function actionAsync(projectRoot: string) {
+  await warnUponCmdExe();
+
+  const { exp, pkg } = getConfig(projectRoot);
+
+  // expo-doctor relies on versioned CLI, which is only available for 44+
+  if (!gteSdkVersion(exp, '44.0.0')) {
+    console.log(
+      chalk.red(
+        `expo-doctor supports Expo SDK +44. Use 'expo-cli doctor' for SDK 43 and lower.`
```
kbrandwijk, line 67:
```suggestion
        `expo-doctor supports Expo SDK 44+. Use 'expo-cli doctor' for SDK 43 and lower.`
```

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+  try {
+    const { stdout } = await spawnAsync('npm', args, {
+      stdio: 'pipe',
+      cwd: projectRoot,
+    });
+
+    return JSON.parse(stdout);
+  } catch (error: any) {
+    if (isSpawnResult(error)) {
+      if (error.stderr.match(/npm ERR! No dependencies found matching/)) {
```
kbrandwijk, line 30: If you use `--json`, why not use the actual json output, instead of parsing human readable output.
```
❯ npm explain bla --json
npm ERR! No dependencies found matching bla
{
  "error": {
    "code": null,
    "summary": "No dependencies found matching bla",
    "detail": ""
  }
}
```

packages/expo-doctor/src/utils/schema.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  if (validateAssets) {
+    try {
+      await validator.validateAssetsAsync(exp);
+    } catch (e: any) {
+      if (e instanceof SchemerError) {
+        assetsErrorMessage = `Error: Problem${
+          e.errors.length > 1 ? '' : 's'
+        } validating asset fields in ${configName}. ${learnMore('https://docs.expo.dev/')}`;
```
kbrandwijk, line 48: Is there a more specific docs link for this?

packages/expo-doctor/src/utils/schema.ts
```diff
@@ -0,0 +1,54 @@
...
+        assetsErrorMessage += e.errors.map(formatValidationError).join('');
+      }
```
kbrandwijk, line 50: This code block is almost identical to the other validation. I would extract it into its own method with topic (field, asset fields) and learnMore parameters.

packages/expo-doctor/src/warnings/windows.ts
```diff
@@ -0,0 +1,43 @@
...
+      // console.log(stderr);
+      return;
+    }
+
+    const [parentProcessName] = stdout.match(/(?<=^").*?(?=",)/) || [''];
+    if (parentProcessName.toLowerCase().includes('cmd.exe')) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        yellow(
+          'WARNING: We recommend using PowerShell or Bash via WSL 2 for development with expo-cli on Windows. You may encounter issues using cmd.exe.\n'
```
kbrandwijk, line 38: `expo-cli` is the old global one.
```suggestion
          'WARNING: We recommend using PowerShell or Bash via WSL 2 for development with Expo CLI on Windows. You may encounter issues using cmd.exe.\n'
```

packages/expo-doctor/src/warnings/windows.ts
```diff
@@ -0,0 +1,43 @@
...
+        )
+      );
+    }
+  })();
```
kbrandwijk, line 42: That's a lot of wrapping. Extract into method and await that. This is too much for inlining in an await() call.

packages/expo-doctor/src/warnings/windows.ts
```diff
@@ -0,0 +1,43 @@
...
+      // console.log(stdout);
+      // console.log(stderr);
+      return;
+    }
+
+    const [parentProcessName] = stdout.match(/(?<=^").*?(?=",)/) || [''];
+    if (parentProcessName.toLowerCase().includes('cmd.exe')) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        yellow(
```
kbrandwijk, line 37: Why not log.warn?

kbrandwijk, line 37: Why not log.warn?

packages/expo-doctor/src/utils/getRemoteVersionsForSdkAsync.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (sdkVersion && sdkVersion in sdkVersions) {
+    const { relatedPackages, facebookReactVersion, facebookReactNativeVersion } = sdkVersions[
+      sdkVersion
+    ];
+    const reactVersion = facebookReactVersion
+      ? {
+          react: facebookReactVersion,
+          'react-dom': facebookReactVersion,
+        }
+      : undefined;
```
kbrandwijk, line 18: This field exists since SDK 14, I think it's safe to simplify this like facebookReactNativeVersion below.

packages/expo-doctor/src/utils/getRemoteVersionsForSdkAsync.ts
```diff
@@ -0,0 +1,26 @@
+import { getVersionsAsync } from '../api/getVersionsAsync';
+
+export type DependencyList = Record<string, string>;
+
+export const getRemoteVersionsForSdkAsync = async (
+  sdkVersion?: string
+): Promise<DependencyList> => {
+  const { sdkVersions } = await getVersionsAsync({ skipCache: true });
+  if (sdkVersion && sdkVersion in sdkVersions) {
```
kbrandwijk, line 9: Why call the endpoint first, and _then_ check if sdkVersion is even passed in. Check `if (sdkVersion)` before calling the endpoint.

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+  return { valid, invalid };
+}
+
+/**
+ * @param pkg
+ * @param explanations
+ * @returns true if all versions of the package satisfy the constraints
+ */
+async function printExplanationsAsync(
```
kbrandwijk, line 75: I do not expect a method called `print...` to also return something, in similar situations in the CLIs we usually split this up so there's no mixed concerns.

packages/expo-doctor/src/checks/IllegalPackageCheck.ts
```diff
@@ -0,0 +1,37 @@
...
+      const illegalPackages = [
+        '@unimodules/core',
+        '@unimodules/react-native-adapter',
+        'react-native-unimodules',
+      ];
+
+      for (const pkg of illegalPackages) {
+        const warning = await warnAboutDeepDependenciesAsync({ name: pkg }, projectRoot);
+        if (warning) {
+          issues.push(warning);
```
kbrandwijk, line 22: I don't understand this logic. `warnAboutDeepDependenciesAsync` returns the result of `printExplanationAsync` (which is either null or a string of concatenated issue lines). That is being pushed into an array here, which is then only used to check its length later. That means this whole issues array is actually just a flag 'issuesFound' that can get set in the loop.

packages/expo-doctor/src/utils/explainDependencies.ts
kbrandwijk: This bubbles up as well. The function that calls this (warnAbout....) also combines printing output (by calling into this method) and bubbling up a return value. This to me was one of the biggest issues with the old implementation. I would seriously consider refactoring this to achieve a clean separation between logic/flow and output.

packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (issues.length) {
+        advice.push(`Upgrade dependencies that are using the invalid package versions.`);
+      }
+    }
+
+    return {
+      isSuccessful: issues.length === 0,
+      issues,
+      advice,
+    };
```
kbrandwijk, line 39: This is a pattern that is used for multiple checks. You start with a list of packages, loop, warn, print a conclusion, return results. I would extract this into a utility function to clean up the individual checks.

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (pkg.version) {
+    return chalk`{bold ${pkg.name}}{cyan @}{${versionColor} ${pkg.version}}`;
+  } else {
+    return chalk`{bold ${pkg.name}}`;
+  }
+}
+
+/**
+ * @param pkg
+ * @returns true if all versions of the package satisfy the constraints
```
kbrandwijk, line 119: incorrect return value doc. Does not return true, returns `null` if valid, otherwise a string.

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+export async function warnAboutDeepDependenciesAsync(
```
kbrandwijk, line 121: method naming (see comment above about separation of concerns). This method doesn't (just) warn, this method actually performs the check, might warn if there's issues, and returns something.

packages/expo-doctor/src/utils/explainDependencies.ts
```diff
@@ -0,0 +1,132 @@
...
+function getInvalidPackagesWarning(
+  pkg: TargetPackage,
+  { explanations }: { explanations: RootNodePackage[] }
+): string {
+  const lines = [];
+  if (pkg.version) {
+    lines.push(`Expected package ${formatPkg(pkg, 'green')}`);
+  } else {
+    lines.push(`Expected to not find any copies of ${formatPkg(pkg, 'green')}`);
+  }
```
kbrandwijk, line 101: This logic is very hard to follow. It depends on a calling convention of the `warnAbout...` method several places up in the stack, that will only include a version for a package if the versions endpoint has that package listed. And even that logic only 'works' for the specific packages that we use the `warnAbout...` method with (illegalPackages - always without a version number, and supportPackages - only with a version number if they were still in use in a particular SDK version).

I think it would be much clearer to explicitly establish the results you are expecting in this call chain. You're looking for a package, and you either don't want to find it, or you want to find a specific version of it. This expectation currently only becomes clear here, 3 levels deep, after passing down through `warnAboutDeepDependenciesAsync` and `printExplanationsAsync`.

packages/expo-doctor/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "prepare": "yarn run clean && yarn run build:prod",
+    "lint": "eslint .",
+    "watch": "yarn run build -w",
+    "test": "jest",
+    "build": "ncc build ./src/index.ts -o build/",
+    "build:prod": "ncc build ./src/index.ts -o build/ --minify --no-cache --no-source-map-register",
+    "clean": "rimraf ./build/"
+  },
+  "engines": {
+    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
```
kbrandwijk, line 36: We dropped support for Node 12 in our tooling afaik (we also dropped Node 12 in our test matrix)

packages/expo-doctor/src/api/getSchemaAsync.ts
```diff
@@ -0,0 +1,9 @@
+import fetch from 'node-fetch';
+
+export async function getSchemaAsync(sdkVersion: string): Promise<any> {
```
kbrandwijk, line 3: A different function that uses sdkVersion - gteSdkVersion (https://github.com/expo/expo-cli/pull/4641/files#diff-6fdc6eae623e54996e6af13c27efa80c6d029dbfede1357467be1233471037b6R4) has a whole check in there for unversioned, here we just assume that `expoConfig.sdkVersion` is not unversioned and valid.

packages/expo-doctor/src/api/getSchemaAsync.ts
```diff
@@ -0,0 +1,9 @@
...
+  const result = await fetch(
+    `https://exp.host/--/api/v2/project/configuration/schema/${sdkVersion}`
+  );
+  const resultJson = await result.json();
+  return resultJson.data.schema;
```
kbrandwijk, line 8: This will crash if sdkVersion is not valid, endpoint output:
```json
{
"errors": [
{
"code": "VALIDATION_ERROR",
"message": "Version '44' is invalid. A version must be a valid semver or 'UNVERSIONED'.",
"isTransient": false
}
]
}
```

packages/expo-doctor/src/api/getVersionsAsync.ts
```diff
@@ -0,0 +1,44 @@
...
+  androidUrl: string;
+  androidVersion: string;
+  iosUrl: string;
+  iosVersion: string;
+  sdkVersions: SDKVersions;
+};
+
+/** Get versions from remote endpoint. */
+export async function getVersionsAsync({
+  skipCache,
```
kbrandwijk, line 39: Unused parameter?

packages/expo-doctor/src/api/getVersionsAsync.ts
```diff
@@ -0,0 +1,44 @@
...
+  beta?: boolean;
+};
+
+export type SDKVersions = Record<string, SDKVersion>;
+
+export type Versions = {
+  androidUrl: string;
+  androidVersion: string;
+  iosUrl: string;
+  iosVersion: string;
```
kbrandwijk, line 33: All these are unused.

packages/expo-doctor/src/api/getVersionsAsync.ts
```diff
@@ -0,0 +1,44 @@
...
+  /** @example "https://dpq5q02fu5f55.cloudfront.net/Exponent-2.17.4.tar.gz" */
+  iosClientUrl?: string;
+  /** @example "https://dev.to/expo/expo-sdk-38-is-now-available-5aa0" */
+  releaseNoteUrl?: string;
+  /** @example "2.17.4" */
+  iosClientVersion?: string;
+  /** @example "https://d1ahtucjixef4r.cloudfront.net/Exponent-2.16.1.apk" */
+  androidClientUrl?: string;
+  /** @example "2.16.1" */
+  androidClientVersion?: string;
```
kbrandwijk, line 16: Unused?

packages/expo-doctor/src/checks/GlobalPrereqsVersionCheck.ts
```diff
@@ -0,0 +1,113 @@
...
+  const cannotDetectWatchmanInstallationMessage =
+    'Cannot detect watchman installation.' + homebrewInstallNote;
+
+  try {
+    watchmanVersion = JSON.parse((await spawnAsync('watchman', ['version'])).stdout.trim()).version;
+  } catch (e: any) {
+    return cannotDetectWatchmanInstallationMessage;
+  }
+
+  // If we can't get the watchman version, `getVersionAsync` will return `null`
```
kbrandwijk, line 63: What is `getVersionAsync`, and if it doesn't exist anymore, do we need watchman at all?

packages/expo-doctor/src/checks/GlobalPrereqsVersionCheck.ts
```diff
@@ -0,0 +1,113 @@
...
+function _isNpmVersionWithinRanges(npmVersion: string, ranges: string[]) {
+  return ranges.some(range => semver.satisfies(npmVersion, range));
+}
+
+async function _checkNpmVersionAsync(): Promise<string | null> {
+  try {
+    try {
+      const yarnVersionResponse = await spawnAsync('yarnpkg', ['--version']);
+      if (yarnVersionResponse.status === 0) {
+        return null;
```
kbrandwijk, line 22: Why do we bail out of the _npm_ version check when yarn is found? All of our checks are hardcoded to use npm now (npm explain etc.), so this seems wrong.

packages/expo-doctor/src/checks/IllegalPackageCheck.ts
```diff
@@ -0,0 +1,37 @@
...
+
+      if (issues.length) {
+        advice.push(`Remove any 'unimodules' packages from your project.`);
+      }
+    }
+
+    return {
+      isSuccessful: !issues.length,
+      advice,
+      issues,
```
kbrandwijk, line 34:
```suggestion
      issues,
      advice,
```
Minor nit, just for consistency with all other checks.

packages/expo-doctor/src/checks/checks.types.ts
```diff
@@ -0,0 +1,19 @@
...
+  /** many checks currently output their own issues, no need to duplicate */
+  issues: string[];
+  // should include at least one bit of actionable advice to fix the issue(s)
+  advice: string[];
+}
+
+export interface DoctorCheckParams {
+  projectRoot: string;
+  exp: any;
+  pkg: any;
```
kbrandwijk, line 18: Could probably tighten down to the types from @expo/config:
```suggestion
  exp: ExpoConfig;
  pkg: PackageJSONConfig;
```

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+    return false;
+  }
+}
+
+export async function actionAsync(projectRoot: string) {
+  await warnUponCmdExe();
+
+  const { exp, pkg } = getConfig(projectRoot);
+
+  // expo-doctor relies on versioned CLI, which is only available for 44+
```
kbrandwijk, line 63: I thought versioned CLI was only officially supported from SDK 46 onwards (included as beta in SDK 45)?

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,103 @@
...
+  ];
+
+  let foundSomeIssues = false;
+
+  const checkParams = { exp, pkg, projectRoot };
+
+  for (const check of checks) {
+    if (!(await runCheckAsync(check, checkParams))) {
+      foundSomeIssues = true;
+    }
```
kbrandwijk, line 90: Similar to my comment about `warnAbout...` it would be great if running the checks, and outputting the result were separated. In my opinion, `runCheckAsync` should just return the results ({ isSuccessful, issues, advice}), and you can properly parse the results here (isSuccessful) and output issues and advice.

This would also work a lot better if you want to parallellize these checks in the future. Logging during checking would mix up the output of different checks together, while running them in parallel and having a single threaded loop to output the results doesn't have this issue.

To me, this is a serious architecture concern throughout.

packages/expo-doctor/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import path from 'path';
+
+import { actionAsync } from './doctor';
+
+const packageJson = () => require('../package.json');
+
+async function run() {
+  const args = process.argv.slice(2);
+
+  if (args.some(arg => ['-V', '-v', '--version'].includes(arg))) {
```
kbrandwijk, line 12: For consistency, our other CLIs (@expo/cli and eas-cli) only support `-v` and `--version`, not `-V`
```suggestion
  if (args.some(arg => ['-v', '--version'].includes(arg))) {
```

packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts
keith-kurak: used LT for global check; individual test checks are moot or now b/c it only works on 46+

packages/expo-doctor/src/doctor.ts
keith-kurak: you are correct! (no idea where I got the idea otherwise...)

packages/expo-doctor/src/api/getSchemaAsync.ts
keith-kurak: cleaned it up a little... an invalid sdkVersion in app.json should fail before this in both old doctor and new

yarn.lock
keith-kurak: do you have any idea what caused this? I don't see anything related to that package in my code

packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
keith-kurak: Now that the outer-try catch lets `Log.exception()` do the printing, we at least get the full exception for unexpected situations. Any advice for doing the debug logging combined with the ora spinner so things don't get out of order? With the regular "issues" output, I just delay printing until the check is done. The extra info from the error here (e.g., `Error: sh exited with non-zero code: 1`) doesn't add much, it'd be good to keep it separate.

yarn.lock
EvanBacon: Run `yarn why @expo/config-plugins` to find which package is depending on the wrong version.

packages/schemer/src/index.ts
keith-kurak: In this use case: yes; in all of them: definitely can't guarantee it. What hoping to see how AJV handles, but there's a lot of layers there.

Feels not great, but absent of reimplementing `get`, this should handle both:
```get(data, lowerFirst(fieldPath)) || get(data, fieldPath)```

File: yarn.lock (hunk 5 above)
Comment (keith-kurak) [reply yarn.lock]: Thanks! So... is it better to keep doctor at `@expo/config` 6.0.24 like the rest of the repo, or upgrade the repo to the latest?

**Comment by keith-kurak** [comment]
@kbrandwijk RE: most of the remaining open comments... a common answer to "why did you write it like that" is... I didn't write it 😄.  I copied a lot from the old Doctor, isolated it from dependencies we didn't want to carry forward, and tweaked it to play nice with new scaffolding. I didn't intend to rewrite all the checks from scratch at this stage, because I would stand a good chance of not fully-understanding their nuance and introducing bugs where we previously had production-battle-tested code, so I limited the scope to safe/ minimal refactors. 

I added some notes at the top about what comes from where (and also added comments to that effect).

File: packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts (hunk 17 above)
Comment (keith-kurak) [reply packages/expo-doctor/src/checks/SupportPackageVersionCheck.ts]: @EvanBacon scratch that, you inspired me to integrate the version checking into the check structure :-). All tests run on compatible projects currently, but the version checking is under test.

File: packages/expo-doctor/src/doctor.ts [hunk 51]
```diff
@@ -0,0 +1,118 @@
...
+    ora.fail();
+    return false;
+  }
+  if (result.isSuccessful) {
+    ora.succeed();
+    return true;
+  }
+  ora.fail();
+  if (result.issues.length) {
+    Log.log(chalk.underline(chalk.yellow(`Issues:`)));
```
EvanBacon, line 49:
```suggestion
    Log.log(chalk.underline.yellow(`Issues:`));
```
I think these can be chained btw. ndb

packages/expo-doctor/src/doctor.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (ltSdkVersion(exp, '46.0.0')) {
+      Log.error(
+        chalk.red(`expo-doctor supports Expo SDK 46+. Use 'expo-cli doctor' for SDK 45 and lower.`)
+      );
+      process.exitCode = 1;
+      return;
+    }
+  } catch (e: any) {
+    Log.exception(e);
+    process.exitCode = 1;
```
EvanBacon, line 83: I like using `Log.exit` for this because it returns `never` so TS knows the function never ends, it can also be easily mocked to determine if `process.exit` was invoked. I usually have `Log.exit` throw an error so tests are aware of unexpected functionality instead of just closing. [example](https://github.com/expo/expo/blob/dce431cd8afa1ffaa945bc2ae9796a560610c357/packages/%40expo/cli/src/__mocks__/log.ts#L17).

packages/expo-doctor/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    logHelpAndExit();
+  }
+
+  // TODO: add offline flag
+
+  const projectRoot = path.resolve(process.cwd(), args[0] ?? process.cwd());
+
+  await fs.access(projectRoot, constants.F_OK).catch((err: any) => {
+    if (err) {
+      console.error(`\x1b[31mProject directory ${projectRoot} does not exist\x1b[0m`);
```
EvanBacon, line 35: Use chalk instead. It's already imported by the import at the top of the file.

packages/expo-doctor/src/utils/log.ts
```diff
@@ -0,0 +1,81 @@
...
+export function warn(...message: string[]): void {
+  console.warn(...message.map(value => chalk.yellow(value)));
+}
+
+export function log(...message: string[]): void {
+  console.log(...message);
+}
+
+/** @deprecated use `debug` package with the `expo:` prefix instead.  */
+export function debug(...message: any[]): void {
```
EvanBacon, line 32: delete in favor of the `debug` package.

EvanBacon (approved):
lgtm

EvanBacon:
Just make sure the tests pass first

PR #4643, opened 2023-02-03
feat(create-expo-app): drop loading spinner when installing dependencies

EvanBacon:
# Why

- A spinner should only be used for short-running, indeterminate tasks.
- @gaearon mentioned that the current behavior is not sparking joy: https://twitter.com/dan_abramov/status/1621355168792039426?s=20&t=T-XihEJP2vTJuHT1igMpcg

# How

- Dropping the spinner in favor of the raw logs because we support three different package managers and it's not entirely obvious how we can create a consistent loading bar UI across all three (`yarn install --json` output differs a lot from `npm install --json`)

# Test Plan

- Ran locally
- Unit tests (minimal overlap)

### DIFF
--- a/packages/create-expo-app/src/createAsync.ts
+++ b/packages/create-expo-app/src/createAsync.ts
@@ -18,7 +18,6 @@ import {
   initializeAnalyticsIdentityAsync,
   track,
 } from './telemetry';
-import { env } from './utils/env';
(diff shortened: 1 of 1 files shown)


EvanBacon:
Available in `create-expo-app@1.2.0`

PR #4644, opened 2023-02-06
feature(create-expo-app): add `--example` flag to initialize from example

byCedric:
# Why

Fixes ENG-7493

# How

- Based on the code from [`create-react-native-app`](https://github.com/expo/create-react-native-app/blob/main/src/Examples.ts), but adopted to this repository

> **Warning**
> I ran into the "premature close" error when testing this: https://github.com/isaacs/node-tar/issues/321. I upgraded `tar@6.1.11` to `tar@6.1.13` to avoid this.

# Test Plan

See tests and help info:

```bash
$ yarn create expo-app --help

# Show list of examples to create from
$ yarn create expo-app --example
# Or create a new example by name
$ yarn create expo-app --example with-router
```

### DIFF
--- a/packages/create-expo-app/src/createAsync.ts
+++ b/packages/create-expo-app/src/createAsync.ts
@@ -1,7 +1,13 @@
 #!/usr/bin/env node
+import chalk from 'chalk';
 import fs from 'fs';
 import path from 'path';
 
+import {
+  downloadAndExtractExampleAsync,
+  ensureExampleExists,
+  promptExamplesAsync,
+} from './Examples';
 import * as Template from './Template';
 import { promptTemplateAsync } from './legacyTemplates';
 import { Log } from './log';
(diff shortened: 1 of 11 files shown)


packages/create-expo-app/src/createAsync.ts
```diff
@@ -78,7 +71,15 @@ async function setupDependenciesAsync(projectRoot: string, props: Pick<Options,
...
-export async function createAsync(inputPath: string, props: Options): Promise<void> {
+export async function createAsync(inputPath: string, options: Options): Promise<void> {
+  if (options.example) {
+    return await createExampleAsync(inputPath, options);
+  }
+
+  return await createTemplateAsync(inputPath, options);
+}
+
+async function createTemplateAsync(inputPath: string, props: Options): Promise<void> {
```
EvanBacon, line 82: Should assert that template and example cannot be used together.

EvanBacon (approved):

packages/create-expo-app/src/createAsync.ts
byCedric: Makes sense, will add it tomorrow!

PR #4645, opened 2023-02-07
fix(xdl): remove unused import from `createBundlesAsync`

byCedric:
# Why

CI is currently failing on all PRs due to an eslint issue

# How

Fixed that eslint issue

# Test Plan

See if CI passes


byCedric:
Note, it seems that there is still an XDL test failure. This PR only intends to unblock the CI for other PRs. Probably have to take a look at it later.

PR #4660, opened 2023-02-24
[doctor] remove watchman check

keith-kurak:
# Why

Resolves https://linear.app/expo/issue/ENG-7606/doctor-dont-require-watchman. Expo works fine without watchman on macOS, so let's not warn about it.

# How

Removed the check

# Test Plan

Tried without watchman installed; no warning

### DIFF
--- a/packages/expo-doctor/src/checks/__tests__/GlobalPrereqsVersionCheck.test.ts
+++ b/packages/expo-doctor/src/checks/__tests__/GlobalPrereqsVersionCheck.test.ts
@@ -1,0 +1,60 @@
+import spawnAsync from '@expo/spawn-async';
+import { execSync } from 'child_process';
+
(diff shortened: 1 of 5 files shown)


EvanBacon (approved):

packages/expo-doctor/src/checks/__tests__/GlobalPrereqsVersionCheck.test.ts
```diff
@@ -0,0 +1,60 @@
...
+// required by runAsync
+const additionalProjectProps = {
+  exp: {
+    name: 'name',
+    slug: 'slug',
+  },
+  pkg: {},
+};
+
+describe(GlobalPrereqsVersionCheck, () => {
```
EvanBacon, line 18: we generally avoid redundant describe blocks (single blocks which match the name of the test file).

PR #4662, opened 2023-02-27
[doctor] check if global cli is installed in project

keith-kurak:
# Why
We now recommend users upgrade by using `npx expo install --fix`, but some have `expo-cli` installed in project for legacy reasons, and that can break the `--fix` flag with `error: unknown option --fix’`.

# How

Added a check, much like the other illegal package checks.

NOTE: I think we'll end up with many illegal package checks, so will likely eventually refactor to reduce that to adding package name and customized advice without so much ceremony, just not yet.

# Test Plan
- Add `expo-cli` to a project, get warning.
- Remove from project, no more warning.


EvanBacon (approved):
lgtm

PR #4672, opened 2023-03-28
fix(doctor): replace commander-based interactivity check with tty/ci check

byCedric:
# Why

Doctor had 2 different parts from legacy and newer code.

- **Legacy**: the `nonInteractive` flag on `commander`'s `program` singleton
- **Newer**: uses `args` instead of `commander`

# How

- Added `isInteractive()` function in **./utils/interactive.ts**
- Added `CI` environment variable in **./utils/env.ts**
- Refactored Ora's `disabled` check to use `isInteractive() || env.EXPO_DEBUG`

# Test Plan

Ora should now be disabled in non-interactive environments.

### DIFF
--- a/packages/expo-doctor/src/utils/env.ts
+++ b/packages/expo-doctor/src/utils/env.ts
@@ -1,6 +1,11 @@
 import { boolish } from 'getenv';
 
 class Env {
+  /** Is running in non-interactive CI mode */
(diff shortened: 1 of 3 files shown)


keith-kurak (commented):
LGTM! thanks!

PR #4678, opened 2023-04-05
Update CLI description to mention expo-router

lekterable:
# Why

CLI still mentions `react-navigation` when in reality `expo-router` is being used

# How

I've updated the description


byCedric:
This PR has been merged through https://github.com/expo/expo-cli/pull/4799

PR #4679, opened 2023-04-07
fix: Create <meta name="description" /> tag in index.html from web.description app config

44px:
# Why

According to table  here https://docs.expo.dev/guides/progressive-web-apps/#usage `<meta name="description" />` should be populated with value of `web.description` or `description` fields in `app.config.js`.

Currently, providing only `web.description` does not create meta tag, because value is read only from top-level `description` field.

# How

Read `web.description` first as it should take precedence in case of PWA manifest generation

# Test Plan

1. Create expo web project
2. Fill only `web.description` in `app.config.js`
3. Export web, check that `<meta name="description" />` is present in `index.html` and filled with value from `web.description` in `app.config.js`


byCedric:
Hi @44px! Thanks for the PR, unfortunately, we have moved all webpack related tooling into https://github.com/expo/expo-webpack-integrations and are not actively maintaining this anymore (we are switching to Metro support for web bundling).

If you still think we should move forward with this PR, please reopen this PR on https://github.com/expo/expo-webpack-integrations.

Hope this helps!

PR #4686, opened 2023-04-18
[cli] remove credentials:manager functionality

quinlanj:
# Why

Since the planned last date expo build will work is January 4, 2023, we don't have any more use for credentials manager. Adding deprecation notice and removing functionality to discourage people from using it.

# How

- add deprecation notice to command description and print it when run

# Test Plan

Screenshot 2023-04-17 at 11 52 43 PM

### DIFF
--- a/packages/expo-cli/src/commands/credentials.ts
+++ b/packages/expo-cli/src/commands/credentials.ts
@@ -1,3 +1,4 @@
+import chalk from 'chalk';
 import type { Command } from 'commander';
 
 import { applyAsyncActionProjectDir } from './utils/applyAsyncAction';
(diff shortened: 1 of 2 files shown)


dsokal (approved):

ide (changes requested):
What about the fetch and push commands?

I also think we'll want to publish two versions of Expo CLI. The first is a minor version (6.4.0) that doesn't break any features but marks the publish and credential commands as superseded and prints out warnings. The second is a major version (7.0.0) that removes support for the publish commands, and maybe deletes the code too.

This way, developers who stay on 6.x can keep the old functionality and learn about the warnings, and everyone else who installs 7.0.0 will not have the old commands at all.

So, I'd suggest splitting this PR into two: one that describes the credential commands as superseded and prints out deprecation warnings, and one that removes the credential commands altogether.

PR #4703, opened 2023-05-09
bump expo/config

EvanBacon:


brentvatne:
this doesn't entirely solve the issue still because npm only considers a peer dep to be satisfied if it is a direct dependency of the project (eg: `expo -> @expo/config` will not satisfy this peer dep). i think we should have a peer dep on the `expo` package instead, and import `@expo/config` from there

EvanBacon:
@brentvatne that won't satisfy the imports to `@expo/config/paths`

brentvatne:
should we expose paths in the public api if we need it in other modules?

PR #4706, opened 2023-05-17
[doctor] remove sh command from dep check for Windows compat

keith-kurak:
# Why
Expo Doctor's dependency version check isn't working at all on Windows right now (other than in WSL) 🙈 .

When calling npx expo install --check, we had to pipe a "y" into the command so it would finish, [necessitating this technique for calling a child process](https://stackoverflow.com/a/39482554). Since this uses sh, it doesn't work in Windows.

Fortunately, there's a WAY easier way to do this

# How
Use the `CI=1` environment variable, which causes the versioned CLI to disable the interactive prompt on `npx expo install --check`.

# Test Plan
- [x] works the same on macOS (same output, etc.)
- [ ] works on Windows (no PC available until next week for me to check this, though I can't imagine it works worse than it does now)

### DIFF
--- a/packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
+++ b/packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
@@ -18,9 +18,10 @@ export class InstalledDependencyVersionCheck implements DoctorCheck {
     try {
       // only way to check dependencies without automatically fixing them is to use interactive prompt
       // In the future, we should add JSON output to npx expo install, check for support, and use that instead
-      await spawnAsync('sh', ['-c', 'echo "n" | npx expo install --check'], {
+      await spawnAsync('npx', ['expo', 'install', '--check'], {
         stdio: 'pipe',
         cwd: projectRoot,
+        env: { ...process.env, CI: '1' },
       });
     } catch (error: any) {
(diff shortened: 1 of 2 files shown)


kbrandwijk:
Ideally, `npx expo install --fix` should support a `--non-interactive` flag, this CI=1 solution now relies on internal logic in a different CLI that turns it non-interactive when set, but I don't think a separate tool should rely on internal logic like that.

keith-kurak:
> Ideally, `npx expo install --fix` should support a `--non-interactive` flag, this CI=1 solution now relies on internal logic in a different CLI that turns it non-interactive when set, but I don't think a separate tool should rely on internal logic like that.

I do plan to add such a flag, but we need it to work for current/ previous SDK versions, as well, so a workaround like this will probably be around for a while. In the category of things we can do to Doctor right now to make it work on Windows without updating current and previous releases of the versioned CLI, do you prefer this PR or [this other one](https://github.com/expo/expo-cli/pull/4705), which avoids the CI flag and uses a pipe to suppress the prompt, like the current version does? I could go either way, but the latter relies on Windows behavior I'm much less certain of, at least when comparing to my confidence in that `CI` flag.

kbrandwijk:
As a workaround, this is fine, as long as that intention exists.

PR #4732, opened 2023-07-07
[doctor] check for sdkVersion in Expo config

keith-kurak:
# Why
Detects when someone's upgrade goes awry due to `sdkVersion` in Expo config overriding the actual installed SDK version. (e.g., https://github.com/expo/expo/issues/23351)

# How
We can't use `exp` from `getExpoConfig()` because this always populates `sdkVersion` (either based on installed version or `expo.sdkVersion`). So, we have to read the actual Expo config to see if `sdkVersion` was set. This is really only practical in Doctor for a static config file. We could possibly attempt some comparison between exp's `sdkVersion` and **package.json**, but there'd be limitations to have to tiptoe around there.

One other option would be to actually change `@expo/config` to expose where `sdkVersion` came from. Not sure if it's worth the effort, though.

@brentvatne I can add some unit tests to this, but wanted to first run the concept by you first to see if we should take this approach or go further in one of the directions described above before I spent more time on it.

@brentvatne 

# Test Plan

The message itself:
image

When you put in the wrong `sdkVersion`, you get a ton of wrong package versions from the step where it checks against **bundledNativeModules**.

### DIFF
--- a/packages/expo-doctor/src/checks/ExpoConfigCommonIssueCheck.ts
+++ b/packages/expo-doctor/src/checks/ExpoConfigCommonIssueCheck.ts
@@ -1,0 +1,34 @@
+import JsonFile from '@expo/json-file';
+
+import { DoctorCheck, DoctorCheckParams, DoctorCheckResult } from './checks.types';
+
+export class ExpoConfigCommonIssueCheck implements DoctorCheck {
+  description = 'Check Expo config for common issues';
+
+  sdkVersionRange = '*';
+
+  async runAsync({ exp }: DoctorCheckParams): Promise<DoctorCheckResult> {
+    const issues: string[] = [];
+
+    // check if SDK version is in Expo config
+    // exp object derives sdkVersion from expo.sdkVersion or installed version, so it will always be populated.
+    // Therefore, we need to check app.json file itself to see if this value is defined.
+    // Unlikely we can also check a dynamic config unless we change @expo/config.
+    const staticConfigPath = exp?._internal?.staticConfigPath;
+
+    // exp provides an empty object if there is no static config
+    if (staticConfigPath && typeof staticConfigPath === 'string') {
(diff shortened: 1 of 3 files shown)


packages/expo-doctor/__mocks__/fs.js
```diff
@@ -1,2 +1,3 @@
-import { fs } from 'memfs';
+import { fs, realpathSync } from 'memfs';
+module.exports = realpathSync;
 module.exports = fs;
```
byCedric, line 3: Does multiple `module.exports = ...` work? I think the `realPathSync` is just getting overwritten when exporting, right?

packages/expo-doctor/src/checks/ExpoConfigCommonIssueCheck.ts
```diff
@@ -0,0 +1,55 @@
...
+    };
+  }
+}
+
+/**
+ * Resolve the Expo SDK Version either from the input Expo config or from the installed
+ * version of the `expo` package.
+ * Adapted from https://github.com/expo/expo/blob/main/packages/%40expo/config/src/getExpoSDKVersion.ts
+ */
+function getExpoSDKVersionFromPackage(projectRoot: string): string | undefined {
```
byCedric, line 39: Is this something we can ask `@expo/config`? If so, I think that might be a better bet. The main reason for that is that we probably want `@expo/config` to handle all configuration loading, including the app manifest.

When we change something in a future version of `@expo/config`, we also need to make that change here. That is, unless we use the installed `@expo/config` package from the project and just use that here.

I do the same for `expo-github-action`, where I pull the configuration file, required to retrieve basic information for the update / preview sub-action. ([see this](https://github.com/expo/expo-github-action/blob/main/src/expo.ts#L146-L160))

byCedric (approved):
LGTM

PR #4734, opened 2023-07-11
[doctor] multiple lock file, unintentionally-bare detection

keith-kurak:
# Why
Multiple "project structure" tests in here (to avoid chicken-and-egg-esque merge mayhem):
- Warn on multiple lock files
- Warn if possibly-unintentionally bare

# How
For the lockfile check, just checking for the existence of more than one lockfile.

For the bare check, right now I'm checking if you have config plugins and ios/ android folders. I know that's not the only situation where you could be unintentionally bare (and you could be intentionally bare and just have left those lines in your Expo config), but not sure of a better way to detect this, and I would really like to detect it, as misunderstanding prebuild and what it means to "use" it comes up pretty regularly in support requests.

## Note on structure
One of the pain points with adding individual checks to doctor right now is that you either have to a) figure out what `Check` class, aka "category" a check goes in, possibly mixing unlike things together, or b) create a new `Check` class. My plan is to address this once we are checking more things, with a pattern like what is [described here](https://github.com/expo/expo-cli/pull/4731), recategorizing by how the check is performed and reworking the UI to not over-explain things.

In the meantime, the "categories" will expand a bit (like they did here). 

# Test Plan
Lockfile check:
image

Bare detection:
image

### DIFF
--- a/packages/expo-doctor/src/checks/ProjectSetupCheck.ts
+++ b/packages/expo-doctor/src/checks/ProjectSetupCheck.ts
@@ -13,7 +13,7 @@ export class ProjectSetupCheck implements DoctorCheck {
     // ** possibly-unintentionally-bare check **
 
     if (
-      (fs.existsSync(`${projectRoot}/ios`) || fs.existsSync(`${projectRoot}/android`)) &&
+      (fs.existsSync(path.join(projectRoot, 'ios')) || fs.existsSync(path.join(projectRoot', 'android'))) &&
       exp.plugins?.length
     ) {
       issues.push(
--- a/packages/expo-doctor/src/checks/ProjectSetupCheck.ts
+++ b/packages/expo-doctor/src/checks/ProjectSetupCheck.ts
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import path from 'path';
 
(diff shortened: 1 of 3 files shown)


packages/expo-doctor/src/checks/ProjectSetupCheck.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  sdkVersionRange = '*';
+
+  async runAsync({ exp, projectRoot }: DoctorCheckParams): Promise<DoctorCheckResult> {
+    const issues: string[] = [];
+
+    // ** possibly-unintentionally-bare check **
+
+    if (
+      (fs.existsSync(`${projectRoot}/ios`) || fs.existsSync(`${projectRoot}/android`)) &&
```
byCedric, line 16: I think this still misses the "Are the `/ios` or `/android` folders gitignored"-check.

Right now, if users get presented with this warning, and add it to their gitignore, the warning will still be visible after making that change right?

Also, prefer `path.join` over self-concatenation (usually for Windows):
```suggestion
      (fs.existsSync(path.join(projectRoot, 'ios')) || fs.existsSync(path.join(projectRoot', 'android'))) &&
```

byCedric (approved):
LGTM!

PR #4736, opened 2023-07-13
chore(webpack-config): fix import name

cyrusduong:
# Why

Looks like a typo... We ran into an issue trying builder.io beta RN build that depends on @expo/webpack-config

# How

Trying to build https://github.com/BuilderIO/builder/blob/main/examples/react-native and following their README instructions. Failing when running expo web resolving `expo/config`. It seems fixed with the changed line.

# Test Plan

Should run against existing tests.... and fix
https://github.com/expo/expo-cli/actions/runs/5083949369/jobs/9135598475#step:5:99


byCedric:
Hi @cyrusduong! This is actually the correct path to [load the `@expo/config` package shipped with that specific `expo` version](https://cdn.jsdelivr.net/npm/expo@49.0.21/config.js). We have multiple references like this, e.g. `expo/metro-config`.

Closing as this was initially correct.

PR #4746, opened 2023-08-16
Revert "fix(webpack-config): fix sourcemap bug with expo camera"

marklawlor:
Reverts expo/expo-cli#4735

We are removing `@koale/useworker`  so these rules can be moved 

https://github.com/expo/expo/pull/23967


EvanBacon:
Probably ok to keep these around for a while since the webpack config isn't versioned with the SDK.

byCedric:
Closing this due to inactivity. `@expo/webpack-config` has been moved to https://github.com/expo/expo-webpack-integrations.

PR #4761, opened 2023-09-20
[image-utils] [ENG-9956] add a function to check whether a PNG is valid

kadikraman:
# Why

We want to check whether the app icon and splash screen images in the codebase are invalid before pre-building so we could show the user an error message and fail the build. [ENG-9956](https://linear.app/expo/issue/ENG-9956/discuss-where-to-validate-app-icons-files-splash-screen-files).

# How

Added an `getPngInfo` function returns the PNG data if a valid PNG is found, and throws otherwise.

Intended usage:

```js
import { getPngInfo } from '@expo/image-utils';

function assertValidPng(filePath: string) {
  try {
    getPngInfo(filePath);
  } catch (error) {
    if (error.code === 'ENOENT') {
      throw new CommandError('IMAGE_NOT_FOUND', `There was no image at ${filePath}`);
    }

    if (error.message.includes('Invalid file signature')) {
      throw new CommandError('IMAGE_CORRUPTED', `The image at ${filePath} is corrupted: ${error.message}`);
    }

    // unknown error
    throw error;
  }
}
```

# Test Plan

Added `packages/image-utils/src/__tests__/image-test.ts` to verify for different file types.


packages/image-utils/src/Image.ts
```diff
@@ -212,3 +214,12 @@ export async function compositeImagesAsync({
     .composite([{ input: foreground, left: x, top: y }])
     .toBuffer();
 }
+
+export async function isValidPng(src: string) {
```
kadikraman, line 218: Alternative method signature:

```js
isValidPng(src: string): { isValid: boolean, errorMessage?: string }
```
or
```js
isValidPng(src: string): { isValid: boolean, error?: Error }
```
or even
```js
isValidPng(src: string): { isValid: boolean, errorCode?: string, errorMessage?: string, error?: Error }
```

For the test cases I added I've seen two errors:
1. `ENOENT: no such file or directory, open 'random/path'`
2. `Invalid file signature`

So I could either map these to error codes or pass them down as is, so we could show a more informative error message ("check <path> is a valid `.png`" vs "file at <path> does not exist".

Not sure where the user-facing error message should live. Thoughts?

byCedric (approved):
Looking good, thanks for working on this!

PR #4767, opened 2023-10-05
update webpack readme

EvanBacon:
# Why

Drop info about exports that we don't want to maintain long term.


byCedric (approved):

byCedric:
Merged through https://github.com/expo/expo-webpack-integrations/pull/3

PR #4772, opened 2023-10-23
Update README.md

thorwebdev:
- Fix broken link.

# Why

# How

# Test Plan


thorwebdev:
Hmm, actually this seems to be something different. Where does `create-expo-app` live? https://www.npmjs.com/package/create-expo-app points to https://github.com/expo/expo-cli but that seems incorrect.

byCedric (approved):

